PR #90, opened 2020-01-27
Improve Package Management to handle reference assemblies

KevinRansom:
This is the interactive version of this pr: https://github.com/dotnet/try/pull/762

1, Clean up generated project a bit
2. New way of computing package root - collect roots for dependent packages
3. Handle references to reference assemblies.
4. Does a better job of handling Framework assembly references

### DIFF
--- a/Microsoft.DotNet.Interactive.Tests/LanguageKernelPackageTests.cs
+++ b/Microsoft.DotNet.Interactive.Tests/LanguageKernelPackageTests.cs
@@ -753,7 +753,7 @@ using XPlot.Plotly;");
             var events = kernel.KernelEvents.ToSubscribedList();
 
             await kernel.SubmitCodeAsync(@"
-#r ""nuget:System.Text.Json""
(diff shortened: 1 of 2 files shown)


Microsoft.DotNet.Interactive.Tests/LanguageKernelPackageTests.cs
```diff
@@ -804,5 +804,48 @@ public async Task Pound_r_nuget_with_no_version_displays_the_version_that_was_in
...
+tInput.Length"
+            };
+
+            await SubmitCode(kernel, source);
+
+            KernelEvents
+                .Should()
+                .ContainSingle(e => e is ReturnValueProduced);
+
+            KernelEvents
```
jonsequitur: These can be combined:

```csharp
KernelEvents
                .Should()
                .ContainSingle<ReturnValueProduced>()
                .Which
                .Value
                .Should()
                .Be(4);
```

PR #102, opened 2020-01-30
Improve kernel routing

jonsequitur:
This has a bunch of changes:

* Improve kernel routing mechanism
* Make `KernelInvocationContext` an `IAsyncDisposable` and add the `OnComplete` method to enable actions to be scheduled for right before the context is completed.
* Introduce `DirectiveCommand` (internal)


Microsoft.DotNet.Interactive.Jupyter/KernelExtensions.cs
```diff
@@ -124,6 +125,38 @@ private static T UseTime<T>(this T kernel)
...
+            static Command time()
+            {
+                return new Command("#!time")
+                {
+                    Handler = CommandHandler.Create((KernelInvocationContext context) =>
+                    {
+                        var timer = new Stopwatch();
+                        timer.Start();
+                        
+                        context.OnComplete(invocationContext =>
```
colombod: love this

PR #103, opened 2020-01-30
kernel server not listening to input

colombod:


Microsoft.DotNet.Interactive/Server/StandardIOKernelServer.cs
```diff
@@ -35,6 +35,10 @@ public class StandardIOKernelServer : IDisposable
 
             _disposables = new CompositeDisposable
             {
+                _input.Subscribe(async line =>
```
jonsequitur: Oops. I believe I removed this because it was not covered by any tests. Can we add a test for this?

colombod: Removed the interface introduced for testing only. Trying to track the "started" state of the server now

PR #106, opened 2020-01-31
make kernel server honor `--log-path` and listen to kernel logging events

brettfo:
The fix was two parts:

1. Call `Program.StartToolLogging()` in `CommandLineParser.cs`.
2. Enable events from the type `KernelBase` in `Program.cs`.

Everything else was to get a proper unit test.

### DIFF
--- a/dotnet-interactive.Tests/CommandLine/CommandLineParserTests.cs
+++ b/dotnet-interactive.Tests/CommandLine/CommandLineParserTests.cs
@@ -1,15 +1,20 @@
-// Copyright (c) .NET Foundation and contributors. All rights reserved.
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
+using System.Collections.Generic;
 using System.CommandLine.Binding;
 using System.CommandLine.IO;
 using System.CommandLine.Parsing;
 using System.IO;
+using System.Threading;
 using System.Threading.Tasks;
 using FluentAssertions;
 using Microsoft.DotNet.Interactive.App.CommandLine;
+using Microsoft.DotNet.Interactive.Commands;
+using Microsoft.DotNet.Interactive.Server;
 using Microsoft.DotNet.Interactive.Telemetry;
+using Microsoft.DotNet.Interactive.Utility;
 using Microsoft.Extensions.DependencyInjection;
 using Xunit;
 using Xunit.Abstractions;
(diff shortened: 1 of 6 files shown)


Microsoft.DotNet.Interactive/Utility/DirectoryUtility.cs
```diff
@@ -85,5 +86,21 @@ private static DirectoryInfo EnsureExists(this DirectoryInfo directory)
 
             return directory;
         }
+
+        public static async Task<bool> WaitForFiles(this DirectoryInfo directory, TimeSpan timeout, Func<FileInfo[], bool> trigger)
```
jonsequitur: It would be better to accept a `CancellationToken` and throw if it elapses.

jonsequitur: Given the way it's being used you might also change the return type to a `FileInfo` and return the specific file.

dotnet-interactive.Tests/CommandLine/CommandLineParserTests.cs
```diff
@@ -146,5 +152,49 @@ public async Task jupyter_returns_error_if_connection_file_path_is_not_passed()
 
             testConsole.Error.ToString().Should().Contain("Required argument missing for command: jupyter");
         }
+
+        [Fact(Timeout = 45000)]
```
jonsequitur: The timeout is unnecessary.

dotnet-interactive/build-and-install-dotnet-interactive.ps1
```diff
@@ -3,11 +3,9 @@ $ErrorActionPreference = "Stop"
 
 $thisDir = Split-Path -Parent $PSCommandPath
 $toolLocation = ""
-$toolVersion = ""
 if (Test-Path 'env:DisableArcade') {
-    dotnet pack "$thisDir\dotnet-interactive.csproj" /p:Version=0.0.0
+    dotnet pack "$thisDir\dotnet-interactive.csproj"
```
jonsequitur: Why do away with the version number?

Installing the specific version ensures that we don't grab a version from a nuget feed if it's higher than the local build, or if the local build is missing.

Microsoft.DotNet.Interactive/Utility/DirectoryUtility.cs
brettfo: Honestly, all of this is a hack to see if I can get the test to consistently pass in CI.  Once I can get it to pass I'll make it look nice.  Naturally it works perfectly on my local machine.

PR #107, opened 2020-01-31
Support sending input_request and receiving input_reply messages via the stdin channel

daxian-dbw:
Make updates to support sending `input_request` and receiving `input_reply` messages via the `stdin` channel.

Major changes:
- Add `string Send(InputRequest message);` to `IJupyterMessageSender` to send message on the `stdin` socket.
- Add `InputRequested` and `PasswordRequested` types that derives from `KernelEventBase`, and handle them in `ExecuteRequestHandler`, since `InputRequested` should only be requested when handling an `execute_request`.
- Add `StdInChannel` and `MessageReceiver`. `StdInChannel` accepts a `MessageSender` and a `MessageReceiver`.
- Add the static method `input(string prompt = "")` for requesting non-password user input in C#.
- Add the static method `password(string prompt = "")` for requesting password user input in C# (thanks @colombod for the help!)
- Update `JupyterMessageSender` and `JupyterRequestContext` to take a `StdInChannel` object.
- Very minor refactoring:
   - rename `_shellSender` and `_ioPubSender` to `_shellChannel` and `_ioPubChannel`
   - fix the parameter indentation in `JupyterRequestContext.cs`

Screen Shot 2020-02-11 at 6 45 01 AM

Screen Shot 2020-02-11 at 6 47 46 AM


daxian-dbw:
> I'd like to add tests, but so far there is no way to make the sub kernels to send a InputRequested event.

Given that the PS kernel PR has been merged, I can work on supporting reading user input in PowerShell (`Read-Host`) as long as the changes to the infrastructure layer roughly make sense 😄.
Then I can add tests to verify this using the PS kernel (of course, fixing the failing tests too).

jonsequitur:
It would be very helpful to include tests at the `Microsoft.DotNet.Interactive` level early on so we can discuss the design. What is the command and event flow that makes sense at this layer and would make sense as well, for example, in kernel-server mode or via a command-line REPL?

Additional things worth discussing:

* Should we create helpers like `input`, `getpass`, and so on? 
* What would requests for other types of input look like and how might we create custom elements to receive that input at the client, e.g. ipywidgets?
* Should we intercept `Console.ReadLine` so that it requests text input?

Microsoft.DotNet.Interactive/StdInChannelNotSupportedException.cs
```diff
@@ -0,0 +1,15 @@
+// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+
+namespace Microsoft.DotNet.Interactive
+{
+    public class StdInChannelNotSupportedException : Exception
```
colombod: This is a jupyter layer object, should be moved there

Microsoft.DotNet.Interactive/Events/InputRequested.cs
```diff
@@ -0,0 +1,41 @@
...
+    {
+        [JsonConstructor]
+        internal InputRequested(string prompt, bool password)
+        {
+            Prompt = prompt;
+            Password = password;
+        }
+
+        public InputRequested(
+            string prompt,
```
colombod: prompt should be optional with default value

Microsoft.DotNet.Interactive/Events/InputRequested.cs
```diff
@@ -0,0 +1,41 @@
...
+            bool password,
```
colombod: this could be easier to understand as isPassword with default as false.

Microsoft.DotNet.Interactive/StdInChannelNotSupportedException.cs
daxian-dbw: I expect the sub-kernels would need to catch this exception, for example the `PowerShellKernel` when executing `PSHost.UI.ReadLine()` or `PSHost.UI.ReadLineAsSecureString()`.
But the sub kernel projects doesn't refer to `Microsoft.DotNet.Interactive.Jupyter` (the jupyter layer depends on the sub kernels), so it would be impossible for individual language kernels to catch the exception if we move the exception definition to the jupyter layer.

Microsoft.DotNet.Interactive/Events/InputRequested.cs
daxian-dbw: What default value shall we use?

Microsoft.DotNet.Interactive/Events/InputRequested.cs
daxian-dbw: Agree. Change the parameter name to `isPassword`.
Do you mean to rename the property `InputRequested.Password` too? If that's the case, would you want me to add `[JsonProperty("Password")]` to that property? (I notice `KernelEventBase` sub-types all have `JsonConstructor`, but not sure if they really got serailized/deserialized anywhere).

Microsoft.DotNet.Interactive/Events/InputRequested.cs
daxian-dbw: Personally, I tend to avoid default parameter values for public APIs. I prefer a callsite to be explicit about the parameter it's using :)
But if this is the code style commonly used in dotnet-interactive, I will add default parameter values to them.

Microsoft.DotNet.Interactive/StdInChannelNotSupportedException.cs
daxian-dbw: I thought about just using `System.NotImplementedException`, but felt having a specific exception type would make it more accurate to handle this situation.

jonsequitur: If this is an expected case for some clients, then it's not really an exception. It would be better notify the user that the given operation isn't supported on the current frontend. 

I'm not sure this requires any changes at the `Microsoft.DotNet.Interactive` layer at this stage.

daxian-dbw: >  It would be better notify the user that the given operation isn't supported on the current frontend.

Agreed. IMO, this exception is only a way to notify the sub-kernel that the stdin channel is not available, and it's the kernel's responsibility to decide what to return from the `ReadLine` method and generate the proper message to notify the user.

daxian-dbw: I removed the `StdInChannelNotSupportedException` exception. Instead, the value of `AllowStdin` is passed to the `InputRequested` instance in `ExecuteRequestHandler`, and `InputRequested.Content` will throw `NotSupportedException` when it's not allowed to use stdin.
Please take another look to see if this is better.

Microsoft.DotNet.Interactive/Events/InputRequested.cs
daxian-dbw: I added the static method `input` to enable requesting input in C# and F#, and default parameter values are used there:
```
        public static string input(
            string prompt = "",
            bool isPassword = false)
```
Given that when calling this public constructor, we will pass in the `prompt` and `isPassword` values anyways, would it make more sense to not have default values here?

Microsoft.DotNet.Interactive/Events/InputRequested.cs
```diff
@@ -0,0 +1,51 @@
...
+            IsPassword = isPassword;
+        }
+
+        [JsonIgnore]
+        private string _content;
+
+        [JsonIgnore]
+        public string Content
+        {
+            get => AllowStdin ? _content : throw new NotSupportedException("The stdin channel is not allowed by the frontend.");
```
jonsequitur: Throwing on a getter can be a gotcha. What's the scenario for allowing this event to be created at all in the case where a stdin channel is not supported? Can we handle this higher up?

daxian-dbw: Hmm, yes, handling this higher up would be better. Let me play this a bit more.

daxian-dbw: @jonsequitur Could you please take another look? I add the `AllowStdin` property to `SubmitCode`, and set that in `ExecuteRequestHandler.Handle` when constructing the `SubmitCode` instance.

jonsequitur: I suspect that the decision to allow stdin would apply to the whole session, based on the frontend, and therefore be more a configuration-level concern than something that would be communicated at the command level. 

We've been thinking for a while about introducing some API to learn more about the client since different clients have different capabilities. This seems related.

daxian-dbw: > I suspect that the decision to allow stdin would apply to the whole session, based on the frontend

I also feel this way. But the fact that the `allow_stdin` is passed along by `execute_request` message makes me wonder if it could be possible a client allows the stdin channel sometimes but not others.

jonsequitur: Interesting. While that's possible, I prefer to avoid letting Jupyter drive the design of the core abstractions. 

Maybe the Jupyter layer could modify this behavior via middleware?

daxian-dbw: A few tests are added to the Jupyter layer to validate the `input_request` message is issued correctly.

Capture our conversation here:
> One approach might be in the `input` helper method, check the current frontend's capabilities, and make a decision there.
Checking the capability of the frontend is something we could access from the `KernelInvocationContext` potentially.
It might be a property bag of sorts, with strong-typed access.
Air code:
>```
>if (!KernelInvocationContext.Current.Get<ClientConfiguration>().IsStdInSupported)
>{
>    await KernelInvocationContext.Current.DisplayErrorAsync("Sorry!");
>}
>```
>No protocol message to check for the frontend capability,
>but there are behavioral differences between frontends that we've considered sniffing somehow in order to have a good experience "out of the box" on Jupyter Lab vs Jupyter Classic vs nteract vs Arcadia, etc.

jonsequitur: Related to that, @colombod is adding an API for grabbing properties like this: #119.

Microsoft.DotNet.Interactive/Commands/SubmitCode.cs
```diff
@@ -16,6 +16,8 @@ public class SubmitCode : KernelCommandBase
             SubmissionType = submissionType;
         }
 
+        public bool AllowStdin { get; set; }
```
jonsequitur: Let's keep this on the context, not the command.

jonsequitur:
An important test will be to make sure that passwords don't appear in diagnostic logs, something like this:

```csharp
var log = new StringBuilder();

using var _ => Pocket.LogEvents.Subscribe(e => log.Append(e.ToLogString()));

// send password input request

log.ToString().Should().NotContain(password);
```

Microsoft.DotNet.Interactive/Commands/SubmitCode.cs
daxian-dbw: It's handled in the context now.

daxian-dbw:
@jonsequitur I will add a test to validate the password scenario. Thanks!

Microsoft.DotNet.Interactive.Jupyter/KernelExtensions.cs
```diff
@@ -35,6 +36,18 @@ public static T UseDefaultMagicCommands<T>(this T kernel)
             return kernel;
         }
 
+        public static CSharpKernel UseJupyterHelpers(
+            this CSharpKernel kernel)
+        {
+            var command = new SubmitCode($@"
```
colombod: this would be safer with 
```
            kernel.DeferCommand(
                new SubmitCode($@"
#r ""{typeof(Kernel).Assembly.Location.Replace("\\", "/")}""
using static {typeof(Kernel).FullName};
"));
```

Microsoft.DotNet.Interactive.Jupyter/JupyterKernelHelper.cs
```diff
@@ -0,0 +1,32 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using Microsoft.DotNet.Interactive.Events;
+
+namespace Microsoft.DotNet.Interactive.Jupyter
+{
+    public static class Kernel
+    {
+        public static string input(
```
colombod: we might need to have a password() gesture so that we can return a PasswordString object and make sure that the formatter for that and ToString() of it gets our encoded or not recorded. In the screenshot you print het content variable and the obfuscation is lost

daxian-dbw: Updated to only cover the non-password input request in this PR. Make `InputRequested` even for non-password input and `PasswordRequested` event for password input. #155 was opened to track the work to add a gesture for get password input.

daxian-dbw:
@colombod Thanks for wiring up the `password` part!
I pushed a commit to update the var names and exception message in the `password` gesture. All rest look good to me!

daxian-dbw:
🎉 🎉 🎉 🎉 🎉 
@colombod @jonsequitur  Thank you for all the help!

mattmcnabb:
Will this feature be included in the released extension any time soon? It appears that although there have been recent preview releases, they still don't include support for prompting via stdin.

jonsequitur:
@mattmcnabb Are you asking about our VS Code extension? This issue refers specifically to the Jupyter integration. 

We have a separate issue for VS Code support: #436.

mattmcnabb:
@jonsequitur Yes 😃 

I'm finding it difficult to distinguish between the two here and in documentation I find on the web. Thanks!

PR #119, opened 2020-02-04
register and expose services on kernel

colombod:


Microsoft.DotNet.Interactive/KernelBase.cs
```diff
@@ -23,6 +23,7 @@ public abstract class KernelBase : IKernel
         private readonly CompositeDisposable _disposables;
         private readonly SubmissionSplitter _submissionSplitter = new SubmissionSplitter();
         private readonly ConcurrentQueue<IKernelCommand> _deferredCommands = new ConcurrentQueue<IKernelCommand>();
+        private readonly ConcurrentDictionary<Type, object> _propertyBag = new ConcurrentDictionary<Type, object>();
```
jonsequitur: Nit: `_properties`

Microsoft.DotNet.Interactive/KernelBase.cs
```diff
@@ -289,5 +292,15 @@ public void RegisterForDisposal(IDisposable disposable)
         public void Dispose() => _disposables.Dispose();
 
         string IKernel.Name => Name;
+
+        public void SetProperty<T>(T service) where T : class
```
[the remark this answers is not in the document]
colombod: I think is better to fail silently on Get but fail on failure to set

Microsoft.DotNet.Interactive/KernelBase.cs
colombod: done

Microsoft.DotNet.Interactive/KernelBase.cs
```diff
@@ -289,5 +292,18 @@ public void RegisterForDisposal(IDisposable disposable)
         public void Dispose() => _disposables.Dispose();
 
         string IKernel.Name => Name;
+
+        public void SetProperty<T>(T property) where T : class
+        {
+            if (!_propertyBag.TryAdd(typeof(T), property))
```
jonsequitur: If you use `Add` you can just let the default exception be thrown.

Microsoft.DotNet.Interactive/KernelBase.cs
```diff
@@ -289,5 +292,18 @@ public void RegisterForDisposal(IDisposable disposable)
...
+            {
+                throw new InvalidOperationException($"Cannot add property with key {typeof(T)}.");
+            }
+        }
+
+        public T GetProperty<T>() where T : class
```
jonsequitur: This should throw if the property is not found.

Microsoft.DotNet.Interactive/KernelBase.cs
colombod: I don’t think there is Add on concurrent dictionary

Microsoft.DotNet.Interactive/KernelBase.cs
colombod: ok, makes sense

Microsoft.DotNet.Interactive/KernelBase.cs
```diff
@@ -187,6 +188,8 @@ private async Task ExecuteCommand(KernelOperation operation)
         private readonly ConcurrentQueue<KernelOperation> _commandQueue =
             new ConcurrentQueue<KernelOperation>();
 
+  
+
```
daxian-dbw: nit: extra new lines that you may want to remove?

Microsoft.DotNet.Interactive/KernelBase.cs
```diff
@@ -289,5 +292,20 @@ public void RegisterForDisposal(IDisposable disposable)
...
+            {
+                throw new InvalidOperationException($"Cannot add property with key {typeof(T)}.");
+            }
+        }
+
+        public T GetProperty<T>() where T : class
+        {
+            return _propertyBag.TryGetValue(typeof(T), out var property) 
+                ? property as T 
+                : throw new InvalidOperationException($"Cannot find property {typeof(T)}.");
```
jonsequitur: Maybe `KeyNotFoundException`?

PR #123, opened 2020-02-05
extension improvements, add ExtensionLab project and samples

jonsequitur:
This adds a laboratory for creating extensions, with a couple of prototype extensions:

* `#!download`, a magic command for downloading files using a URI
* `#!record`, a magic command that records a transcript of incoming commands to a specified file, in order to allow them to be replayed in the actual order they were submitted, which is not reproducible from an `.ipynb` file

The goal is to provide a place for people to iterate on extension ideas that might be useful as separate extension packages, or might eventually mature to be in-the-box features of `dotnet-interactive`, while also exercising the extension APIs.


Microsoft.DotNet.Interactive.ExtensionLab/DownloadExtension.cs
```diff
@@ -0,0 +1,76 @@
...
+    {
+        public Task OnLoadAsync(IKernel kernel)
+        {
+            if (kernel is KernelBase kernelBase)
+            {
+                var download = new Command("#!download")
+                {
+                    new Option<Uri>("--uri"),
+                    new Option<FileInfo>(
+                        "--path",
```
colombod: maybe ```--output```? curl uses ```-o / --output <file>```

Microsoft.DotNet.Interactive.ExtensionLab/Microsoft.DotNet.Interactive.ExtensionLab.csproj
```diff
@@ -0,0 +1,18 @@
...
+  <PropertyGroup>
+    <TargetFramework>netcoreapp3.1</TargetFramework>
+  </PropertyGroup>
+
+ <PropertyGroup>
+    <IsPackable>true</IsPackable>
+    <PackageDescription>Experimental kernel extensions for dotnet-interactive</PackageDescription>
+    <PackageTags>dotnet-interactive</PackageTags>
+  </PropertyGroup>
+
```
colombod: here you need something like
```
  <ItemGroup>
    <None Include="$(OutputPath)\$(AssemblyName).dll" Pack="true" PackagePath="interactive-extensions/dotnet" Visible="false" />
  </ItemGroup>
```

The path ```interactive-extensions/dotnet``` is the location the composite kernel uses for probing

Microsoft.DotNet.Interactive.ExtensionLab/Microsoft.DotNet.Interactive.ExtensionLab.csproj
```diff
@@ -0,0 +1,18 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp3.1</TargetFramework>
```
colombod: here you need this
```
<IncludeBuildOutput>false</IncludeBuildOutput>
```

Microsoft.DotNet.Interactive.ExtensionLab/RecordTranscriptExtension.cs
```diff
@@ -0,0 +1,51 @@
...
+        public Task OnLoadAsync(IKernel kernel)
+        {
+            if (kernel is CompositeKernel kernelBase)
+            {
+                var record = new Command(
+                    "#!record",
+                    "Records a replayable transcript of code submissions.")
+                {
+                    new Option<FileInfo>(
+                        "--path",
```
colombod: same question here is not ```--output``` more clear?

Microsoft.DotNet.Interactive.ExtensionLab/DownloadExtension.cs
```diff
@@ -0,0 +1,76 @@
...
+
+        private static readonly HttpClient _httpClient = new HttpClient();
+
+        private async Task Download(Uri uri, FileInfo path, KernelInvocationContext context)
+        {
+            var response = await _httpClient.GetAsync(uri);
+            response.EnsureSuccessStatusCode();
+
+            await using var responseStream = await response.Content.ReadAsStreamAsync();
+            await using var fileStream = File.Create(path.FullName);
```
eerhardt: (nit) will this work if the path is for a directory that doesn't exist yet?

Microsoft.DotNet.Interactive.ExtensionLab/DownloadExtension.cs
```diff
@@ -0,0 +1,76 @@
...
+
+            responseStream.Seek(0, SeekOrigin.Begin);
```
eerhardt: Why seek it to the beginning? It should already be in the right place (and potentially some streams may not be seekable).

eerhardt:
> using System;

(nit) copyright

---
Refers to: Microsoft.DotNet.Interactive.ExtensionLab/RecordTranscriptExtension.cs:1 in b5548e4. [](commit_id = b5548e4d82062a10c60e48e0bd61fecb9d103009, deletion_comment = False)

jonsequitur:
I'm working out the nuget packaging details now. Previously, our extensions were included alongside a library that the user might want to write code against. These are pure extensions, which we haven't build before.

PR #140, opened 2020-02-07
create integration test leg

brettfo:
Creates a separate set of unit tests in `dotnet-interactive.IntegrationTests` that assume the global tool `dotnet interactive` has already been installed.  This will give us much better assurance that the tool package installs and executes as expected.

The new test project has no references to other projects and that is by design to ensure that the machine isn't tainted by building product code and inadvertently satisfying some runtime requirement that won't be present on end user machines.

Extending a request to @colombod and @jonsequitur to try their local workflows to make sure I don't break anything.

### DIFF
--- a/dotnet-interactive.IntegrationTests/DirectoryTestHelper.cs
+++ b/dotnet-interactive.IntegrationTests/DirectoryTestHelper.cs
@@ -3,7 +3,7 @@ using System.IO;
 using System.Linq;
 using System.Threading.Tasks;
 
-namespace Microsoft.DotNet.Interactive.App.Tests
+namespace Microsoft.DotNet.Interactive.App.IntegrationTests
 {
     public static class DirectoryTestHelper
     {
--- a/dotnet-interactive.IntegrationTests/IntegrationTestAttribute.cs
+++ b/dotnet-interactive.IntegrationTests/IntegrationTestAttribute.cs
@@ -1,0 +1,22 @@
+﻿using System;
+using Xunit;
+
+namespace Microsoft.DotNet.Interactive.App.IntegrationTests
+{
+    /// <summary>
+    /// Signifies that a unit test should only be run during the integration test leg and not as part of a normal test
+    /// run or during the regular dev loop.
+    /// 
+    /// To run locally, [<see cref="IntegrationTestAttribute"/>] will need to be replaced with [<see cref="FactAttribute"/>].
+    /// </summary>
+    public class IntegrationTestAttribute : FactAttribute
+    {
+        public IntegrationTestAttribute()
+        {
+            if (string.IsNullOrEmpty(Environment.GetEnvironmentVariable("INTEGRATION_TEST_RUN")))
+            {
+                Skip = "Only run as an integration test";
+            }
+        }
+    }
+}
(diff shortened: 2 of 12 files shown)


brettfo:
Integration test passed and I've confirmed that the generated `.trx` log was appropriately reported to the parent test run.

dotnet-interactive.IntegrationTests/LoggingTests.cs
```diff
@@ -0,0 +1,62 @@
...
+using FluentAssertions;
+using Microsoft.DotNet.Interactive.App.IntegrationTests.Utility;
+using Microsoft.DotNet.Interactive.Utility;
+using Xunit;
+
+namespace Microsoft.DotNet.Interactive.App.IntegrationTests
+{
+    public class LoggingTests
+    {
+        [Fact]
```
jonsequitur: So that these won't run as part of the normal dev loop, let's use a derived `FactAttribute` that skips unless, e.g., it's a release build or it detects that we're running in a CI build.

brettfo: Added a commit with exactly this.  I confirmed locally that the VS Test Explorer excluded this test.  I also added a doc comment explaining that to run it locally the `[IntegrationTest]` attribute will need to be replaced with `[Fact]`.

I'll confirm that the test still runs during the integration leg and that the `.trx` file was appropriately added to the test run.

dotnet-interactive.IntegrationTests/DirectoryTestHelper.cs
```diff
@@ -3,7 +3,7 @@
 using System.Linq;
 using System.Threading.Tasks;
 
-namespace Microsoft.DotNet.Interactive.App.Tests
+namespace Microsoft.DotNet.Interactive.App.IntegrationTests
```
jonsequitur: Do we also want to move these into an equivalent folder?

dotnet-interactive.IntegrationTests/IntegrationTestAttribute.cs
```diff
@@ -0,0 +1,22 @@
...
+
+namespace Microsoft.DotNet.Interactive.App.IntegrationTests
+{
+    /// <summary>
+    /// Signifies that a unit test should only be run during the integration test leg and not as part of a normal test
+    /// run or during the regular dev loop.
+    /// 
+    /// To run locally, [<see cref="IntegrationTestAttribute"/>] will need to be replaced with [<see cref="FactAttribute"/>].
+    /// </summary>
+    public class IntegrationTestAttribute : FactAttribute
```
jonsequitur: Maybe `IntegrationFactAttribute`?

dotnet-interactive.IntegrationTests/DirectoryTestHelper.cs
brettfo: Done.

dotnet-interactive.IntegrationTests/IntegrationTestAttribute.cs
brettfo: Done.

PR #150, opened 2020-02-10
Emptypackages

KevinRansom:
So ...

we can't detect package roots for packages that don't participate in the build.  A nuget package that has no dll's to reference is a content assembly.  Content assemblies do participate in the build.  So if no assemblies are referenced it must provide content.

E.g ...
````
<package xmlns="http://schemas.microsoft.com/packaging/2012/06/nuspec.xsd">
  <metadata>
    <id>EmptyPackage</id>
    <version>1.0.0</version>
   ...
  </metadata>
  <files>
    <file src="TextFile1.txt" target="contentFiles/any/any/TextFile1.txt" />
  </files>
</package>
````


jonsequitur:
We were hoping to provide a package authoring experience that doesn't force people to use a nuspec.

PR #152, opened 2020-02-10
add frontedEnvironment to InvocationContext

colombod:


Microsoft.DotNet.Interactive/AutomationEnvironment.cs
```diff
@@ -0,0 +1,11 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+
+namespace Microsoft.DotNet.Interactive
+{
+    public class AutomationEnvironment : FrontendEnvironmentBase
```
jonsequitur: Are we using this for anything?

Microsoft.DotNet.Interactive/KernelInvocationContext.cs
```diff
@@ -138,5 +139,11 @@ public ValueTask DisposeAsync()
             // This method is not async because it would prevent the setting of _current.Value to null from flowing up to the caller.
             return new ValueTask(Task.CompletedTask);
         }
+
+        public FrontendEnvironmentBase FrontendEnvironment
+        {
+            get => _frontendEnvironment?? new AutomationEnvironment();
```
jonsequitur: I don't think `AutomationEnvironment` should be the default.

PR #154, opened 2020-02-10
[WIP] build/test with DisableArcade=1

brettfo:
Add a CI leg that runs build/test with the environment variable `DisableArcade=1` to mimic some local environments.

### DIFF
--- a/Directory.Build.targets
+++ b/Directory.Build.targets
@@ -11,6 +11,9 @@
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="16.2.0" />
     <PackageReference Include="xunit" Version="2.4.1" />
     <PackageReference Include="xunit.runner.visualstudio" Version="2.4.1" />
+    <None Include="$(MSBuildThisFileDirectory)xunit.runner.json"
+          CopyToOutputDirectory="PreserveNewest"
+          Visible="false" />
   </ItemGroup>
 
   <!-- Consolidate FSharp package versions -->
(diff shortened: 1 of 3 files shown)


jonsequitur:
I'm not sure we should delete the empty test project. I expect there's going to be test coverage specific to it over time, as with the C# kernel.

brettfo:
The test runner is hanging and it was happening on that empty project, so it was an experiment to see if that helped.  It didn't.

Microsoft.DotNet.Interactive.CSharp/ScriptExecutionExtensions.cs
```diff
@@ -25,7 +25,8 @@ internal static class ScriptExecutionExtensions
                                     return (ScriptState<object>) null;
                                 }, cancellationToken));
 
-            return completed.Result;
+            var result = await completed;
```
jonsequitur: This shouldn't be necessary because the task has already been awaited once.

Were you seeing an issue?

brettfo: I'm exploring to try to figure out why `dotnet test` is hanging in CI and one thread suggested not fully-awaited tasks, but that doesn't seem to be the case.  I'll keep looking.

brettfo:
@jonsequitur I got the `DisableArcade` leg to be green; resetting it again just to check.  The fix was to make xunit run in serial (whole bunch of threads about this, all pointing to the same solution.)  What are your thoughts on this solution?

brettfo:
Not currently a priority.

PR #162, opened 2020-02-12
Replace RestoreSources with #i

jonsequitur:
This replaces `#r "nuget:RestoreSources=/path/to/nuget/source"` with `#i "nuget:/path/to/nuget/source"` for the C# kernel. It still supports both URLs and directory paths.

The goal is to make this gesture work eventually in both C# script and FSI but we can pre-polyfill this for now in .NET Interactive's middleware to try it out and work out the details.

### DIFF
--- a/Microsoft.DotNet.Interactive.CSharp/CSharpKernelExtensions.cs
+++ b/Microsoft.DotNet.Interactive.CSharp/CSharpKernelExtensions.cs
@@ -63,15 +63,20 @@ using static {typeof(Kernel).FullName};
 
         private static Command i(PackageRestoreContext restoreContext)
         {
-            return new Command("#i")
+            var iDirective = new Command("#i")
             {
-
+                new Argument<string>("source")
             };
+            iDirective.Handler = CommandHandler.Create<string,KernelInvocationContext>((source,context) =>
+            {
+                restoreContext.AddRestoreSource(source);
+            });
+            return iDirective;
         }
 
         private static Command r(PackageRestoreContext restoreContext)
         {
-            var poundR = new Command("#r")
+            var rDirective = new Command("#r")
             {
                 new Argument<PackageReference>(
                     result =>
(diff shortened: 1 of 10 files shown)


Microsoft.DotNet.Interactive.CSharp/CSharpKernelExtensions.cs
```diff
@@ -84,8 +59,51 @@ public static CSharpKernel UseNugetDirective(this CSharpKernel kernel)
...
+
+        private static Command i(PackageRestoreContext restoreContext)
+        {
+            var iDirective = new Command("#i")
+            {
+                new Argument<string>("source")
+            };
+            iDirective.Handler = CommandHandler.Create<string, KernelInvocationContext>((source, context) =>
+            {
+                restoreContext.AddRestoreSource(source.Replace("nuget:", ""));
```
colombod: I think here we call PackageRestoreContext a NuGetPacakgeRestoreContext. if we want to be pluggable we should rename this type to express that we are talking about Nuget sources and NuGet packages.

Microsoft.DotNet.Interactive/Events/DisplayedValue.cs
```diff
@@ -2,18 +2,17 @@
...
 {
-    public class DisplayedValue 
+    public class DisplayedValue
     {
         private readonly string _displayId;
         private readonly string _mimeType;
+        private readonly KernelInvocationContext _context;
 
-        public DisplayedValue(string displayId, string mimeType)
+        public DisplayedValue(string displayId, string mimeType, KernelInvocationContext context)
```
colombod: how will this impact the serialization of events?

jonsequitur: This type is not serialized.

Microsoft.DotNet.Interactive.CSharp/CSharpKernelExtensions.cs
jonsequitur: I don't think we know what generalization looks like until we try to build a second implementation so it's fine for "package" to mean "NuGet package" for now. Much of the code in `PackageRestoreContext` would be the same for other package managers.

PR #220, opened 2020-03-02
format string as text/plain by default

jonsequitur:


daxian-dbw:
This change will result in another conflict to #202, which I will resolve after this PR gets merged.

jonsequitur:
@daxian-dbw Does this imply a missing test case?

daxian-dbw:
@jonsequitur Not sure I understand what you mean. The conflict comes from changing of the test name in that PR as well as the input data (`User:` --> `User: `)

PR #231, opened 2020-03-04
deferred commands can force output

colombod:


Microsoft.DotNet.Interactive/KernelCommandExtensions.cs
```diff
@@ -11,6 +11,34 @@ namespace Microsoft.DotNet.Interactive
     public static class KernelCommandExtensions
     {
         internal const string TokenKey = "token";
+        internal const string ForcePublishEvents = "forcePublishEvents";
+
+        public static void SetForcePublishEvents(
```
jonsequitur: We should keep this internal for now.

Microsoft.DotNet.Interactive/KernelCommandExtensions.cs
```diff
@@ -11,6 +11,34 @@ namespace Microsoft.DotNet.Interactive
     public static class KernelCommandExtensions
     {
         internal const string TokenKey = "token";
+        internal const string ForcePublishEvents = "forcePublishEvents";
```
jonsequitur: Something like "publish internal command events" might be more accurate.

Microsoft.DotNet.Interactive.Jupyter/RequestHandlerBase.cs
```diff
@@ -46,7 +46,7 @@ protected RequestHandlerBase(IKernel kernel, IScheduler scheduler, JupyterFronte
 
             bool FilterByToken(IKernelEvent e)
             {
-                return e.Command?.GetToken() == context.Token;
+                return (e.Command?.GetToken() == context.Token) || e.Command.GetForcePublishEvents();
```
jonsequitur: The change in logic makes the function name inaccurate.

PR #236, opened 2020-03-04
add stubs for lsp over http

brettfo:
This adds all of the stubs necessary to invoke the `textDocument/hover` LSP function, which is analogous to QuickInfo in Visual Studio.  Currently only the C# kernel responds to this and it's responding with a hard-coded string.  The work to produce the _real_ content is forthcoming, but I needed the plumbing, first.

The LSP work is mostly split between two files.  The first is `lsp.js` which handles the LSP-over-HTTP business.  Not much to look at here.

The next part is specific to CodeMirror and how it's exposed through `jupyter notebook`.  I've limited this PR to that case because it's simpler because any given web page has only one notebook open.  That works by detecting the global JavaScript variables `CodeMirror` and `Jupyter` (not set in `jupyter lab`).  Since CodeMirror doesn't natively understand a text hover operation, I've tied into CodeMirror cell creation and hook into `mousemove` and `mouseleave` to register timer-based callbacks that ultimately result in `textDocument/hover` being called if the mouse is in the same position for 500ms.

CodeMirror also doesn't have the ability to get the line/column of the character under the mouse cursor, so I've had to get creative.  I do that by first iterating over the DOM elements that represent the lines of the editor.  If one of those lines' bounding rectangles contains the mouse cursor, then I start to look for the column.  Finding the column is much harder because the DOM elements don't represent anything useful, so I simply check each one in order, again checking the bounding rectangle.  The end result is that the column that I detect is really the column of the **first** character of an arbitrary text span, but since CodeMirror is splitting the text on reasonable word boundaries, this will always yield the start position of what any programmer would call an identifier.

In the example below, the following markdown is returned as a hard-coded string from the C# LSP provider:

``` markdown
textDocument/hover at position (0, 9) with `markdown`
```

image

@colombod I re-worked the `requirejs.config()` call in `HttpApiBootstrapper.cs` so that the prefix `dotnet-interactive` points to `http://localhost:<port>/resources`.  This makes it so that I can later call the following nicely:

``` javascript
require(['dotnet-interactive/dotnet-interactive'], ...); // returns http://localhost:<port>/resources/dotnet-interactive.js
require(['dotnet-interactive/lsp'], ...); // returns http://localhost:<port>/resources/lsp.js
require(['dotnet-interactive/code-mirror-lsp'], ...); // returns http://localhost:<port>/resources/code-mirror-lsp.js
```


Microsoft.DotNet.Interactive.Tests/Language.cs
```diff
@@ -19,7 +19,7 @@ public static string LanguageName(this Language language)
             {
                 Language.CSharp => "csharp",
                 Language.FSharp => "fsharp",
-                Language.PowerShell => "pwsh",
+                Language.PowerShell => "powershell",
```
colombod: this will alter also magic commands in notebooks samples

Microsoft.DotNet.Interactive/KernelBase.cs
```diff
@@ -167,6 +168,11 @@ command switch
 
         public abstract bool TryGetVariable(string name, out object value);
 
+        public virtual Task<JObject> LspMethod(string methodName, JObject request)
```
colombod: are we expecting the api to be opt in or otp out? @jonsequitur should this be abstract?

dotnet-interactive/HttpApiBootstrapper.cs
```diff
@@ -10,36 +10,38 @@ internal static class HttpApiBootstrapper
     {
         public static string GetJSCode(Uri apiRoot)
         {
-            var id = $"{Process.GetCurrentProcess().Id}.{apiRoot.Port}";
             var template = @"#!javascript
-if ((typeof(requirejs) !==  typeof(Function)) || (typeof(requirejs.config) !== typeof(Function))) { 
-    let script = document.createElement(""script""); 
-    script.setAttribute(""src"", ""https://cdnjs.cloudflare.com/ajax/libs/require.js/2.3.6/require.min.js""); 
+if ((typeof(require) !==  typeof(Function)) || (typeof(require.config) !== typeof(Function))) {
```
colombod: this will impact jupyter lab experience

dotnet-interactive/HttpApiBootstrapper.cs
```diff
@@ -10,36 +10,38 @@ internal static class HttpApiBootstrapper
...
     };
-    document.getElementsByTagName(""head"")[0].appendChild(script); 
+    document.getElementsByTagName('head')[0].appendChild(script);
 }
 else {
     loadDotnetInteractiveApi();
 }
 
 function loadDotnetInteractiveApi(){
-    let apiRequire = requirejs.config({context:""dotnet-interactive.$SEED$"",paths:{dotnetInteractive:""$API_URL$""}});
```
colombod: restarting kernel will change port, this seed into the name is to prevent cache and avoid restarting the browser tag completely and loosing the work.

colombod: double check since juoyter lab has require but not require js and require.js ints require (if not dfined) and requirejs. in the original code requirejs was the safe api to invoke.

dotnet-interactive/resources/code-mirror-lsp.js
```diff
@@ -0,0 +1,131 @@
+﻿define({
+    init: function (global) {
+        // used to display textDocument/hover
+        var hoverDiv = document.createElement('div');
```
colombod: I think document should be passed in or looked up onto global parameter, this code could be hard to test without always running inside a browser.

dotnet-interactive/resources/dotnet-interactive.js
```diff
@@ -47,16 +49,24 @@
...
                     }
                 });
-        }
+        };
 
         global.interactive = client;
 
         client.loadKernels();
 
+        require([
```
colombod: this can be done by setting the dependencies in the define above and then `lsp,cm` variable in the function defined.
Something like
```js
define(["dotnet-interactive/lsp", "dotnet-interactive/code-mirror-lsp"], function(lsp, cm) {
....
}
```

Microsoft.DotNet.Interactive.Tests/Language.cs
brettfo: This is a test-only change, so it won't affect any magic commands.  I found that executing `/variables/pwsh/x` wouldn't find the right kernel, but `/variables/powershell/x` would, so I had to change this because I created it wrong in the beginning.

Microsoft.DotNet.Interactive/KernelBase.cs
brettfo: Ultimately I don't care either way.  I made the method `virtual` simply so F#/pwsh don't have to implement this method simply to return `null` if they don't support LSP.

dotnet-interactive/HttpApiBootstrapper.cs
brettfo: I can go back to `requirejs`.  I was running into issues with `require(['dotnet-interactive/foo'], ...)` resolving with the `context` and `$SEED$` values, but if it's a cache buster then I'll go back to it and try to get it working correctly.

dotnet-interactive/resources/code-mirror-lsp.js
brettfo: Good call, will do.

dotnet-interactive/resources/dotnet-interactive.js
brettfo: I did it down here because I'm treating `dotnet-interactive.js` to be the root object/script and `lsp.js` (currently) requires the `global.interactive` variable to be set and I wasn't sure if it would break by trying to import lsp/code-mirror _before_ dotnet-interactive.js.

colombod: I see, this makes sense!

dotnet-interactive/HttpApiBootstrapper.cs
colombod: I had issues with xplot injecting the right require api in jupyter lab.

brettfo:
Re-did this PR to take in recent changes.  In short:

1. Editor sniffing is now one layer up and it only loads the CodeMirror component if necessary.
2. The element that is detected as the notebook root is decorated in two ways:
  A. The class `dotnet-interactive-notebook-root` is added.
  B. The key/value pair `dotnet-interactive-host`=`http://localhost:<port>/` is added to the node.

  This was done so that the LSP methods have an easy way to look up the tree in the DOM to know what end point to issue the LSP commands to.

3. `apiRequire()` was renamed to `dotnet_require` and added to the root object so it's available wherever necessary.
4. Used `MutationObserver` to detect when new cells are added to a notebook; this is easier than using the built-in `CodeMirror.defineInitHook()` function, because that function is called before the editor is attached to the DOM, so we have no way of knowing if the new editor is part of our notebook or another one.  The `MutationObserver` method solves this.

Microsoft.DotNet.Interactive/KernelBase.cs
jonsequitur: `JObject` isn't an ideal return type. We should use strong types here and let serialization happen higher up.

dotnet-interactive.Tests/KernelHttpApiTests.cs
```diff
@@ -66,23 +81,69 @@ public async Task variable_is_returned_with_application_json_content_type(Langua
...
+        }
+
+        [Theory]
+        [InlineData(Language.FSharp)]
+        [InlineData(Language.PowerShell)]
+        public async Task unimplemented_lsp_handlers_return_empty_string(Language language)
+        {
+            // language kernels that don't implement any LSP handling
+            var response = await GetServer(language).HttpClient.PostJsonAsync($"/lsp/textDocument/hover", new JObject());
+            response.EnsureSuccessStatusCode();
```
jonsequitur: This should not be called in a test because the error won't contain useful information, e.g. why it failed and what was actually returned.

We have a `ShouldSucceed()` extension method in dotnet/try that's much more useful here.

dotnet-interactive/KernelExtensions.cs
```diff
@@ -76,22 +76,13 @@ public static T UseHttpApi<T>(this T kernel, StartupOptions startupOptions)
...
-                        string value =
-                            script[type: "text/javascript"](
-
-                                    scriptContent.ToHtmlContent())
-                                .ToString();
-
+                        var html = HttpApiBootstrapper.GetHtmlInjection(new Uri($"http://localhost:{startupOptions.HttpPort}"));
                         context.Publish(new DisplayedValueProduced(
-                            scriptContent,
+                            null,
```
jonsequitur: This whole call could be replaced by `context.Display()`.

We shouldn't pass null as the value in any case.

dotnet-interactive/resources/code-mirror-lsp.js
```diff
@@ -0,0 +1,154 @@
...
+                        hoverDiv.style.top = `${event.pageY}px`;
+                        hoverDiv.style.left = `${event.pageX}px`;
+                        hoverDiv.style.display = 'inline';
+                        if (result.contents.kind === "markdown") {
+                            require(['components/marked/lib/marked'], function (marked) {
+                                // this should work for jupyter notebook
+                                hoverDiv.innerHTML = marked(result.contents.value);
+                            }, function (_err) {
+                                // couldn't load `marked`; this will happen in jupyter lab
+                                hoverDiv.innerHTML = `<pre>${escapeString(result.contents.value)}</pre>`;
```
jonsequitur: You should do something like this, which will handle the encoding more correctly than the bespoke `escapeString` method:

```js
let pre = document.createElement('pre');
pre.text = result.contents.value;
hoverDiv.appendChild(pre);
```

dotnet-interactive/resources/lsp.js
```diff
@@ -0,0 +1,39 @@
+define(function () {
+
+    function getHostFromElement(element) {
+        let notebookRoot = element.closest('.dotnet-interactive-notebook-root');
+        return notebookRoot.data['dotnet-interactive-host'];
+    }
+
+    return {
+        init: function (global) {
+            var lsp = {};
```
colombod: this as var will be hoisted, use let

dotnet-interactive.Tests/KernelHttpApiTests.cs
brettfo: I think `EnsureSuccessStatusCode()` is sufficient here.  In dotnet/try I found [`ShouldSucceedWithOutput()`](https://github.com/dotnet/try/blob/ffec59d135585039792217ed14da25d7f7858d08/WorkspaceServer.Tests/RunResultExtensions.cs#L24) and all that does is verify that a 200 code was returned.  I temporarily made a test fail to see the output of `EnsureSuccessStatusCode()` and it was

```
System.Net.Http.HttpRequestException : Response status code does not indicate success: 404 (Not Found).
```

Which I think is as expressive as we can hope for here.

Microsoft.DotNet.Interactive/KernelBase.cs
brettfo: I just added a commit titled `strongly type LSP responses` that does just this; take a look.

dotnet-interactive.Tests/KernelHttpApiTests.cs
jonsequitur: In this case, at this point in time, maybe, but it won't be ideal as the code grows. As an API gets more interesting, the return code is rarely very informative, e.g. we'll want to know why a 400 or 500 was returned. The test should pinpoint the problem, not just indicate that there was one.

Microsoft.DotNet.Interactive/LanguageService/MarkupContent.cs
```diff
@@ -0,0 +1,11 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.DotNet.Interactive.LanguageService
+{
+    public class MarkupContent
+    {
+        public MarkupKind Kind { get; set; }
```
jonsequitur: How about a mime type? :smile:

brettfo: I'm trying to stick to the LSP spec from [here](https://microsoft.github.io/language-server-protocol/specification#textDocument_hover).

dotnet-interactive.Tests/KernelHttpApiTests.cs
brettfo: The only thing I can think of is to also output the response body if there was one.  If that's what we'll want I can add that helper.

brettfo: Just added commit `report more helpful errors` that logs the return code as well as the response content (if any).

Microsoft.DotNet.Interactive.CSharp/CSharpKernel.cs
```diff
@@ -24,8 +24,8 @@
 
 namespace Microsoft.DotNet.Interactive.CSharp
 {
-    public class CSharpKernel : 
-        KernelBase, 
+    public partial class CSharpKernel :
```
jonsequitur: Let's avoid partial classes.

dotnet-interactive.Tests/KernelHttpApiTests.cs
```diff
@@ -76,35 +90,83 @@ public async Task variable_is_returned_with_application_json_content_type(Langua
...
+        ""character"": 2
+    }
+}
+");
+            var response = await GetServer(language).HttpClient.PostJsonAsync("/lsp/textDocument/hover", request);
+            await response.ShouldSucceed();
+            var responseJson = await response.Content.ReadAsStringAsync();
+            var json = JObject.Parse(responseJson);
+            var responseType = json["contents"]["kind"].Value<string>();
+            responseType.Should().Be("markdown");
```
jonsequitur: In cases where there are multiple assertions, you can make sure we get feedback on all of them by doing this before beginning the assertions:

```csharp
using var _ = new AssertionScope();
```

dotnet-interactive/HttpApiBootstrapper.cs
```diff
@@ -9,12 +9,17 @@ namespace Microsoft.DotNet.Interactive.App
...
-        public static string GetJSCode(Uri[] probingUris, string seed)
+        public static string GetHtmlInjection(Uri[] probingUris, string seed)
         {
             var apiCacheBuster = $"{Process.GetCurrentProcess().Id}.{seed}";
-            var template = @"// ensure `requirejs` is available
-if ((typeof (requirejs) !== typeof (Function)) || (typeof (requirejs.config) !== typeof (Function))) 
-{
+            var template = @"
+<div>
+    <div id='dotnet-interactive-this-cell-$SEED$' style='display: none'>
```
colombod: this value is stored in the notebook as output, when a saved notebook loads this will be evaluated, then on the first run  a new div with new id will be added to the dom.  what will happen?

brettfo: Once this cell is re-evaluated, it's output will be overwritten.  Should be harmless.

Microsoft.DotNet.Interactive.CSharp/CSharpKernel.cs
```diff
@@ -38,6 +42,12 @@ public class CSharpKernel :
 
         private WorkspaceFixture _fixture;
 
+        private readonly JsonSerializer _jsonSerializer = JsonSerializer.CreateDefault(new JsonSerializerSettings()
+        {
+            ContractResolver = new CamelCasePropertyNamesContractResolver(),
```
colombod: is this convention going to be shared with all kernels doing lsp?

brettfo: Probably, but for now only `CSharpKernel.cs` does anything with LSP.  I'll move this up a level.

brettfo: Done.

PR #251, opened 2020-03-06
allow multiple http endpoints in a single document DOM

brettfo:
Previously a global variable named `interactive` was created to allow calls such as `interactive.csharp.getVariable('x')`, but in Jupyter Lab where potentially many notebooks might be open at the same time, there would be a collision where one notebook would be getting another's http endpoint.

In the old world, the following transformation took place:

```
#!javascript
alert('x is ' + interactive.csharp.getVariable('x'));
```
V
```
<script type="text/javascript">
alert('x is ' + interactive.csharp.getVariable('x'));
</script>
```

But now the result is this:

```
<script type="text/javascript">
createDotnetInteractiveClient('http://localhost:1234').then((interactive) => {
alert('x is ' + interactive.csharp.getVariable('x'));
});
</script>
```

The result is that the submitted code has its own private `interactive` variable that points to the appropriate http endpoint.  I've verified the behavior by having two different notebooks open in Jupyter Lab and after setting `var x = 1;` in tab 1 and `var x = 2;` in tab 2, the calls to `interactive.csharp.getVariable('x')` always returned the value corresponding to the given notebook/kernel.

Note for @colombod:
We spoke about prefixing the code with a simple `let interactive = createDotnetInteractiveClient(...)`, but that could lead to multiple `let interactive = ...` lines in the HTML document which is invalid.  Similarly we didn't want to use `var interactive = ...` because that essentially creates a global variable.  The solution was to wrap everything in a function so the given `interactive` variable is unique to that code submission.


dotnet-interactive/HttpApiBootstrapper.cs
```diff
@@ -12,34 +12,50 @@ public static string GetJSCode(Uri apiRoot)
...
-    apiRequire(['dotnetInteractive'], 
-    function(api) {       
-        api.createDotnetInteractiveClient(""$HOST$"", window);
-    },
-    function(error){
-        console.log(error);
-    });
+function loadDotnetInteractiveApi() {
+    // ensure interactive helpers are loaded
+    let interactiveHelperElement = document.getElementById('dotnet-interactive-script-loaded');
```
brettfo: This sentinel element is just so that we don't create the global function `createDotnetInteractiveClient` repeatedly.

brettfo:
@jonsequitur I just saw your changes to have separate HTML/JS kernels.  Once this is reviewed I'll rebase and move the script rewriting to `JavascriptKernel.js` as appropriate; I'm too lazy right now.

dotnet-interactive/resources/dotnet-interactive.js
```diff
@@ -1,62 +1,58 @@
 ﻿define({
```
brettfo: I recommend you diff this file with whitespace disabled.  The shape was changed so that everything is essentially indented one level deeper.

Microsoft.DotNet.Interactive/JavaScriptKernel.cs
```diff
@@ -12,8 +12,11 @@ public class JavaScriptKernel : KernelBase
     {
         public const string DefaultKernelName = "javascript";
 
-        public JavaScriptKernel() : base(DefaultKernelName)
+        private string _hostAddress;
+
+        public JavaScriptKernel(string hostAddress) : base(DefaultKernelName)
```
jonsequitur: I think the address should be moved to the `FrontEndEnvironment`. It shouldn't be the responsibility of someone creating the kernel to understand details of the environment like this, and the wrapper script(s) will likely need to vary in other ways by frontend anyway.

brettfo: What are your thoughts on threading the value through?  `JavaScriptKernel.cs` is in the `Microsoft.DotNet.Interactive` project, but `JupyterFrontendEnvironment` (which seems like the logical place for this) is one level up in `Microsoft.DotNet.Interactive.Jupyter`.  The other option is to add `HttpEndpoint` to `FrontendEnvironmentBase`, but that seems heavy-handed to force all front ends to support the existence of an http endpoint.

jonsequitur: The point of the `FrontEndEnvironment` is to vary by capabilities, so it could be added to the `JupyterFrontEndEnvironment` but not to `FrontendEnvironmentBase`.

I don't think the `Microsoft.DotNet.Interactive` layer should be aware of the HTTP APIs.

colombod: The port is the port of dotnet interactive, it is not related to the frontend. Not so sure the value should be there. If it does then is part of the base class not just jupyter

jonsequitur: The purpose of the `FrontEndEnvironment` (though it has no significant behavior currently) was to allow the kernel to detect front-end capabilities and vary output accordingly, e.g. script loading capabilities. This is very much in line with that goal.

brettfo: My point above is that we can't add it to `JupyterFrontEndEnvironment` because of this:

``` C#
// JavaScriptKernel.cs in top-level project `Microsoft.DotNet.Interactive`
if (context.FrontendEnvironment is JupyterFrontendEnvironment j) // <- this doesn't work because `JupyterFrontendEnvironment is defined (appropriately) in `Microsoft.DotNet.Interactive.Jupyter`
{
    // ...
}
```

We either have to:

1. Move `JavaScriptKernel.cs` to `Microsoft.DotNet.Interactive.Jupyter` so that having a reference to `JupyterFrontendEnvironment` makes sense, or
2. Add the host (e.g., `http://localhost:1234`) to `FrontenvEnvironmentBase`, or
3. Add `class FooFrontEndEnvironment : FrontendEnvironmentBase { string Host {get; set;} }` to `Microsoft.DotNet.Interactive`.

Option (1) possibly makes sense because we don't actually have a JavaScript kernel; it's dependent on the host understanding `<script>` tags.  The case where this doesn't make sense is when we have a browser-based front end that's not Jupyter (e.g., Azure notebooks).  Options (2) or (3) I think I like best because it's entirely conceivable for a command-line interactive environment to have http language services.

I'm leaning towards option 3.

PR #268, opened 2020-03-12
discover ip that can be accessed from other machine

colombod:


Microsoft.DotNet.Interactive.Jupyter/JupyterFrontendEnvironment.cs
```diff
@@ -0,0 +1,14 @@
...
+
+
+using System;
+
+namespace Microsoft.DotNet.Interactive.Jupyter
+{
+    public class JupyterFrontendEnvironment : FrontendEnvironmentBase
+    {
+        public bool AllowStandardInput { get; set; }
+        public Uri Host { get; set; }
```
jonsequitur: Do these setters need to exist or be public?

PR #285, opened 2020-03-17
return proper hover info for C#

brettfo:
Return proper C# QuickInfo for the LSP method `textDocument/hover` to remove the hard-coded placeholder.

The `hover` method requires a document URI, but in notebooks, there's no meaningful definition of 'document', so I utilized the URI field to data-encode the cell contents.  Given that there are no execution order guarantees in notebooks, this seemed the best way to ensure that `textDocument/hover` always did something meaningful.

The Roslyn LSP implementation is written as a `net472` console app that communicates over stdio, so it couldn't be repurposed for this, but the code was small so I copied where appropriate from [here](https://github.com/dotnet/roslyn/blob/e0a1787d544c8e7c5fef2007dbc8ef6592d8c028/src/Features/LanguageServer/Protocol/Handler/Hover/HoverHandler.cs).

image

### DIFF
--- a/Microsoft.DotNet.Interactive.CSharp/CSharpKernel.cs
+++ b/Microsoft.DotNet.Interactive.CSharp/CSharpKernel.cs
@@ -137,20 +137,10 @@ namespace Microsoft.DotNet.Interactive.CSharp
                 return null;
             }
 
-            var markdown = info.ToMarkdownString();
-            if (string.IsNullOrWhiteSpace(markdown))
-            {
-                return null;
-            }
-
             var linePosSpan = text.Lines.GetLinePositionSpan(info.Span);
             return new TextDocumentHoverResponse()
             {
-                Contents = new MarkupContent()
-                {
-                    Kind = MarkupKind.Markdown,
-                    Value = markdown,
-                },
+                Contents = info.ToMarkupContent(),
                 Range = linePosSpan.ToLspRange(),
             };
         }
--- a/Microsoft.DotNet.Interactive.CSharp/CSharpKernel.cs
+++ b/Microsoft.DotNet.Interactive.CSharp/CSharpKernel.cs
@@ -12,8 +12,10 @@ using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.Completion;
 using Microsoft.CodeAnalysis.CSharp;
 using Microsoft.CodeAnalysis.CSharp.Scripting;
+using Microsoft.CodeAnalysis.QuickInfo;
 using Microsoft.CodeAnalysis.Recommendations;
 using Microsoft.CodeAnalysis.Scripting;
+using Microsoft.CodeAnalysis.Text;
 using Microsoft.DotNet.Interactive.Commands;
 using Microsoft.DotNet.Interactive.Events;
 using Microsoft.DotNet.Interactive.Extensions;
@@ -102,7 +104,7 @@ namespace Microsoft.DotNet.Interactive.CSharp
             return false;
         }
 
-        public override Task<LspResponse> LspMethod(string methodName, JObject request)
+        public override async Task<LspResponse> LspMethod(string methodName, JObject request)
         {
             LspResponse result;
             switch (methodName)
@@ -110,25 +112,46 @@ namespace Microsoft.DotNet.Interactive.CSharp
                 case "textDocument/hover":
                     // https://microsoft.github.io/language-server-protocol/specification#textDocument_hover
                     var hoverParams = request.ToLspObject<HoverParams>();
-                    result = TextDocumentHover(hoverParams);
+                    result = await TextDocumentHover(hoverParams);
                     break;
                 default:
                     result = null;
                     break;
             }
 
-            return Task.FromResult(result);
+            return result;
         }
 
-        public TextDocumentHoverResponse TextDocumentHover(HoverParams hoverParams)
+        public async Task<TextDocumentHoverResponse> TextDocumentHover(HoverParams hoverParams)
         {
+            var documentContents = hoverParams.TextDocument.Uri.DecodeDocumentFromDataUri();
+            var document = await GetDocumentWithCodeAsync(documentContents);
+
+            var text = await document.GetTextAsync();
+            var position = text.Lines.GetPosition(new LinePosition(hoverParams.Position.Line, hoverParams.Position.Character));
+
+            var service = QuickInfoService.GetService(document);
+            var info = await service.GetQuickInfoAsync(document, position);
+            if (info == null)
+            {
+                return null;
+            }
+
+            var markdown = info.ToMarkdownString();
+            if (string.IsNullOrWhiteSpace(markdown))
+            {
+                return null;
+            }
+
+            var linePosSpan = text.Lines.GetLinePositionSpan(info.Span);
             return new TextDocumentHoverResponse()
             {
                 Contents = new MarkupContent()
                 {
                     Kind = MarkupKind.Markdown,
-                    Value = $"textDocument/hover at position ({hoverParams.Position.Line}, {hoverParams.Position.Character}) with `markdown`",
+                    Value = markdown,
                 },
+                Range = linePosSpan.ToLspRange(),
             };
         }
 
@@ -253,27 +276,12 @@ namespace Microsoft.DotNet.Interactive.CSharp
             string code,
             int cursorPosition)
         {
-            var compilation = ScriptState.Script.GetCompilation();
-            var originalCode =
-                ScriptState?.Script.Code ?? string.Empty;
-
-            var buffer = new StringBuilder(originalCode);
-            if (!string.IsNullOrWhiteSpace(originalCode) && !originalCode.EndsWith(Environment.NewLine))
-            {
-                buffer.AppendLine();
-            }
-
-            buffer.AppendLine(code);
-            var fullScriptCode = buffer.ToString();
+            var document = await GetDocumentWithCodeAsync(code);
+            var text = await document.GetTextAsync();
+            var fullScriptCode = text.ToString();
             var offset = fullScriptCode.LastIndexOf(code, StringComparison.InvariantCulture);
             var absolutePosition = Math.Max(offset, 0) + cursorPosition;
 
-            if (_fixture == null || ShouldRebuild())
-            {
-                _fixture = new WorkspaceFixture(compilation.Options, compilation.References);
-            }
-
-            var document = _fixture.ForkDocument(fullScriptCode);
             var service = CompletionService.GetService(document);
 
             var completionList = await service.GetCompletionsAsync(document, absolutePosition);
@@ -292,11 +300,6 @@ namespace Microsoft.DotNet.Interactive.CSharp
             var items = completionList.Items.Select(item => item.ToModel(symbolToSymbolKey, document)).ToArray();
 
             return items;
-
-            bool ShouldRebuild()
-            {
-                return compilation.References.Count() != _fixture.MetadataReferences.Count();
-            }
         }
 
         public async Task LoadExtensionsFromDirectoryAsync(
@@ -309,12 +312,43 @@ namespace Microsoft.DotNet.Interactive.CSharp
                 context);
         }
 
+        private async Task<Document> GetDocumentWithCodeAsync(string code)
+        {
+            if (ScriptState == null)
+            {
+                ScriptState = await CSharpScript.RunAsync(string.Empty, ScriptOptions);
+            }
+
+            var compilation = ScriptState.Script.GetCompilation();
+            var originalCode = ScriptState.Script.Code ?? string.Empty;
+
+            var buffer = new StringBuilder(originalCode);
+            if (!string.IsNullOrWhiteSpace(originalCode) && !originalCode.EndsWith(Environment.NewLine))
+            {
+                buffer.AppendLine();
+            }
+
+            buffer.AppendLine(code);
+            var fullScriptCode = buffer.ToString();
+
+            if (_fixture == null || ShouldRebuild())
+            {
+                _fixture = new WorkspaceFixture(compilation.Options, compilation.References);
+            }
+
+            var document = _fixture.ForkDocument(fullScriptCode);
+            return document;
+
+            bool ShouldRebuild()
+            {
+                return compilation.References.Count() != _fixture.MetadataReferences.Count();
+            }
+        }
+
         private bool HasReturnValue =>
             ScriptState != null &&
             (bool)_hasReturnValueMethod.Invoke(ScriptState.Script, null);
 
         internal NativeAssemblyLoadHelper NativeAssemblyLoadHelper { get; }
     }
-
-
 }
--- a/Microsoft.DotNet.Interactive.CSharp/CSharpKernel.cs
+++ b/Microsoft.DotNet.Interactive.CSharp/CSharpKernel.cs
@@ -124,7 +124,11 @@ namespace Microsoft.DotNet.Interactive.CSharp
 
         public async Task<TextDocumentHoverResponse> TextDocumentHover(HoverParams hoverParams)
         {
-            var documentContents = hoverParams.TextDocument.Uri.DecodeDocumentFromDataUri();
+            if (!hoverParams.TextDocument.Uri.TryDecodeDocumentFromDataUri(out var documentContents))
+            {
+                return null;
+            }
+
             var document = await GetDocumentWithCodeAsync(documentContents);
 
             var text = await document.GetTextAsync();
--- a/Microsoft.DotNet.Interactive.CSharp/CSharpKernel.cs
+++ b/Microsoft.DotNet.Interactive.CSharp/CSharpKernel.cs
@@ -111,23 +111,25 @@ namespace Microsoft.DotNet.Interactive.CSharp
                 return null;
             }
 
-            var document = await GetDocumentWithCodeAsync(documentContents);
-
+            var (document, offset) = GetDocumentWithOffsetFromCode(documentContents);
             var text = await document.GetTextAsync();
-            var position = text.Lines.GetPosition(new LinePosition(hoverParams.Position.Line, hoverParams.Position.Character));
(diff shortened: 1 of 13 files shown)


Microsoft.DotNet.Interactive.CSharp/CSharpKernel.cs
```diff
@@ -102,33 +104,54 @@ public Task<bool> IsCompleteSubmissionAsync(string code)
             return false;
         }
 
-        public override Task<LspResponse> LspMethod(string methodName, JObject request)
+        public override async Task<LspResponse> LspMethod(string methodName, JObject request)
```
jonsequitur: Can we use a more specific, made-for-purpose type for the request?

brettfo: We can't because each request is a JSON object that's unique to each request type, so there's no deserialization we can do ahead of time that's meaningful.

Microsoft.DotNet.Interactive.CSharp/LanguageServiceExtensions.cs
```diff
@@ -0,0 +1,50 @@
...
+        public static Range ToLspRange(this LinePositionSpan linePosSpan)
+        {
+            return new Range()
+            {
+                Start = linePosSpan.Start.ToLspPosition(),
+                End = linePosSpan.End.ToLspPosition(),
+            };
+        }
+
+        public static string ToMarkdownString(this QuickInfoItem info)
```
jonsequitur: How about introducing a return type that indicates that this is in fact a Markdown string?

Microsoft.DotNet.Interactive.CSharp/CSharpKernel.cs
brettfo: The other alternative would be to strongly-type methods and their requests, but that would force each kernel to implement something like this:

``` C#
override TextDocumentHoverResponse Hover(HoverParams hoverParams) { }
override TextDocumentCompletionResponse Completion(CompletionParams completion) { }
override TextDocumentSignatureHelpResponse SignatureHelp(SignatureHelpParams sigHelp() { }
// ... there are a ridiculous amount of methods that we'd have to add/override
```

The other alternative I considered is adding `ILspKernel` with the above methods using default interface methods to simply return `null`.

Microsoft.DotNet.Interactive/LanguageService/LanguageServiceExtensions.cs
```diff
@@ -28,5 +30,39 @@ public static T ToLspObject<T>(this JObject obj)
         {
             return obj.ToObject<T>(_jsonSerializer);
         }
+
+        private const string DataUriPrefix = "data:";
+
+        public static string DecodeDocumentFromDataUri(this string dataUri)
```
jonsequitur: I'd recommend introducing types here to encapsulate validation and parsing logic.

Microsoft.DotNet.Interactive.CSharp/LanguageServiceExtensions.cs
brettfo: I'll switch it to return `MarkupContent` which specifies its type as `markdown` or `plaintext`.

Microsoft.DotNet.Interactive.CSharp/CSharpKernel.cs
jonsequitur: It's generally a good idea to keep this layer ignorant of serialization. (The `Newtonsoft.Json` reference exists only for the sake of `JsonFormatter` and `Microsoft.DotNet.Interactive.Server`, not for core functionality.)

So the latter approach will be more flexible over time. I would go even further, in fact, and decompose into one interface per method, because they won't all make sense for all kernels. Something like this?

```csharp
public interface ISupportHoverText
{
    TextDocumentHoverResponse Hover(HoverParams hoverParams);
}
public interface ISupportCompletion
{
    TextDocumentCompletionResponse Completion(CompletionParams completion);
}
public interface ISupportSignatureHelp
{
    TextDocumentSignatureHelpResponse SignatureHelp(SignatureHelpParams sigHelp();
}
```

brettfo: (but of course F# can't implement an interface with default methods, yet)

Microsoft.DotNet.Interactive/LanguageService/LanguageServiceExtensions.cs
brettfo: Done.

Microsoft.DotNet.Interactive.CSharp/CSharpKernel.cs
brettfo: Done.

Microsoft.DotNet.Interactive.CSharp/CSharpKernel.cs
```diff
@@ -309,12 +312,43 @@ bool ShouldRebuild()
                 context);
         }
 
+        private async Task<Document> GetDocumentWithCodeAsync(string code)
+        {
+            if (ScriptState == null)
+            {
+                ScriptState = await CSharpScript.RunAsync(string.Empty, ScriptOptions);
```
jonsequitur: Let's move this lazy initialization into a single location rather than let it be repeated. Ideally, an external caller should never see a null `ScriptState`. There's a skipped test for this somewhere.

Microsoft.DotNet.Interactive/LanguageService/LanguageServiceExtensions.cs
```diff
@@ -28,5 +30,39 @@ public static T ToLspObject<T>(this JObject obj)
...
+        {
+            // from https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/Data_URIs, basic format is
+            // data:[<mime_type>][;base64],<data>
+            if (!dataUri.StartsWith(DataUriPrefix))
+            {
+                throw new FormatException($"Expected string to start with '{DataUriPrefix}'");
```
jonsequitur: This should be an `ArgumentException`.

Microsoft.DotNet.Interactive/LanguageService/LanguageServiceExtensions.cs
```diff
@@ -28,5 +30,39 @@ public static T ToLspObject<T>(this JObject obj)
...
+            }
+
+            var headerEnd = dataUri.IndexOf(',');
+            if (headerEnd < 0)
+            {
+                throw new FormatException("Expected ',' before data.");
```
jonsequitur: Ditto.

Microsoft.DotNet.Interactive/LanguageService/LanguageServiceExtensions.cs
brettfo: Resolved when I renamed the method to `Try...`.

Microsoft.DotNet.Interactive.CSharp/CSharpKernel.cs
brettfo: The best I can think of is to turn this:

``` C#
public ScriptState ScriptState { get; private set; }
```

into this:

``` C#
private ScriptState _scriptState;
public ScriptState ScriptState
{
    get
    {
        if (_scriptState == null)
        {
            _scriptState = CSharpScript.RunAsync(string.Empty, ScriptOptions).Result;
        }

        return _scriptState;
    }
    private set
    {
        _scriptState = value;
    }
}
```

If we're willing to remove the `ScriptState` property and turn it into `Task<ScriptState> GetScriptState()` then we can get rid of the `.Result` call in the `get` method, but if we want to keep the property the only way I can think of is to keep it synchronous; that's the only way to guarantee that a null property isn't accessed too early.

I searched through the tests and couldn't find one that was disabled due to a `null` ScriptState, but I found the `null` error from the LSP test, so maybe that's good enough.

jonsequitur: Looks like the test got deleted at some point.

One approach to avoiding the need to call `.Result` is to call `DeferCommand`. The command will get executed later in an async context.

brettfo: We can't wrap it in `DeferCommand` because the hover method can happen outside of kernel commands; it's just a method on `CSharpKernel`, it needs to be callable from anywhere that a C# kernel exists.

brettfo: I experimented with `Task<ScriptState> GetScriptState()` but that sets off a cascade of `async` that wreaks havoc all over the place, so I added a commit with the `.Result` I mentioned above to see how it feels.

brettfo:
@jonsequitur Take a look at the last commit ('ensure all deferred commands have run before LSP methods').  I had to make `KernelBase.RunDeferredCommandsAsync()` public, but otherwise it seems to work nicely.  I figured you wouldn't like `InternalsVisibleTo`.

dotnet-interactive.Tests/HttpApiTests.cs
```diff
@@ -246,6 +248,72 @@ public async Task unimplemented_lsp_handlers_return_empty_string(Language langua
...
+        [Fact]
+        public async Task lsp_methods_run_deferred_commands()
+        {
+            // declare a variable in deferred code
+            var kernelBase = GetServer().Kernel as KernelBase;
+            kernelBase.DeferCommand(new SubmitCode("var one = 1;"));
+
+            // it's not defined
+            var response = await GetServer().HttpClient.GetAsync($"/variables/{Language.CSharp.LanguageName()}/one");
+            response.StatusCode.Should().Be(HttpStatusCode.NotFound);
```
jonsequitur: This sanity check is not needed.

brettfo: Why not?  I want to ensure the test is in a valid state and that we haven't accidentally executed `var one = 1;` at this time.

dotnet-interactive.Tests/HttpApiTests.cs
```diff
@@ -246,6 +248,72 @@ public async Task unimplemented_lsp_handlers_return_empty_string(Language langua
...
+
+            // ensure we can get http info about it
+            var encoded = Convert.ToBase64String(Encoding.UTF8.GetBytes("Console.WriteLine(one);"));
```
jonsequitur: This trivia shouldn't be in a test. Did we encapsulate this into a class somewhere?

brettfo: No, because this is always handled by the client (`lsp.js`), but we have to fake that here because we don't have a client.

dotnet-interactive.Tests/HttpApiTests.cs
```diff
@@ -246,6 +248,72 @@ public async Task unimplemented_lsp_handlers_return_empty_string(Language langua
...
+            hover.Contents.Value.Should().Be("(field) int one");
+
+            // and it now exists
+            response = await GetServer().HttpClient.GetAsync($"/variables/{Language.CSharp.LanguageName()}/one");
+            var responseContent = await response.Content.ReadAsStringAsync();
+            responseContent.Should().BeJsonEquivalentTo(1);
+        }
+
+        [Fact]
+        public async Task lsp_textDocument_hover_handles_growing_script_offsets()
```
jonsequitur: This kind of test should be done closer to the domain layer, as the thing being tested isn't the serialization or HTTP API.

brettfo: Will do.

brettfo: Check commit 'create separate LSP (de)serialization...` where I use the object model directly in the HTTP tests, and add separate JSON serialization/deserialization tests.

Microsoft.DotNet.Interactive.CSharp/CSharpKernel.cs
```diff
@@ -12,23 +12,25 @@
...
 using XPlot.Plotly;
 using Task = System.Threading.Tasks.Task;
 
 namespace Microsoft.DotNet.Interactive.CSharp
 {
     public class CSharpKernel :
         KernelBase,
-        IExtensibleKernel
+        IExtensibleKernel,
+        ISupportHover
```
jonsequitur: Another option here that might work is something like:

```csharp
public interface ILanguageServer<HoverText> { }
public interface ILanguageServer<Completion> { }
```

...etc.

The return types would also need to be represented somehow but it relates these different aspects of the overall set of LSP functionality. Maybe useful?

Microsoft.DotNet.Interactive.CSharp/CSharpKernel.cs
```diff
@@ -102,33 +104,32 @@ public Task<bool> IsCompleteSubmissionAsync(string code)
...
             {
-                case "textDocument/hover":
-                    // https://microsoft.github.io/language-server-protocol/specification#textDocument_hover
-                    var hoverParams = request.ToLspObject<HoverParams>();
-                    result = TextDocumentHover(hoverParams);
-                    break;
-                default:
-                    result = null;
-                    break;
+                return null;
```
jonsequitur: Do we want to return something other than `null`, e.g. a `TextDocumentHoverResponse.Empty`?

Microsoft.DotNet.Interactive.CSharp/CSharpKernel.cs
```diff
@@ -170,24 +171,22 @@ public TextDocumentHoverResponse TextDocumentHover(HoverParams hoverParams)
...
-                                                           catchException: e =>
-                                                           {
-                                                               exception = e;
-                                                               return true;
-                                                           },
-                                                           cancellationToken: context.CancellationToken)
-                                                       .UntilCancelled(context.CancellationToken);
-                    }
+
+                    ScriptState = await ScriptState.ContinueWithAsync(
```
jonsequitur: Why remove the `else` and add a run?

jonsequitur: And is the null check even needed if we're going to guarantee that `ScriptState` is always initialized? It looks like `GetDocumentWithOffsetFromCode` is already making that assumption.

Microsoft.DotNet.Interactive.Tests/LanguageServiceSerializationTests.cs
```diff
@@ -0,0 +1,57 @@
...
+using Xunit;
+
+namespace Microsoft.DotNet.Interactive.Tests
+{
+    public class LanguageServiceSerializationTests
+    {
+        private T Parse<T>(string json)
+        {
+            var jo = JObject.Parse(json);
+            return jo.ToLspObject<T>();
```
jonsequitur: This will use default serialization settings. What we're doing in `Microsoft.DotNet.Interactive.Server` is encapsulating the serialization (e.g. no explicitly dependency on `NewtonSoft.Json`) and considering serialization and deserialization as part of our API and contract.

Microsoft.DotNet.Interactive.CSharp/CSharpKernel.cs
brettfo: `HoverText` would likely be specified like `class HoverText : ILanguageServiceParams<HoverParams, HoverResponse>`, but what about something like diagnostics, `Diagnostics : ILanguageServiceParams<void, DiagnosticsResult>`?  It feels better to specifically have `ISupportHover` that requires it's own very specific methods along with a separate (in the future) `ISupportDiagnostics` which will have a very different API due to the push nature, and wouldn't fit nicely into the `ILanguageServer<Diagnostics>` model.

Microsoft.DotNet.Interactive.CSharp/CSharpKernel.cs
brettfo: A hover response needs content and an optional range and I can't think of a value that's valid according to the spec that could represent `empty` except for `null`, unless we define it like `public static TextDocumentHoverResponse Empty => null;` which doesn't buy us much.

Microsoft.DotNet.Interactive.CSharp/CSharpKernel.cs
brettfo: I removed the else because if you're executing your first cell, the `catchException` property can't be specified, so it's possible that an error could be swallowed.  This way `catchException` is always used.

`GetDocumentWithOffsetFromCode` removed the null-check and initialization; that's instead handled by the fact that the deferred commands will have caused the initialization by the time `GetDocument...` is called.

Microsoft.DotNet.Interactive.Tests/LanguageServiceSerializationTests.cs
brettfo: LSP types have a strict expectation on member casing that `ToLspObject()` honors; it wouldn't be valid anywhere to serialize it any differently.

Microsoft.DotNet.Interactive.CSharp/CSharpKernel.cs
jonsequitur: Shhh! We were trying to hide those exceptions. 😝 

But what if there are no deferred commands?

dotnet-interactive.Tests/HttpApiTests.cs
jonsequitur: That should be a different test, then. Multiple assertions per test tends to reduce the usefulness of the tests for diagnosing problems.

dotnet-interactive/HttpRouting/LspRouter.cs
```diff
@@ -44,16 +44,31 @@ public async Task RouteAsync(RouteContext context)
...
 
                     if (targetKernel is KernelBase kernelBase)
                     {
+                        await kernelBase.RunDeferredCommandsAsync();
                         var methodName = segments[1];
                         var lspBodyReader = new StreamReader(context.HttpContext.Request.Body);
                         var stringBody = await lspBodyReader.ReadToEndAsync();
                         var request = JObject.Parse(stringBody);
-                        var response = await kernelBase.LspMethod(methodName, request);
+
```
jonsequitur: Same concern here: We should use a method that encapsulates which configuration will be used, rather than going directly to `Newtonsoft.Json`.

brettfo:
Implemented differently in #348.

PR #286, opened 2020-03-18
Integrate PowerShell cloud shell with the PowerShell kernel

daxian-dbw:
Add `Enter-AzShell` cmdlet to `Microsoft.Dotnet.Interactive.PowerShell` module to integrate Azure cloud shell with the PowerShell kernel.

Running `Enter-AzShell` will authenticate you to Azure and establish a PowerShell cloud shell session for you. After that, all subsequent code submitted from the cells are executed in that cloud shell session.
When you are done with the cloud shell, just type and run `exit` from a cell to end the session, and after that, all subsequent code submitted from the cells will again be executed using the local PowerShell kernel.

/cc @SteveL-MSFT 

Mar-18-2020 00-39-31

### DIFF
--- a/Microsoft.DotNet.Interactive.PowerShell/Commands/EnterAzShellCommand.cs
+++ b/Microsoft.DotNet.Interactive.PowerShell/Commands/EnterAzShellCommand.cs
@@ -1,0 +1,41 @@
+// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Management.Automation;
+using System.Management.Automation.Runspaces;
+
+namespace Microsoft.DotNet.Interactive.PowerShell.Commands
+{
+    /// <summary>
+    /// Connect to Azure PowerShell for code execution.
+    /// </summary>
+    [Cmdlet(VerbsCommon.Enter, "AzShell")]
+    public sealed class EnterAzShellCommand : PSCmdlet
+    {
+        /// <summary>
+        /// ProcessRecord override.
+        /// </summary>
+        protected override void ProcessRecord()
+        {
+            var context = KernelInvocationContext.Current;
+            if (context?.CurrentKernel is PowerShellKernel psKernel)
+            {
+                if (Runspace.DefaultRunspace.Id != psKernel.DefaultRunspaceId)
+                {
+                    var ex = new InvalidOperationException("'Enter-AzShell' should run from the default Runspace of the PowerShell kernel.");
+                    var error = new ErrorRecord(
+                        exception: ex,
+                        errorId: "ShouldRunFromDefaultKernelRunspace",
+                        errorCategory: ErrorCategory.InvalidOperation,
+                        targetObject: null);
+                    this.ThrowTerminatingError(error);
+                }
+
+                psKernel.AzShell = new AzShellConnectionUtils(context);
+                var windowSize = Host.UI.RawUI.WindowSize;
+                psKernel.AzShell.ConnectAndInitializeAzShell(windowSize.Width, windowSize.Height).Wait();
+            }
+        }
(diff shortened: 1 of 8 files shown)


daxian-dbw:
@SteveL-MSFT Can you please review this PR when you get a chance? You are most familiar with the involved changes.

daxian-dbw:
We could do something similar to light up `Enter-PSSession` in PowerShell kernel -- run submitted code in a remote PSSession.
And we could abstract out the remote code run interface, to cover both the `Enter-PSSession` and `Enter-AzShell` scenarios.
See  https://github.com/dotnet/interactive/issues/290 for details.

Microsoft.DotNet.Interactive.PowerShell/Commands/EnterAzShellCommand.cs
```diff
@@ -0,0 +1,53 @@
...
+        [Parameter(ParameterSetName = DefaultSetName)]
+        public SwitchParameter Reset;
+
+        /// <summary>
+        /// ProcessRecord override.
+        /// </summary>
+        protected override void ProcessRecord()
+        {
+            var context = KernelInvocationContext.Current;
+            if (context?.CurrentKernel is PowerShellKernel psKernel)
```
jonsequitur: This API is confusing, I know, but this should probably be `HandlingKernel`, not `CurrentKernel`.

Microsoft.DotNet.Interactive.PowerShell/Host/VTColorUtils.cs
```diff
@@ -9,6 +9,7 @@ internal static class VTColorUtils
     {
         internal const string EscapeCharacters = "\x1b[";
         internal const string ResetColor = "\x1b[0m";
+        internal const string BoldColor = "\x1b[1m";
```
jonsequitur: We've added a reference to `System.CommandLine.Rendering` so we might want to refactor to use that for VT. I'm happy to help with the the refactor, as I'd love feedback on the APIs.

Microsoft.DotNet.Interactive.PowerShell/PowerShellKernel.cs
```diff
@@ -150,72 +156,120 @@ public override bool TryGetVariable(string name, out object value)
...
+            }
 
-            var completionList = completion.CompletionMatches.Select(c => new CompletionItem(
-                displayText: c.CompletionText,
-                kind: c.ResultType.ToString(),
-                documentation: c.ToolTip
-            ));
+            if (shouldDispose)
+            {
+                AzShell.Dispose();
```
jonsequitur: You might also want to do `base.RegisterForDisposal` for `AzShell` when it's initialized. For example, do we care if the `PowerShellKernel` gets disposed without someone having done an explicit `exit`?

Microsoft.DotNet.Interactive.PowerShell/Commands/EnterAzShellCommand.cs
daxian-dbw: Adding comment to `HandlingKernel` and `CurrentKernel` properties should be sufficient.
`CurrentKernel` is the one the notebook running against initially, `HandlingKernel` could be changed when using `#!pwsh` for example, is this understanding correct?

Microsoft.DotNet.Interactive.PowerShell/Host/VTColorUtils.cs
daxian-dbw: Would love to refactor and use that library! Opened #299 to track the work.

Microsoft.DotNet.Interactive.PowerShell/PowerShellKernel.cs
daxian-dbw: Good point. Will address this.

PR #300, opened 2020-03-19
build package without magic targets

### LINKED ISSUE #272: Package Microsoft.DotNet.Interactive.PowerShell has duplicate files that cause a failure to build

brettfo:
Another attempt and building the package without MSBuild magic.

Also re-enable building the PowerShell package.  Fixes #272.

### DIFF
--- a/Microsoft.DotNet.Interactive.PowerShell/Microsoft.DotNet.Interactive.PowerShell.csproj
+++ b/Microsoft.DotNet.Interactive.PowerShell/Microsoft.DotNet.Interactive.PowerShell.csproj
@@ -1,4 +1,4 @@
-﻿<Project Sdk="Microsoft.NET.Sdk" InitialTargets="GetNuGetPackageRoot;CopyGalleryModules;">
+﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
     <TargetFramework>netcoreapp3.1</TargetFramework>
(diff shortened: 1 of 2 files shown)


Microsoft.DotNet.Interactive.PowerShell/Microsoft.DotNet.Interactive.PowerShell.csproj
```diff
@@ -18,76 +18,43 @@
...
-    <PackageReference Include="PowerShellGet" Version="$(PowerShellGetVersion)">
-      <PrivateAssets>all</PrivateAssets>
-    </PackageReference>
-    <PackageReference Include="Microsoft.PowerShell.Archive" Version="$(MicrosoftPowerShellArchiveVersion)">
-      <PrivateAssets>all</PrivateAssets>
-    </PackageReference>
-    <PackageReference Include="ThreadJob" Version="$(ThreadJobVersion)">
-      <PrivateAssets>all</PrivateAssets>
-    </PackageReference>
+    <PackageReference Include="PackageManagement" Version="1.4.6" GeneratePathProperty="true" />
```
jonsequitur: Why do away with `<PrivateAssets>all</PrivateAssets>`/

brettfo: Because that wasn't copying the files where we need, which is under the `Modules/` directory.

brettfo: NM, I misunderstood what PrivateAssets was doing here.  Will restore in a future PR.

PR #306, opened 2020-03-19
add kernel execution smoke tests

brettfo:
This adds an integration test that smoke-tests the stdio kernel.  It currently executes `1+1` in both the C# and F# kernels.  The PowerShell kernel isn't included in this test because it doesn't return a value of `2`, rather it reports a stdout value of `2\r\n`, so that'll have to be added later/separately.

The purpose of these tests is to ensure that the just built global tool can execute; it's a really high-level way of ensuring that the NuGet package was laid out correctly.

In the failure case, all of the `JObject` values that were received are output to the log to make diagnosing failures easier.

### DIFF
--- a/src/Microsoft.DotNet.Interactive/Server/KernelEventEnvelope.cs
+++ b/src/Microsoft.DotNet.Interactive/Server/KernelEventEnvelope.cs
@@ -58,6 +58,7 @@ namespace Microsoft.DotNet.Interactive.Server
                 [nameof(IncompleteCodeSubmissionReceived)] = typeof(KernelEventEnvelope<IncompleteCodeSubmissionReceived>),
                 [nameof(InputRequested)] = typeof(KernelEventEnvelope<InputRequested>),
                 [nameof(HoverTextProduced)] = typeof(KernelEventEnvelope<HoverTextProduced>),
+                [nameof(KernelReady)] = typeof(KernelEventEnvelope<KernelReady>),
                 [nameof(PackageAdded)] = typeof(KernelEventEnvelope<PackageAdded>),
                 [nameof(PasswordRequested)] = typeof(KernelEventEnvelope<PasswordRequested>),
                 [nameof(ReturnValueProduced)] = typeof(KernelEventEnvelope<ReturnValueProduced>),
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -289,6 +289,10 @@ stages:
         pool:
           vmImage: ubuntu-16.04
         variables:
+        - name: DOTNET_TRY_CLI_TELEMETRY_OPTOUT
+          value: 1
+        - name: DOTNET_TRY_SKIP_FIRST_TIME_EXPERIENCE
+          value: 1
         # specify tool version on public CI builds
         - ${{ if eq(variables['System.TeamProject'], 'public') }}:
           - name: ToolVersionArgs
--- a/src/Microsoft.DotNet.Interactive.Js/src/dotnet-interactive/contracts.ts
+++ b/src/Microsoft.DotNet.Interactive.Js/src/dotnet-interactive/contracts.ts
@@ -85,6 +85,7 @@ export const ErrorProducedType = "ErrorProduced";
 export const HoverTextProducedType = "HoverTextProduced";
 export const IncompleteCodeSubmissionReceivedType = "IncompleteCodeSubmissionReceived";
 export const InputRequestedType = "InputRequested";
+export const KernelReadyType = "KernelReady";
 export const PackageAddedType = "PackageAdded";
 export const PasswordRequestedType = "PasswordRequested";
 export const ReturnValueProducedType = "ReturnValueProduced";
@@ -105,6 +106,7 @@ export type KernelEventType =
     | typeof HoverTextProducedType
     | typeof IncompleteCodeSubmissionReceivedType
     | typeof InputRequestedType
+    | typeof KernelReadyType
     | typeof PackageAddedType
     | typeof PasswordRequestedType
     | typeof ReturnValueProducedType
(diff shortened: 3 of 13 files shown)


dotnet-interactive.IntegrationTests/KernelSmokeTests.cs
```diff
@@ -0,0 +1,83 @@
...
+                    {
+                        return true;
+                    }
+                }
+
+                return false;
+            }, TimeSpan.FromSeconds(5));
+
+            foundIt.Should().BeTrue("Unable to find expected result in:\n" + string.Join("\n", outputMessages.Select(o => o.ToString())));
+            errorLines.Should().BeEmpty();
```
jonsequitur: We have `ToSubscribedList` for gathering up events, and the event definitions under `Microsoft.DotNet.Interactive.Events` are intended to be serialized, with public serialization APIs (instead of `JObject`). If we use those APIs we'll get better maintainability (if the contract changes, for example) and better test feedback. One example where the current test can fail without being informative (which is all the more important in a test that's typically running only on the server): 

* `foundIt` is false but we can't see why.
* We don't know if `errorLines` was empty, and we can't see the errors that were returned.

dotnet-interactive.IntegrationTests/Utility/ProcessHelper.cs
```diff
@@ -61,5 +61,18 @@ public static class ProcessHelper
...
             return process;
         }
+
+        public static Process StartInteractiveServer(
+            string kernelName,
+            Action<string> output = null,
+            Action<string> error = null)
+        {
+            return Start(
+                command: "dotnet",
```
jonsequitur: Technically the command is `dotnet-interactive`.

brettfo: This was intentional; most user invocations will be of the form `dotnet interactive ...` which I wanted to mimic.

brettfo:
@jonsequitur I've rebased and recreated this using the strongly-type command/event envelopes, but I'm interested in your feedback.  Since we're executing out-of-process, we have no choice but to wait until we eventually get the expected event back, and we can handle this in 2 ways.  I've duplicated the same test with one change:

`EventuallyContainSingle<T>(TimeSpan timeout)` will wait up to `timeout` for the collection to have the requested item.  **Good:** the test feels fluent; **Bad:** it requires a timeout at the callsite and it uses `Thread.Sleep()`.

`EventuallyContainSingle2<T>(Action<...> continuation)` honors the test's specified timeout.  **Good:** test-based timeout feels better, `await Task.Delay()` is generally better; **Bad:** we can't use fluent APIs across an async boundary without involving extra syntax.  I approached this one by passing in a continuation for the rest of the test assertions.

There's a third possibility with two varieties that would look something like this, but I avoided it because it feels like it's breaking up the test awkwardly:

Option A:

``` csharp
var x = await events
    .Should()
    .EventuallyContainSingle<DisplayEventBase>()

x
    .Which
    .FormattedValues
    .Should()
    .Contain(fv => fv.Value.Trim() == "2"));
```

Option B:

``` csharp
(await events                                     //
    .Should()                                     //
    .EventuallyContainSingle<DisplayEventBase>()) // note that we're wrapping this so we can await it
    .Which
    .FormattedValues
    .Should()
    .Contain(fv => fv.Value.Trim() == "2"));
```

jonsequitur:
@colombod and I were just writing a similar assertion, and decided to do a `Task.Run(...).Result` within the custom assertion to allow the test to be synchronous (without risking a deadlock). The `async` isn't really providing any benefit to the test itself. 

I like `EventuallyContain`. Another possibility here is to write a custom assertion on `IObservable<T>` which could include the default timeout and allow this syntax to be more natural than `async`.

brettfo:
I think I'm finally at the end of making these smoke tests stable.  Ultimately it came down two 2 separate issues with the STDIO client:

1. If this is the first launch of the tool, the telemetry information is displayed which can't be parsed as an event envelope.  The fix for this was twofold:
   1. For the integration tests, set the `DOTNET_TRY_CLI_TELEMETRY_OPTOUT` and `DOTNET_TRY_SKIP_FIRST_TIME_EXPERIENCE` variables to prevent this display (and the telemetry); it doesn't make sense to collect telemetry for an integration test.
   2. When debugging locally I can't guarantee that the first-run message won't happen, so we may still get the message that breaks parsing.  The fix here was to swallow `JsonReaderExtension` when parsing the event envelope.  This is a pattern we follow elsewhere and if something else is broken, the test will still fail as expected.
2. We hit a race on the very first launch of the global tool in the integration test where the `SubmitCode` command was sent before the server was ready to listen to STDIN, so it was getting lost.  Possibly related to JIT, but I can't be sure.  This isn't an issue on dev machines, just the integration VMs.  The fix for this was to emit a `Ready` event when the STDIO server is listening and have the test client wait for that before continuing.  This same race could also happen in a VS Code notebook and a subsequent PR will wait for the `Ready` event.  We can't make that change yet, though, because a version of the tool that outputs that event hasn't yet been published so the client would never get that event.

jonsequitur:
Could item 2 be the cause for #537 (which is still intermittently happening to me on my dev machine)?

brettfo:
@jonsequitur Possibly, as a separate work item I'll look at how the VS Code client handles unparsable lines.

jonsequitur:
Another unparseable output I've been seeing is the kestrel output when HTTP ports are configured. 

image

src/Microsoft.DotNet.Interactive/Server/KernelEventEnvelope.cs
```diff
@@ -61,6 +61,7 @@ public static void ResetToDefaults()
                 [nameof(HoverTextProduced)] = typeof(KernelEventEnvelope<HoverTextProduced>),
                 [nameof(PackageAdded)] = typeof(KernelEventEnvelope<PackageAdded>),
                 [nameof(PasswordRequested)] = typeof(KernelEventEnvelope<PasswordRequested>),
+                [nameof(Ready)] = typeof(KernelEventEnvelope<Ready>),
```
jonsequitur, line 64: Maybe `KernelReady`?

brettfo: Done.

PR #310, opened 2020-03-22
using targets to avoid parallel execution in powershell project

colombod:


jonsequitur:
Does this solve the local build problem?

colombod:
No, I can’t see the module folder in the output of the build

brettfo:
The `DependsOnTargets="Restore"` will just move the slowness to later in the build.  I had some luck locally simply wrapping the entire `<ItemGroup>` with `<Target Name="GatherModulesContent" BeforeTargets="Build">`.

Can you try that locally to see what happens?

Microsoft.DotNet.Interactive.PowerShell/Microsoft.DotNet.Interactive.PowerShell.csproj
```diff
@@ -31,39 +31,43 @@
              Link="Modules\%(RecursiveDir)%(FileName)%(Extension)"
              PackagePath="contentFiles/any/any/Modules"
              PackageCopyToOutput="true"
-             CopyToOutputDirectory="PreserveNewest" />
+             CopyToOutputDirectory="PreserveNewest"/>
```
brettfo: This hurts.

PR #315, opened 2020-03-24
update SDK sha and add comment

brettfo:
Update missing SDK sha and add helpful comment.

### DIFF
--- a/Binder Dependencies/Dockerfile
+++ b/Binder Dependencies/Dockerfile
@@ -27,10 +27,12 @@ RUN apt-get install -y --no-install-recommends \
 RUN rm -rf /var/lib/apt/lists/*
 
 # Install .NET Core SDK
+
+# When updating the SDK version, the sha512 value a few lines down must also be updated.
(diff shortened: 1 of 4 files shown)


brettfo:
Adding a commit to build the Dockerfiles as part of CI to catch this in the future.  The commit has been poisoned to verify that it fails appropriately.  Once that's good, I'll back out the wrong values.

brettfo:
With a bad SHA, the build failed with:

```
sha512sum: WARNING: 1 computed checksum did NOT match
```

Restored appropriate SHA, so once green it should be good to go.

brettfo:
@colombod Is there a special syntax used in `Binder Dependencies/Dockerfile`?  The build of the image is failing with:

```
Step 17/36 : COPY ./notebooks/ ${HOME}/notebooks/
COPY failed: stat /var/lib/docker/tmp/docker-builder468519462/notebooks: no such file or directory
```

Or are there extra arguments we need to specify?

PR #321, opened 2020-03-27
fetch mock for api tests

colombod:


jonsequitur:
>     <TypescriptInputs Include="$(JsApiRoot)package.json" />

IS this common that mocha.opts would be part of the product package itself?

---
Refers to: dotnet-interactive/dotnet-interactive.csproj:130 in 933e775. [](commit_id = 933e775d8c5fead883551254f7aec5390fa4ec4d, deletion_comment = False)

Microsoft.DotNet.Interactive.Js/tests/variableApi.spec.ts
```diff
@@ -0,0 +1,62 @@
...
+
+describe("dotnet-interactive", () => {
+    const rootUrl = "https://dotnet.interactive.com:999";
+    beforeEach(() => {
+        const expectedKernels = require("./Responses/kernlesResponse.json");
+        fetchMock.get(`${rootUrl}/kernels`, expectedKernels);
+    })
+    afterEach(fetchMock.restore);
+    describe("Variable api", () => {
+        it("car load multiple variables in a single request", async () => {
```
jonsequitur: >car [](start = 12, length = 3)

"can"

Microsoft.DotNet.Interactive.Js/tests/variableApi.spec.ts
```diff
@@ -0,0 +1,62 @@
...
+
+            fetchMock.post(`${rootUrl}/variables`, require("./Responses/variableBundleResponse01.json"));
+
+            let client = await createDotnetInteractiveClient(rootUrl);
+            let variables = await client.getVariables(varaibleRequest);
+
+            expect(variables).to.deep.eq(expectedVariables);
+        });
+
+        it("car load variable using kernel client", async () => {
```
jonsequitur: >car [](start = 12, length = 3)

"can"

Microsoft.DotNet.Interactive.Js/tests/testSupprot.ts
```diff
@@ -0,0 +1,5 @@
+import { DotnetInteractiveClient, KernelClientContainer } from "../src/dotnet-interactive/dotnet-interactive-interfaces";
+
+export function asKernelClientContainer(client: DotnetInteractiveClient): KernelClientContainer {
+    return <KernelClientContainer><any>client;
+}
```
jonsequitur: File name is misspelled.

Microsoft.DotNet.Interactive.Js/src/dotnet-interactive/KernelClientImpl.ts
```diff
@@ -61,4 +67,35 @@ export class KernelClientImpl implements DotnetInteractiveClient {
...
+}
+
+export async function createDotnetInteractiveClient(address: string, clientFetch: ClientFetch = null): Promise<DotnetInteractiveClient> {
+
+    let rootUrl = address;
+    if (!address.endsWith("/")) {
+        rootUrl = `${rootUrl}/`;
+    }
+
+    async function defaukltClientFetch(input: string, requestInit: RequestInit = null): Promise<Response> {
```
jonsequitur: >defauklt [](start = 19, length = 8)

Spelling

Microsoft.DotNet.Interactive.Js/src/dotnet-interactive.ts
```diff
@@ -1 +1 @@
-export * from "./dotnet-interactive/KernelClient";
\ No newline at end of file
+export * from "./dotnet-interactive/librayInit";
```
jonsequitur: >librayInit [](start = 36, length = 10)

Spelling

Microsoft.DotNet.Interactive.Js/tests/client.spec.ts
```diff
@@ -3,26 +3,56 @@
...
-            .equal('function');
+            expect(typeof (global.createDotnetInteractiveClient))
+                .to
+                .equal('function');
         });
-       
+    });
+
+    describe("kernel discovery", () => {
+        it("creates kernel clients for all discovered kernesl", async () => {
```
jonsequitur: >kernesl [](start = 54, length = 7)

Spelling

Microsoft.DotNet.Interactive.Js/tests/client.spec.ts
```diff
@@ -3,26 +3,56 @@
...
+            const rootUrl = "https://dotnet.interactive.com:999";
+            const expectedKernels = require("./Responses/kernlesResponse.json");
```
jonsequitur: >kernlesResponse [](start = 57, length = 15)

Spelling

Microsoft.DotNet.Interactive.Js/tests/Responses/variableBundleResponse01.json
```diff
@@ -0,0 +1,11 @@
...
+  "csharp":
+  {
+      "a": 1,
+      "b": "2"
+  },
+  "fsharp": {
+      "c": 3,
+      "d": false
+  }
+}
```
jonsequitur: Can we give this file a more intention-revealing name?

Do these responses need to be in separate files?

dotnet-interactive.IntegrationTests/HttpApiTests.cs
```diff
@@ -0,0 +1,63 @@
...
+
+            using var client = new HttpClient();
+
+            listeningAddress.Should().NotBeNullOrWhiteSpace();
+
+            var response = await client.GetAsync($"{listeningAddress}/resources/dotnet-interactive.js");
+
+            await response.ShouldSucceed();
+
+            response.Content.Headers.ContentType.MediaType.Should().Be("application/javascript");
```
brettfo: Add `kernelServer.Process.Kill()` and `.WaitForExit().Should().Be().True()`, otherwise the `dotnet-interactive.exe` process will live forver.

dotnet-interactive/dotnet-interactive.csproj
```diff
@@ -126,14 +126,13 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <TypescriptInputs Include="$(JsApiRoot)mocha.opts" />
```
brettfo: If changes to this file can affect the resulting .js files, then it should remain in the item group.  If not, e.g., test only, then the removal is fine.

brettfo: And to add a clarifying point for @jonsequitur, inclusion in the `@(TypescriptInputs)` group does not mean the file will end up in the package, it simply means that changes to any of these files invalidate the output of the TypeScript compilation.

colombod: that file contains settings for running tests only

dotnet-interactive.IntegrationTests/HttpApiTests.cs
jonsequitur: This should be done in test teardown I suspect, not within the test itself, since the same would be true if an earlier assertion prevents the test method from running to completion.

---
In reply to: [399446642](https://github.com/dotnet/interactive/pull/321#discussion_r399446642) [](ancestors = 399446642)

dotnet-interactive.IntegrationTests/HttpClientTestExtensions.cs
```diff
@@ -0,0 +1,35 @@
...
+        public static async Task<HttpResponseMessage> PostJsonAsync(this HttpClient client, string requestUri,
+            JObject requestBody)
+        {
+            var content = new StringContent(requestBody.ToString(), Encoding.UTF8, "application/json");
+            var response = await client.PostAsync(requestUri, content);
+            return response;
+        }
+
+        public static async Task ShouldSucceed(this HttpResponseMessage response)
+        {
```
jonsequitur: The approach we took to this in dotnet/try was more specific, I believe. We should assert against the actual status code. That can be an optional parameter that defaults to 200, but a change in that status code to a different success code should still fail the test.

PR #325, opened 2020-03-31
use --http-port-range

colombod:
--http-port-range can be used to constrain the range of ports to use for http api.
the option is available at runtime and install time
```>dotnet interactive jupyter install --http-port-range 1000-2000```

### DIFF
--- a/dotnet-interactive.Tests/CommandLine/CommandLineParserTests.cs
+++ b/dotnet-interactive.Tests/CommandLine/CommandLineParserTests.cs
@@ -6,6 +6,7 @@ using System.CommandLine.Binding;
 using System.CommandLine.IO;
 using System.CommandLine.Parsing;
 using System.IO;
+using System.Linq;
 using System.Threading.Tasks;
 using FluentAssertions;
 using Microsoft.DotNet.Interactive.App.CommandLine;
@@ -73,6 +74,49 @@ namespace Microsoft.DotNet.Interactive.App.Tests.CommandLine
                 .Verbose
                 .Should()
                 .BeTrue();
+        }
+
+        [Fact]
+        public void It_parses_port_range_option()
+        {
+            var result = _parser.Parse($"jupyter --http-port-range 3000-4000 {_connectionFile}");
+
+            var binder = new ModelBinder<StartupOptions>();
+
+            var options = (StartupOptions)binder.CreateInstance(new BindingContext(result));
+
+            options
+                .HttpPortRange
+                .Should()
+                .BeEquivalentTo(new PortRange{ Start = 3000, End = 4000 });
+        }
+
+        [Fact]
+        public void jupyter_install_parses_port_range_option()
+        {
+            var result = _parser.Parse("jupyter install --http-port-range 3000-4000");
+
+            var binder = new ModelBinder<StartupOptions>();
+
+            var options = (StartupOptions)binder.CreateInstance(new BindingContext(result));
+
+            options
+                .HttpPortRange
+                .Should()
+                .BeEquivalentTo(new PortRange { Start = 3000, End = 4000 });
+        }
+
+        [Theory]
+        [InlineData("jupyter --http-port 8000 --http-port-range 3000-4000")]
+        [InlineData("jupyter --http-port-range 3000-4000 --http-port 8000")]
+        public void port_range_and_port_cannot_be_specified_together(string commandLine)
+        {
+            var result =  _parser.Parse(commandLine);
+
+           result.Errors
+               .Select(e => e.Message)
+                .Should()
+                .Contain(errorMessage => errorMessage == "Cannot specify both http-port and http-port-range together");
         }
 
         [Fact]
--- a/dotnet-interactive.Tests/CommandLine/CommandLineParserTests.cs
+++ b/dotnet-interactive.Tests/CommandLine/CommandLineParserTests.cs
@@ -116,7 +116,7 @@ namespace Microsoft.DotNet.Interactive.App.Tests.CommandLine
             result.Errors
                 .Select(e => e.Message)
                  .Should()
-                 .Contain(errorMessage => errorMessage == "Cannot specify both http-port and http-port-range together");
+                 .Contain(errorMessage => errorMessage == "Cannot specify both --http-port and --http-port-range together");
         }
 
         [Fact]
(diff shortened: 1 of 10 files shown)

### CODE REVIEW

File: dotnet-interactive.Tests/CommandLine/CommandLineParserTests.cs [hunk 1]
```diff
@@ -75,6 +76,49 @@ public async Task It_parses_verbose_option()
                 .BeTrue();
         }
 
+        [Fact]
+        public void It_parses_port_range_option()
```
jonsequitur: This test name should call out specifically which command is being tested.

dotnet-interactive.Tests/JupyterInstallCommandTests.cs
```diff
@@ -4,12 +4,29 @@
...
+        [Fact]
+        public async Task Appends_http_port_range_arguments()
+        {
+            var console = new TestConsole();
+            var kernelSpec = new  InMemoryJupyterKernelSpec(true, null);
+            var jupyterCommandLine = new JupyterInstallCommand(console, kernelSpec , new PortRange{ Start = 100, End = 400});
+
+            await jupyterCommandLine.InvokeAsync();
+
+            foreach (var installedKernelSpec in kernelSpec.InstalledKernelSpecs)
```
jonsequitur: This test will pass if the collection is empty. It's safer to do something like `kernelSpec.InstalledKernelSpecs.Should().All...`

dotnet-interactive/CommandLine/CommandLineParser.cs
```diff
@@ -99,6 +100,51 @@ public static class CommandLineParser
...
+                        pr.End = end;
+                        if (start > end)
+                        {
+                            result.ErrorMessage = "STart port must be lower then end port";
+                            return null;
+                        }
+                        return pr;
+                    }
+
+                    result.ErrorMessage = "Must specify a port range as 1000-3000";
```
jonsequitur: It's better not to use specific numbers here.

dotnet-interactive/Program.cs
```diff
@@ -113,6 +115,27 @@ int GetFreePort()
...
+                for (var currentPort = httpPortRange.Start; currentPort <= httpPortRange.End; currentPort++)
+                {
+                    try
+                    {
+                        var l = new TcpListener(IPAddress.Loopback, currentPort);
+                        l.Start();
+                        var port = ((IPEndPoint)l.LocalEndpoint).Port;
+                        l.Stop();
+                        return currentPort;
+                    }
```
jonsequitur: Rather than duplicate this code, can we consolidate these into a single method with an optional parameter for the port range?

dotnet-interactive/CommandLine/PortRange.cs
```diff
@@ -0,0 +1,11 @@
+// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.DotNet.Interactive.App.CommandLine
+{
+    public class PortRange
+    {
+        public int Start { get; set; }
+        public int End { get; set; }
+    }
```
jonsequitur: It looks like this can be made immutable.

dotnet-interactive/CommandLine/CommandLineParser.cs
```diff
@@ -99,6 +100,51 @@ public static class CommandLineParser
...
+                    }
+
+                    var pr = new PortRange();
+
+                    var source = result.Tokens[0].Value;
+
+                    if (string.IsNullOrWhiteSpace(source))
+                    {
+                        result.ErrorMessage = "Must specify a port range";
+                        return null;
```
jonsequitur: Can this actually get hit?

dotnet-interactive/CommandLine/CommandLineParser.cs
```diff
@@ -99,6 +100,51 @@ public static class CommandLineParser
...
+                    }
+
+                    if (int.TryParse(parts[0], out var start) && int.TryParse(parts[1], out var end))
+                    {
+                        pr.Start = start;
+                        pr.End = end;
+                        if (start > end)
+                        {
+                            result.ErrorMessage = "STart port must be lower then end port";
```
jonsequitur: >STart [](start = 51, length = 5)

Weird casing.

dotnet-interactive/CommandLine/CommandLineParser.cs
```diff
@@ -99,6 +100,51 @@ public static class CommandLineParser
...
+                        return null;
+                    }
+
+                    var parts = source.Split(new[] {"-"}, StringSplitOptions.RemoveEmptyEntries);
+
+                    if (parts.Length != 2)
+                    {
+                        result.ErrorMessage = "Must specify a port range";
+                        return null;
+                    }
```
jonsequitur: Can this be cleaned up so the error cases aren't scattered around the method? This one should be able to be consolidate into the one below, which is more informative because it also tells people the form that the argument needs to take.

dotnet-interactive.Tests/InMemoryJupyterKernelSpec.cs
```diff
@@ -25,12 +26,20 @@ public class InMemoryJupyterKernelSpec : IJupyterKernelSpec
             _error = error;
         }
 
+        public List<string> InstalledKernelSpecs
+        {
+            get => _kernelSpecs;
+        }
```
jonsequitur: Backing field and method names should match. Or you can use a get-only property instead of a backing field since you expose `List<string>` as the type.

jonsequitur:
>     public class InMemoryJupyterKernelSpec : IJupyterKernelSpec

This class has enough logic in it, and no tests comparing its behavior to the "real" version, that it makes me suspicious of the value of the tests using it.

---
Refers to: dotnet-interactive.Tests/InMemoryJupyterKernelSpec.cs:12 in 4edb913. [](commit_id = 4edb9138e1810a79795e815f7422bfc9afc94dc6, deletion_comment = False)

colombod:
> > ```
> > public class InMemoryJupyterKernelSpec : IJupyterKernelSpec
> > ```
> 
> This class has enough logic in it, and no tests comparing its behavior to the "real" version, that it makes me suspicious of the value of the tests using it.
> 
> Refers to: dotnet-interactive.Tests/InMemoryJupyterKernelSpec.cs:12 in [4edb913](https://github.com/dotnet/interactive/commit/4edb9138e1810a79795e815f7422bfc9afc94dc6). [](commit_id = [4edb913](https://github.com/dotnet/interactive/commit/4edb9138e1810a79795e815f7422bfc9afc94dc6), deletion_comment = False)

This class is capturing the files the real commandline will see.

Issue #157, opened 2020-02-11
.NET kernel installation failed with error: Could not find jupyter kernelspec module


romancerny:
https://devblogs.microsoft.com/dotnet/net-interactive-is-here-net-notebooks-preview-2/#comment-4582

Above link instructs to: "Next, in an ordinary console, install the dotnet interactive global tool:"
`> dotnet tool install --global Microsoft.dotnet-interactive`

By ordinary console I understood Win 10 cmd. After running following command tool installed without errors and confirmed that it was installed successfully.

`> dotnet tool install --global Microsoft.dotnet-interactive --version 1.0.110520`

Following the next step: "Install the .NET kernel by running the following within your Anaconda Prompt:"

`> dotnet interactive jupyter install`

Returns following error. Below are the details of my command execution:

`(base) C:\Users\roman.cerny>jupyter kernelspec list`
Available kernels:
python3 C:\Program Files (x86)\Microsoft Visual Studio\Shared\Anaconda3_64\share\jupyter\kernels\python3

`(base) C:\Program Files (x86)\Microsoft Visual Studio\Shared\Anaconda3_64\share>dotnet interactive jupyter install`
.NET kernel installation failed with error: Could not find jupyter kernelspec module

The same happens even after restarting my PC

Please advise

Thank you,
Roman

romancerny:
Thank you @jonsequitur , would there be any ETA for this bug fix?

colombod:
PR #337 should address the issue. If the kernelspec module is not found the setup uses the path as indicated in jupyter docs. It is also possible to specify a path at install time

Issue #148, opened 2020-02-09
Custom install location of jupyter kernel


MikaelUmaN:
**Is your feature request related to a problem? Please describe**

I use docker for my development and have an image that I like to add net kernels to.

Usually I mount a folder as the home folder of the user in order to get things like ssh keys and configs in the container while the image remains general. I imagine this is quite common.

While I can install dotnet where I like and dotnet interactive as well, it seems “dotnet interactive install jupyter” always picks the home folder.

Thus when I volume mount the .dotnet folder is overwritten and doesn’t show up.

**Describe the solution you'd like**

Cmd option to point to a path, e.g /opt/.dotnet

**Describe alternatives you've considered**

Various workarounds which are OK but are invasive to my colleagues that use the same image.

colombod:
You might have to change the way jupyter looks up the folders for kernel specs. With the pr #337  you can specify where you want the kernel spec file installed, just make sure jupyter or jupyter lab can read from there

Issue #135, opened 2020-02-07
dotnet-interactive jupyter install should work without anaconda...


Jaykul:
### The problem
Currently the `dotnet interactive jupyter install` can only be run in an environment where Python with Jupyter is working. Why?

There are many implementations of the jupyter kernel.

Obviously you know about [nteract](https://nteract.io), since you've linked to in your article. It's not the only alternative implementation that does not need Python to be installed, nor the python jupyter.

Right now, if I have installed `Microsoft.dotnet-interactive` and `nteract` on a computer that doesn't have Python, all I have to do is copy the kernel.json files from another computer to make them work, so all we need is for the install command to work!

### The install command should have a default location

The "install" command should, if it can't find Juptyer, just install the kernels to the well-known default location that Jupyter (and nteract) uses. On Windows, that's `%AppData%\jupyter\kernels`

#### Alternatively
For Windows users, it would actually be really awesome if you packaged up nteract and the dotnet kernels with the dotnet runtime, and bundled them in an installer 😁 so we could download an actual desktop Windows app and get everything running.

daxian-dbw:
@jonsequitur Is there a plan to improve the installation as suggested in this issue?
I think a more straightforward local installation can help the adoption of dotnet-interactive. If you think the work is relatively scoped, maybe we PowerShell team can help on this one 😄

jonsequitur:
Absolutely agreed. Anaconda is not a requirement nor a dependency, it's just the simplest solution for most people to install Jupyter. The installer clearly needs to handle more cases.

@colombod and I had a discussion around what we might do here. A couple of ideas:

* If `jupyter kernelspec` is unavailable (whether not installed or not on the path), list a number of well-known paths as @Jaykul suggested and ask if we should install to those locations. (When available, `jupyter kernelspec install` which we currently use is the recommended and preferred approach.)
* Provide an option to specify the install location (e.g. `dotnet interactive jupyter install --location /path/to/kernelspecs`

Thoughts?

daxian-dbw:
I like both ideas. When the `--location` flag is not specified, the installer by default leverages `jupyter kernelspec`, or list the well-known paths if the former is not available. When `--location` is specified, just simply install to the location.

colombod:
PR #337 adds new behaviour for the install command

PR #337, opened 2020-04-08
change strategy for kernel spec install

colombod:
Addressing #148 , #157 , #135 

Adding option to install the kernels to a specific directory using 
```
> dotnet interactive jupyter install --location [my folder]
```

When installing using the command
```
> dotnet interactive jupyter install 
```
if the jupyter kernelspec module is not accessible the command will try to locate the following paths

Windows | linux | macos 
------------ | ------------- | -------------
 %AppData%/jupyter/kernels | ~/.local/share/jupyter/kernels |~/Library/Jupyter/kernels

As documented [here](https://jupyter-client.readthedocs.io/en/stable/kernels.html)


daxian-dbw:
@colombod This is awesome! ❤️

dotnet-interactive/CommandLine/CommandLineParser.cs
```diff
@@ -153,6 +153,11 @@ public static class CommandLineParser
                 "--log-path",
                 "Enable file logging to the specified directory");
 
+            var locationOption = new Option<DirectoryInfo>(
+                "--location",
```
jonsequitur, line 157: This should be `--path` for consistency.

dotnet-interactive/CommandLine/CommandLineParser.cs
```diff
@@ -269,8 +275,8 @@ Task<int> JupyterHandler(StartupOptions startupOptions, JupyterOptions options,
                     return jupyter(startupOptions, console, startServer, context);
                 }
 
-                Task<int> InstallHandler(IConsole console, InvocationContext context, PortRange httpPortRange) =>
-                    new JupyterInstallCommand(console, new JupyterKernelSpec(), httpPortRange).InvokeAsync();
+                Task<int> InstallHandler(IConsole console, InvocationContext context, PortRange httpPortRange, DirectoryInfo location) =>
+                    new JupyterInstallCommand(console, new JupyterKernelSpec(), httpPortRange, location).InvokeAsync();
```
jonsequitur, line 279: `location` -> `path`

dotnet-interactive/JupyterInstallCommand.cs
```diff
@@ -48,19 +53,49 @@ public async Task<int> InvokeAsync()
...
+                    {
+                        _console.Out.WriteLine($"Installing kernels at location {_location.FullName}");
+                    }
+                    else if (_jupyterKernelSpec.CanInstall)
+                    {
+                        _console.Out.WriteLine("Installing kernels using jupyter kernelspec module");
+                    }
+                    else
+                    {
+                        _console.Out.WriteLine("jupyter kernelspec module is not available, installing kernels default directory");
```
daxian-dbw:
```suggestion
                        _console.Out.WriteLine("jupyter kernelspec module is not available, installing kernels at the default directory");
```

daxian-dbw: Would it be useful to print out the default directory as well?

dotnet-interactive/JupyterInstallCommand.cs
```diff
@@ -69,6 +104,107 @@ public async Task<int> InvokeAsync()
...
+                    directory =  new DirectoryInfo(Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData),"jupyter","kernels"));
+                    break;
+                case PlatformID.Unix:
+                    directory = new DirectoryInfo("~/.local/share/jupyter/kernels");
+                    break;
+                case PlatformID.MacOSX:
+                    directory = new DirectoryInfo("~/Library/Jupyter/kernels");
+                    break;
+                default:
+                    throw new ArgumentOutOfRangeException();
```
daxian-dbw: Maybe a `PlatformNotSupportedException` is better?

dotnet-interactive.Tests/CommandLine/CommandLineParserTests.cs
```diff
@@ -91,6 +93,18 @@ public void jupyter_command_parses_port_range_option()
...
+        [Fact]
+        public void jupyter_install_command_parses_path_option()
+        {
+            Directory.CreateDirectory(_kernelSpecInstallPath.FullName);
+            
+            var result = _parser.Parse($"jupyter install --path {_kernelSpecInstallPath}");
+
+            var option = result.CommandResult.OptionResult("--path");
+
+            option.Should().NotBeNull();
```
jonsequitur, line 105: We should also verify that the path was captured correctly. Something like this:

```csharp

result.FindResultFor(option).GetValueOrDefault().Should().Be(...)

```

dotnet-interactive/JupyterKernelSpecInstaller.cs
```diff
@@ -0,0 +1,155 @@
...
+                    directory = new DirectoryInfo(Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData), "jupyter", "kernels"));
+                    break;
+                case PlatformID.Unix:
+                    directory = new DirectoryInfo("~/.local/share/jupyter/kernels");
+                    break;
+                case PlatformID.MacOSX:
+                    directory = new DirectoryInfo("~/Library/Jupyter/kernels");
+                    break;
+                default:
+                    throw new ArgumentOutOfRangeException();
```
daxian-dbw, line 149: Maybe use `PlatformNotSupportedException`?

dotnet-interactive/JupyterKernelSpecInstaller.cs
```diff
@@ -0,0 +1,155 @@
...
+                    }
+                }
+
+                var notAvailable = "The kernelspec module is not available.";
+
+                var location = GetDefaultDirectory();
+                
+                if (!location.Exists)
+                {
+                    return new  KernelSpecInstallResult(false, string.Join('\n', notAvailable, $"The kernelspec path ${location.FullName} does not exist."));
```
daxian-dbw, line 54: Why not create the default path and then install there? As a user, this allows me to install `dotnet-interactive` first and then install `nteract`, right?

dotnet-interactive/CommandLine/CommandLineParser.cs
```diff
@@ -153,6 +153,11 @@ public static class CommandLineParser
                 "--log-path",
                 "Enable file logging to the specified directory");
 
+            var pathOption = new Option<DirectoryInfo>(
+                "--path",
+                "Installs the kernelspecs to the specified directory")
+                .ExistingOnly();
```
daxian-dbw, line 159: Is `ExistingOnly` really needed?  Would it be more consistent to behave like `--log-path`? I guess the log directory will be created if it doesn't exist.

dotnet-interactive/JupyterKernelSpecInstaller.cs
jonsequitur: Interesting idea. Do you think this makes things easier and/or increases the potential for things to go wrong?

(Also, just to be nitpicky, these paths aren't specific to nteract.)

dotnet-interactive/CommandLine/CommandLineParser.cs
jonsequitur: Ideally this directory exists and was created by some flavor of Jupyter install. If the directory doesn't exist, and we create it on the fly, it seems like it increases the likelihood that Jupyter won't find the kernelspecs.

dotnet-interactive.Tests/InMemoryJupyterKernelSpecInstaller.cs
```diff
@@ -4,55 +4,55 @@
...
             _shouldInstallSucceed = shouldInstallSucceed;
             _shouldUninstallSucceed = shouldUninstallSucceed;
-            _error = error;
+            _message = message;
         }
 
         public List<string> InstalledKernelSpecs { get; } = new List<string>();
 
-        public Task<CommandLineResult> InstallKernel(DirectoryInfo directory)
+        public Task<KernelSpecInstallResult> InstallKernel(DirectoryInfo kernelSpecPath, DirectoryInfo destination = null)
```
jonsequitur, line 28: Does this method need to return `Task`?

dotnet-interactive.Tests/InMemoryJupyterKernelSpecInstaller.cs
```diff
@@ -4,55 +4,55 @@
...
+                        string.Join('\n', _message, 
+                        $"[InstallKernelSpec] Installed kernelspec {kernelSpecPath.Name} in {installPath}" )));
             }
 
-            return Task.FromResult(new CommandLineResult(1, error: _error));
+            return Task.FromResult(new KernelSpecInstallResult(false,  _message));
         }
 
-        public Task<CommandLineResult> UninstallKernel(DirectoryInfo directory)
+        public Task<KernelSpecInstallResult> UninstallKernel(DirectoryInfo directory)
```
jonsequitur, line 47: Does this method need to return `Task`?

dotnet-interactive.Tests/JupyterInstallCommandTests.cs
```diff
@@ -26,37 +27,41 @@ public async Task Appends_http_port_range_arguments()
...
+
             await installCommand.InvokeAsync();
-            
-            console.Error.ToString().Should()
-                   .Contain(".NET kernel installation failed")
-                   .And
-                   .Contain("Could not find jupyter kernelspec module");
+            var consoleError = console.Error.ToString();
+            using var scope = new AssertionScope();
+            consoleError.Should().Contain("Failed to install \".NET (F#)\" kernel. The error was: Could not find jupyter kernelspec module.");
```
jonsequitur, line 44: You can also do `Should().ContainAll(...)` and not need the `AssertionScope`.

dotnet-interactive/JupyterKernelSpecInstaller.cs
daxian-dbw: I think the flexibility is nice to have. I cannot think of how this can make things go wrong, but I'm not as familiar to the Jupyter scenarios like you guys, so I might be wrong.

dotnet-interactive.Tests/JupyterKernelSpecTests.cs
```diff
@@ -35,24 +35,43 @@ public async Task Returns_success_output_when_kernel_installation_succeeded()
             var kernelDir = CreateDirectory();
 
             var result = await kernelSpec.InstallKernel(kernelDir);
-            result.Error.Should().BeEmpty();
-            result.ExitCode.Should().Be(0);
+            using var scope = new AssertionScope();
+            result.Succeeded.Should().BeTrue();
 
             _kernelInstallations.Add(new DirectoryInfo(kernelDir.Name));
```
jonsequitur, line 41: I just noticed this pattern. Adding cleanup steps after assertions will prevent the cleanup from happening if the assertion fails.

dotnet-interactive/CommandLine/CommandLineParser.cs
daxian-dbw: If the user specifies the path, then we probably should trust the user knows what he is doing :) Otherwise, the user should depends on the default path.

dotnet-interactive/JupyterKernelSpecInstaller.cs
jonsequitur: It's possible, though it seems pretty unlikely, that some Jupyter installer could fail if the directory is already present or not empty.

dotnet-interactive/CommandLine/CommandLineParser.cs
jonsequitur: Creating a directory is pointless if Jupyter doesn't know about it. Maybe I want to run the `dotnet interactive jupyter install` before I've installed Jupyter but it seems like an edge case, won't block anyone, and adds complexity, so I'm happy to keep it simpler until we have more feedback.

dotnet-interactive/JupyterKernelSpecInstaller.cs
colombod: Good idea

dotnet-interactive.Tests/InMemoryJupyterKernelSpecInstaller.cs
colombod: this is used in the command to print out to console

dotnet-interactive.Tests/InMemoryJupyterKernelSpecInstaller.cs
colombod: this is used in the command to print out to console

jonsequitur: I mean if all return paths need do `Task.FromRestult` then it looks like there's no need to wrap the `KernelSpecInstallResult` in a `Task`.

PR #345, opened 2020-04-14
More variable sharing

jonsequitur:
Here I'm experimenting with how to access a variable in one subkernel from another subkernel. 

Please have a look at `VariableSharingTests.cs` and share your thoughts.

The next step in the experiment is to do something like this:

```fsharp
#!fsharp
let x = 123
```

```csharp
#!csharp
fsharp.TryGetVariable("x", out var x);
x
```

EDIT: The latter approach isn't working yet as it requires cross-kernel dependency resolution for types.


Microsoft.DotNet.Interactive.Tests/VariableSharingTests.cs
```diff
@@ -0,0 +1,53 @@
...
+namespace Microsoft.DotNet.Interactive.Tests
+{
+    public class VariableSharingTests
+    {
+        [Theory]
+        [InlineData(
+            "fsharp",
+            "let x = 123",
+            "csharp",
+            "GetKernel(\"fsharp\").TryGetVariable(\"x\", out var x);\nx")]
```
colombod, line 23: The api for ```TryGetVariable``` looks good to me. Should we use same gesture for getting kernel ?

Microsoft.DotNet.Interactive.PowerShell/Microsoft.DotNet.Interactive.PowerShell.csproj
```diff
@@ -33,41 +33,7 @@
              PackageCopyToOutput="true"
              CopyToOutputDirectory="PreserveNewest" />
 
-    <!-- PackageManagement -->
```
colombod, line 36: revert this changes :)

PR #348, opened 2020-04-16
rework kernel lsp handling and properly implement C# hover text

brettfo:
Kind of a re-work of #285, but with a bunch of changes I discussed privately with @jonsequitur.  Leaving the other PR up temporarily as a reference point for the next tasks (see below.)

This changes the language service handling to happen purely at the command/event layer along with shape/interface checking for capabilities on the kernels themselves.  When F#/PowerShell are ready to support hover text, they simply need to implement the `IKernelLanguageService<RequestHoverTextCommand>` interface; the rest comes for free.  Future language service capabilities (e.g., completion) will be handled in the same way.

The LSP protocol layers are kept completely separate in the `dotnet-interactive/Lsp` directory, including LSP-specific serialization, and that is even kept mostly separate from the `LspRouter` which delegates all LSP responsibility elsewhere.

Next tasks: re-connect the editor integration via `lsp.js`.


dotnet-interactive.Tests/LspSerializationContractTests.cs
```diff
@@ -0,0 +1,71 @@
...
+
+        public LspSerializationContractTests()
+        {
+            _configuration = new Configuration()
+                .UsingExtension("json");
+
+            _configuration = _configuration.SetInteractive(true);
+        }
+
+        private void Descent<T>(T expected, JsonSerializer serializer, [CallerMemberName] string testName = null, [CallerFilePath] string filePath = null)
```
brettfo, line 28: I needed the exact opposite of an Assent test, and since I didn't want to mess with escaping double quotes inline in the code, I cheated and did what (I assume) Assent does, but backwards by finding an appropriate file on disk that will be deserialized and compared to a known object.  This method doesn't, however, show the diff like Assent does.

dotnet-interactive/Lsp/LspHandlerExtensions.cs
```diff
@@ -0,0 +1,92 @@
...
+        /// <typeparam name="TRequestCommand">The command type that will be submitted to <see cref="KernelBase.SendAsync(IKernelCommand, System.Threading.CancellationToken)"/>.</typeparam>
+        /// <typeparam name="TResultEvent">The expected resultant event produced by the kernel.  May not be returned if the specified kernel doesn't support the given LSP method.</typeparam>
+        /// <typeparam name="TResult">The type that will be serialized back into <see cref="JObject"/></typeparam>
+        /// <param name="kernel">The kernel that the <typeparamref name="TRequestCommand"/> is submitted to.</param>
+        /// <param name="request">The JSON-parsed request.</param>
+        /// <param name="commandCtor">Function to construct an appropriate <typeparamref name="TRequestCommand"/> from the <typeparamref name="TRequest"/>.</param>
+        /// <param name="resultDtor">Function to deconstruct the <typeparamref name="TResultEvent"/> into the final <typeparamref name="TResult"/>.</param>
+        private static async Task<JObject> GetLanguageServiceResultOrDefault<TRequest, TRequestCommand, TResultEvent, TResult>(
```
brettfo, line 53: Due to XML escaping, this looks frightening, but the doc comment for this method renders as:

```
Completes the path from `JObject` -> `TRequest` -> `TRequestCommand` -> `KernelBase.SendAsync(IKernelCommand, System.Threading.CancellationToken)` -> `TResultEvent` -> `TResult` -> `JObject`.
```

Microsoft.DotNet.Interactive.Tests/LanguageServiceTests.cs
```diff
@@ -0,0 +1,105 @@
...
+        private Task<IKernelCommandResult> SendHoverRequest(KernelBase kernel, string code, int line, int character)
+        {
+            var command = new RequestHoverTextCommand(RequestHoverTextCommand.MakeDataUriFromContents(code), new Position(line, character));
+            return kernel.SendAsync(command);
+        }
+
+        [Theory]
+        [InlineData(Language.FSharp)]
+        [InlineData(Language.PowerShell)]
+        public async Task HoverOnUnsupportedLanguageServiceReturnsNothing(Language language)
```
colombod, line 30: WE should use the fake kernel here to ensure this behaviour is retained even after  Powershell and Fsharp gets the api

colombod, line 30: can we align the naming pattern to the other tests like

```lsp_textDocument_hover_returns_expected_result```

File: dotnet-interactive/Lsp/LspSerializer.cs [hunk 4]
```diff
@@ -0,0 +1,54 @@
...
+
+        static LspSerializer()
+        {
+            JsonSerializerSettings = CreateDefaultSettings();
+            JsonSerializer = JsonSerializer.Create(JsonSerializerSettings);
+        }
+
+        private static JsonSerializerSettings CreateDefaultSettings()
+        {
+            var settings = new JsonSerializerSettings
```
colombod, line 25: can we store this value?

dotnet-interactive/Lsp/LspSerializer.cs
```diff
@@ -0,0 +1,54 @@
...
+        {
+            // TODO: is there a faster way to do this?
+            var foundError = false;
+            var settings = CreateDefaultSettings();
+            settings.Error = (sender, e) =>
+            {
+                foundError = true;
+                e.ErrorContext.Handled = true;
+            };
+            var serializer = JsonSerializer.Create(settings);
```
colombod, line 48: why building this if you have the static properties?

brettfo: I need to add the `.Error` handler, but only when this `TryToObject<T>()` method is called, so I can't modify the static value or else it'll pollute all other (de)serialization calls.  I couldn't think of another way to have local-only error swallowing without `catch (JsonSerializationException)`.

dotnet-interactive/Lsp/LspSerializer.cs
brettfo: (See comment below about the `.Error` handler.)

Microsoft.DotNet.Interactive.Tests/LanguageServiceTests.cs
brettfo: Will do.

brettfo: Will do.

dotnet-interactive.Tests/LspSerializationContractTests.cs
jonsequitur: Ok... but the opposite of Assent is Dissent.

---
In reply to: [409881202](https://github.com/dotnet/interactive/pull/348#discussion_r409881202) [](ancestors = 409881202)

Microsoft.DotNet.Interactive.Tests/LanguageServiceTests.cs
```diff
@@ -0,0 +1,105 @@
...
+                .ContainSingle<LanguageServiceHoverResponseProduced>()
+                .Which
+                .Contents
+                .Value
+                .Should()
+                .Contain(expected);
+        }
+
+        [Theory]
+        [InlineData(Language.CSharp, "var x = 1; // hovering in a comment", 0, 20)]
```
jonsequitur, line 62: >0, 20 [](start = 76, length = 5)

For this kind of thing you can use MarkupTestFile, which is more readable and maintainable than specifying the ints separately.

dotnet-interactive/Lsp/LspSerializer.cs
jonsequitur: Are you trying to implement partial success? Why are errors occurring?

---
In reply to: [409909923](https://github.com/dotnet/interactive/pull/348#discussion_r409909923) [](ancestors = 409909923)

dotnet-interactive.Tests/HttpApiTests.cs
```diff
@@ -198,53 +202,71 @@ public async Task When_subkernel_does_not_exist_then_it_returns_404()
...
-}
-");
-            var response = await GetServer(language).HttpClient.PostJsonAsync("/lsp/textDocument/hover", request);
+            var request = new
+            {
+                SomeField = 1,
+                SomeOtherField = "test",
+            };
+            var response = await GetServer(language).HttpClient.PostObjectAsJsonAsync("/lsp/textDocument/hover", request);
             await response.ShouldSucceed();
```
jonsequitur, line 232: >   await response.ShouldSucceed(); [](start = 9, length = 34)

Why not return a 400?

dotnet-interactive.Tests/HttpApiTests.cs
```diff
@@ -198,53 +202,71 @@ public async Task When_subkernel_does_not_exist_then_it_returns_404()
...
+        [Theory]
+        [InlineData(Language.CSharp)]
+        [InlineData(Language.FSharp)]
+        [InlineData(Language.PowerShell)]
+        public async Task unimplemented_lsp_methods_return_no_result(Language language)
         {
-            // language kernels that implement LSP handling, but not for the specified method
-            var response = await GetServer(language).HttpClient.PostJsonAsync($"/lsp/not/a/method", new JObject());
+            var response = await GetServer(language).HttpClient.PostObjectAsJsonAsync($"/lsp/not/a/method", new object()); // payload doesn't matter
             await response.ShouldSucceed();
```
jonsequitur, line 255: > await response.ShouldSucceed(); [](start = 11, length = 32)

Maybe a 404?

dotnet-interactive.Tests/LspSerializationContractTests.cs
brettfo: Good call, I was interpreting this as "Ascent".  Will fix.

dotnet-interactive.Tests/HttpApiTests.cs
brettfo: The regular stdio LSP protocol dictates that a value must be returned and eventually the `null` returned could (should?) be more descriptive error info that might be displayed to the user (e.g., when you F12 on a comment in C# you get a dialog "Cannot navigate to the symbol under the caret.")  It seemed more appropriate to return 200/OK along with an actionable payload, even if it's an error case.

_However_, there's no spec for LSP over HTTP, so we're blazing a trail here.  Ultimately I don't feel too strongly about it and I'll yield to you on this.

dotnet-interactive/Lsp/LspSerializer.cs
brettfo: I'm trying to prevent leaking internal implementation details if the user submits bad or wrong JSON.  If they payload isn't correct, the deserialization will throw and that seemed sloppy, but ultimately I don't feel too strongly about it.

Microsoft.DotNet.Interactive.Tests/LanguageServiceTests.cs
brettfo: Thanks, I've updated most of the tests to do this.  There are some out-of-bounds tests that still must manually specify a line/character, though.

dotnet-interactive.Tests/HttpApiTests.cs
brettfo: See my other comment about handling HTTP LSP issues in general.

dotnet-interactive/Lsp/LspSerializer.cs
colombod: I am talking about the fact that we keep rebuilding the serialization settings and the serializer using the exact same code that the type initializer is using to create the static property JsonSerializer  and the static field JsonSerializerSettings. should we reuse those objects isntead of keep on building them at each serialization?

brettfo: There are two things happening here.  The static settings (`.JsonSerializerSettings`) and static serializer (`.JsonSerializer`) are populated once in the static constructor, and they're regularly reused and never recreated.

I think the part you're concerned about is the `.TryToObject<T>()` method, and that's because we need to know, only in this function call, if there was a deserialization error we need to mark the error as handled to prevent a `JsonSerializationException` from being thrown so that we can instead return a `bool` value.  I can't set the `.Error` property on the static settings object because then any subsequent call to the regular static serializer/settings will have the `.Error` handler set, and that's not what's wanted.

The alternative would be to use the static values and instead make the function body look like this:

``` C#
public static bool TryToObject<T>(this JToken token, out T result)
{
    try
    {
        result = token.ToObject<T>(JsonSerializer);
        return true;
    }
    catch (JsonSerializationException)
    {
        result = default(T);
        return false;
    }
}
```

I opted away from this because throwing/catching exceptions is expensive and @jonsequitur was concerned about serialization/deserialization perf.

dotnet-interactive.Tests/HttpApiTests.cs
jonsequitur: There are two clients we're targeting right now:

1) A custom web client for use with our Jupyter shims, which we'll write. This isn't strict LSP, and we can potentially even use the Try .NET JavaScript code for this. Since a malformed request is a programming error on our part in any case, a 400 is valuable and appropriate. 

2) VS Code, which doesn't talk over HTTP.

I think following LSP strictly is useful for the second case only.

Thoughts?

Microsoft.DotNet.Interactive/Commands/RequestHoverTextCommand.cs
```diff
@@ -0,0 +1,27 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Text;
+using Microsoft.DotNet.Interactive.LanguageService;
+
+namespace Microsoft.DotNet.Interactive.Commands
+{
+    public class RequestHoverTextCommand : LanguageServiceCommandBase
```
jonsequitur, line 10: You can leave `Command` off the type name.

Microsoft.DotNet.Interactive/Events/LanguageServiceHoverResponseProduced.cs
```diff
@@ -0,0 +1,21 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Microsoft.DotNet.Interactive.Commands;
+using Microsoft.DotNet.Interactive.LanguageService;
+
+namespace Microsoft.DotNet.Interactive.Events
+{
+    public class LanguageServiceHoverResponseProduced : KernelEventBase
```
jonsequitur, line 9: I think calling this `HoverTextProduced` is simpler and at least as informative.

Microsoft.DotNet.Interactive/Extensions/SerializerExtensions.cs
```diff
@@ -0,0 +1,26 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.IO;
+using Newtonsoft.Json;
+
+namespace Microsoft.DotNet.Interactive.Extensions
+{
+    public static class SerializerExtensions
```
jonsequitur, line 9: These should not be public.

Microsoft.DotNet.Interactive/Extensions/SerializerExtensions.cs
```diff
@@ -0,0 +1,26 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.IO;
+using Newtonsoft.Json;
+
+namespace Microsoft.DotNet.Interactive.Extensions
```
jonsequitur, line 7: The `Extensions` namespace refers to kernel extensions, not extension methods.

Serialization stuff might make more sense in the `Server` namespace.

Microsoft.DotNet.Interactive/LanguageService/IKernelLanguageService.cs
```diff
@@ -1,11 +1,13 @@
 ﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System.Threading.Tasks;
+using Microsoft.DotNet.Interactive.Commands;
+
 namespace Microsoft.DotNet.Interactive.LanguageService
 {
-    public class HoverParams
+    public interface IKernelLanguageService<TCommand> where TCommand: LanguageServiceCommandBase
```
jonsequitur, line 9: I think it makes sense to generalize this to all commands, e.g. `IKernelCommandHandler<T>`. There's nothing special about language services.

Microsoft.DotNet.Interactive/LanguageService/Position.cs
```diff
@@ -5,7 +5,20 @@ namespace Microsoft.DotNet.Interactive.LanguageService
 {
     public struct Position
```
jonsequitur, line 6: Should this go into the `Parsing` namespace?

Microsoft.DotNet.Interactive/LanguageService/Range.cs
```diff
@@ -5,7 +5,13 @@ namespace Microsoft.DotNet.Interactive.LanguageService
 {
     public class Range
```
jonsequitur, line 6: Any reason not to use `TextSpan`? The new parsing code uses this and a few other types from `Microsoft.CodeAnalysis.Text`.

dotnet-interactive.Tests/LspSerializationContractTests.cs
```diff
@@ -0,0 +1,71 @@
...
+
+        public LspSerializationContractTests()
+        {
+            _configuration = new Configuration()
+                .UsingExtension("json");
+
+            _configuration = _configuration.SetInteractive(true);
+        }
+
+        private void Dissent<T>(T expected, JsonSerializer serializer, [CallerMemberName] string testName = null, [CallerFilePath] string filePath = null)
```
jonsequitur, line 28: After looking at it more closely, I think this makes the Assent tests confusing (since some of the `.approved` files no longer involve an "approval" process), and it makes the tests where this is used harder to understand (because the file contents are important to understanding the test, so why put them in a separate file?).

Let's just inline the JSON into the tests for these. If escaping is a pain you can use anonymous objects e.g.

```csharp
var json = new {
    contents = new {
        kind = "markdown",
        value=  "content",
        range= new {
            start = new { line = 1, character= 2 },
            end = new { line= 3, character= 4 }
    }.ToJson();
```

Microsoft.DotNet.Interactive/LanguageService/Range.cs
jonsequitur, line 6: What's the relationship of this to `TextSpan` (which we use a bunch in the `Parsing` namespace)? And how should we think about the split between types (like `HoverParam`) that are in `Microsoft.DotNet.Interactive.LanguageService` versus `Microsoft.DotNet.Interactive.App.Lsp`?

dotnet-interactive/Lsp/LspHandlerExtensions.cs
```diff
@@ -0,0 +1,92 @@
...
+            this KernelBase kernel,
+            JObject request,
+            Func<TRequest, TRequestCommand> commandCtor,
+            Func<TResultEvent, TResult> resultDtor)
+            where TRequestCommand : IKernelCommand
+        {
+            // JObject -> TRequest
+            if (!request.TryToObject<TRequest>(out var requestParams))
+            {
+                // couldn't deserialize, shouldn't be fatal
```
jonsequitur, line 63: This looks like a programmer error, not a user error, so why not just throw?

dotnet-interactive/Lsp/LspSerializer.cs
brettfo: As per Jon's other comments and what we discussed this morning, this will change to throw/404/500/etc. in case of a malformed request, because this is a situation where the person doing the wrong thing is either us or somebody calling our API and _not_ the end user.

Microsoft.DotNet.Interactive/LanguageService/Range.cs
brettfo: While the types are similar, `TextSpan` is a Roslyn-specific type and I didn't want F# and PowerShell to be forced to use those types as well as force them to take a (indirect) dependency on Roslyn.  I also didn't want to leak Roslyn-specific stuff out into the language service layer which shouldn't know or care about Roslyn.

jonsequitur: The submission parser is using some types from `Microsoft.CodeAnalysis.Text` at the upper layer because they're robust, fast, and have a lot of functionality. We'll need *some* common abstraction at the polyglot syntax level so why not use these?

jonsequitur: That said, whatever that common abstraction is, let's consolidate it in `Microsoft.DotNet.Interactive.Parsing` and not up in the `dotnet-interactive` tool code.

brettfo: In that case I'll fall back to using `TextSpan` and only convert to/from LSP `Range` at the appropriate serialization layers.

Microsoft.DotNet.Interactive/KernelBase.cs
```diff
@@ -386,11 +380,37 @@ public void RegisterForDisposal(IDisposable disposable)
...
+            {
+                command.Handler = handlerGenerator(handlingKernel);
+            }
+            else
+            {
+                // default behavior is to return no result
+                command.Handler = (command, context) =>
+                {
+                    context.Publish(new LanguageServiceNoResultProduced(command));
+                    return Task.CompletedTask;
```
jonsequitur, line 409: Do we need an event that says we didn't do anything? Why not just ... not do anything?

Microsoft.DotNet.Interactive/LanguageService/ContextExtensions.cs
```diff
@@ -0,0 +1,23 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Microsoft.DotNet.Interactive.Commands;
+using Microsoft.DotNet.Interactive.Events;
+
+namespace Microsoft.DotNet.Interactive.LanguageService
+{
+    public static class ContextExtensions
```
jonsequitur, line 9: >ContextExtensions [](start = 24, length = 17)

`KernelInvocationContextExtensions`

Microsoft.DotNet.Interactive/LanguageService/IKernelCommandHandler.cs
```diff
@@ -1,11 +1,13 @@
...
+
 namespace Microsoft.DotNet.Interactive.LanguageService
 {
-    public class HoverParams
+    public interface IKernelCommandHandler<TCommand> where TCommand: IKernelCommand
     {
-        public TextDocument TextDocument { get; set; }
-        public Position Position { get; set; }
+        Task Handle(TCommand command, KernelInvocationContext context);
     }
```
jonsequitur, line 12: I suppose `HandleAsync` is the recommended naming.

Microsoft.DotNet.Interactive/LanguageService/MarkupKind.cs
```diff
@@ -1,13 +1,8 @@
...
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
-using Newtonsoft.Json;
-using Newtonsoft.Json.Converters;
-using Newtonsoft.Json.Serialization;
-
 namespace Microsoft.DotNet.Interactive.LanguageService
 {
-    [JsonConverter(typeof(StringEnumConverter), converterParameters: typeof(CamelCaseNamingStrategy))]
     public enum MarkupKind
```
jonsequitur, line 6: >MarkupKind [](start = 16, length = 10)

This doesn't belong at the domain layer, as it seems like an artifact of LSP. We should favor behaviors/types over enums here.

brettfo: What about `abstract class ContentBase` and `class PlainTextContent : ContentBase`/`class MarkdownContent : ContentBase`?  I'm open to naming suggestions if you don't like those.  Then the `MarkupKind` enum will exist only at the LSP serialization layer.

Microsoft.DotNet.Interactive.CSharp/LanguageServiceExtensions.cs
```diff
@@ -23,7 +22,7 @@ public static LinePositionSpan SubtractLineOffset(this LinePositionSpan linePosS
                 linePosSpan.End.SubtractLineOffset(offset));
         }
 
-        public static MarkupContent ToMarkupContent(this QuickInfoItem info)
+        public static string ToMarkdownString(this QuickInfoItem info)
```
jonsequitur, line 25: Should the return type not be `MarkdownContent` instead of `string`?

brettfo: No.  `MarkdownContent` only exists at the LSP serialization layer.  This extension method feeds into `HoverMarkdownProduced` (vs. `HoverPlainTextProduced`) so a regular `string` is appropriate here.  See the modified `HoverTextProduced` type.

jonsequitur: I think it's useful to have a `MarkdownContent` type in the core domain. The goal of avoiding strings there is so that it's clear what kind of string it is, in any context, not just in the context of an event. 

I wasn't sure if that applies in this part of the code.

brettfo: Having an `abstract ContentBase` caused a bunch of problems with event (de)serialization; there was no way to determine the type.  E.g., consider:

``` C#
abstract class ContentBase
{
    string Value;
}

class PlainTextContent : ContentBase { }

class MarkdownContent : ContentBase { }

class HoverTextProduced
{
    ContentBase Content;
}
```

Then when deserializing the `HoverTextProduced` event from the string `..."content": {"value": "is this plaintext or markdown?"},...`, how is the `content` field to be interpreted?  Without making the serializer much more complicated (1) so it can handle the property that's an abstract class, and (2) so it also carries type information with it (e.g., `"content": {"type":"MarkdownContent", "value": "..."}`).

This felt much cleaner and more in line with the existing (de)serialization work.

brettfo: Yes, a core markdown/plaintext content type would be useful, but is outside the scope of this PR, and given that upcoming work is blocked on this, I thought it best to move forward.  We can revisit/add a backlog item to figure out a markdown editor/builder.

PR #357, opened 2020-04-24
Update first time experience environment variable name.

suhsteve:


suhsteve:
@colombod @jonsequitur if you can take a look.

jonsequitur:
This library is also consumed by `dotnet-try`, and there's some discussion of further generalizing it, so ideally we should make this parameterizable.

PR #373, opened 2020-05-04
improve hover request/result

brettfo:
1. Only return one event type for hover requests and allow it to contain multiple (mimeType/content) pairs.  This mimics `ReturnValueProduced` and allows the different clients to display whatever content type they choose.
2. Move the base64 encoding requirement for hover code contents to be at the LSP layer.  This way the `RequestHoverText` command simply supplies the current state of the code, just like `RequestCompletion`.


Microsoft.DotNet.Interactive/Events/HoverTextProduced.cs
```diff
@@ -1,19 +1,26 @@
...
+        public IReadOnlyCollection<FormattedValue> Content { get; }
         public LinePositionSpan? Range { get; set; }
 
-        public HoverTextProduced(IKernelCommand command, string content, LinePositionSpan? range = null)
+        public HoverTextProduced(IKernelCommand command, IReadOnlyCollection<FormattedValue> content, LinePositionSpan? range = null)
             : base(command)
         {
+            if (content.Count == 0)
+            {
+                throw new InvalidOperationException("At least one content required.");
```
jonsequitur, line 21: This should be `ArgumentException`.

dotnet-interactive/Lsp/HoverResponse.cs
```diff
@@ -19,15 +27,30 @@ public HoverResponse(MarkupContent contents, Range range = null)
 
         public static HoverResponse FromHoverEvent(HoverTextProduced responseEvent)
         {
-            var markupContent = responseEvent switch
+            var preferredContent = responseEvent.Content.OrderBy(c => PriorityFromMimeType(c.MimeType)).First();
+            return new HoverResponse(new MarkupContent(MarkupKindFromMimeType(preferredContent.MimeType), preferredContent.Value));
+        }
+
+        private static int PriorityFromMimeType(string mimeType)
```
jonsequitur, line 34: This seems related to content negotiation. Would a more generalized approach make sense in the formatting layer?

brettfo: Not here.  This is the LSP layer where _only_ plaintext or markdown is allowed, not both, and most LSP clients prefer markdown, but given that we plan on removing the LSP shim anyway, this is merely a stopgap until an hour from now when I properly remove LSP.

PR #375, opened 2020-05-04
refactoring repo file structure

colombod:


eng/build.ps1
```diff
@@ -19,8 +19,8 @@ try {
 
     # directly invoke npm tests
     # if (($null -ne $args) -and ($args.Contains("-test") -or $args.Contains("-t"))) {
-    #     TestUsingNPM "$PSScriptRoot\..\Microsoft.DotNet.Try.Client"
-    #     TestUsingNPM "$PSScriptRoot\..\Microsoft.DotNet.Try.js"
+    #     TestUsingNPM "$PSScriptRoot\..\src\Microsoft.DotNet.Try.Client"
+    #     TestUsingNPM "$PSScriptRoot\..\src\Microsoft.DotNet.Try.js"
```
jonsequitur, line 23: Do we actually use these files in some way or is this vestigial?

eng/build.sh
```diff
@@ -40,6 +40,6 @@ function TestUsingNPM() {
 
 # directly invoke npm tests
 # if [[ "$run_tests" == "true" ]]; then
-#     TestUsingNPM "$scriptroot/../Microsoft.DotNet.Try.Client"
-#     TestUsingNPM "$scriptroot/../Microsoft.DotNet.Try.js"
+#     TestUsingNPM "$scriptroot/../src/Microsoft.DotNet.Try.Client"
+#     TestUsingNPM "$scriptroot/../src/Microsoft.DotNet.Try.js"
```
jonsequitur, line 44: Ditto.

src/Microsoft.DotNet.Interactive/Events/HoverMarkdownProduced.cs
```diff
@@ -0,0 +1,16 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
```
jonsequitur, line 1: Looks like you need a rebase.

eng/build.ps1
colombod: checking with @brettfo  as i think we moved this part to yaml tasks

src/dotnet-interactive-vscode/src/tests/unit/testClientAdapter.ts
```diff
@@ -0,0 +1,25 @@
+import { Observable } from "rxjs";
```
colombod, line 1: @brettfo is this right? I had this after updating master with your latest merge.

brettfo: After my last PR `testClientAdapter.ts` looks like this:

``` typescript
import { Observable } from "rxjs";
import { EventEnvelope } from "../../events";
import { Command } from "../../commands";

// Replays all events given to it
export class TestClientTransport {
    constructor(readonly fakedEvents: { [key: string]: {eventType: string, event: any}[] }) {
    }

    submitCommand(commandType: string, command: Command): Observable<EventEnvelope> {
        let eventsToReturn = this.fakedEvents[commandType];
        return new Observable<EventEnvelope>(subscriber => {
            for (let event of eventsToReturn) {
                subscriber.next({
                    eventType: event.eventType,
                    event: event.event,
                    cause: {
                        token: ''
                    }
                });
            }

            subscriber.complete();
        });
    }
}
```

eng/build.ps1
brettfo: `npm test` (or similar) is invoked via YAML in the following directories:

- `Microsoft.DotNet.Interactive.Js/`
- `dotnet-interactive-vscode/`

Otherwise those comments are from the `/try` repo and can be deleted.

src/dotnet-interactive-vscode/src/tests/unit/testClientAdapter.ts
brettfo: And the location looks correct, too.

colombod: @brettfo  so the file is consistent with master.

eng/build.ps1
colombod: removed them

eng/build.sh
colombod: removed

src/Microsoft.DotNet.Interactive/Events/HoverMarkdownProduced.cs
colombod: done

PR #378, opened 2020-05-05
add tool to generate TypeScript interfaces for all commands and events

brettfo:
Includes generated `contracts.ts` in both the HTTP API and VSCode notebook extension.  Also includes unit tests to verify the contents of the `contracts.ts` files.

Upcoming work: Update VSCode extension to use new `KernelTransport` interface.


src/interface-generator/GeneratorExtensions.cs
```diff
@@ -0,0 +1,38 @@
...
+            if (baseType == typeof(object) || baseType == typeof(ValueType))
+            {
+                // don't navigate up to these types
+                return null;
+            }
+
+            return baseType;
+        }
+
+        public static string CamlCase(this string value)
```
jonsequitur, line 23:
```suggestion
        public static string CamelCase(this string value)
```

src/interface-generator/InterfaceGenerator.cs
```diff
@@ -0,0 +1,215 @@
...
+
+        private static HashSet<Type> AlwaysEmitTypes = new HashSet<Type>()
+        {
+            typeof(KernelCommandBase),
+            typeof(KernelEventBase)
+        };
+
+        private static HashSet<string> OptionalFields = new HashSet<string>()
+        {
+            "KernelCommandBase.TargetKernelName",
```
jonsequitur, line 41: I'd suggest using `nameof` for these.

PR #382, opened 2020-05-07
Log directive update

jonsequitur:


src/Microsoft.DotNet.Interactive/Events/DiagnosticLogEntryProduced.cs
```diff
@@ -5,9 +5,9 @@
 
 namespace Microsoft.DotNet.Interactive.Events
 {
-    public class DiagnosticLogEventProduced : DiagnosticEventBase
+    public class DiagnosticLogEntryProduced : DiagnosticEventBase
```
colombod, line 8: since this event is renamed you must run the contract generator to emit the typescript files again

PR #386, opened 2020-05-07
refactor vscode to use `KernelTransport` interface

brettfo:
This also has the benefit of removing the dependency on `rxjs` which takes the size of the VSIX from 780k to 18k.


src/dotnet-interactive-vscode/src/tests/unit/notebook.test.ts
```diff
@@ -34,46 +36,51 @@ describe('Notebook tests', () => {
...
         let code = `
 Console.WriteLine(1);
 Console.WriteLine(1);
 Console.WriteLine(1);
 `;
-        let clientMapper = new ClientMapper(() => new TestClientTransport({
+        let clientMapper = new ClientMapper(() => new TestKernelTransport({
             'SubmitCode': [
                 {
                     eventType: 'CodeSubmissionReceived',
```
colombod, line 72: ditto

src/dotnet-interactive-vscode/src/tests/unit/notebook.test.ts
```diff
@@ -112,29 +121,36 @@ Console.WriteLine(1);
...
-            },
-            {
-                outputKind: CellOutputKind.Text,
-                text: '2\r\n'
-            },
-            {
-                outputKind: CellOutputKind.Text,
-                text: '3\r\n'
+        let outputs: Array<CellOutput> = [];
+        await client.execute('csharp', code, cellOutput => {
```
colombod, line 136: usually we follow the convention `(code:string, targetKenelName: string)` is this our type or vscode interace?

brettfo: This is our own interface; I'll reverse it to match.

PR #396, opened 2020-05-09
ensure appropriate version of `dotnet-interactive` is present

brettfo:
`extension.ts` will now ensure that an appropriate version of `dotnet` is installed, or else it will acquire one for the user.  If one couldn't be acquired, it is a fatal error and nothing can be done.

A similar check is also done for `dotnet-interactive`, but it currently only continues if a valid version is present and errors otherwise.  An upcoming PR will be to acquire a local tool version of it if necessary.  If the appropriate version is not present, an error is shown in the bottom corner or the window:
image

The minimum good version of `dotnet-interactive` is the one with the recent hover text updates from a few days ago.  That version has not yet been published to NuGet.org, so when we do publish again, it should be at least that version.  A workaround is that all locally-built arcade versions of the tool are allowed (`*-dev`).

Also included are some other minor improvements like not returning from `execute()` until either `CommandHandled` or `CommandFailed` has returned.  This makes the notebook UI in VS Code much nicer.

N.b., the changes to `interactiveClient.ts` are mostly whitespace, so disabling that in the GitHub view is recommended for that file.

### DIFF
--- a/src/Microsoft.DotNet.Interactive.Js/src/dotnet-interactive/contracts.ts
+++ b/src/Microsoft.DotNet.Interactive.Js/src/dotnet-interactive/contracts.ts
@@ -240,11 +240,14 @@ export interface KernelEventEnvelopeObserver {
     (eventEnvelope: KernelEventEnvelope): void;
 }
 
-export interface DisposableSubscription {
+export interface Disposable {
     dispose(): void;
 }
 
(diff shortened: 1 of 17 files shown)


brettfo:
@colombod I've added a `dispose()` method to `KernelTransport` so that the process can be killed when VS Code disposes the notebook.

PR #401, opened 2020-05-11
simplify HTTP options

colombod:
Aligning api to comply with issue #444

### DIFF
--- a/src/dotnet-interactive.Tests/CommandLine/CommandLineParserTests.cs
+++ b/src/dotnet-interactive.Tests/CommandLine/CommandLineParserTests.cs
@@ -219,7 +219,7 @@ namespace Microsoft.DotNet.Interactive.App.Tests.CommandLine
         }
 
         [Fact]
-        public async Task jupyter_command_does_enable_http_api_by_default()
+        public async Task jupyter_command_enables_http_api_by_default()
         {
             await  _parser.InvokeAsync($"jupyter {_connectionFile}");
 
--- a/src/dotnet-interactive.Tests/CommandLine/CommandLineParserTests.cs
+++ b/src/dotnet-interactive.Tests/CommandLine/CommandLineParserTests.cs
@@ -47,6 +47,11 @@ namespace Microsoft.DotNet.Interactive.App.Tests.CommandLine
                     _startOptions = startupOptions;
                     return Task.FromResult(1);
                 },
+                startHttp: (startupOptions, console, startServer, context) =>
+                {
+                    _startOptions = startupOptions;
+                    return Task.FromResult(1);
+                },
             telemetry: new FakeTelemetry(),
                 firstTimeUseNoticeSentinel: new NopFirstTimeUseNoticeSentinel());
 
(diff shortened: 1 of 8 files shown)


src/dotnet-interactive.Tests/CommandLine/CommandLineParserTests.cs
```diff
@@ -192,32 +219,46 @@ public void jupyter_parses_connection_file_path()
         }
 
         [Fact]
-        public async Task jupyter_does_not_enable_http_api_by_default()
+        public async Task jupyter_command_does_enable_http_api_by_default()
```
jonsequitur, line 222:
```suggestion
        public async Task jupyter_command_enables_http_api_by_default()
```

src/dotnet-interactive.Tests/CommandLine/CommandLineParserTests.cs
```diff
@@ -192,32 +219,46 @@ public void jupyter_parses_connection_file_path()
...
             var result = _parser.Parse($"jupyter --default-kernel bsharp {Path.GetTempFileName()}");
             var binder = new ModelBinder<JupyterOptions>();
             var options = (JupyterOptions)binder.CreateInstance(new BindingContext(result));
+
             options.DefaultKernel.Should().Be("bsharp");
         }
 
         [Fact]
-        public async Task jupyter_returns_error_if_connection_file_path_does_not_exits()
+        public async Task jupyter_command_returns_error_if_connection_file_path_does_not_exits()
```
jonsequitur, line 261:
```suggestion
        public async Task jupyter_command_returns_error_if_connection_file_path_does_not_exist()
```

src/dotnet-interactive/CommandLine/CommandLineParser.cs
```diff
@@ -291,30 +265,51 @@ Command StdIO()
...
-                        }
-
-                        return startStdIO(
-                            startupOptions,
-                            CreateKernel(options.DefaultKernel, new BrowserFrontendEnvironment(), startupOptions, null),
-                            console);
-                    });
+                    (startupOptions, options, console, context) => startStdIO(
+                        startupOptions,
+                        CreateKernel(options.DefaultKernel, new BrowserFrontendEnvironment(), startupOptions, null),
```
jonsequitur, line 274: Should we inject the frontend environment?

src/dotnet-interactive/CommandLine/readme.md
```diff
@@ -9,21 +9,21 @@ A number of APIs are available over HTTP in `dotnet-interactive`. You can enable
 dotnet interactive http                                        | Enabled on auto-selected port
 dotnet interactive http --http-port *                          | Enabled on auto-selected port
 dotnet interactive http --http-port 1234                       | Enabled on port 1234
-dotnet interactive http --http-port-range 8001-9000            | Enabled on a port between 8001 and 9000
+dotnet interactive http --http-port-range 8001-9000            | UNSUPPORTED
```
jonsequitur, line 12: Not sure if you plan was to delete the ones that are unsupported, but I don't think documenting them is necessary.

colombod: removing them

src/dotnet-interactive/CommandLine/CommandLineParser.cs
colombod: there is no need here to inject, we need to provide one to the create kernel function. That is the composition part for stdio server

PR #406, opened 2020-05-11
acquire `dotnet-interactive` as a local tool if necessary

brettfo:
This feels complicated, but it's not too bad.

In `extension.ts` the command `dotnet --version` is run.  If that errors out or if the version is less than the setting `dotnet-interactive.minimumDotNetSdkVersion` (specified in `package.json`), then a version of .NET will be acquired for you.  This is all using the `ms-dotnettools.vscode-dotnet-runtime` extension.

Next, dotnet is run with `tool run dotnet-interactive -- --version`.  If that fails (because the local tool isn't installed), or if the version is less than the setting `dotnet-interactive.minimumInteractiveToolVersion`, then the local tool version will be acquired for you using the package feed `dotnet-interactive.interactiveToolSource`.

All of those settings can be overridden by a user.

Finally, the args passed to dotnet to start an interactive session can be overrided by the `dotnet-interactive.kernelTransportArgsOverride` setting.  If not specified or empty, the default args are `['tool', 'run', 'dotnet-interactive', '--', 'stdio']`.  To run a local dev build you could set it to `['path/to/Microsoft.DotNet.Interactive.App.dll', '--', 'stdio']` (or you could do something even more crazy).

### DIFF
--- a/src/dotnet-interactive-vscode/package.json
+++ b/src/dotnet-interactive-vscode/package.json
@@ -22,7 +22,9 @@
     "Other"
   ],
   "activationEvents": [
-    "onNotebookEditor:dotnet-interactive"
+    "onNotebookEditor:dotnet-interactive",
+    "onCommand:dotnet-interactive.acquire",
+    "onCommand:dotnet-interactive.reportInteractiveVersion"
   ],
   "main": "./out/extension.js",
   "extensionDependencies": [
@@ -45,6 +47,14 @@
     ],
     "commands": [
       {
+        "command": "dotnet-interactive.acquire",
+        "title": "Install .NET Interactive"
+      },
+      {
+        "command": "dotnet-interactive.reportInteractiveVersion",
+        "title": "Report installed version for .NET Interactive"
+      },
+      {
         "command": "dotnet-interactive.exportAsJupyterNotebook",
         "title": "DotNet Interactive: Export as Jupyter Notebook"
       }
@@ -64,18 +74,22 @@
   },
   "devDependencies": {
     "@types/chai": "4.2.11",
+    "@types/chai-fs": "2.0.2",
     "@types/glob": "7.1.1",
     "@types/mocha": "7.0.2",
     "@types/node": "13.11.0",
+    "@types/tmp": "0.2.0",
     "@types/vscode": "1.44.0",
     "@typescript-eslint/eslint-plugin": "2.26.0",
     "@typescript-eslint/parser": "2.26.0",
     "chai": "4.2.0",
+    "chai-fs": "2.0.0",
     "eslint": "6.8.0",
     "glob": "7.1.6",
     "mocha": "7.1.1",
     "mocha-multi-reporters": "1.1.7",
     "mocha-trx-reporter": "3.2.4",
+    "tmp": "0.2.1",
     "typescript": "3.8.3",
     "vsce": "1.75.0",
     "vscode-test": "1.3.0"
--- a/src/dotnet-interactive-vscode/package.json
+++ b/src/dotnet-interactive-vscode/package.json
@@ -48,6 +48,11 @@
     "configuration": {
       "title": ".NET Interactive Notebook",
       "properties": {
+        "dotnet-interactive.kernelTransportArgsOverride": {
+          "type": "array",
+          "default": [],
+          "description": "Override command line arguments passed to start an interactive session."
+        },
         "dotnet-interactive.interactiveToolSource": {
           "type": "string",
           "default": "https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-tools/nuget/v3/index.json",
--- a/src/dotnet-interactive-vscode/package.json
+++ b/src/dotnet-interactive-vscode/package.json
@@ -45,6 +45,26 @@
         ]
       }
     ],
+    "configuration": {
+      "title": ".NET Interactive Notebook",
+      "properties": {
+        "dotnet-interactive.interactiveToolSource": {
+          "type": "string",
+          "default": "https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-tools/nuget/v3/index.json",
+          "description": "The NuGet source to use when acquiring the .NET Interactive tool."
+        },
+        "dotnet-interactive.minimumDotNetSdkVersion": {
+          "type": "string",
+          "default": "3.1",
+          "description": "The minimum required version of the .NET SDK."
+        },
+        "dotnet-interactive.minimumInteractiveToolVersion": {
+          "type": "string",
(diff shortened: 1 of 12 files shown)


src/dotnet-interactive-vscode/src/acquisition.ts
```diff
@@ -0,0 +1,129 @@
...
+        const { code } = await processExitCodeAndOutput(dotnetPath, ['new', 'tool-manifest', '--output', globalStoragePath]);
+        if (code !== 0) {
+            throw new Error('Unable to create local tool manifest file.');
+        }
+    }
+
+    // then acquire the tool and add it to the manifest
+    vscode.window.showInformationMessage('Acquiring latest `dotnet-interactive` tool.');
+    let toolSource = 'https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-tools/nuget/v3/index.json';
+    const result = await processExitCodeAndOutput(dotnetPath, ['tool', 'install', '--add-source', toolSource, 'Microsoft.dotnet-interactive', '--tool-manifest', localToolManifestFile]);
```
colombod, line 102: could we have a flag, like previewFeed, or the like in the extension configuration so users can decide what feed to use? this way we do not need to release a package again to change this source.

colombod:
> This feels complicated, but it's not too bad.
> 
> First, the acquisition of the .NET SDK moved to the new file `acquisition.ts`. Nothing there has changed.
> 
> Now on to the good bits:
> 
> 1. We need to check the version of `dotnet-interactive` via the `--version` switch, but that gets odd because `dotnet` might exist system-wide, or it might have been acquired for us, which means it has a different path and would only succeed as a local tool.  Because of this we try twice, first simply with the args `--version`, then again with `tool run dotnet-interactive -- --version`.
>    
>    * To handle the case where a local tool was already acquired and up to date (e.g., you did this 10 minutes ago), we have to set the working directory to the location of the local tool manifest, **BUT** if we've never done a local tool install, that directory won't exist and trying to launch a process with a non-existent working directory is bad, so we only set the working directory if it exists.  This is fine because in that case, the version check will fail and we'll proceed onto step 3 below.
> 2. If any one of those doesn't fail and has an up-to-date version, we return the appropriate args.
> 3. If that failed, we install a local tool to the extension's global storage so we don't stomp on any system-wide settings.  We install from the internally managed but publicly available `dotnet-tools` feed, which gets builds out of `master`.  **If we want to change this to be NuGet.org, we need to publish a newer package there, because the current one is 2 months old.**
>    
>    * In the event that the local tool install failed, you're out of luck; there's nothing we can do so we throw.
> 4. We then pass the path to `dotnet` and the args necessary to invoke the `interactive` sub-command to `StdioKernelTransport` which does the regular setup.
> 
> Check the function `getArgsToLaunchInteractive()` for the real magic.

We should push a new one to nuget i suppose

src/dotnet-interactive-vscode/src/acquisition.ts
```diff
@@ -0,0 +1,129 @@
...
+    if (result.code !== 0) {
```
colombod, line 103: we should communicate both failure and success otherwise the users will never understand if we completed this or not

README.md
```diff
@@ -35,7 +35,7 @@ First, make sure you have the following installed:
 - Next, in an ordinary console, install the `dotnet interactive` global tool:
 
 ```console
-> dotnet tool install -g --add-source "https://dotnet.myget.org/F/dotnet-try/api/v3/index.json" Microsoft.dotnet-interactive
+> dotnet tool install -g --add-source "https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-tools/nuget/v3/index.json" Microsoft.dotnet-interactive
```
colombod, line 38: is this the path to use instead of myget?

jonsequitur: The MyGet feed URL appears in a number of other locations.

brettfo: There should be a separate PR to update the URLs.  The MyGet feed hasn't been updated since January.

src/dotnet-interactive-vscode/src/tests/unit/acquisition.test.ts
```diff
@@ -0,0 +1,333 @@
...
+        throw new Error('This function should have never been called.');
+    }
+
+    function report() {
+        // noop
+    }
+
+    it("global storage doesn't exist", async () => {
+        // this tests the scenario where the extension launches for the first time and nothing exists
+        await withFakeGlobalStorageLocation(false, async globalStoragePath => {
```
jonsequitur, line 81: > and nothing exists

This is pretty heavy for a unit test.

src/dotnet-interactive-vscode/src/tests/unit/acquisition.test.ts
```diff
@@ -0,0 +1,333 @@
...
+            expect(globalStoragePath).to.not.be.a.path(); // sanity check that it doesn't exist
```
jonsequitur, line 82: Should the sanity check be a separate test (of the test infra itself)? It would make this check and others like it in other tests unnecessary.

src/dotnet-interactive-vscode/src/tests/unit/acquisition.test.ts
```diff
@@ -0,0 +1,333 @@
...
+
+    function installInteractiveToolThatAlwaysThrows(args: InstallInteractiveArgs, globalStoragePath: string): Promise<void> {
+        throw new Error('This function should have never been called.');
+    }
+
+    function report() {
+        // noop
+    }
+
+    it("global storage doesn't exist", async () => {
```
jonsequitur, line 79: The name expresses just a precondition. It would be more helpful if it expressed an expected outcome, like `it("throws existential panic when nothing exists"`.

I think the expectation in this case is it uses a default/pre-installed version?

src/dotnet-interactive-vscode/src/tests/unit/acquisition.test.ts
brettfo: I agree, but the integration tests aren't (can't be) run as part of CI and this was messy enough to figure out I wanted to make sure it wouldn't break again. (and the filesystem work was part of the messiness that I wanted to ensure continued to work.)

src/dotnet-interactive-vscode/package.json
```diff
@@ -45,6 +45,26 @@
         ]
       }
     ],
+    "configuration": {
+      "title": ".NET Interactive Notebook",
+      "properties": {
+        "dotnet-interactive.interactiveToolSource": {
+          "type": "string",
```
colombod, line 57: This is good but even just "use latest" or somethign along that line to switch from nuget.org to this feed

PR #408, opened 2020-05-13
Feature/enable variable set

colombod:


src/Microsoft.DotNet.Interactive.FSharp/FSharpKernel.fs
```diff
@@ -159,15 +140,16 @@ type FSharpKernel() as this =
...
+        | Some cv ->
+            value <- cv.Value.ReflectionValue :?> 'a
             true
-        | None ->
+        | _ ->
             false
 
     override _.SetVariableAsync(name: string, value: Object) : Task = 
-        raise (NotImplementedException())
+        script.Value.Fsi.AddBoundValue(name, value) |> ignore
```
TIHan, line 151: This will return a `FSharpErrorInfo` array. Should you report potential errors that occur?

PR #409, opened 2020-05-13
More docs improvements

jonsequitur:


samples/readme.md
```diff
@@ -0,0 +1,11 @@
+# .NET Interactive Samples
+
+This folder contains a number of samples that you can use to explore features of .NET Interactive.
+
+* [`docker-image`](docker-image/readme.md) contains a Dockerfile that generates an image with the latest .NET Interactive and Jupyter. This lets you try out .NET Interactive's Jupyter experience without needing to install Jupyter directly.
```
LadyNaggaga, line 5: Should we say docker file instead? Just to set user expectations in the meantime.

jonsequitur: @colombod Any build-related fallout if we rename this folder to capture the purpose of this Docker setup vs the Binder one?

docs/extending-dotnet-interactive.md
```diff
@@ -0,0 +1,78 @@
+# Extending .NET Interactive
+
+You can create extensions for .NET Interactive in order to create custom experiences including custom visualizations, new magic commands, code execution pipeline behaviors, and more. Extensions can be distributed using NuGet packages
```
LadyNaggaga, line 3: Instead of code execution pipeline behavior could we say pipeline execution?

src/dotnet-interactive-vscode/README.md
```diff
@@ -1,21 +1,17 @@
...
-1. Install this extension from the [marketplace](https://marketplace.visualstudio.com/items?itemName=ms-dotnettools.dotnet-interactive-vscode).
-1. Launch VS Code Insiders like this: `code-insiders --enable-proposed-api ms-dotnettools.dotnet-interactive-vscode`
-1. Open an existing file with the `.dotnet-interactive` extension.
-
-## Development
+2. To edit an existing Jupyter notebook, open a file with a `.ipynb` extension,
+   
+   _or_
+   
+   To create a new .NET Interactive notebook, create a file with the extension `.dotnet-interactive` and open it.
```
LadyNaggaga, line 16: I think this should be step 2. Step 2 should be be the Or

src/dotnet-interactive-vscode/README.md
```diff
@@ -1,21 +1,17 @@
...
+This extension is currently **_in preview_**.  [Visual Studio Code Insiders](https://code.visualstudio.com/insiders/)  is required.
 
 ---
 
 # .NET Interactive Notebooks
 
-This extension adds support for using the `dotnet-interactive` global tool from within VS Code in a notebook-like environment.
+This extension adds support for using .NET Interactive in a Visual Studio Code notebook.
 
 ## Getting Started
```
LadyNaggaga, line 10: 1.  Install the latest VS Code Insiders
2.  Install the .NET Interactive Notebook extension from the market place. https://marketplace.visualstudio.com/items?itemName=ms-dotnettools.dotnet-interactive-vscode
3.  Create a new folder `notebooks` and navigate to it
4.  Launch VS Code Insiders like this `code-insiders --enable-proposed-api ms-dotnettools.dotnet-interactive-vscode .`
image
5. Inside notebooks create `foo.dotnet-interactive` and you all set 
image

Or 
To edit an existing Jupyter notebook, open a file with a `.ipynb` extension,

samples/readme.md
colombod: yes this will break one of `Dockerfile_official_image` jon in the yaml file https://github.com/dotnet/interactive/blob/69f49f4e21a197100884456daa1442525fd73339/azure-pipelines.yml#L215

src/dotnet-interactive-vscode/README.md
colombod: I'd say `Launch VS Code Insiders like this` in `Launch VS Code Insiders in the current directory with the command`

samples/readme.md
colombod: the goal of such file is to 
`This lets you try out .NET Interactive's Jupyter experience without needing to install Jupyter directly. It can be used as base for packaging and redistribute docker images inside organization where end users are not allowed to install the components locally`

jonsequitur: I'm going to shave this yak another day, then, since it has a build impact. That justifies a separate PR I think.

PR #411, opened 2020-05-14
add command/event pair to change the current working directory

brettfo:
This is useful for VS Code notebooks that do any local file IO; it's a reasonable assumption that file access will be performed relative to the notebook file.  I found this issue trying to port the HousingML notebook, and I noticed that downloading the CSV didn't place the file where I'd hoped/expected.

As for testing from the kernel, we can't really unit test this because either it works or it doesn't and we can't pollute the current test environment; it'll destroy everything else.  Testing from the notebook also doesn't make much sense, because we can simply fire-and-forget the command.  If it worked, it worked; if it failed, it failed, and there's nothing actionable the user can do about it.

We also can't start the stdio server in the notebook directory, because to get the local tool scenario to work, the initial working directory has to be in the extensions global storage location.

The only other possibility I can think of would be to launch with arguments like the following:

`dotnet interactive stdio --working-directory C:\directory\that\contains\notebooks`

I'm certainly open to doing that, but it seemed more useful to have the possibility of changing the working directory multiple times during notebook execution.

If we merge this I'll have to come back later and update the min required tool version where this command/event pair exist.

### DIFF
--- a/src/Microsoft.DotNet.Interactive/Commands/ChangeWorkingDirectory.cs
+++ b/src/Microsoft.DotNet.Interactive/Commands/ChangeWorkingDirectory.cs
@@ -1,0 +1,15 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.DotNet.Interactive.Commands
+{
+    public class ChangeWorkingDirectory : KernelCommandBase
+    {
+        public string WorkingDirectory { get; }
+
+        public ChangeWorkingDirectory(string workingDirectory)
+        {
+            WorkingDirectory = workingDirectory;
+        }
+    }
+}
--- a/src/Microsoft.DotNet.Interactive/Commands/ChangeWorkingDirectory.cs
+++ b/src/Microsoft.DotNet.Interactive/Commands/ChangeWorkingDirectory.cs
@@ -1,13 +1,15 @@
 ﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System.IO;
+
 namespace Microsoft.DotNet.Interactive.Commands
 {
     public class ChangeWorkingDirectory : KernelCommandBase
     {
-        public string WorkingDirectory { get; }
+        public DirectoryInfo WorkingDirectory { get; }
 
-        public ChangeWorkingDirectory(string workingDirectory)
+        public ChangeWorkingDirectory(DirectoryInfo workingDirectory)
         {
             WorkingDirectory = workingDirectory;
         }
--- a/src/Microsoft.DotNet.Interactive/Events/WorkingDirectoryChanged.cs
+++ b/src/Microsoft.DotNet.Interactive/Events/WorkingDirectoryChanged.cs
@@ -10,7 +10,7 @@ namespace Microsoft.DotNet.Interactive.Events
     {
         public DirectoryInfo WorkingDirectory { get; }
 
-        public WorkingDirectoryChanged(IKernelCommand command, DirectoryInfo workingDirectory)
+        public WorkingDirectoryChanged(DirectoryInfo workingDirectory, IKernelCommand command = null)
             : base(command)
         {
             WorkingDirectory = workingDirectory;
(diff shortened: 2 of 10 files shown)


src/Microsoft.DotNet.Interactive/Commands/ChangeWorkingDirectory.cs
```diff
@@ -0,0 +1,15 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.DotNet.Interactive.Commands
+{
+    public class ChangeWorkingDirectory : KernelCommandBase
+    {
+        public string WorkingDirectory { get; }
+
+        public ChangeWorkingDirectory(string workingDirectory)
```
jonsequitur, line 10: This should be a `DirectoryInfo` rather than a string.

src/Microsoft.DotNet.Interactive/Events/WorkingDirectoryChanged.cs
```diff
@@ -0,0 +1,18 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Microsoft.DotNet.Interactive.Commands;
+
+namespace Microsoft.DotNet.Interactive.Events
+{
+    public class WorkingDirectoryChanged : KernelEventBase
+    {
+        public string WorkingDirectory { get; }
```
jonsequitur, line 10: This should be a `DirectoryInfo`.

jonsequitur:
This makes good sense. I think a command line argument to do this might also make sense. If we do that, we might want it to be an argument (which feels more common) and allow it to eventually be used as well to open files or projects, e.g.

```console
> dotnet interactive stdio /path/to/someplace
```

src/dotnet-interactive-vscode/src/vscode/notebookProvider.ts
```diff
@@ -34,7 +34,9 @@ export class DotNetInteractiveNotebookContentProvider implements vscode.Notebook
 
         let client = this.clientMapper.getOrAddClient(uri);
         let notebookPath = path.dirname(uri.fsPath);
-        client.changeWorkingDirectory(notebookPath);
+        client.changeWorkingDirectory(notebookPath).catch(() => {
+            vscode.window.showInformationMessage(`Unable to set notebook working directory to '${notebookPath}'.`);
```
jonsequitur, line 38: Should this information and message not be coming from the kernel in the form of a `CommandFailed`?

brettfo: Changed `reject()` to `reject(err)` and included the message.

src/Microsoft.DotNet.Interactive/Events/WorkingDirectoryChanged.cs
```diff
@@ -0,0 +1,19 @@
...
+using System.IO;
+using Microsoft.DotNet.Interactive.Commands;
+
+namespace Microsoft.DotNet.Interactive.Events
+{
+    public class WorkingDirectoryChanged : KernelEventBase
+    {
+        public DirectoryInfo WorkingDirectory { get; }
+
+        public WorkingDirectoryChanged(IKernelCommand command, DirectoryInfo workingDirectory)
```
jonsequitur, line 13: In most cases the `IKernelCommand` parameter is not the first parameter and is optional.

src/dotnet-interactive-vscode/src/vscode/notebookProvider.ts
```diff
@@ -31,14 +32,25 @@ export class DotNetInteractiveNotebookContentProvider implements vscode.Notebook
                 break;
         }
 
-        this.clientMapper.getOrAddClient(uri);
+        let client = this.clientMapper.getOrAddClient(uri);
+        let notebookPath = path.dirname(uri.fsPath);
+        client.changeWorkingDirectory(notebookPath).catch((err) => {
+            let message = `Unable to set notebook working directory to '${notebookPath}'.`;
+            if (err && err.message) {
+                message += '\r\n' + err.message.toString();
```
jonsequitur, line 40: Any reason to use the Windows-specific `\r\n`?

brettfo: muscle memory, I'll make it a simple `\n` since it doesn't matter

PR #417, opened 2020-05-15
don't crash the kernel server if a command wasn't found

brettfo:
While testing I discovered that submitting a command to the kernel that doesn't exist results in a crash which shouldn't happen.  This fixes that behavior by throwing a very specific exception when a command doesn't exist and returns a diagnostic event.

I verified this in VS Code by submitting a command that doesn't exist and verifying that the server remained active and subsequent (valid) commands were honored.

### DIFF
--- a/src/Microsoft.DotNet.Interactive.Tests/Server/StandardIOKernelServerTests.cs
+++ b/src/Microsoft.DotNet.Interactive.Tests/Server/StandardIOKernelServerTests.cs
@@ -198,6 +198,31 @@ namespace Microsoft.DotNet.Interactive.Tests.Server
             stdOut.Should().Contain(guid);
         }
 
+        [Fact]
+        public async Task missing_command_returns_a_diagnostic_event()
+        {
+            // since it's not a valid command we have to manually create it
+            var command = new
+            {
+                CommandType = "ThisIsNotAValidCommandType",
+                Command = new object(), // body doesn't matter
+            };
+            var writer = new StringWriter();
+            Serializer.JsonSerializer.Serialize(writer, command);
+            var commandText = writer.ToString();
+
+            await _standardIOKernelServer.WriteAsync(commandText);
+
+            _kernelEvents
+                .Should()
+                .ContainSingle<KernelEventEnvelope<DiagnosticLogEntryProduced>>()
+                .Which
+                .Event
+                .Message
+                .Should()
+                .Be("Command 'ThisIsNotAValidCommandType' not found.");
+        }
+
         public void Dispose()
         {
             _disposables.Dispose();
--- a/src/Microsoft.DotNet.Interactive/Server/Serializer.cs
+++ b/src/Microsoft.DotNet.Interactive/Server/Serializer.cs
@@ -6,7 +6,7 @@ using Newtonsoft.Json.Serialization;
 
 namespace Microsoft.DotNet.Interactive.Server
 {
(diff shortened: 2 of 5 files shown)


src/Microsoft.DotNet.Interactive/Server/Serializer.cs
```diff
@@ -6,7 +6,7 @@
 
 namespace Microsoft.DotNet.Interactive.Server
 {
-    internal static class Serializer
+    public static class Serializer
```
brettfo, line 9: This was made public so the unit test could use the same serializer.

src/Microsoft.DotNet.Interactive.Tests/Server/StandardIOKernelServerTests.cs
```diff
@@ -198,6 +198,31 @@ public async Task it_produces_values_when_executing_Console_output()
...
+            };
+            var writer = new StringWriter();
+            Serializer.JsonSerializer.Serialize(writer, command);
+            var commandText = writer.ToString();
+
+            await _standardIOKernelServer.WriteAsync(commandText);
+
+            _kernelEvents
+                .Should()
+                .ContainSingle<KernelEventEnvelope<DiagnosticLogEntryProduced>>()
```
colombod, line 218: I think this should be at least a commandFailed event @jonsequitur  what are your thoughts?

brettfo: We can't do command failed event, because that requires returning the command that failed, so then on the other side, the client would try to deserialize another invalid command object:

``` json
{
  "eventType": "CommandFailed",
  "event": "...",
  "command": {
    "commandType": "NotACommand",
   "...": "this will fail to parse according to `contracts.ts`",
  }
}
```

colombod: we can roundtrip the evnelope

colombod: this is a logic that does not make it to the command pipeline, is done at the transport layer when commands are deserialised

jonsequitur: A diagnostic event isn't forceful enough. 

One possibility is the command on the `CommandFailed` event could be null. What do we do when there's a serialization error?

brettfo: I've been experimenting and we can return a `CommandFailed` event with a null `command` field, but since that field is `null` there's no token, so we can't pair the failed event with the client-side command that triggered it.  The end result is that the client never knows that a `CommandFailed` event comes back so it's forever waiting for a result.

It seems like we have 2 options at this point:

1. Populate a fake/shim `command` field so we can regurgitate the token so the client get notified that something terrible happened, or
2. Drop this PR since we now tightly couple the kernel server and client extension, so we're much less likely to get into a situation where the client submits a command that it's not prepared to handle.

jonsequitur: I think #2 is fine. This is a programming error, so handling it within the domain doesn't make a lot of sense.

PR #425, opened 2020-05-16
update installation instructions

brettfo:


src/dotnet-interactive-vscode/README.md
```diff
@@ -10,28 +10,12 @@ This extension adds support for using the `dotnet-interactive` global tool from
...
 ## Getting Started
 
-1. Install the [.NET SDK 3.1](https://dotnet.microsoft.com/download/visual-studio-sdks).
-1. Install the latest [.NET Interactive Global Tool](https://www.nuget.org/packages/Microsoft.dotnet-interactive/).
 1. Install the latest [VS Code Insiders](https://code.visualstudio.com/insiders/).
-1. Open this directory with `code-insiders` and F5
-1. Open a file with the `.dotnet-interactive` extension.
+1. Install this extension from the [marketplace](https://marketplace.visualstudio.com/items?itemName=ms-dotnettools.dotnet-interactive-vscode).
+1. Launch VS Code Insiders like this: `code-insiders --enable-proposed-api ms-dotnettools.dotnet-interactive-vscode`
+1. Open an existing file with the `.dotnet-interactive` extension.
```
jonsequitur, line 16: When do we get to step 2?

jonsequitur: I'd love to come up with a better extension sooner rather than later.

brettfo:
``` basic
1 DoSomething
2 GOTO 1
3 Profit!
```

Not sure, I can't seem to get past this part.

brettfo: Re: extension:  What about `.interactive`?  I think something like `.int` is too short and not descriptive enough and the VS Code team might come after us if we used `.code`.

PR #426, opened 2020-05-16
use same update object to display progress

colombod:
This fixes the issue of progress update on another line when importing packages


samples/vs-code/covid-19.dotnet-interactive
```diff
@@ -118,7 +118,7 @@ notebookScope.plot = function (plotTarget) {
                         name: "Recovered",
                         type: "scattergeo",
                         mode: "markers",
-                        geo: "geo1",
```
jonsequitur, line 121: Do we want to make these same changes in the equivalent `.ipynb`?

jonsequitur:
Maybe the VS Code notebooks should go under `/samples/notebooks`, alongside the Jupyter examples? They're still notebooks.

samples/vs-code/covid-19.dotnet-interactive
colombod: let me do that

colombod: done

colombod:
> Maybe the VS Code notebooks should go under `/samples/notebooks`, alongside the Jupyter examples? They're still notebooks.

they will be copied / mounted to the my binder image, I put those files in their own folder as I am not even sure we should have the ipynb files in a folder we open with vscode

PR #429, opened 2020-05-16
improve experience with jupyter notebooks in vs code

### LINKED ISSUE #427: VS Code overwrites opened .ipynb file contents with .NET Interactive format
### Describe the bug
Please provide as much information as you can.

After opening and saving an `.ipynb`:

image

Which version of .NET Interactive were you using? You can find this out in a notebook by running the `#!about` magic command or, at the command line, you can run `dotnet interactive --version`.

Version: 1.0.126511+f4313f53dc1b70447ef17c2116588ee27b6bac25

 - Frontend
    - [ ] Jupyter Notebook
    - [ ] Jupyter Lab
    - [ ] nteract
    - [x] Visual Studio Code
    - [ ] Other (please specify)

### LINKED ISSUE #428: When opening an .ipynb in VS Code, there are empty cells.
Empty cells when opening an `.ipynb`. In the screen shot here, I opened `/samples/notebooks/Extensions.ipynb` and the notebook contains empty cells, though not missing code.

Version: 1.0.126511+f4313f53dc1b70447ef17c2116588ee27b6bac25

image

brettfo:
Fixes #427.
Fixes #428.

### DIFF
--- a/src/dotnet-interactive-vscode/src/interactiveNotebook.ts
+++ b/src/dotnet-interactive-vscode/src/interactiveNotebook.ts
@@ -1,7 +1,22 @@
 import { RawNotebookCell } from "./interfaces";
 import { trimTrailingCarriageReturn } from './utilities';
 
-export const editorLanguages = ['csharp', 'fsharp', 'html', 'javascript', 'markdown', 'powershell'];
+let languageAliases = new Map<string, string>();
+// full language names always map to themselves
+languageAliases.set('csharp', 'csharp');
+languageAliases.set('fsharp', 'fsharp');
+languageAliases.set('html', 'html');
+languageAliases.set('javascript', 'javascript');
+languageAliases.set('markdown', 'markdown');
+languageAliases.set('powershell', 'powershell');
+// short aliases
+languageAliases.set('cs', 'csharp');
+languageAliases.set('fs', 'fsharp');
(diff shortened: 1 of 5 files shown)


src/dotnet-interactive-vscode/src/interactiveNotebook.ts
```diff
@@ -1,7 +1,22 @@
 import { RawNotebookCell } from "./interfaces";
 import { trimTrailingCarriageReturn } from './utilities';
 
-export const editorLanguages = ['csharp', 'fsharp', 'html', 'javascript', 'markdown', 'powershell'];
+let languageAliases = new Map<string, string>();
```
jonsequitur, line 4: Are these intended to be for code coloring / language service selection on the VS Code side, or for subkernel routing on the .NET Interactive side?

If the latter, we don't currently support `cs` and `fs` but maybe we should.

brettfo: These are to normalize `cs` to `csharp`, etc., so the kernel should never see `cs`, but I agree, adding `cs`/`c#`/`C#`/etc. would be good.

jonsequitur: We support `C#` and `c#` currently as kernel name aliases.

PR #437, opened 2020-05-18
move jupyter integration to be an explicit conversion step

brettfo:
Removes the `.ipynb` file association.  This is good because we don't want to accidentally stomp on an existing Jupyter notebook.  Instead, the user explicitly invokes the convert command that will write out a new `.dotnet-interactive` file which then can then use as normal.  There is an existing command to export as a Jupyter notebook to complete the cycle.

### DIFF
--- a/src/dotnet-interactive-vscode/package.json
+++ b/src/dotnet-interactive-vscode/package.json
@@ -25,6 +25,7 @@
   "activationEvents": [
     "onNotebookEditor:dotnet-interactive",
     "onCommand:dotnet-interactive.acquire",
+    "onCommand:dotnet-interactive.convertJupyterNotebook",
     "onCommand:dotnet-interactive.reportInteractiveVersion"
   ],
   "main": "./out/extension.js",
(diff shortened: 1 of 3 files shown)


src/dotnet-interactive-vscode/package.json
```diff
@@ -92,7 +90,11 @@
       },
       {
         "command": "dotnet-interactive.reportInteractiveVersion",
-        "title": "Report installed version for .NET Interactive"
+        "title": ".NET Interactive: Report installed version for .NET Interactive"
```
jonsequitur, line 93:
```suggestion
        "title": ".NET Interactive: Report installed version of .NET Interactive"
```

src/dotnet-interactive-vscode/package.json
```diff
@@ -92,7 +90,11 @@
...
+      },
+      {
+        "command": "dotnet-interactive.convertJupyterNotebook",
+        "title": ".NET Interactive: Convert Jupyter notebook to .NET Interactive notebook"
```
jonsequitur, line 97:
```suggestion
        "title": ".NET Interactive: Convert Jupyter notebook (.ipynb) to .NET Interactive notebook"
```

PR #448, opened 2020-05-19
append git sha to vscode package

brettfo:


src/dotnet-interactive-vscode/package.json
```diff
@@ -1,7 +1,8 @@
 {
   "name": "dotnet-interactive-vscode",
   "displayName": ".NET Interactive Notebooks",
-  "description": ".NET Interactive Notebooks for VS Code",
+  "//description": "During build, the description gets the git SHA appended to it.",
```
jonsequitur, line 4: Did you mean to leave this in?

brettfo: Yes, because if you run the script locally you'll have edits to this file that shouldn't get checked in; this is a sad attempt at documenting this.

PR #449, opened 2020-05-19
Update tool version for vs code extension

colombod:


src/dotnet-interactive-vscode/src/interactiveClient.ts
```diff
@@ -49,8 +49,9 @@ export class InteractiveClient {
 
             let reportOutputs = () => {
                 if (this.deferredOutput.length > 0) {
-                    outputs = [...this.deferredOutput, ...outputs];
```
brettfo, line 52: This is incorrect.  `observer` is expected to be called with a complete set of outputs since they can change.  By calling it twice like this the second invocation will completely overwrite the first.

brettfo: A quick check locally showed this to work:

``` typescript
let reportOutputs = () => {
    if (this.deferredOutput.length > 0) {
        outputs.push(...this.deferredOutput);
        this.deferredOutput = [];
    }

    observer(outputs);
};
```

colombod: done

PR #451, opened 2020-05-20
add custom grammars notebook languages and magic commands

brettfo:
A change from before; this registers new languages in VS Code, e.g., instead of `csharp` we register `dotnet-interactive.csharp`.  This fixes 3 things:

1. We can ensure that when we enable language services (hover, completion) we don't override an official language extension like Omnisharp.
2. We can ensure that official language extensions like Omnisharp don't get asked to provide language services (hover, completion) for notebook cells, because we have all the context, they don't.
3. We can specifically mark all lines that aren't language specifiers as magic commands.  Currently I'm choosing to mark magic commands as comments for 2 reasons:
    1. The regular C# grammar treats all `#` lines as keywords and I want magic commands to be obviously different.
    2. The idea of a comment seemed appropriate, in that it's something the language executor ignores and only has special meaning somewhere else.

All of this has one pretty major consequence: we have to tell VS Code that a C# cell is of the type `dotnet-interactive.csharp` and to do that properly we had to register a completely new language.  I'm currently calling it `C# Interactive` (similarly for F#, PowerShell, etc.)  This does open the issue of confusion between the actual C# interactive project and this.  I'm open to naming suggestions.

The backing file format still uses the short name specifiers, like `#!csharp`; only the IDE is told the lie that it's really `dotnet-interactive.csharp`.

image

Work not included in this PR: language services for magic commands; that requires more kernel work.


src/dotnet-interactive-vscode/src/interactiveNotebook.ts
```diff
@@ -5,6 +5,8 @@ import { RawNotebookCell } from "./interfaces";
 import { trimTrailingCarriageReturn } from './utilities';
 
 let languageAliases = new Map<string, string>();
+// polyglot language cell type
+languageAliases.set('dotnet-interactive', 'dotnet-interactive');
```
colombod, line 9: a polyglot cell targets the .NET kernel itself (composite), right?

brettfo: Currently backed this part out, so the only thing this PR allows you to do is set the language mode on a text file to `.NET Interactive Script` and it'll get the polyglot coloring.

src/dotnet-interactive-vscode/package.json
```diff
@@ -109,6 +109,87 @@
...
+      {
+        "id": "dotnet-interactive.fsharp",
+        "aliases": [
+          "F# Interactive"
+        ]
+      },
+      {
+        "id": "dotnet-interactive.html",
+        "aliases": [
+          "HTML Interactive"
```
jonsequitur, line 129:
```suggestion
          "HTML (.NET Interactive)"
```

src/dotnet-interactive-vscode/package.json
```diff
@@ -109,6 +109,87 @@
         "command": "dotnet-interactive.stopAllNotebookKernels",
         "title": ".NET Interactive: Stop all notebook kernels"
       }
+    ],
+    "languages": [
+      {
+        "id": "dotnet-interactive.csharp",
+        "aliases": [
+          "C# Interactive"
```
jonsequitur, line 117:
```suggestion
          "C# (.NET Interactive)"
```

src/dotnet-interactive-vscode/package.json
```diff
@@ -109,6 +109,87 @@
...
+      {
+        "id": "dotnet-interactive.html",
+        "aliases": [
+          "HTML Interactive"
+        ]
+      },
+      {
+        "id": "dotnet-interactive.javascript",
+        "aliases": [
+          "JavaScript Interactive"
```
jonsequitur, line 135:
```suggestion
          "JavaScript (.NET Interactive)"
```

src/dotnet-interactive-vscode/package.json
```diff
@@ -109,6 +109,87 @@
...
+      {
+        "id": "dotnet-interactive.markdown",
+        "aliases": [
+          "Markdown Interactive"
+        ]
+      },
+      {
+        "id": "dotnet-interactive.powershell",
+        "aliases": [
+          "PowerShell Interactive"
```
jonsequitur, line 153:
```suggestion
          "PowerShell (.NET Interactive)"
```

src/dotnet-interactive-vscode/package.json
```diff
@@ -109,6 +109,87 @@
...
+      {
+        "id": "dotnet-interactive.magic-commands",
+        "aliases": [
+          "Magic Commands Interactive"
+        ]
+      },
+      {
+        "id": "dotnet-interactive.markdown",
+        "aliases": [
+          "Markdown Interactive"
```
jonsequitur, line 147:
```suggestion
          "Markdown (.NET Interactive)"
```

src/dotnet-interactive-vscode/package.json
```diff
@@ -109,6 +109,87 @@
...
+      {
+        "id": "dotnet-interactive.javascript",
+        "aliases": [
+          "JavaScript Interactive"
+        ]
+      },
+      {
+        "id": "dotnet-interactive.magic-commands",
+        "aliases": [
+          "Magic Commands Interactive"
```
jonsequitur, line 141:
```suggestion
          "Magic Command (.NET Interactive)"
```

PR #458, opened 2020-05-21
enable wallaby for vscode extension development

colombod:
enable wallaby js for fast iteration and coverage


src/dotnet-interactive-vscode/src/tests/integration/suite/extension.test.ts
```diff
@@ -26,13 +26,13 @@ suite('Extension Test Suite', () => {
         let code = '1+1';
         let result: Array<CellOutput> = [];
         await client.execute(code, 'csharp', outputs => result = outputs);
-        expect(result).to.deep.equal([
+        expect(result).to.deep.include(
```
brettfo, line 29: Why `include` instead of `equal`?  It seems valuable to have this test fail if there's extra output.

PR #468, opened 2020-05-22
don't return the notebook data until it's ready for execution

brettfo:
Previously, if notebook cells were executed too quickly they would fail because the deferred commands had not yet run which meant helper functions we inject, like `display()`, would fail to bind and all subsequent calls to those would fail.  The workaround was to restart the kernel and wait for a bit and hope you weren't too fast.

Since the stdio kernel transport always executes a ChangeWorkingDirectory command, that's enough to ensure the kernel is in a ready state, so the change was to turn `ClientMapper.getOrAddClient` to instead return a Promise; that way we can properly `await` the CWD.  Only on client creation is the `await` actually awaiting anything; all subsequent calls return immediately.

Since a constructor can't be async/Promise I made the .ctor private and added a static `create` method with the appropriate signature and moved the CWD invocation there.  The rest of the changes on this PR were simply to add `await` where necessary when calling `getOrAddClient`.

### DIFF
--- a/src/dotnet-interactive-vscode/src/stdioKernelTransport.ts
+++ b/src/dotnet-interactive-vscode/src/stdioKernelTransport.ts
@@ -73,7 +73,7 @@ export class StdioKernelTransport {
                             let failed = <CommandFailed>envelope.event;
                             let message = `Unable to set notebook working directory to '${notebookPath}'.\n${failed.message}`;
                             diagnosticChannel.appendLine(message);
-                            reject();
+                            resolve(kernelTransport);
                             disposable.dispose();
                             break;
                         case CommandHandledType:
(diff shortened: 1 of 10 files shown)


src/dotnet-interactive-vscode/src/stdioKernelTransport.ts
```diff
@@ -58,29 +58,36 @@ export class StdioKernelTransport {
...
+            // immediately clean up afterwards because we'll never do this again
+            let token = 'change-working-directory-token';
+            let disposable = kernelTransport.subscribeToKernelEvents(envelope => {
+                if (envelope.command?.token === token) {
+                    switch (envelope.eventType) {
+                        case CommandFailedType:
+                            let failed = <CommandFailed>envelope.event;
+                            let message = `Unable to set notebook working directory to '${notebookPath}'.\n${failed.message}`;
+                            diagnosticChannel.appendLine(message);
+                            reject();
```
brettfo, line 76: In this scenario, the initial CWD command failed.  Should we still call `resolve()` and pretend it was OK, or do we fail and not even open the notebook?

brettfo: Nevermind; I tried this scenario where CWD fails and with the `reject()` call the notebook immediately closes with no actionable message to the user.  That doesn't seem very useful so I changed this to `resolve(...)` so the notebook can still be opened, and the failure message at least appears in the diagnostic log below.

PR #473, opened 2020-05-24
Completion improvements

jonsequitur:


src/Microsoft.DotNet.Interactive.Tests/LanguageServices/CompletionTests.cs
```diff
@@ -0,0 +1,117 @@
...
+                .Which
+                .CompletionList
+                .Should()
+                .Contain(item => item.DisplayText == variableName);
+        }
+
+        [Theory]
+        [InlineData(Language.FSharp)]
+        [InlineData(Language.CSharp)]
+        public async Task Completions_are_available_for_symbols_declared_in_a_previous_submission_ending_in_a_trailing_expression(Language language)
```
brettfo, line 92: What about symbols being available 2+ submissions back, each with trailing expressions?  I want to make sure we don't miss that scenario, because it's very useful to have each cell end with an expression to see the state as you're building your notebook.

src/Microsoft.DotNet.Interactive.CSharp/Microsoft.DotNet.Interactive.CSharp.csproj
```diff
@@ -31,6 +31,7 @@
     <PackageReference Include="Microsoft.CodeAnalysis.CSharp.Features" Version="$(MicrosoftCodeAnalysisCommonVersion)" />
     <PackageReference Include="Microsoft.CodeAnalysis.Scripting.Common" Version="$(MicrosoftCodeAnalysisCommonVersion)" />
     <PackageReference Include="Microsoft.CodeAnalysis.Workspaces.Common" Version="$(MicrosoftCodeAnalysisCommonVersion)" />
+    <PackageReference Include="System.ServiceModel.Syndication" Version="4.7.0" />
```
brettfo, line 34: What functionality is this providing?

jonsequitur: I have no idea how that got there.

src/Microsoft.DotNet.Interactive.Tests/LanguageServices/CompletionTests.cs
jonsequitur: I'll add more scenarios to these tests once the current ones are passing. Also on the list, submissions containing magic commands.

PR #484, opened 2020-05-28
allow DisplayedValueUpdated to update across submissions

### LINKED ISSUE #480: Cannot update display from another cell
### Describe the bug
I write in a cell:
```csharp
var updater = Display(12);
updater.Update(145);
```
the output below the cell displays the value 145
Then in another cell I write
```csharp
updater?.Update(30);
```
I would expected the output cell to update the display to 30. But is not changed

1.0.127302

- OS
    - [X] Windows 10
    - [ ] macOS
    - [ ] iOS
    - [ ] Android
 - Browser 
    - [ ] Chrome
    - [ ] Edge
    - [ ] Firefox
    - [ ] Safari
 - Frontend
    - [ ] Jupyter Notebook
    - [ ] Jupyter Lab
    - [ ] nteract
    - [X] Visual Studio Code
    - [ ] Other (please specify)

### Screenshots
image

brettfo:
Fixes #480.

Allows a display updater to be used across submissions.  Once this is merged and the signed build is complete, there will be another PR to update the VS Code tool version so that it contains the kernel updates.

Thanks to @colombod for doing the kernel work/tests.

### DIFF
--- a/src/Microsoft.DotNet.Interactive.Tests/LanguageKernelFormattingTests.cs
+++ b/src/Microsoft.DotNet.Interactive.Tests/LanguageKernelFormattingTests.cs
@@ -139,7 +139,7 @@ namespace Microsoft.DotNet.Interactive.Tests
                     v.Value.ToString().Contains("<b>world</b>"));
         }
 
-        [Theory(Timeout = 45000)]
+        [Theory]
         [InlineData(Language.CSharp)]
         [InlineData(Language.FSharp)]
         public async Task Displayed_value_can_be_updated_from_later_submissions(Language language)
@@ -153,19 +153,16 @@ namespace Microsoft.DotNet.Interactive.Tests
             };
 
             await kernel.SubmitCodeAsync(submissions[0]);
-            await kernel.SubmitCodeAsync(submissions[1]);
 
-            KernelEvents
-                .OfType<DisplayedValueProduced>()
-                .SelectMany(v => v.FormattedValues)
+            var updateCommandResult = await kernel.SubmitCodeAsync(submissions[1]);
+
+            updateCommandResult
+                .KernelEvents
+                .ToSubscribedList()
                 .Should()
-                .ContainSingle(v =>
-                    v.MimeType == "text/html" &&
-                    v.Value.ToString().Contains("<b>hello</b>"));
-
-            KernelEvents
-                .OfType<DisplayedValueUpdated>()
-                .SelectMany(v => v.FormattedValues)
+                .ContainSingle<DisplayedValueUpdated>()
+                .Which
+                .FormattedValues
                 .Should()
                 .ContainSingle(v =>
                     v.MimeType == "text/html" &&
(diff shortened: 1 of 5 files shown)


src/Microsoft.DotNet.Interactive.Tests/LanguageKernelFormattingTests.cs
```diff
@@ -139,6 +139,39 @@ public async Task Displayed_value_can_be_updated(Language language)
                     v.Value.ToString().Contains("<b>world</b>"));
         }
 
+        [Theory(Timeout = 45000)]
```
jonsequitur, line 142: The timeout shouldn't be needed.

src/Microsoft.DotNet.Interactive.Tests/LanguageKernelFormattingTests.cs
```diff
@@ -139,6 +139,39 @@ public async Task Displayed_value_can_be_updated(Language language)
...
+                    v.MimeType == "text/html" &&
+                    v.Value.ToString().Contains("<b>hello</b>"));
+
+            KernelEvents
+                .OfType<DisplayedValueUpdated>()
+                .SelectMany(v => v.FormattedValues)
+                .Should()
+                .ContainSingle(v =>
+                    v.MimeType == "text/html" &&
+                    v.Value.ToString().Contains("<b>world</b>"));
```
jonsequitur, line 172: The test feedback for these will be better if you do something like this:

```csharp
 KernelEvents
                .Should()
                .ContainSingle<DisplayedValueUpdated>()
                .Which
                .FormattedValues
                .Should()
                .ContainSingle(v =>   ...      );
```

src/Microsoft.DotNet.Interactive.Tests/LanguageKernelFormattingTests.cs
```diff
@@ -139,6 +139,39 @@ public async Task Displayed_value_can_be_updated(Language language)
...
+            await kernel.SubmitCodeAsync(submissions[0]);
+            await kernel.SubmitCodeAsync(submissions[1]);
+
+            KernelEvents
+                .OfType<DisplayedValueProduced>()
+                .SelectMany(v => v.FormattedValues)
+                .Should()
+                .ContainSingle(v =>
+                    v.MimeType == "text/html" &&
+                    v.Value.ToString().Contains("<b>hello</b>"));
```
jonsequitur, line 164: This first assertion is effectively a sanity check, not really the subject of this test, and it duplicates other test coverage.

src/Microsoft.DotNet.Interactive/Events/DisplayedValue.cs
```diff
@@ -35,6 +36,11 @@ public void Update(object updatedValue)
                 _mimeType,
                 updatedValue.ToDisplayString(_mimeType));
 
+            if (KernelInvocationContext.Current?.Command is SubmitCode)
+            {
+                _context = KernelInvocationContext.Current;
```
jonsequitur, line 41: This is kind of weird. 

What happens in Jupyter?

colombod: The behaviour is consistent and the data is updated on the frontend. We tested that too

brettfo:
@jonsequitur Updated test from feedback.

PR #517, opened 2020-06-09
account for submission splitting when requesting language services

brettfo:
When language service requests are received, parse/split the submission and re-create with appropriate line offsets.  In the example screenshot below, the F# kernel is told to compute completion items on line 2.  It has no idea that the actual incoming request was for line 5.

The resultant events are also massaged; F# can specify that the computed result is on line 2, but when an event listener (e.g., VS Code) gets the result, they only see line 5.

image

Note to @jonsequitur and @colombod: We talked about making the incoming commands mutable, but that prevents us from using the handy trick of checking the parent command on the resulting events, so we have to stick with re-creating new commands so we have an original to compare against when computing the line offsets on the way back out.

### DIFF
--- a/src/Microsoft.DotNet.Interactive.Tests/LanguageKernelTestBase.cs
+++ b/src/Microsoft.DotNet.Interactive.Tests/LanguageKernelTestBase.cs
@@ -8,7 +8,6 @@ using Microsoft.DotNet.Interactive.Commands;
 using Microsoft.DotNet.Interactive.CSharp;
 using Microsoft.DotNet.Interactive.Events;
 using Microsoft.DotNet.Interactive.FSharp;
-using Microsoft.DotNet.Interactive.Jupyter;
 using Microsoft.DotNet.Interactive.PowerShell;
 using Microsoft.DotNet.Interactive.Tests.Utility;
 using Pocket;
@@ -16,6 +15,8 @@ using Recipes;
 using Xunit.Abstractions;
 using Serilog.Sinks.RollingFileAlternate;
 using SerilogLoggerConfiguration = Serilog.LoggerConfiguration;
+using System.Collections.Generic;
+using Microsoft.DotNet.Interactive.Jupyter;
 
 namespace Microsoft.DotNet.Interactive.Tests
 {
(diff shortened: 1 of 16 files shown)


src/Microsoft.DotNet.Interactive/Parsing/SubmissionParser.cs
```diff
@@ -38,26 +38,26 @@ public SubmissionParser(KernelBase kernel)
...
-        public PolyglotSyntaxTree Parse(string code)
+        public PolyglotSyntaxTree Parse(string code, string defaultLanguage = null)
         {
             var sourceText = SourceText.From(code);
 
             var parser = new PolyglotSyntaxParser(
-                sourceText, 
-                DefaultLanguage, 
+                sourceText,
+                defaultLanguage ?? DefaultLanguage,
```
brettfo, line 47: Incoming language service requests might have a different language specified, so we can't assume the default.

src/Microsoft.DotNet.Interactive.Tests/LanguageKernelTestBase.cs
```diff
@@ -74,33 +75,40 @@ public void Dispose()
             _lockReleaser.Dispose();
         }
 
+        protected CompositeKernel CreateFullKernel(Language defaultKernelLanguage = Language.CSharp)
```
jonsequitur, line 78:
```suggestion
        protected CompositeKernel CreateCompositeKernel(Language defaultKernelLanguage = Language.CSharp)
```

It's good to avoid introducing alternative terms for an already-named concept.

src/Microsoft.DotNet.Interactive/Events/CompletionRequestCompleted.cs
```diff
@@ -10,19 +10,43 @@ namespace Microsoft.DotNet.Interactive.Events
...
             CompletionList = completionList ?? throw new ArgumentNullException(nameof(completionList));
-            Range = range;
+            _range = range;
         }
 
         /// <summary>
         /// The range of where to replace in a completion request.
         /// </summary>
-        public LinePositionSpan? Range { get; set; }
+        public LinePositionSpan? Range
```
jonsequitur, line 27: Can we dedupe this implementation with the one in `HoverTextProduced`?

src/Microsoft.DotNet.Interactive/KernelBase.cs
```diff
@@ -179,6 +183,60 @@ private async Task SetKernel(IKernelCommand command, KernelInvocationContext con
...
         }
 
+        private async Task HandleLinePositionLanguageServiceRequests(
+            LanguageServiceCommandBase languageServiceCommand,
+            KernelInvocationContext context,
+            KernelPipelineContinuation continueOnCurrentPipeline)
+        {
+            var tree = SubmissionParser.Parse(languageServiceCommand.Code, languageServiceCommand.TargetKernelName);
+            var nodes = tree.GetRoot().ChildNodes.ToArray();
+            if (nodes.Length == 1)
```
jonsequitur, line 193: Rather than inferring this (as it might not be future-proof as the parser evolves), we might generalize the approach taken by `SubmitCode.SuppressSplit`.

src/Microsoft.DotNet.Interactive/Commands/LanguageServiceCommandBase.cs
```diff
@@ -0,0 +1,22 @@
...
+        public LinePosition Position { get; protected set; }
+
+        protected LanguageServiceCommandBase(string code, LinePosition position, string targetKernelName = null, IKernelCommand parent = null)
+            : base(targetKernelName, parent)
+        {
+            Code = code;
+            Position = position;
+        }
+
+        internal abstract LanguageServiceCommandBase WithCodeAndPosition(string code, LinePosition position);
```
jonsequitur, line 20: Is this intended to return a new instance? The `With___` naming kind of implies that.

brettfo: Yes, a new instance is intended.

src/Microsoft.DotNet.Interactive/KernelBase.cs
```diff
@@ -179,6 +183,60 @@ private async Task SetKernel(IKernelCommand command, KernelInvocationContext con
...
+            {
+                // no splitting happened, continue like normal
+                await continueOnCurrentPipeline(languageServiceCommand, context);
```
jonsequitur, line 196: All of this logic is pretty awkward and I'm super hesitant to duplicate it, especially since we already know this also needs to differ between completions (i.e. magic commands need to broadcast to all subkernels) vs hover text.

We should discuss how we can simplify.

src/Microsoft.DotNet.Interactive/Parsing/SubmissionParser.cs
jonsequitur: Can we remove the default property then, so this is less confusing?

brettfo:
@jonsequitur No, line remappings for diagnostics won't work with this because the kernels simply `.ToString()` their errors.  If we add a `DiagnosticEvent` or similar, we can add a `Range` property of type `LinePositionSpan` and do the same one-line hack we use in `HoverTextProduced` and `CompletionRequestCompleted`.  I think that should probably be the next PR after this.

jonsequitur:
Maybe the kernels can pass richer error objects that let the final stringification happen elsewhere?

PR #518, opened 2020-06-09
initial magic command completion support

jonsequitur:


src/Microsoft.DotNet.Interactive/Parsing/SubmissionParser.cs
```diff
@@ -258,31 +256,39 @@ public void AddDirective(Command command)
...
+                ICommand _ => "Method",
+                _ => throw new ArgumentOutOfRangeException(nameof(symbol))
+            };
+
+            var helpBuilder = new DirectiveHelpBuilder(
+                new TestConsole(),
+                parser.Configuration.RootCommand.Name);
+
+            return new CompletionItem(
+                name,
```
brettfo, line 279: Specify argument names.

src/Microsoft.DotNet.Interactive.Tests/LanguageServices/CompletionTests.cs
```diff
@@ -144,5 +151,44 @@ public async Task Completions_are_available_for_symbols_declared_in_a_submission
                 .Should()
                 .Contain(item => item.DisplayText == variableName);
         }
+
+        [Theory]
+        [InlineData("[|#!c|]", "#!csharp", Skip = "Composite kernel magic command completions not working yet")]
+        [InlineData("[|#!w|]", "#!who,#!whos")]
+        [InlineData("[|#!w|]\n", "#!who,#!whos")]
+        [InlineData("[|#!w|] \n", "#!who,#!whos")]
+        public async Task Completions_are_available_for_magic_commands(
```
brettfo, line 160: Add test for just `#!` (unless it already exists and I just missed it.)

jonsequitur: I'll add this in the next round of changes. There are a bunch of additional test cases (e.g. for documentation and options) but the size of the refactor made me want to merge sooner rather than later.

PR #521, opened 2020-06-09
Improve telemetry

colombod:
consolidate app insight libraries
introduce directive to track frontend telemetry
[vscode] stdio will produce a frontend property with value vscode 
frontend properties are not hashed


src/dotnet-interactive/ContentFiles/kernels/.net-csharp/kernel.json
```diff
@@ -2,6 +2,7 @@
   "argv": [
     "dotnet",
     "interactive",
+    "[jupyter]",
```
jonsequitur, line 5: This should be unnecessary, since we already know the frontend based on the verb.

colombod: `dotnet interactive jupyter install` will have jupyter as verb but no fronted while the kernel spec change is explicit about it.

Maybe processing that verb to decide about the production of a frontend proeprty for the juptyer verb when is not a jupyter install gesture?

jonsequitur: > Maybe processing that verb to decide about the production of a frontend proeprty for the juptyer verb when is not a jupyter install gesture?

That should work, no?

colombod: will do that then

PR #527, opened 2020-06-10
refactor `SubmitCode` handling to use IKernelCommandHandler<> interface

brettfo:
Required a refactor of the F# kernel to allow interface implementations with different generic instantiations.  See dotnet/fsharp#2867 for details.

In doing this I wanted to prevent the old F# kernel from being instantiated so that we don't get into a situation where we call `new FSharpKernel()` and it doesn't properly handle commands.  To do this I (1) changed the name to `FSharpKernelInternal`, (2) set the constructor to private, and (3) forced creation through a `CreateKernelForWrapper()` method.  This way it's very explicit when and where we instantiate this.  Spoiler alert, it only happens in the C# wrapper class, now.

Second, I created a wrapper in C# that simply forwards everything to the underlying kernel as appropriate and rewrites the `HandleAsync` methods to `Handle{specific-thing}Async`.

Third, I wanted to keep the contents of the F# kernel extensions, but allow them to be applied to the new C# wrapper.  To accomplish this I made the method generic with the constraint `IFSharpKernel and KernelBase`.  This keeps the extension methods strongly typed to F# and will prevent the C#/PowerShell kernels from accidentally injecting F#-specific helper code.

Finally, it was done this way so that when the time comes for us to remove the C# wrapper, the steps will be:
1. Remove C# wrapper `FSharpKernel`.
2. Rename `FSharpKernelInternal` -> `FSharpKernel` and make the constructor public again.
3. Remove/replace all instances of the `Microsoft.DotNet.Interactive.FSharp.Wrapper` namespace.

Otherwise everything is basically unchanged.

### DIFF
--- a/src/Microsoft.DotNet.Interactive/KernelBase.cs
+++ b/src/Microsoft.DotNet.Interactive/KernelBase.cs
@@ -348,10 +348,6 @@ namespace Microsoft.DotNet.Interactive
             _disposables.Add(disposable);
         }
 
-        protected abstract Task HandleSubmitCode(
-            SubmitCode command, 
-            KernelInvocationContext context);
-
         private protected void SetHandler(
             IKernelCommand command,
             KernelInvocationContext context)
@@ -363,10 +359,10 @@ namespace Microsoft.DotNet.Interactive
                     switch (command)
                     {
                         case SubmitCode submitCode:
-                            submitCode.Handler = (_, invocationContext) =>
+                            if (this is IKernelCommandHandler<SubmitCode> submitHandler)
                             {
-                                return HandleSubmitCode(submitCode, context);
-                            };
+                                SetHandler(submitHandler, submitCode);
+                            }
                             break;
 
                         case RequestCompletion requestCompletion:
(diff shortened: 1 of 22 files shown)


src/Microsoft.DotNet.Interactive/KernelBase.cs
```diff
@@ -348,10 +348,6 @@ public void RegisterForDisposal(IDisposable disposable)
             _disposables.Add(disposable);
         }
 
-        protected abstract Task HandleSubmitCode(
-            SubmitCode command, 
-            KernelInvocationContext context);
-
         private protected void SetHandler(
```
colombod, line 351: This could be
```
        private protected void SetHandler(
            IKernelCommand command,
            KernelInvocationContext context)
        {
            if (command is KernelCommandBase kb)
            {
                if (kb.Handler == null)
                {
                    switch(command, this)
                    {
                        case (SubmitCode submitCode, IKernelCommandHandler<SubmitCode> submitHandler) :
                            SetHandler(submitHandler, submitCode);
                            break;
                        case (RequestCompletion requestCompletion, IKernelCommandHandler<RequestCompletion> completionHandler):
                            SetCompletionHandler(requestCompletion, completionHandler);
                            break;
                        case (RequestHoverText hoverCommand, IKernelCommandHandler<RequestHoverText> hoverHandler):
                            SetHandler(hoverHandler, hoverCommand);
                            break;
                        case (ChangeWorkingDirectory cwdCommand, _):
                            cwdCommand.Handler = (_, __) =>
                            {
                                Directory.SetCurrentDirectory(cwdCommand.WorkingDirectory.FullName);
                                context.Publish(new WorkingDirectoryChanged(cwdCommand.WorkingDirectory, cwdCommand));
                                return Task.CompletedTask;
                            };
                            break;
                    }
                }
            }
        }
```

brettfo: I'll deflect to @jonsequitur on this one; I have no preference.

Issue #535, opened 2020-06-12
Intellisense not showing loaded extensions


colombod:
### Describe the bug
After loading an extension that introduces a new magic command the new command is not part of completion list.

The extension adds a directive to the composite kernel

**Which version of .NET Interactive are you using?** 
1.0.131103

- OS
    - [X] Windows 10
    - [ ] macOS
    - [ ] Linux (Please specify distro)
    - [ ] iOS
    - [ ] Android
 - Browser 
    - [ ] Chrome
    - [ ] Edge
    - [ ] Firefox
    - [ ] Safari
 - Frontend
    - [ ] Jupyter Notebook
    - [ ] Jupyter Lab
    - [ ] nteract
    - [X] Visual Studio Code
    - [ ] Other (please specify)

### Screenshots
![image](https://user-images.githubusercontent.com/375556/84473853-ac517680-ac81-11ea-94ef-977eca93375c.png)

PR #536, opened 2020-06-12
Completion list should contain directives added dynamically

colombod:
Related to #535


src/Microsoft.DotNet.Interactive.Tests/KernelExtensions.cs
```diff
@@ -0,0 +1,21 @@
+// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+
+namespace Microsoft.DotNet.Interactive.Tests
+{
+    public static class KernelExtensions
+    {
+        public static IKernel FindKernel(this IKernel kernel, Language name)
```
jonsequitur, line 10: There's a public method in `Microsoft.DotNet.Interactive` to do this.

jonsequitur: Oh, I see, this is just shimming the enum. 

You can dedupe the string conversion logic by calling `name.LanguageName()`.

src/Microsoft.DotNet.Interactive.Tests/LanguageServices/CompletionTests.cs
```diff
@@ -190,5 +192,66 @@ public async Task Completions_are_available_for_symbols_declared_in_a_submission
...
+                Handler = CommandHandler.Create(() => { })
+            });
+
+                kernelToExtend.AddDirective(new Command("#!twilight")
+            {
+                Handler = CommandHandler.Create(() => { })
+            });
+
+
+            foreach (var position in Enumerable.Range(span.Start, span.Length + 1))
```
jonsequitur, line 237: Rather than duplicate this, I wonder if we shouldn't extract a custom assertion. There are more completion tests to be written still, and we'll also want to add verifications for things like documentation and item kind.

src/Microsoft.DotNet.Interactive.Tests/KernelExtensions.cs
colombod: perfect

src/Microsoft.DotNet.Interactive.PowerShell.Tests/Microsoft.DotNet.Interactive.PowerShell.Tests.v3.ncrunchproject
```diff
@@ -0,0 +1,15 @@
+﻿<ProjectConfiguration>
```
jonsequitur, line 1: Please exclude this.

PR #538, opened 2020-06-13
expand grammar for magic commands

brettfo:
Instead of classifying the whole line `#!share --from csharp x` as a comment, we can now mark the magic command itself as different from arguments that get passed to it, etc.

image

No functional change, purely display only.

### DIFF
--- a/src/dotnet-interactive-vscode/syntaxes/source.dotnet-interactive.magic-commands.tmGrammar.json
+++ b/src/dotnet-interactive-vscode/syntaxes/source.dotnet-interactive.magic-commands.tmGrammar.json
@@ -4,19 +4,27 @@
     "patterns": [
         {
             "name": "comment.line.magic-commands",
-            "match": "^#!(?!([cf](#|s(harp)?)|powershell|pwsh|html|javascript|js|markdown|md).*$)"
-        },
-        {
-            "include": "#magic-command-name"
-        },
-        {
-            "include": "#flag-argument"
-        },
-        {
-            "include": "#parameter"
-        },
-        {
-            "include": "#strings"
+            "begin": "^(#!)(?!([cf](#|s(harp)?)|powershell|pwsh|html|javascript|js|markdown|md))",
+            "end": "(?<=$)",
+            "beginCaptures": {
+                "1": {
+                    "name": "comment.line.magic-commands.hash-bang"
+                }
+            },
+            "patterns": [
+                {
+                    "include": "#magic-command-name"
+                },
+                {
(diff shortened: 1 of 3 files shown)


src/dotnet-interactive-vscode/syntaxes/source.dotnet-interactive.magic-commands.tmGrammar.json
```diff
@@ -1,9 +1,63 @@
...
+    "repository": {
+        "magic-command-name": {
+            "patterns": [
+                {
+                    "name": "keyword.control.magic-commands",
+                    "match": "(?<=^#!)[a-zA-Z0-9_-]+"
+                }
+            ]
+        },
+        "flag-argument": {
```
jonsequitur, line 31: In GNU/POSIX command line terminology and thus in System.CommandLine this is an "option". The Windows equivalent includes e.g. `/p` so you should add `/` as a prefix to this regex.

src/dotnet-interactive-vscode/syntaxes/source.dotnet-interactive.magic-commands.tmGrammar.json
```diff
@@ -1,9 +1,63 @@
...
+            "patterns": [
+                {
+                    "name": "constant.language.magic-commands",
+                    "match": "--?[^\\s-]+"
+                }
+            ]
+        },
+        "parameter": {
```
jonsequitur, line 39: In GNU/POSIX this would be an argument.

src/dotnet-interactive-vscode/syntaxes/source.dotnet-interactive.magic-commands.tmGrammar.json
```diff
@@ -1,9 +1,63 @@
...
+            "patterns": [
+                {
+                    "name": "variable.parameter.magic-commands",
+                    "match": "[_a-zA-Z0-9]+"
```
jonsequitur, line 43: This looks overly restrictive.

brettfo: What about any non-whitespace and non-quote character so we don't match on `abcd"efg`?

brettfo: Update: I've changed the regex to be any non-quote, non-whitespace character: `[^\s\"]+`

PR #540, opened 2020-06-15
doc updates

jonsequitur:


jonsequitur:
I realized we have developer instructions in three different documents. Consolidating...

PR #544, opened 2020-06-16
support notebook backup

brettfo:
Trying to stay on top of a notebook breaking change [here](https://github.com/microsoft/vscode/issues/93265#issuecomment-644443400).

`backupNotebook()` is periodically called by VS Code to save the state of the notebook.  The context object passed in is a file whose parent directory may or may not exist, so once we ensure it's present, we call into our existing `save()` function and provide a mechanism to remove the temporary file.


jonsequitur:
Tests?

rebornix:
FYI, we decided (for now) that when users choose to revert the file, we will invoke `NotebookContentProvider#openNotebook` for the notebook data instead of asking `NotebookContentProvider` to make modifications directly to the notebook editor/document.

Issue #512, opened 2020-06-06
Fix up API names and serialization formats


jonsequitur:
It would be good to get some breaking name changes done before more widely documenting direct API-level use of .NET Interactive. This includes NuGet packages and the messaging APIs. 

_**If you have suggestions for naming changes, please let us know.**_

Here is a non-exhaustive list of types where renaming could be improved. Typically, these types have either evolved such that their original purpose is no longer well-described by their names, or because the naming was inconsistent with other naming conventions.

### Commands

- [x] ~`IKernelCommand`~
- [x] `KernelCommandBase` -> `KernelCommand`
- [x] `RequestCompletion.Position` is actually a `LinePosition`

### Events
- [x] ~`IKernelEvent`~
- [x] `KernelEventBase` -> `KernelEvent`
- [x] `CommandHandled` -> `CommandSucceeded` (to better match `CommandFailed`, the other terminal event
- [x] `CompletionRequestCompleted`. For consistency this would become `CompletionsProduced`, but...
- [ ] Consider changing `____Produced` events, e.g. `HoverTextProduced` -> `HoverText____`  "Produced" is kind of stilted.
- [x] `CompletionRequestCompleted.Range` is actually a `LinePositionSpan`
- [x] Remove `Value` property from serialized forms of `DisplayedValueProduced`, `DisplayedValueUpdated`, `ErrorProduced`, `ReturnValueProduced`, `StandardOutputValueProduced`, and `StandardErrorValueProduced`
- [x] Remove `CompletionRequestReceived`
- [x] Stop using `FileInfo` and `DirectoryInfo` in event classes.

### Kernel names

- [x] ~`IKernel`~ 
- [x] `KernelBase` -> `Kernel`

### Other types

- [ ] `ISupportNuget` can be made a bit clearer and more specific.

PR #551, opened 2020-06-17
Fixed how PSCustomObject instances are displayed with OutDisplay

justinytchen:
When running something like `[pscustomobject]@{
    a = 'b';
    c = 'd'
  } | Out-Display` nothing is shown because of how the InputObject is used in OutDisplayCommand.

Created fix by checking for the case that the InputObject is a PSCustomObject and converting it to a PSObject to find the properties.

### DIFF
--- a/src/Microsoft.DotNet.Interactive.PowerShell.Tests/PowerShellKernelTests.cs
+++ b/src/Microsoft.DotNet.Interactive.PowerShell.Tests/PowerShellKernelTests.cs
@@ -249,7 +249,7 @@ for ($j = 0; $j -le 4; $j += 4 ) {
                 e => e.Should().BeOfType<CodeSubmissionReceived>(),
                 e => e.Should().BeOfType<CompleteCodeSubmissionReceived>(),
                 e => e.Should().BeOfType<DisplayedValueProduced>().Which.Value.Should().BeEquivalentTo(props),
-                e => e.Should().BeOfType<CommandHandled>()
+                e => e.Should().BeOfType<CommandSucceeded>()
             );
         }
     }
--- a/src/Microsoft.DotNet.Interactive.PowerShell.Tests/PowerShellKernelTests.cs
+++ b/src/Microsoft.DotNet.Interactive.PowerShell.Tests/PowerShellKernelTests.cs
@@ -245,10 +245,10 @@ for ($j = 0; $j -le 4; $j += 4 ) {
             var result = await kernel.SendAsync(new SubmitCode("[pscustomobject]@{ prop1 = 'value1'; prop2 = 'value2'; prop3 = 'value3' } | Out-Display"));
             var outputs = result.KernelEvents.ToSubscribedList();
 
-            Assert.Collection(outputs,
+            outputs.Should().SatisfyRespectively(
                               e => e.Should().BeOfType<CodeSubmissionReceived>(),
                               e => e.Should().BeOfType<CompleteCodeSubmissionReceived>(),
-                              e => e.Should().BeOfType<DisplayedValueProduced>().Which.FormattedValues.Should().Equals(props),
+                              e => e.Should().BeOfType<DisplayedValueProduced>().Which.Value.Should().BeEquivalentTo(props),
                               e => e.Should().BeOfType<CommandHandled>()
                              );
         }
--- a/src/Microsoft.DotNet.Interactive.PowerShell.Tests/PowerShellKernelTests.cs
+++ b/src/Microsoft.DotNet.Interactive.PowerShell.Tests/PowerShellKernelTests.cs
@@ -55,7 +55,6 @@ for ($j = 0; $j -le 4; $j += 4 ) {
 
             var events = result.KernelEvents.ToSubscribedList();
 
-            var formattedValues = ((DisplayEventBase)(result.KernelEvents.ToSubscribedList()[2])).FormattedValues;
             Assert.Collection(events,
                                        e => e.Should().BeOfType<CodeSubmissionReceived>(),
                                        e => e.Should().BeOfType<CompleteCodeSubmissionReceived>(),
--- a/src/Microsoft.DotNet.Interactive.PowerShell.Tests/PowerShellKernelTests.cs
+++ b/src/Microsoft.DotNet.Interactive.PowerShell.Tests/PowerShellKernelTests.cs
@@ -232,25 +232,24 @@ for ($j = 0; $j -le 4; $j += 4 ) {
         }
 
         [Fact]
-        public async Task Powershell_customobject_is_formatted_for_outdisplay()
+        public async Task PowerShell_Custom_Object_OutDisplay_Passed()
         {
-            var obj = new PSObject();
-            obj.Properties.Add(new PSNoteProperty("prop1", "value1"));
-            obj.Properties.Add(new PSNoteProperty("prop2", "value2"));
-            obj.Properties.Add(new PSNoteProperty("prop3", "value3"));
-            var propDict = obj.Properties.ToDictionary(x => x.Name, x => x.Value);
+            var props = (new Dictionary<string, object>
+                {
+                    { "prop1", "value1" },
+                    { "prop2", "value2" },
+                    { "prop3", "value3" }
+                });
 
             var kernel = CreateKernel(Language.PowerShell);
             var result = await kernel.SendAsync(new SubmitCode("[pscustomobject]@{ prop1 = 'value1'; prop2 = 'value2'; prop3 = 'value3' } | Out-Display"));
             var outputs = result.KernelEvents.ToSubscribedList();
 
             outputs.Should().SatisfyRespectively(
-                                  e => e.Should().BeOfType<CodeSubmissionReceived>(),
-                                  e => e.Should().BeOfType<CompleteCodeSubmissionReceived>(),
-                                  e => e.Should().BeOfType<DisplayedValueProduced>().
-                                        Which.Value.Should().BeOfType<PSObject>().
-                                        Which.Properties.ToDictionary(x => x.Name, x => x.Value).Should().BeEquivalentTo(propDict),
-                                  e => e.Should().BeOfType<CommandHandled>()
+                              e => e.Should().BeOfType<CodeSubmissionReceived>(),
+                              e => e.Should().BeOfType<CompleteCodeSubmissionReceived>(),
+                              e => e.Should().BeOfType<DisplayedValueProduced>().Which.Value.Should().BeEquivalentTo(props),
+                              e => e.Should().BeOfType<CommandHandled>()
                              );
         }
     }
--- a/src/Microsoft.DotNet.Interactive.PowerShell.Tests/PowerShellKernelTests.cs
+++ b/src/Microsoft.DotNet.Interactive.PowerShell.Tests/PowerShellKernelTests.cs
@@ -232,23 +232,20 @@ for ($j = 0; $j -le 4; $j += 4 ) {
         }
 
         [Fact]
-        public async Task Powershell_customobject_is_parsed_for_outdisplay()
+        public async Task Powershell_customobject_is_formatted_for_outdisplay()
         {
-            var props = (new Dictionary<string, object>
-            {
-                { "prop1", "value1" },
-                { "prop2", "value2" },
-                { "prop3", "value3" }
-            });
-
             var kernel = CreateKernel(Language.PowerShell);
             var result = await kernel.SendAsync(new SubmitCode("[pscustomobject]@{ prop1 = 'value1'; prop2 = 'value2'; prop3 = 'value3' } | Out-Display"));
             var outputs = result.KernelEvents.ToSubscribedList();
 
+            string mimeType = "text/html";
+            string formattedHtml = "<table><thead><tr><th><i>key</i></th><th>value</th></tr></thead><tbody><tr><td>prop1</td><td>value1</td></tr><tr><td>prop2</td><td>value2</td></tr><tr><td>prop3</td><td>value3</td></tr></tbody></table>";
+            FormattedValue fv = new FormattedValue(mimeType, formattedHtml);
+
             outputs.Should().SatisfyRespectively(
                 e => e.Should().BeOfType<CodeSubmissionReceived>(),
                 e => e.Should().BeOfType<CompleteCodeSubmissionReceived>(),
-                e => e.Should().BeOfType<DisplayedValueProduced>().Which.Value.Should().BeEquivalentTo(props),
+                e => e.Should().BeOfType<DisplayedValueProduced>().Which.FormattedValues.ElementAt(0).Should().BeEquivalentTo(fv),
                 e => e.Should().BeOfType<CommandSucceeded>()
             );
         }
--- a/src/Microsoft.DotNet.Interactive.PowerShell.Tests/PowerShellKernelTests.cs
+++ b/src/Microsoft.DotNet.Interactive.PowerShell.Tests/PowerShellKernelTests.cs
@@ -4,6 +4,7 @@
 using System;
 using System.IO;
 using System.Management.Automation;
+using System.Collections.Generic;
 using System.Threading.Tasks;
 using FluentAssertions;
 using System.Linq;
@@ -54,6 +55,7 @@ for ($j = 0; $j -le 4; $j += 4 ) {
 
             var events = result.KernelEvents.ToSubscribedList();
 
+            var formattedValues = ((DisplayEventBase)(result.KernelEvents.ToSubscribedList()[2])).FormattedValues;
             Assert.Collection(events,
                                        e => e.Should().BeOfType<CodeSubmissionReceived>(),
                                        e => e.Should().BeOfType<CompleteCodeSubmissionReceived>(),
@@ -228,6 +230,28 @@ for ($j = 0; $j -le 4; $j += 4 ) {
 
                 File.Delete(_allUsersCurrentHostProfilePath);
             }
+        }
+
+        [Fact]
+        public async Task PowerShell_Custom_Object_OutDisplay_Passed()
+        {
+            var props = (new Dictionary<string, object>
+                {
+                    { "prop1", "value1" },
+                    { "prop2", "value2" },
+                    { "prop3", "value3" }
+                });
+
+            var kernel = CreateKernel(Language.PowerShell);
+            var result = await kernel.SendAsync(new SubmitCode("[pscustomobject]@{ prop1 = 'value1'; prop2 = 'value2'; prop3 = 'value3' } | Out-Display"));
+            var outputs = result.KernelEvents.ToSubscribedList();
+
+            Assert.Collection(outputs,
+                              e => e.Should().BeOfType<CodeSubmissionReceived>(),
+                              e => e.Should().BeOfType<CompleteCodeSubmissionReceived>(),
+                              e => e.Should().BeOfType<DisplayedValueProduced>().Which.FormattedValues.Should().Equals(props),
+                              e => e.Should().BeOfType<CommandHandled>()
+                             );
         }
     }
 }
--- a/src/Microsoft.DotNet.Interactive.PowerShell.Tests/PowerShellKernelTests.cs
+++ b/src/Microsoft.DotNet.Interactive.PowerShell.Tests/PowerShellKernelTests.cs
@@ -232,24 +232,25 @@ for ($j = 0; $j -le 4; $j += 4 ) {
         }
 
         [Fact]
-        public async Task PowerShell_Custom_Object_OutDisplay_Passed()
+        public async Task Powershell_customobject_is_formatted_for_outdisplay()
         {
-            var props = (new Dictionary<string, object>
-                {
-                    { "prop1", "value1" },
-                    { "prop2", "value2" },
-                    { "prop3", "value3" }
-                });
+            var obj = new PSObject();
+            obj.Properties.Add(new PSNoteProperty("prop1", "value1"));
+            obj.Properties.Add(new PSNoteProperty("prop2", "value2"));
+            obj.Properties.Add(new PSNoteProperty("prop3", "value3"));
+            var propDict = obj.Properties.ToDictionary(x => x.Name, x => x.Value);
 
             var kernel = CreateKernel(Language.PowerShell);
             var result = await kernel.SendAsync(new SubmitCode("[pscustomobject]@{ prop1 = 'value1'; prop2 = 'value2'; prop3 = 'value3' } | Out-Display"));
             var outputs = result.KernelEvents.ToSubscribedList();
 
             outputs.Should().SatisfyRespectively(
-                              e => e.Should().BeOfType<CodeSubmissionReceived>(),
-                              e => e.Should().BeOfType<CompleteCodeSubmissionReceived>(),
-                              e => e.Should().BeOfType<DisplayedValueProduced>().Which.Value.Should().BeEquivalentTo(props),
-                              e => e.Should().BeOfType<CommandHandled>()
+                                  e => e.Should().BeOfType<CodeSubmissionReceived>(),
+                                  e => e.Should().BeOfType<CompleteCodeSubmissionReceived>(),
+                                  e => e.Should().BeOfType<DisplayedValueProduced>().
+                                        Which.Value.Should().BeOfType<PSObject>().
+                                        Which.Properties.ToDictionary(x => x.Name, x => x.Value).Should().BeEquivalentTo(propDict),
+                                  e => e.Should().BeOfType<CommandHandled>()
                              );
         }
     }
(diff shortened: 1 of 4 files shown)


justinytchen:
@TylerLeonhardt

jonsequitur:
Thanks, @justinytchen! Can you add a test?

src/Microsoft.DotNet.Interactive.PowerShell/Commands/OutDisplayCommand.cs
```diff
@@ -54,7 +55,21 @@ public sealed class OutDisplayCommand : PSCmdlet
         /// </summary>
         protected override void ProcessRecord()
         {
+            
             object obj = InputObject is PSObject psObject ? psObject.BaseObject : InputObject;
+
+            // If object is PSCustomObject, convert to PSObject to find the properties to display
+            if (InputObject is PSObject && ((PSObject)InputObject).BaseObject is PSCustomObject)
```
TylerLeonhardt, line 62: you should be able to do something like:

```suggestion
            if (InputObject is PSObject psObj && psObj.BaseObject is PSCustomObject)
```

which will create a `psObj` variable that you can use later if `InputObject is PSObject`

src/Microsoft.DotNet.Interactive.PowerShell/Commands/OutDisplayCommand.cs
```diff
@@ -54,7 +55,21 @@ public sealed class OutDisplayCommand : PSCmdlet
...
+            {
+                Dictionary<string, object> table = new Dictionary<string, object>();
+                var props = ((PSObject)InputObject).Properties;
```
TylerLeonhardt, line 65: With the above change, you can simplify this to:

```suggestion
                var props = psObj.Properties;
```

src/Microsoft.DotNet.Interactive.PowerShell/Commands/OutDisplayCommand.cs
```diff
@@ -54,7 +55,21 @@ public sealed class OutDisplayCommand : PSCmdlet
         /// </summary>
         protected override void ProcessRecord()
         {
+            
             object obj = InputObject is PSObject psObject ? psObject.BaseObject : InputObject;
```
TylerLeonhardt, line 59: what if we broke up this ternary operator and did:

```csharp
object obj;

if (InputObject is PSObject psObject)
{
    if (psObject is PSCustomObject)
    {
          // do your code here
    }
    else
    {
        obj = psObject.BaseObject;
    }
}
else
{
    obj = InputObject
}
```

TylerLeonhardt:
@daxian-dbw you mentioned to me that we should consider doing this for _all_ `PSObject`s because `PSObject`s can have extra properties added to them due to PowerShell's Extended Type System (a great example is `$profile`? Can you elaborate on that?

src/Microsoft.DotNet.Interactive.PowerShell/Commands/OutDisplayCommand.cs
TylerLeonhardt: Yeah my bad that should be `psObject.BaseObject`

TylerLeonhardt:
But what would you expect the Formatter to show? Exactly what PowerShell shows today?

dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/interactive?pullRequest=551) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/interactive?pullRequest=551) before we can accept your contribution.<br/><br/>:x: justinytchen [sign now](https://cla.dotnetfoundation.org/dotnet/interactive?pullRequest=551)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/interactive?pullRequest=551) it.</sub>

src/Microsoft.DotNet.Interactive.PowerShell.Tests/PowerShellKernelTests.cs
```diff
@@ -216,5 +217,27 @@ public async Task VerifyAllUsersProfileRuns()
...
+                    { "prop1", "value1" },
+                    { "prop2", "value2" },
+                    { "prop3", "value3" }
+                });
+
+            var kernel = CreateKernel(Language.PowerShell);
+            var result = await kernel.SendAsync(new SubmitCode("[pscustomobject]@{ prop1 = 'value1'; prop2 = 'value2'; prop3 = 'value3' } | Out-Display"));
+            var outputs = result.KernelEvents.ToSubscribedList();
+
+            Assert.Collection(outputs,
```
jonsequitur, line 235: For a more purely FluentAssertions approach, you can change this line to:

```csharp
outputs.Should().SatisfyRespectively(
```

src/Microsoft.DotNet.Interactive.PowerShell.Tests/PowerShellKernelTests.cs
```diff
@@ -216,5 +217,27 @@ public async Task VerifyAllUsersProfileRuns()
...
+                              e => e.Should().BeOfType<CodeSubmissionReceived>(),
+                              e => e.Should().BeOfType<CompleteCodeSubmissionReceived>(),
+                              e => e.Should().BeOfType<DisplayedValueProduced>().Which.FormattedValues.Should().Equals(props),
```
jonsequitur, line 238: This is a common gotcha. `Equals` here is `object.Equals` and if it returns false, the test will still pass.

```suggestion
                              e => e.Should().BeOfType<DisplayedValueProduced>().Which.FormattedValues.Should().Be(props),
```

src/Microsoft.DotNet.Interactive.PowerShell.Tests/PowerShellKernelTests.cs
```diff
@@ -216,5 +217,27 @@ public async Task VerifyAllUsersProfileRuns()
                 File.Delete(_allUsersCurrentHostProfilePath);
             }
         }
+
+        [Fact]
+        public async Task PowerShell_Custom_Object_OutDisplay_Passed()
```
jonsequitur, line 222: I would rename this. All tests "should pass" so "passed" is not adding information.

What specifically can we expect about the output?

daxian-dbw:
I don't know much about how that formatter works, so it's just a question for you to think if that can help.
In general, what would you expect to display for an object with ETS properties attached? Take the `Process` object returned from `Get-Process` as an example, how would you expect those extended properties to be used in display? 

I don't know how the `Display` method works, so assuming it depends on type matching and reflection, then is it possible to have new code handle the reflection of PSObject (namely, as opposed to reflection, use the `Properties` member of a `PSObject`)?

jonsequitur:
You can add a custom formatter that outputs whatever would be the most sensible. Here are some examples of defaults:

https://github.com/dotnet/interactive/blob/51cac3166e3f4413c5b044229f9e10744a0b9651/src/Microsoft.DotNet.Interactive.Formatting/DefaultPlainTextFormatterSet.cs#L60

TylerLeonhardt:
I think this change is sufficient for now cause it does fix a bug (`pscustomobject`'s not showing up)... however if @justinytchen would like to mess around with this some more, can you try:

* Creating a `Formatter` for `PSObject` (that iterates over the Properties like you're doing now)
* Out-Display then simply passes the `PSObject` directly into `display()`

Then Get-Process | Out-GridView should look closer to this:

```
 NPM(K)    PM(M)      WS(M)     CPU(s)      Id  SI ProcessName
 ------    -----      -----     ------      --  -- -----------
      0     0.00      82.27     285.23     592   1 Spotlight
      0     0.00      20.09       1.55   81655 …85 SqlToolsResourc
      0     0.00       0.43       2.01     703   1 ssh-agent
      0     0.00       0.79       0.33     520   1 storeaccountd
      0     0.00       0.30       0.22    9827   1 storedownloadd
      0     0.00       9.02      10.70    9825   1 storeuid
      0     0.00      47.55     331.04     584   1 Stream Deck
      0     0.00      13.13      25.90   48902   1 studentd
      0     0.00      37.11     170.95     543   1 suggestd
      0     0.00       2.64     108.80     632   1 swcd
      0     0.00      33.48     754.50     553   1 SystemUIServer
      0     0.00      10.75       9.79     490   1 talagent
      0     0.00       5.81      29.77     467   1 tccd
      0     0.00     220.05     502.79   61330   1 Teams
      0     0.00     130.24      33.05   85422   1 Telegram
      0     0.00       8.77      24.04     581   1 TextInputMenuAg
      0     0.00       1.09       2.44   45058   1 transparencyd
      0     0.00      13.68     530.40     461   1 trustd
      0     0.00       2.25       2.30     483   1 TrustedPeersHel
      0     0.00       9.02       9.76   16504   1 universalAccess
      0     0.00       2.85       2.90   23344   1 UsageTrackingAg
      0     0.00       0.11       0.54   15691   1 USBAgent
      0     0.00       4.97     109.34     512   1 useractivityd
      0     0.00       6.96     140.35     454   1 UserEventAgent
      0     0.00      10.74      48.82     476   1 usernoted
      0     0.00       5.32     132.23     493   1 ViewBridgeAuxil
      0     0.00      16.21      13.04   43113   1 vsls-agent
      0     0.00      43.79       1.84   79756   1 vsls-agent
      0     0.00       0.55       0.31   22538 …38 VTDecoderXPCSer
      0     0.00       3.68       0.07   61355 …55 VTDecoderXPCSer
      0     0.00       1.44       5.92   67071 …71 VTDecoderXPCSer
      0     0.00       7.21     602.40   67697 …97 VTDecoderXPCSer
      0     0.00       3.48       0.07   81589 …89 VTDecoderXPCSer
      0     0.00      26.18     985.84     492   1 WiFiAgent
      0     0.00       1.12       1.11     604   1 WiFiVelocityAge
      0     0.00       1.51       8.99   20207 …07 XprotectService
      0     0.00      18.02      34.22   62998   1 Zappy
```

Instead of the giant mess that it is now lol

justinytchen:
@TylerLeonhardt 
I can create a separate branch/PR for the changes specifically for PSObject (and leave the PSCustomObject changes on this one)

src/Microsoft.DotNet.Interactive.PowerShell/PowerShellKernel.cs
```diff
@@ -70,6 +72,32 @@ static PowerShellKernel()
             // Add accelerators that exist in other namespaces.
             addAccelerator.Invoke(null, new object[] { "Layout", typeof(Layout.Layout) });
             addAccelerator.Invoke(null, new object[] { "Chart", typeof(Chart) });
+
+            //TODO: Is this declared anywhere else?
+            string[] mimeTypes = new string[] { HtmlFormatter.MimeType, JsonFormatter.MimeType, PlainTextFormatter.MimeType };
```
justinytchen, line 77: not sure about this

TylerLeonhardt:
@justinytchen can you add a screen shot of the before and after if PSCustomObject and:

```
Get-Process | Out-Display
```

src/Microsoft.DotNet.Interactive.PowerShell/PowerShellKernel.cs
```diff
@@ -70,6 +72,32 @@ static PowerShellKernel()
...
+                            table.Add(p.Name, p.Value);
+                        }
+                        obj = table;
+                    }
+                    else
+                    {
+                        obj = psObj.BaseObject;
+                    }
+                    obj.FormatTo(writer, mt);
+                }, mimeType: mt);
```
daxian-dbw, line 99: Why moving this here? I think it fits better in the `Out-Display` command.
This formatting doesn't do anything special that cannot be done in general before you call `display` in `Out-Display`.

daxian-dbw: Let's fix the `PSCustomObject` only in this PR and defer the fix for all ETS extended properties to a separate PR.

TylerLeonhardt: Why do it in Out_Display when .NET Interactive already has this Formatter concept? It seems like we should leverage that instead.

daxian-dbw: I don't see any benefit to use the formatter for the current code. It's fine to do this if it's a fix for PSObject in general, but it's not.
Actually, now I double the formatter is the right tool to solve the PSObject ETS issue in general.
Besides, do you know where those formatter will be invoked? Maybe on a different thread that doesn't have a default Runspace, then `p.Value` will fail if the property is a `ScriptProperty`.

TylerLeonhardt: Reread the code - 100% agree. @justinytchen and I synced on this and we're all on the same page.

daxian-dbw: Great! Also, we we do need a formatter for anything, then it's better to do it using the extension method to `Kernel`, just like how we did for invoking the profile.

TylerLeonhardt: @daxian-dbw I didn't see your longer comment before commenting but exactly. Justin's gonna change this PR back but we'll defer the generic solution for now. He's gonna focus on `Out-GridView` :)

daxian-dbw: That's right, focus on `Out-GridView`. The PSObject ETS member issue is a _hard and not urgent_ one to fix.

src/Microsoft.DotNet.Interactive.PowerShell/Commands/OutDisplayCommand.cs
```diff
@@ -1,6 +1,7 @@
 // Copyright (c) .NET Foundation and contributors. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 using static Microsoft.DotNet.Interactive.Kernel;
+using System.Collections.Generic;
```
daxian-dbw, line 5: nit: can you also move the following 2 using directives out of the namespace block?
```
using System.Management.Automation;
using Microsoft.DotNet.Interactive.Events;
```

src/Microsoft.DotNet.Interactive.PowerShell.Tests/PowerShellKernelTests.cs
```diff
@@ -216,5 +217,27 @@ public async Task VerifyAllUsersProfileRuns()
...
+            var kernel = CreateKernel(Language.PowerShell);
+            var result = await kernel.SendAsync(new SubmitCode("[pscustomobject]@{ prop1 = 'value1'; prop2 = 'value2'; prop3 = 'value3' } | Out-Display"));
+            var outputs = result.KernelEvents.ToSubscribedList();
+
+            outputs.Should().SatisfyRespectively(
+                              e => e.Should().BeOfType<CodeSubmissionReceived>(),
+                              e => e.Should().BeOfType<CompleteCodeSubmissionReceived>(),
+                              e => e.Should().BeOfType<DisplayedValueProduced>().Which.Value.Should().BeEquivalentTo(props),
+                              e => e.Should().BeOfType<CommandHandled>()
+                             );
```
TylerLeonhardt, line 240:
```suggestion
            outputs.Should().SatisfyRespectively(
                e => e.Should().BeOfType<CodeSubmissionReceived>(),
                e => e.Should().BeOfType<CompleteCodeSubmissionReceived>(),
                e => e.Should().BeOfType<DisplayedValueProduced>().Which.Value.Should().BeEquivalentTo(props),
                e => e.Should().BeOfType<CommandHandled>()
            );
```

src/Microsoft.DotNet.Interactive.PowerShell.Tests/PowerShellKernelTests.cs
```diff
@@ -216,5 +217,27 @@ public async Task VerifyAllUsersProfileRuns()
...
+
+        [Fact]
+        public async Task Powershell_customobject_is_parsed_for_outdisplay()
+        {
+            var props = (new Dictionary<string, object>
+                {
+                    { "prop1", "value1" },
+                    { "prop2", "value2" },
+                    { "prop3", "value3" }
+                });
```
TylerLeonhardt, line 229:
```suggestion
            var props = (new Dictionary<string, object>
            {
                { "prop1", "value1" },
                { "prop2", "value2" },
                { "prop3", "value3" }
            });
```

TylerLeonhardt:
I _just_ ran into a place where we can apply the same fix...

https://github.com/dotnet/interactive/blob/master/src/Microsoft.DotNet.Interactive.PowerShell/PowerShellKernel.cs#L126-L134

when you use:

```
#!pwsh
[pscustomobject]@{
    foo = "asdf"
}

#!csharp
#!share --from pwsh foo

foo["foo"]
```

it fails saying:

image

We just this this same check for pscustomobject in this place as well. I would refactor the code that turns a pscustomobject into a dictionary into a private method so it can be reused.

jonsequitur:
@justinytchen This will need to be rebased onto master. We renamed a few types.

src/Microsoft.DotNet.Interactive.PowerShell/PowerShellKernel.cs
```diff
@@ -125,7 +126,19 @@ public override bool TryGetVariable<T>(string name, out T value)
...
                     case PSObject psobject:
-                        value = (T) psobject.BaseObject;
+                        if (psobject.BaseObject is PSCustomObject)
+                        {
+                            Dictionary<string, object> table = new Dictionary<string, object>();
+                            foreach (var p in psobject.Properties)
+                            {
+                                table.Add(p.Name, p.Value);
+                            }
+                            value = (T) ((Object)table);
```
daxian-dbw, line 136: This is likely going to fail, unless `T` is `object`.

colombod: casting to object is the only safe way to then cast to T, might introduce boxing and unboxing

daxian-dbw: I meant `table` here is of the type `Dictionary<string, object>`, so unless `T` itself is `object` or `Dictionary<string, object>`, the casting will fail at runtime.
But I don't know if this is a real problem, maybe `T` will always be `object` when calling `TryGetVariable`?

TylerLeonhardt: I agree with Dongbo. For a `TryGetVariable` method, it's probably best to wrap that in a try/catch or use `is`/`as`.

what do you think of this?

```csharp
public override bool TryGetVariable<T>(string name, out T value)
{
    var variable = pwsh.Runspace.SessionStateProxy.PSVariable.Get(name);
    if (variable != null)
    {
        object output;
        switch (variable.Value)
        {
            case PSObject psobject:
                if (psobject.BaseObject is PSCustomObject)
                {
                    Dictionary<string, object> table = new Dictionary<string, object>();
                    foreach (var p in psobject.Properties)
                    {
                        table.Add(p.Name, p.Value);
                    }
                    output = table;
                }
                else
                {
                    output = psobject.BaseObject;
                }
                break;
            default:
                output = variable.Value;
                break;
        }

        if (output is T tval)
        {
            value = tval;
            return true;
        }
    }

    value = default;
    return false;
}

```

src/Microsoft.DotNet.Interactive.PowerShell/PowerShellKernel.cs
```diff
@@ -122,17 +123,34 @@ public override bool TryGetVariable<T>(string name, out T value)
 
             if (variable != null)
             {
+                object outVal;
```
daxian-dbw, line 126: nit: ideally the duplicate code should be moved to a utility method. Also, the switch statement can be avoid to simply the code.
```c#
            if (variable != null)
            {
                object outVal = variable.Value;
                if (outVal is PSObject psobject)
                {
                    if (psobject.BaseObject is PSCustomObject)
                    {
                        Dictionary<string, object> table = new Dictionary<string, object>();
                        foreach (var p in psobject.Properties)
                        {
                            table.Add(p.Name, p.Value);
                        }
                        outVal = table;
                    }
                    else
                    {
                        outVal = psobject.BaseObject;
                    }
                }

                if(outVal is T tObj)
                {
                    value = tObj;
                    return true;
                }
            }
```

src/Microsoft.DotNet.Interactive.PowerShell/PowerShellKernel.cs
```diff
@@ -232,6 +244,15 @@ public override Task SetVariableAsync(string name, object value)
             return Task.CompletedTask;
         }
 
+        public static Dictionary<String, Object> GetPSObjectProperties(PSObject psObject){
```
daxian-dbw, line 247: Please don't add the utility method in `PowerShellKernel.cs`. Instead, add it to `PowerShellExtensions.cs` as an extension method.
Also, move all the logic that deal with PSObject in the extension method, not just the logic that handles `PSCustomObject`.
Example:
```c#
public static object Unwrap(this PSObject psobj)
{
    object obj = psObj.BaseObject;
    if (obj is PSCustomObject)
    {
        ....
        obj = table;
    }

    return obj;
}
```

In `out-display`, use it like this (similarly in `TryGetVariable`):

```c#
object obj = InputObject is PSObject psObj ? psObj.Unwrap() : InputObject;
DisplayedValue displayedValue = display(obj, MimeType);
```

src/Microsoft.DotNet.Interactive.PowerShell/PowerShellExtensions.cs
```diff
@@ -96,5 +97,32 @@ internal static SecureString GetSecureStringPassword(this PasswordString pwdStri
 
             return secure;
         }
+
+        internal static object Unwrap(this PSObject inputObject)
+        {
+            object obj;
+            if (inputObject is PSObject psObj)
```
daxian-dbw, line 104: `inputObject` is already a `PSObject`.
Please refactor this method to something like the following:
```c#
public static object Unwrap(this PSObject psobj)
{
    object obj = psObj.BaseObject;
    if (obj is PSCustomObject)
    {
        ....
        obj = table;
    }

    return obj;
}
```

src/Microsoft.DotNet.Interactive.PowerShell/PowerShellExtensions.cs
```diff
@@ -96,5 +97,21 @@ internal static SecureString GetSecureStringPassword(this PasswordString pwdStri
 
             return secure;
         }
+
+        internal static object Unwrap(this PSObject psObj)
+        {
+            object obj = psObj.BaseObject;
+            if (psObj.BaseObject is PSCustomObject)
```
daxian-dbw, line 104:
```suggestion
            if (obj is PSCustomObject)
```

src/Microsoft.DotNet.Interactive.PowerShell/PowerShellKernel.cs
```diff
@@ -2,6 +2,7 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
+using System.Collections.Generic;
```
daxian-dbw, line 5: This can be removed now.

src/Microsoft.DotNet.Interactive.PowerShell.Tests/PowerShellKernelTests.cs
```diff
@@ -216,5 +217,27 @@ public async Task VerifyAllUsersProfileRuns()
...
+            });
+
+            var kernel = CreateKernel(Language.PowerShell);
+            var result = await kernel.SendAsync(new SubmitCode("[pscustomobject]@{ prop1 = 'value1'; prop2 = 'value2'; prop3 = 'value3' } | Out-Display"));
+            var outputs = result.KernelEvents.ToSubscribedList();
+
+            outputs.Should().SatisfyRespectively(
+                e => e.Should().BeOfType<CodeSubmissionReceived>(),
+                e => e.Should().BeOfType<CompleteCodeSubmissionReceived>(),
+                e => e.Should().BeOfType<DisplayedValueProduced>().Which.Value.Should().BeEquivalentTo(props),
```
jonsequitur, line 238: The `Value` property will no longer be serialized soon (#512), so it's better here to verify the `FormattedValues`, e.g.

```cs
 e => e.Should().BeOfType<DisplayedValueProduced>()
    .Which                      
    .FormattedValues
    .Should()
    .ContainSingle(f => f.Value == ... )
```

colombod:
Could you rebase the PR on master? I'd be happy to merge after that.

Thank you for your contribution

src/Microsoft.DotNet.Interactive.PowerShell/PowerShellKernel.cs
```diff
@@ -69,6 +69,32 @@ static PowerShellKernel()
             // Add accelerators that exist in other namespaces.
             addAccelerator.Invoke(null, new object[] { "Layout", typeof(Layout.Layout) });
             addAccelerator.Invoke(null, new object[] { "Chart", typeof(Chart) });
+
+            //TODO: Is this declared anywhere else?
+            string[] mimeTypes = new string[] { HtmlFormatter.MimeType, JsonFormatter.MimeType, PlainTextFormatter.MimeType };
+
+            foreach (string mt in mimeTypes)
```
daxian-dbw, line 76: It seems the rebase broke the changes. The formater code should be removed already.

justinytchen: I forgot to fetch the upstream master; will be redoing the rebase

PR #552, opened 2020-06-17
directive nodes can use parent kernel to load completions

colombod:


colombod:
this has been addressed in #559

PR #577, opened 2020-06-25
refactor for connect and proxy kernel

colombod:
Goals
make #!connect extensible
converge stdio server and abstract the stream construction to allow usage of console streams, namedpipes and http


src/Microsoft.DotNet.Interactive.Tests/Server/KernelServerTests.cs
```diff
@@ -38,18 +38,22 @@ public StandardIOKernelServerTests(ITestOutputHelper output)
                     .UseDefaultMagicCommands()
             };
 
-            _standardIOKernelServer = new StandardIOKernelServer(
+            var inputStream = new TextReaderInputStream(new StreamReader(new MemoryStream()));
```
jonsequitur, line 41: Do these streams need to be disposed?

PR #583, opened 2020-06-26
use signalR client to expose http conenction for proxy kernel

colombod:


src/Microsoft.DotNet.Interactive/KernelExtensions.cs
```diff
@@ -174,7 +174,44 @@ public static CompositeKernel UseProxyKernelWithNamedPipe(this CompositeKernel k
...
+                        .Build();
+
+                    await connection.StartAsync();
+
+                    var client = connection.CreateKernelClient();
+                    var proxyKernel = new ProxyKernel(kernelName, client);
+                    await connection.SendAsync("connect");
+
+                    proxyKernel.RegisterForDisposal(client);
+                    proxyKernel.RegisterForDisposal(async () =>
```
jonsequitur, line 204: This looks like an async void and should be avoided. Looks like we should add `IAsyncDisposable` support to `PocketDisposable`.

src/Microsoft.DotNet.Interactive/SignalR/SignalRInputTextStream.cs
```diff
@@ -0,0 +1,37 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Reactive.Disposables;
+using System.Reactive.Subjects;
+using Microsoft.AspNetCore.SignalR.Client;
+
+namespace Microsoft.DotNet.Interactive.SignalR
```
jonsequitur, line 9: This should be in the `Server` namespace.

src/Microsoft.DotNet.Interactive/SignalR/SignalRInputTextStream.cs
```diff
@@ -0,0 +1,37 @@
...
+{
+    public class SignalRInputTextStream : IInputTextStream
```
jonsequitur, line 11: Does this need to be public?

src/Microsoft.DotNet.Interactive/SignalR/SignalROutputTextStream.cs
```diff
@@ -0,0 +1,22 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using Microsoft.AspNetCore.SignalR.Client;
+
+namespace Microsoft.DotNet.Interactive.SignalR
```
jonsequitur, line 7: This should be in the `Server` namespace.

src/Microsoft.DotNet.Interactive/SignalR/SignalROutputTextStream.cs
```diff
@@ -0,0 +1,22 @@
...
+{
+    public class SignalROutputTextStream : OutputTextStream
```
jonsequitur, line 9: Does this need to be public?

src/dotnet-interactive/CommandLine/CommandLineParser.cs
```diff
@@ -478,6 +478,7 @@ private static IServiceCollection RegisterKernelInServiceCollection(IServiceColl
                          .UseDefaultMagicCommands()
                          .UseLog()
                          .UseAbout()
+                         .UseProxyKernelWithSignalR()
```
jonsequitur, line 481: Can we come up with better naming than `UseProxy*`?

src/Microsoft.DotNet.Interactive/SignalR/SignalROutputTextStream.cs
colombod: no

src/Microsoft.DotNet.Interactive/SignalR/SignalRInputTextStream.cs
colombod: No

src/dotnet-interactive/CommandLine/CommandLineParser.cs
colombod: Proposing

```
.UseConnectionOverNamedPipe()
.UseConnectionOverSignalR()
```

src/Microsoft.DotNet.Interactive/KernelExtensions.cs
colombod: ok but that one  is using rx CompositeDisposable.

jonsequitur: It's an implementation detail, and since we can change the code for `PocketDisposable`, this seems pretty straightforwrd.

src/dotnet-interactive/CommandLine/CommandLineParser.cs
jonsequitur: For each transport we have two APIs, one to expose a kernel as a server, and the other to snap in clients as subkernels. How can we name them so it's obvious which end of the connection we're talking about?

PR #587, opened 2020-06-27
#!connect with remote composite kernels

colombod:
when remotign a command the targetKernelName is serialisedacross the boundaries causing the remote compositeKernel to fail routing

### DIFF
--- a/src/Microsoft.DotNet.Interactive.Tests/NamedPipeKernelTests.cs
+++ b/src/Microsoft.DotNet.Interactive.Tests/NamedPipeKernelTests.cs
@@ -30,105 +30,47 @@ namespace Microsoft.DotNet.Interactive.Tests
         }
 
         [FactSkipLinux]
-        public async Task can_address_remote_composite_kernel_specifying_a_remote_targetKernelName()
-        {
-            var fSharpKernel = new FSharpKernel();
-            using var kernel = new CompositeKernel
-            {
-                fSharpKernel
-            }.UseConnectionOverNamedPipe();
-
-            kernel.DefaultKernelName = fSharpKernel.Name;
-
-            var pipeName = Guid.NewGuid().ToString();
-            var remoteKernelInvoked = false;
-
-            var defaultRemoteKernel = new FakeKernel("defaultRemote")
-            {
-                Handle = (command, context) => Task.CompletedTask
-            };
-
-            var expectedRemoteKernel = new FakeKernel("expectedRemote")
-            {
-                Handle = (command, context) =>
-                {
-                    remoteKernelInvoked = true;
-                    return Task.CompletedTask;
-                }
-            };
-
-            using var remoteKernel = new CompositeKernel
-            {
-                defaultRemoteKernel,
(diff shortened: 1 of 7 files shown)


src/Microsoft.DotNet.Interactive.Tests/NamedPipeKernelTests.cs
```diff
@@ -27,7 +27,110 @@ public NamedPipeKernelTests(ITestOutputHelper output)
...
+            var defaultRemoteKernel = new FakeKernel("defaultRemote")
+            {
+                Handle = (command, context) =>
+                {
+                    remoteKernelInvoked = true;
+                    return Task.CompletedTask;
+                }
+            };
+
+            using var remoteKernel = new CompositeKernel
```
jonsequitur, line 109: Maybe name this `remoteCompositeKernel` for clarity? Lots of kernels in these tests.

src/Microsoft.DotNet.Interactive.Tests/NamedPipeKernelTests.cs
```diff
@@ -27,7 +27,110 @@ public NamedPipeKernelTests(ITestOutputHelper output)
...
+
+            var proxyCommand = new SubmitCode($"#!connect named-pipe test {pipeName}");
+
+            await kernel.SendAsync(proxyCommand);
+
+            var proxyCommand2 = new SubmitCode(@"
+var x = 1 + 1;
+x", targetKernelName: "test");
+
+            await kernel.SendAsync(proxyCommand2);
```
jonsequitur, line 129: Do we need to send two commands to prove this is working?

src/Microsoft.DotNet.Interactive/KernelExtensions.cs
```diff
@@ -180,35 +182,36 @@ public static CompositeKernel UseConnectionOverNamedPipe(this CompositeKernel ke
 
         public static CompositeKernel UseConnectionOverSignalR(this CompositeKernel kernel)
         {
-            var connectionCommand = new Command("signalr");
+            var connectionCommand = new Command("signalr", "Connects to a kernel using signal R");
```
jonsequitur, line 185:
```suggestion
            var connectionCommand = new Command("signalr", "Connects to a kernel using SignalR");
```

src/Microsoft.DotNet.Interactive/KernelExtensions.cs
```diff
@@ -180,35 +182,36 @@ public static CompositeKernel UseConnectionOverNamedPipe(this CompositeKernel ke
...
+
                     var connection = new HubConnectionBuilder()
                         .WithUrl(hubUrl)
                         .Build();
 
                     await connection.StartAsync();
 
                     var client = connection.CreateKernelClient();
-                    var proxyKernel = new ProxyKernel(kernelName, client);
+                    var proxyKernel = new ProxyKernel(kernelName, client, remoteKernelName);
```
jonsequitur, line 203: Why do we need to specify a remote name?

PR #599, opened 2020-06-30
[wip] performance improvements

colombod:


src/Microsoft.DotNet.Interactive.CSharp/CSharpKernel.cs
```diff
@@ -135,6 +138,8 @@ public async Task HandleAsync(RequestHoverText command, KernelInvocationContext
             var service = QuickInfoService.GetService(document);
             var info = await service.GetQuickInfoAsync(document, cursorPosition);
 
+            GC.Collect();
```
jonsequitur, line 141: 😒

colombod: Look the other way :)

jonsequitur: 😭

PR #603, opened 2020-07-01
us gc memory pressure to control memory allocation

colombod:


src/Microsoft.DotNet.Interactive.CSharp/GCPressure.cs
```diff
@@ -0,0 +1,24 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+
+using System;
+
+namespace Microsoft.DotNet.Interactive.CSharp
```
jonsequitur, line 7: This isn't specific to C#.

Issue #609, opened 2020-07-02
Additional output rendered when executing code on conencted kernel


colombod:
### Describe the bug
When VSCode extension is connected to a kernel via signalR and DisplayEvents have been generate on the remote kernel they are handled later on as deferred command outputs and appended to the cell output n vscode

Actual behaviour
![image](https://user-images.githubusercontent.com/375556/86341545-59e6f280-bc4e-11ea-9237-b8bcf8c26720.png)

Expected
![image](https://user-images.githubusercontent.com/375556/86341575-64a18780-bc4e-11ea-9411-39393378114a.png)

### Please complete the following:

**Which version of .NET Interactive are you using?** 
1.0.135104

- OS
    - [X] Windows 10
    - [ ] macOS
    - [ ] Linux (Please specify distro)
    - [ ] iOS
    - [ ] Android
 - Browser 
    - [ ] Chrome
    - [ ] Edge
    - [ ] Firefox
    - [ ] Safari
 - Frontend
    - [ ] Jupyter Notebook
    - [ ] Jupyter Lab
    - [ ] nteract
    - [X] Visual Studio Code
    - [ ] Other (please specify)

### Screenshots
![image](https://user-images.githubusercontent.com/375556/86341198-e1803180-bc4d-11ea-9213-13c239780e5b.png)

PR #610, opened 2020-07-02
Connected kernels do not buffer events not produced by the user

colombod:
Address #609


src/dotnet-interactive-vscode/src/interactiveClient.ts
```diff
@@ -228,7 +228,7 @@ export class InteractiveClient {
                 for (let listener of listeners) {
                     listener(eventEnvelope);
                 }
-            } else {
+            } else if(token.startsWith("deferredCommand::")) {
```
brettfo, line 231: Space between `if` and `(`.

colombod: missed it but now is done

PR #611, opened 2020-07-02
add diagnostics

brettfo:
Returns diagnostics from the kernel.

I briefly went down the path of attaching diagnostics to the `CommandFailed` event, but even successfully executing code can produce diagnostics (e.g., warnings) so it made more sense to have a separate event specifically for diagnostics.  Diagnostics have their ranges appropriately adjusted for magic commands and subkernels.

image

Diagnostics in polyglot scenarios don't work well because the first kernel to complete execution by returning either a `CommandSucceeded` or `CommandFailed` event will prevent subsequent kernels from getting an opportunity to return diagnostics.  Given that polyglot cells are split into separate cells on open and given how useful diagnostics are, I thought it best to continue with the current implementation and add a skipped test so it can be revisited later.  Example polyglot execution that doesn't properly return diagnostics from both kernels:

```
#!fsharp
[1;2;3] |> List.summmmmmm
//              ^^^^^^^^^ only this diagnostic is returned

#!csharp
Console.WriteLineeeeee();
//      ^^^^^^^^^^^^^^ this one is never returned because the `SubmitCode` handler is never invoked after the F# kernel fails
```

A future PR will add live-typing diagnostics to VS Code.


src/Microsoft.DotNet.Interactive.Tests/LanguageKernelTests.cs
```diff
@@ -246,9 +246,9 @@ void f()
         }
 
         [Theory]
-        [InlineData(Language.CSharp)]
-        [InlineData(Language.FSharp)]
-        public async Task it_returns_diagnostics(Language language)
+        [InlineData(Language.CSharp, "CS0103", "The name 'aaaadd' does not exist in the current context", "(1,1): error CS0103: The name 'aaaadd' does not exist in the current context")]
+        [InlineData(Language.FSharp, "FS0039", "The value or constructor 'aaaadd' is not defined.", "input.fsx (1,1)-(1,7) typecheck error The value or constructor 'aaaadd' is not defined.")]
+        public async Task it_returns_diagnostics_on_command_failed(Language language, string code, string diagnosticMessage, string errorMessage)
```
colombod, line 251: This test should be "when_code_contains_compile_time_error_diagniostics_are_produced"

src/Microsoft.DotNet.Interactive.Tests/LanguageKernelTests.cs
```diff
@@ -269,19 +269,136 @@ public async Task it_returns_diagnostics(Language language)
...
+                new LinePosition(0, 6));
 
             KernelEvents
                 .Should()
                 .ContainSingle<CommandFailed>()
                 .Which
                 .Message
                 .Should()
-                .Be(error);
+                .Be(errorMessage);
```
colombod, line 282: don't think we need to assert on command filed for compile error, but if you do multiple assertions use the assertion scope

src/Microsoft.DotNet.Interactive.Tests/LanguageKernelTests.cs
```diff
@@ -269,19 +269,136 @@ public async Task it_returns_diagnostics(Language language)
...
+    };
+}
+
+", "The switch expression does not handle all possible values of its input type")]
+        [InlineData(Language.FSharp, @"
+let x n =
+    match n with
+    | 0 -> ()
+", "Incomplete pattern matches on this expression.")]
+        public async Task it_returns_diagnostics_on_command_succeeded(Language language, string code, string diagnosticText)
```
colombod, line 329: This one again is a test on diagnostics produced for compile time warning. no need to assert on command succeeded or failuer

src/Microsoft.DotNet.Interactive.Tests/LanguageKernelTests.cs
```diff
@@ -269,19 +269,136 @@ public async Task it_returns_diagnostics(Language language)
...
+                .Should()
+                .ContainSingle<DiagnosticsProduced>(d => d.Diagnostics.Count > 0)
+                .Which
+                .Diagnostics
+                .Should()
+                .ContainSingle(diag => diag.Message.Contains(diagnosticText));
+        }
+
+        [Fact(Skip = "The first failed sub-command cancels all subsequent command executions; the second kernel doesn't get a chance to report.")]
+        public async Task diagnostics_can_be_returned_from_multiple_subkernels()
```
colombod, line 349: generally diagnostics are produced, not returned

src/Microsoft.DotNet.Interactive.Tests/LanguageKernelTests.cs
```diff
@@ -269,19 +269,136 @@ public async Task it_returns_diagnostics(Language language)
...
+                .ContainSingle(d => d.Code.StartsWith("CS"))
+                .And
+                .ContainSingle(d => d.Code.StartsWith("FS"));
+        }
+
+        [Theory]
+        [InlineData(Language.CSharp, "Console.WritLin();")]
+        [InlineData(Language.FSharp, "printfnnn \"\"")]
+        [InlineData(Language.PowerShell, "::()")]
+        public async Task returned_diagnostics_are_remapped_to_the_appropriate_span(Language language, string languageSpecificCode)
```
colombod, line 376: same here, diagnostics are mapped to spans.

src/dotnet-interactive-vscode/src/tests/unit/notebook.test.ts
```diff
@@ -297,6 +309,150 @@ Console.WriteLine(1);
         ]);
     });
 
+    it('diagnostics are reported on CommandFailed', async () => {
```
colombod, line 312: non compiling code produces diagnostics same observation as for the .NET part, we want to assert that diagnostics are done on compiler error and warnings

brettfo: I'd like to leave the events in these tests because the events are faked and it doesn't really matter what the contents of `SubmitCode` actually is; what really matters is the `CommandSucceeded`/`CommandFailed` aspect and I wanted to test exactly that.

PR #618, opened 2020-07-06
Forward undefined directives to language kernels

colombod:
Addresses #613


src/Microsoft.DotNet.Interactive/Parsing/SubmissionParser.cs
```diff
@@ -78,19 +78,27 @@ private IReadOnlyList<KernelCommand> SplitSubmission(KernelCommand originalComma
...
-                                {
-                                    var message =
-                                        string.Join(Environment.NewLine,
-                                                    parseResult.Errors
-                                                               .Select(e => e.ToString()));
-
-                                    context.Fail(message: message);
-                                    return Task.CompletedTask;
-                                }, parent: originalCommand.Parent));
+                            if (parseResult.Errors.All(e => e.SymbolResult == null) && directiveNode is ActionDirectiveNode actionDirectiveNode)
```
jonsequitur, line 81: I'm not sure what `SymbolResult == null` is supposed to indicate or whether it will behave consistently the way you expect.

colombod: that is my question, an unknown directive is producing an error without any symbols. I am actually looking for a way to discriminate by something used in the wrong way (other tests are covering scenarios with wrong options and args) and something that is unknown

jonsequitur: I suspect a heuristic approach will be needed, and a bunch of tests covering various kinds of parsing errors.

src/Microsoft.DotNet.Interactive.CSharp/InteractiveWorkspace.cs
```diff
@@ -61,6 +62,24 @@ public void AddSubmission(ScriptState scriptState)
             var debugName = assemblyName;
             var projectId = ProjectId.CreateNewId(debugName: debugName);
 
+            solution = CreateProjectAndAddToSolution(projectId, debugName, assemblyName, compilationOptions, solution,
+                projectReferenceProjectId);
+
+            var workingDocumentId = DocumentId.CreateNewId(
+                projectId,
+                debugName: $"working document for {submission}");
```
jonsequitur, line 70: Why have a `debugName` variable and also this?

colombod: because that is the debug name of the document, not of the project

colombod: using variable only

src/Microsoft.DotNet.Interactive/Parsing/LanguageNodeExtensions.cs
```diff
@@ -0,0 +1,17 @@
...
+using System.Linq;
+
+namespace Microsoft.DotNet.Interactive.Parsing
+{
+    internal static class LanguageNodeExtensions
+    {
+        public static bool IsUnknownActionDirective(this DirectiveNode directiveNode)
+        {
+            var parseResult = directiveNode.GetDirectiveParseResult();
+            return parseResult.Errors.All(e => e.SymbolResult == null) && directiveNode is ActionDirectiveNode;
```
jonsequitur, line 13: Nit: There's a small perf benefit to flipping the order of these two evaluations.

PR #624, opened 2020-07-08
I missed a point

colombod:


src/dotnet-interactive/CommandLine/CommandLineParser.cs
```diff
@@ -119,9 +119,9 @@ public static class CommandLineParser
                     {
                         switch (directive.Key)
                         {
-                            case "vscode":
-                            case "jupyter":
-                            case "synapse":
+                            case "debug":
+                                break;
+                            default:
```
jonsequitur, line 124: Oh. This needs to be a strict allow list. The number of possible directives that we don't want to report is effectively infinite.

colombod: ok then previous Pr is good

Issue #208, opened 2020-02-28
 Import table from same Notebook


Ayanmullick:
**Is your feature request related to a problem? Please describe.**
As an Azure Architect, one of the notebook features that sounded most promising to me was the ability to put markdown documentation and PowerShell deployment automation script in one file. 

However, since there is no a way to reference a text block from a code block one would need to import a tabled set of parameter values from an external file.

**Describe the solution you'd like**
It'd be convenient if there was a way for one to populate a variable in a code block with content from a text block without having to use a second file. 
For eg. : How could one populate the `$Data` variable with the contents of the table in the text block without using another file in the notebook below? 
[https://github.com/Ayanmullick/test/blob/master/testfolder/test.ipynb](https://github.com/Ayanmullick/test/blob/master/testfolder/test.ipynb)

**Describe alternatives you've considered**
I considered `Import-CSV`, `Get-Content`, `ConvertFrom-Json` and `ConvertFrom-SourceTable`. However, [I couldn't figure out the value for the `-Path` parameter to point it to a text block in the same file.](https://stackoverflow.com/questions/60264332/import-table-from-same-notebook)

jonsequitur:
You can pipe strings to `Out-Display` and specify a mime type, which should be more robust than parsing the `.ipynb` file.

![image](https://user-images.githubusercontent.com/547415/75587605-ac5b9c80-5a2b-11ea-8797-01eb69a94b02.png)

We're also working on some scenarios for sharing data between cells containing different languages. But `dotnet-interactive` is currently unaware of the contents of any Markdown cells.

Ayanmullick:
@jonsequitur , My purpose of importing a markdown table to a PowerShell variable is so it can be consumed by other cmdlets not just `Out-Display`.

|Location| RG     | Name |
|--------|--------|------|
| EastUS | EastRG |EastVM|
| WestUS | WestRG |WestVM|


`$b=Import-Csv -Path <>  -Delimiter '|'`
`$b| Foreach-Object {New-AzVM -Location $PSItem.Location -ResourceGroupName $PSItem.RG -Name $PSItem.Name}`

Something like that.

jonsequitur:
[...]

PR #627, opened 2020-07-08
KeyValueStoreKernel

jonsequitur:
This implements a `KeyValueStoreKernel` as a way to paste literal text into a cell and then access it from other subkernels via `#!share`.

It uses this capability to add the ability to customize the kernel chooser directive, so you can store a value using this syntax:

```
#!value --name myCsvData
one,two,three
1,2,3

#!share --from value myCsvData
var df = DataFrame.LoadCsv(myCsvData);
```

As a next step, I'm considering adding support for `--mime-type` so that the value can also be immediately displayed as the specified mime type.

Suggestions on the workflow or naming?


jonsequitur:
I believe this addresses #208. Thoughts, @ayanmullick?

src/Microsoft.DotNet.Interactive.Tests/KeyValueStoreKernelTests.cs
```diff
@@ -0,0 +1,68 @@
...
+        [Fact]
+        public async Task SubmitCode_is_not_valid_without_a_value_name()
+        {
+            using var kernel = CreateKernel();
+
+            var storedValue = "1,2,3";
+
+            var result = await kernel.SubmitCodeAsync(
+                             @$"
+#!value
```
brettfo, line 23: There's something funny going on with your line endings in this file.

PR #629, opened 2020-07-09
reset cell execution state on kernel termination

brettfo:
If a kernel has to be restarted because it's stuck, we need to set each cell's state as idle so it's executable again after restart.

### DIFF
--- a/src/dotnet-interactive-vscode/src/clientMapper.ts
+++ b/src/dotnet-interactive-vscode/src/clientMapper.ts
@@ -36,9 +36,7 @@ export class ClientMapper {
         }
     }
 
-    closeAllClients() {
-        for (let fsPath of this.clientMap.keys()) {
-            this.closeClient({ fsPath });
-        }
+    isDotNetClient(uri: Uri): boolean {
+        return this.clientMap.has(uri.fsPath);
     }
 }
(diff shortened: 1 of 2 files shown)


jonsequitur:
How do we test this?

brettfo:
We don't in a meaningful way.  Everything under the `src/vscode` directory can't be directly unit-tested because unit tests can't get a VSCode runtime environment.  We can in theory add integration tests that _do_ get a VSCode runtime environment, but we can't run those in CI.  We can add `Cell` and `CellMetadata` interfaces to verify that we set those appropriately, but that seems heavy-handed, but if you want it I'll add it.

jonsequitur:
What about a manually-runnable sanity check suite for local development for now?

brettfo:
I've been exploring this and it seems that integration tests can't (aren't allowed to?) use the notebook APIs.  I've modified the integration runner to test against an insiders build and pass `--enable-proposed-api ms-dotnettools.dotnet-interactive-vscode`, but in the test itself when I call `vscode.notebook.activeNotebookEditor` it fails with `Proposed API is only available when running out of dev or with the following command line switch: --enable-proposed-api nullExtensionDescription`.  I've sent off some probing emails to try to discover what the issue is.

Until then I've filed #632 to track adding integration tests.

Issue #633, opened 2020-07-09
Live diagnostics cause connect to fail


colombod:
### Describe the bug
When editing text in a cell with the ```#!connect```
![image](https://user-images.githubusercontent.com/375556/87100280-e774a480-c243-11ea-8a4d-7f5def8645b6.png)

probably related to #631 

### Please complete the following:

**Which version of .NET Interactive are you using?** 
1.0.135802

- OS
    - [X] Windows 10
    - [ ] macOS
    - [ ] Linux (Please specify distro)
    - [ ] iOS
    - [ ] Android
 - Browser 
    - [ ] Chrome
    - [ ] Edge
    - [ ] Firefox
    - [ ] Safari
 - Frontend
    - [ ] Jupyter Notebook
    - [ ] Jupyter Lab
    - [ ] nteract
    - [X] Visual Studio Code
    - [ ] Other (please specify)

Issue #631, opened 2020-07-09
Live diagnostics can cause multiple incorrect NuGet package restore attempts


brettfo:
In a new cell start slowly typing `#r "nuget: Newtonsoft.Json"`.  When you finally commit the cell, multiple NuGet packages are added to the restore list.  E.g.,

```
Installing package Newtonsoft.Json.....
Installing package Newtons.....
Installing package Newt.....
```

Resubmitting the cell has the same effect.

Workaround: restart the notebook kernel and execute a cell with the text already present.

jonsequitur:
This is probably related to anonymous commands generated during submission splitting.

brettfo:
Fixed by #635.

PR #635, opened 2020-07-10
can suppress command execution

colombod:
When handling diagnostics request command execution is suppressed

Adresses
#633
#631

### DIFF
--- a/src/Microsoft.DotNet.Interactive/Commands/KernelCommand.cs
+++ b/src/Microsoft.DotNet.Interactive/Commands/KernelCommand.cs
@@ -10,8 +10,6 @@ namespace Microsoft.DotNet.Interactive.Commands
 {
     public abstract class KernelCommand
     {
-        private bool _suppress;
-
         protected KernelCommand(string targetKernelName = null, KernelCommand parent = null)
         {
             if (parent is null)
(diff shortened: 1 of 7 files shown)


src/Microsoft.DotNet.Interactive/Commands/KernelCommand.cs
```diff
@@ -47,5 +60,10 @@ public virtual async Task InvokeAsync(KernelInvocationContext context)
 
             await Handler(this, context);
         }
+
+        internal void SuppressExecution()
```
jonsequitur, line 64: Why take this approach instead of just not generating the commands in the first place?

colombod: we do need kernels choosing commands at least, right?

colombod: will try that way

jonsequitur: Yes. I think we can pretty safely filter based on command types.

jonsequitur: Much cleaner now.

jonsequitur:
So does this fix #633?

colombod:
> So does this fix #633?

yes, it seems

PR #647, opened 2020-07-14
ensure stdio server sets the working directory

brettfo:
A recent refactoring stopped passing the `--working-dir` option to the kernel server.

### DIFF
--- a/src/Microsoft.DotNet.Interactive.Tests/NamedPipeConnectionTests.cs
+++ b/src/Microsoft.DotNet.Interactive.Tests/NamedPipeConnectionTests.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
+using System.IO;
 using System.Threading.Tasks;
 using FluentAssertions;
 using Microsoft.DotNet.Interactive.Commands;
@@ -73,6 +74,6 @@ x");
                                       .BeTrue();
         }
 
-        void StartServer(Kernel remoteKernel, string pipeName) => Task.Run(() => { remoteKernel.EnableApiOverNamedPipe(pipeName); });
+        void StartServer(Kernel remoteKernel, string pipeName) => Task.Run(() => { remoteKernel.EnableApiOverNamedPipe(pipeName, new DirectoryInfo(Environment.CurrentDirectory)); });
     }
 }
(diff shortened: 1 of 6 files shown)


brettfo:
@jonsequitur The test failure is from [this test](https://github.com/dotnet/interactive/blob/93e54340503fa327a01f313c25784f5f0a66bc38/src/dotnet-interactive.Tests/CommandLine/CommandLineParserTests.cs#L321).  The issue is that the test process is actually changing its working dir (as requested) so all subsequent tests are operating from there so the test directory can never be cleaned up once the end of method (end of `using` block) is hit.  I propose three options, in order of preference:

1. Drop that unit test since we have an integration test that covers that exact scenario.  The downside is that if the parser is changed locally, the developer won't know until they submit a PR and see it fail there; the integration tests aren't part of the regular local workflow.
2. Rewrite this test to use a different, but valid directory to pass to `--working-dir`.  This would require a pre-test check similar to this:

``` csharp
var someOtherDirectory = typeof(JupyterRequestContext).Assembly.Location; // could be anything, just not bin/debug of the current process
Environment.CurrentDirectory.Should().Not().Be(someOtherDirectory)

parser.InvokeAsync($"--working-dir {someOtherDirectory}");

_startOptions.WorkingDir.FullName.Should().Be(someOtherDirectory);
```

The downside to this is that it might mess with subsequent tests in the future.  It doesn't seem to affect it currently, but I can't guarantee that now.

3. Refactor so that we can abstract away the actual changing of the directory, perhaps add a method `virtual void ChangeWorkingDirectory(string dir) {...}` on `KernelServer` and for the test add `class TestKernelServer : KernelServer { override ChangeWorkingDirectory() { /*noop*/ } }`.

jonsequitur:
These tests are parser tests and none of them are expected to actually trigger the underlying behaviors. They do this by replacing the default implementations with a shim:

https://github.com/dotnet/interactive/blob/93e54340503fa327a01f313c25784f5f0a66bc38/src/dotnet-interactive.Tests/CommandLine/CommandLineParserTests.cs#L41-L63

If the working directory is being changed, then the shim must not be working as expected. 

But you'll also notice not all of them call `InvokeAsync` in the first place. You can also just use `Parse`:

https://github.com/dotnet/interactive/blob/93e54340503fa327a01f313c25784f5f0a66bc38/src/dotnet-interactive.Tests/CommandLine/CommandLineParserTests.cs#L99-L112

PR #666, opened 2020-07-27
Wpf example

Keboo:
This is an initial implementation showing an example of using `#!connect` with a WPF app.

This also includes some custom formatters for a `SolidColorBrush` and `FrameworkElements`. 
This is largely based on the work done on a [live stream](https://www.youtube.com/watch?v=L_qtrCukwzk) with @jonsequitur and @colombod.  

This is largely a proof of concept.


samples/connect-wpf/WpfFormatterMixins.cs
```diff
@@ -0,0 +1,81 @@
...
+            UseFrameworkElementFormatter();
+            return kernel;
+        }
+
+        private static void UseSolidColorBrushFormatter()
+        {
+            Formatter<SolidColorBrush>.Register((brush, writer) =>
+            {
+                var color = brush.Color;
+                string stringValue = $"#{color.R:X2}{color.G:X2}{color.B:X2}{color.A:X2}";
```
jonsequitur, line 25: I think the "DSL style" for `PocketView` is more readable.

Keboo: Agreed, fixed

samples/connect-wpf/WpfConnect.csproj
```diff
@@ -0,0 +1,14 @@
+﻿<Project Sdk="Microsoft.NET.Sdk.WindowsDesktop">
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp3.1</TargetFramework>
+    <OutputType>WinExe</OutputType>
+    <UseWPF>true</UseWPF>
+    <DisableArcade>1</DisableArcade>
```
Keboo, line 7: Not sure if there is a better way to handle this. Had some compiler error with Arcade.

samples/connect-wpf/README.dib
```diff
@@ -0,0 +1,94 @@
...
+#!wpf
+#!dispatcher
+using System.Windows.Media;
+using System.Windows.Controls;
+using System.Windows;
+
+
+var grid = (Grid)App.MainWindow.Content;
+grid.Background = new SolidColorBrush(Colors.Blue);
+grid
```
Keboo, line 34: The first time this is run, you wont see the blue background. I believe this is because the brush is applied but has not yet been rendered. Subsequent runs will properly render with the blue background.

jonsequitur: Interesting. What triggers rendering?

samples/connect-wpf/WpfConnect.csproj
brettfo: Curious, what errors with arcade were you seeing?  It _should_ just work by running `build.cmd` once then regular work with `dotnet-interactive.sln`.

PR #674, opened 2020-07-30
enable jupyter notebooks to be opened with this extension

brettfo:
Since we don't want to 'steal' the `.ipynb` file association from the Python extension, we implement Jupyter Notebook support via a new command `dotnet-interactive.openNotebook`.  We also register a new notebook content type `dotnet-interactive-jupyter` that won't forcefully take away the file association.

There is also the new command `dotnet-interactive.saveNotebook` which will allow migrating between `.dib` and `.ipynb`.  The regular user action of Ctrl+S to save will detect the file extension and do the correct thing.

The bulk of this PR was refactoring file read/write to switch on the file extension, and that included a much-needed rewrite of the unit tests.


src/dotnet-interactive-vscode/src/fileFormats/interactive.ts
```diff
@@ -0,0 +1,113 @@
...
+        let lastNonBlank = findIndexReverse(cellLines, line => line.length > 0);
+        if (firstNonBlank >= 0 && lastNonBlank >= 0) {
+            lines.push(`#!${getSimpleLanguage(cell.language)}`);
+            lines.push('');
+            lines.push(...cellLines.slice(firstNonBlank, lastNonBlank + 1));
+            lines.push('');
+        }
+    }
+
+    return lines.join('\r\n');
```
jonsequitur, line 100: Is this potentially changing line endings? Also, is this going to run on Linux and macOS?

brettfo: > Is this potentially changing line endings?

Yes.

> is this going to run on Linux and macOS?

Yes.  This is what we currently do, so there'll be no change with what's currently there.

I considered adding something like `const lineSeparator = isWindows() ? '\r\n' : '\n';`, but my next PR will be to move file serialization/parsing to the kernel and that adds another wrinkle to the mix: it's possible that we may be running the kernel on a different architecture than where the notebook is open, so it's probably best for the notebook to request that the kernel serialize the notebook in a given format with the given line ending.

jonsequitur: Do we see the user's line ending preferences from VS Code?

brettfo: Added a commit to read the user setting which is either `\n`, `\r\n`, or `auto` (or `undefined` if they don't have the setting).  If it's not a known value, we check if they're on a Windows machine or not to determine.

The helper function added is guaranteed to _only_ return either `\n` or `\r\n`.

Issue #657, opened 2020-07-22
Missing /nbextensions/widgets/notebook/js/extension.js


darrenjrobinson:
### Describe the bug
#!javascript results in 404 GET /nbextensions/widgets/notebook/js/extension.js

### Please complete the following:

**Which version of .NET Interactive are you using?** 

```
Version: 1.0.136005+eb038ecb4877981ad1b17a7f6830868c0041aa11
Build date: 2020-07-10T18:53:44.0000000Z
```

![image](https://user-images.githubusercontent.com/384733/88121508-44d80080-cc09-11ea-9bd5-0c0d8a544c26.png)

- OS
    - [ ] Windows 10
    - [ ] macOS
    - [X ] Linux (Please specify distro) Ubuntu 18.04 LTS
    - [ ] iOS
    - [ ] Android
 - Browser 
    - [X] Chrome
    - [ ] Edge
    - [ ] Firefox
    - [ ] Safari
 - Frontend
    - [X ] Jupyter Notebook
    - [ ] Jupyter Lab
    - [ ] nteract
    - [ ] Visual Studio Code
    - [ ] Other (please specify)

```
You are using Jupyter notebook.

The version of the notebook server is: 5.7.8
The server is running on this version of Python:
Python 3.7.3 | packaged by conda-forge | (default, Jul  1 2019, 21:52:21) 
[GCC 7.3.0]
```

### Screenshots

![image](https://user-images.githubusercontent.com/384733/88121426-03dfec00-cc09-11ea-813b-205fb4375a0e.png)

![image](https://user-images.githubusercontent.com/384733/88121473-2245e780-cc09-11ea-9f8e-a1fe3226dd23.png)

colombod:
I was able to execute it and use both promsie and async approach

![image](https://user-images.githubusercontent.com/375556/88186881-b9249b00-cc2d-11ea-8265-a7bfa0183f9a.png)


The only difference I have is the js code
```
#!js

(async () => {
    let output = document.getElementById("output");
    output.innerHTML = await interactive.pwsh.getVariable("hi");
})();
```

darrenjrobinson:
It works for @jonsequitur too. But not for me. 
Only reason I can see is the error in the console

```
[W 21:25:47.058 NotebookApp] 404 GET /notebooks/work/undefinedresources/dotnet-interactive.js (172.17.0.1): No such file or directory: work/undefinedresources/dotnet-interactive.js
[...]

PR #682, opened 2020-08-04
js formatter waits for apiUri

colombod:
Addressing #657

### DIFF
--- a/src/dotnet-interactive.Tests/FormatterConfigurationTests.cs
+++ b/src/dotnet-interactive.Tests/FormatterConfigurationTests.cs
@@ -115,7 +115,7 @@ alert('hello');
                 .Throw<TimeoutException>()
                 .Which
                 .Message
-                .Should().Be("Timeout resolving api uri, please try again.");
+                .Should().Be("Timeout resolving the kernel's HTTP endpoint. Please try again.");
         }
 
         [Fact]
--- a/src/dotnet-interactive.Tests/FormatterConfigurationTests.cs
+++ b/src/dotnet-interactive.Tests/FormatterConfigurationTests.cs
@@ -84,7 +84,7 @@ namespace Microsoft.DotNet.Interactive.App.Tests
         {
             var frontendEnvironment = new HtmlNotebookFrontedEnvironment(new Uri("http://12.12.12.12:4242"));
 
-            CommandLineParser.SetUpFormatters(frontendEnvironment, new StartupOptions(httpPort: new HttpPort(4242)), 1.Seconds());
+            CommandLineParser.SetUpFormatters(frontendEnvironment, new StartupOptions(httpPort: new HttpPort(4242)), 5.Seconds());
             var script = new ScriptContent("alert('hello');");
             var mimeType = Formatter.PreferredMimeTypeFor(script.GetType());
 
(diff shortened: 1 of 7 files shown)


src/dotnet-interactive.Tests/FormatterConfigurationTests.cs
```diff
@@ -97,18 +93,35 @@ public void ScriptContent_type_is_wrapped_when_http_and_the_frontendEnvironment_
...
+            CommandLineParser.SetUpFormatters(frontendEnvironment, new StartupOptions(httpPort: new HttpPort(4242)), 1.Seconds());
+            var script = new ScriptContent("alert('hello');");
+            var mimeType = Formatter.PreferredMimeTypeFor(script.GetType());
+
+            Action formatting = () =>  script.ToDisplayString(mimeType);
+            formatting.Should()
+                .Throw<TimeoutException>()
+                .Which
+                .Message
+                .Should().Be("Timeout resolving api uri, please try again.");
```
jonsequitur, line 118:
```suggestion
                .Should().Be("Timeout resolving the kernel's HTTP endpoint. Please try again.");
```

src/dotnet-interactive/CommandLine/CommandLineParser.cs
```diff
@@ -507,15 +507,24 @@ public static void SetUpFormatters(FrontendEnvironment frontendEnvironment, Star
                     {
                         Formatter<ScriptContent>.Register((script, writer) =>
                         {
-                            var fullCode = $@"if (typeof window.createDotnetInteractiveClient === typeof Function) {{
-createDotnetInteractiveClient('{frontedEnvironment.DiscoveredUri.AbsoluteUri}').then(function (interactive) {{
-let notebookScope = getDotnetInteractiveScope('{frontedEnvironment.DiscoveredUri.AbsoluteUri}');
+                            if (!Task.Run(() =>
+                            {
+                                var apiUri = frontedEnvironment.GetApiUriAsync().Result;
```
jonsequitur, line 512: The `Task.Run` call can be async.

src/dotnet-interactive/CommandLine/CommandLineParser.cs
```diff
@@ -507,15 +507,24 @@ public static void SetUpFormatters(FrontendEnvironment frontendEnvironment, Star
...
 }}";
-                            IHtmlContent content =
-                                PocketViewTags.script[type: "text/javascript"](fullCode.ToHtmlContent());
-                            content.WriteTo(writer, HtmlEncoder.Default);
+                                IHtmlContent content =
+                                    PocketViewTags.script[type: "text/javascript"](fullCode.ToHtmlContent());
+                                content.WriteTo(writer, HtmlEncoder.Default);
+                            }).Wait(apiUriTimeout))
+                            {
+                                throw new TimeoutException("Timeout resolving api uri, please try again.");
```
jonsequitur, line 525: Is this throwing even when the timeout hasn't expired?

colombod: only if we complete without timeout Task.Wait doesn't throw if we sue a timeout, it returns true if we completed within the timeout interval, false otherwise.

Issue #645, opened 2020-07-14
F# - Too many namespaces opened by default


dsyme:
I'm looking over the source for the F# kernel and I notice that `FSharpPocketViewTags` is opened by default, e.g. through this list in `UseDefaultFormatting`.  This brings into scope a swathe of HTML tags like `a` and `area`. 

We should not be doing this in the default programming model and it won't create a programming experience that will last the test of time, even if it marginally improves demos.  People won't know where these values are coming from or what they are for, and they might very easily collide with other programming models.  

In any case, F# basically already has its own point of view on default-opened namespaces for a programming model.  It's ok to add `FSharpKernelHelpers` to this but I would expect pretty much everything else to be explicitly opened.

For C# I think you should discuss with the C# designers - this is an issue for the basic concept of the C# scripting programming model and the defaults really impact on the reusability of script-like assets across different settings.

Here is the current list of explicit opens:

    // From UseDefaultNamespaces
    open System
    open System.Text
    open System.Threading.Tasks
    open System.Linq

    // From UseKernelHelpers
    open Microsoft.DotNet.Interactive.FSharp.FSharpKernelHelpers

    // From UseDefaultFormatting
    open Microsoft.AspNetCore.Html 
    open Microsoft.DotNet.Interactive.FSharp.FSharpPocketViewTags 
    open XPlot.Plotly
    open Microsoft.DotNet.Interactive.Formatting

Of these, I feel they categorise as follows:

##### Definitely not:

    open Microsoft.AspNetCore.Html 
    open Microsoft.DotNet.Interactive.FSharp.FSharpPocketViewTags 
    open XPlot.Plotly
    open Microsoft.DotNet.Interactive.Formatting

##### Probably not (for F# code)

    open System.Threading.Tasks
    open System.Linq

##### Probably ok (for F# code)

    open System
    open System.IO
    open System.Text

##### Definitely ok:

    open Microsoft.DotNet.Interactive.FSharp.FSharpKernelHelpers

gbaydin:
> People won't know where these values are coming from or what they are for, and they might very easily collide with other programming models.

I can confirm. I've been confused about where the list of names I keep seeing in autocomplete come from.

KevinRansom:
Don has already fixed this,

For F# the default namespaces are:
````
System
System.IO
System.Text
Microsoft.DotNet.Interactive.Formatting
````

For C# the default namespaces are:
````
System
System.IO
System.Text
Microsoft.DotNet.Interactive.Formatting
````

PR #688, opened 2020-08-07
Trim back on default namespaces for F#

dsyme:
As mentioned in #645 , there are too many namespaces open by default for F#, and in particular littering the unqualified namespace of the programming model with `a`, `p` etc of the HTML tags is a real problem - I personally had no idea what was going on with the type errors I got when accidentally using these names in code that has nothing to do with HTML.  This is particularly a problem in F# I think but I'm pretty sure you want to pull back on this in C# too

In any case, for F# only the standard namespaces of the F# scripting model should be open by default, plus the one extra default namespace (and anything directly associated with it) for things like `display`

    open Microsoft.DotNet.Interactive.FSharp.FSharpKernelHelpers

The namespaces `System`, `System.Linq` etc. should not be open by default for F# code.  F# already opens `FSharp.Core`, `FSharp.Core.Operators` etc. and that is its standard scripting model.

### DIFF
--- a/src/Microsoft.DotNet.Interactive.FSharp/FSharpKernelExtensions.fs
+++ b/src/Microsoft.DotNet.Interactive.FSharp/FSharpKernelExtensions.fs
@@ -17,8 +17,8 @@ open XPlot.Plotly
 [<AbstractClass; Extension; Sealed>]
 type FSharpKernelExtensions private () =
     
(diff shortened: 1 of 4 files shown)


dsyme:
On my machine I got one failing test for C# on master `hover_text_commands_have_offsets_normalized_after_switching_to_the_same_language` but it is unrelated to these changes, I'm not sure what's up

dsyme:
I've included this in #689 and perhaps it's better to just focus on that PR

Issue #687, opened 2020-08-07
Move to a more standard style of HTML DSL for F# #689


dsyme:
I'd like us to review the F# FSharpPocketViewTags programming model (though the PocketView implementation can be used underneath)

The F# community have a fairly standard way of doing HTML DSLs, which is the same across many F# web programming technologies like Fable (client-side HTML) and Giraffe (server-side HTML).  Roughly speaking the end result is to always look like this

    div 
      [ ] 
      [ p [ todoStyle ] [ str item.Description ]
        button [ ClassName "button is-info"; dispatchToggle ] [ str toggleText ]
        divider
        button [ ClassName "button is-danger"; dispatchDelete ] [ str "Delete" ] ]

That is, each HTML element is `tag [...attributes...] [ ...content...]`

This is how we should be descrbing HTML in the F# .NET Interactive programming model.  It doesn't have to be precisely the same as every other HTML DSL in F# (nor share code) but this is the standard syntax in the community and the one we should be using for window dressing.  It's also very, very effective as a DSL syntax, I and others absolutely love using it, particularly because it jives so well with F# list expressions.

I'd also like to make sure the namespace is not open by default, see #645, and that the way of opening is fairly simple. Currently it is

    open Microsoft.DotNet.Interactive.FSharp.FSharpPocketViewTags

which is not great  - I'd like any of

    open FSharpPocketViewTags
    open PocketViewTags
    open HtmlTags

I'll experiment with options

jonsequitur:
I think given that great HTML DSLs already exist in F#, we should remove the F# implementation of `PocketViewTags`. 

The C# version is likely to support extension authors more often than notebook users and there as well I agree we shouldn't open the namespace by default.

dsyme:
> I think given that great HTML DSLs already exist in F#, we should remove the F# implementation of PocketViewTags.

At the moment my feeling is we should replace it by an implementation of an F# DSL that is surface syntax for PocketViewTags and conforms to the HTML element part of [Giraffe.ViewEngine](https://github.com/giraffe-fsharp/Giraffe.ViewEngine) exterior API. I'm trialling that right now.

I'm loathe to diverge on the actual HTML object because I suspect 
1. there's devil in the detail 
2. there may be c#-f# interop on this
3. Giraffe.ViewEngine is not yet something I'd  wouldn't want to take an actual dependency  on 

BTW here's an example of the code (compare the C# code with explicit use of ResizeArray due to lack of list expression syntax):

```
open Html

[...]

PR #689, opened 2020-08-07
move to a more standard DSL for F# HTML

dsyme:
This adjusts the HTML DSL for F# as described in https://github.com/dotnet/interactive/issues/687

Also includes #688 to trim back on default namespaces

Typical F# html code looks like this

```fsharp
open Html  // opens Microsoft.DotNet.Interactive.FSharp.FSharpKernelHelpers.Html

let register (df:DataFrame) (writer:TextWriter)
    let take = 20
    table [] [
      thead [] [
        th [] [i [ str "index"]]
        for c in df.Columns -> th [] [c.Name]
      ]
      tbody [] [
        for i in 0 .. min take (int df.Rows.Count - 1) do
          td [] [ object i ]
          for o in df.Rows.[int64 i] do
            td [] [ object o ]
      ]
    ]
    |> writer.Write
Formatter<DataFrame>.Register( (fun df writer -> register df writer), mimeType = "text/html")
```

TODO:
* [x] please check design
* [x] check the notebooks I adjusted by hand


dsyme:
This is ready

src/Microsoft.DotNet.Interactive.FSharp.Tests/ApiTests.fs
```diff
@@ -3,35 +3,39 @@
...
 
     [<Fact>]
-    member __.``inner HTML from content``() =
-        Assert.Equal("<div>d</div>", div.innerHTML("d").ToString())
+    member __.``inner HTML from string``() =
+        Assert.Equal("<div>d</div>", (div [] [str "d"]).ToString())
+
+    [<Fact>]
+    member __.``inner HTML from object``() =
+        Assert.Equal("<div>11</div>", (div [] [object 11]).ToString())
```
jonsequitur, line 25: A more interesting example here would be an object of a type that leads to additional formatting.

dsyme: I've added quite a number of more tests here

dsyme:
I went through all the F# docs for formatting to make sure things were adding up.

Along the way I merged the "HTML.ipynb" into the "Displaying Outputs" and also renamed the notebook files so they don't have spaces in their names, to help me get the cross-references right.

I also added a fix and testing for #703.

dsyme:
I also relented on a little on the question of which namespaces are open for F# and also included these by default:

    open System
    open System.IO
    open System.Text

I'll add this to the documentation.

dsyme:
OK, I've added a document on the F# notebook programming model.

dsyme:
@jonsequitur @colombod ready again

src/Microsoft.DotNet.Interactive.FSharp/FSharpHtml.fs
```diff
@@ -0,0 +1,511 @@
...
+                | Text s -> encoder.Encode(writer, s)
+                | Tagged  p -> p.WriteTo(writer, encoder)
+                | Obj o -> encoder.Encode(writer, string o)
+
+    type HtmlAttribute = string * obj
+
+    [<AutoOpen>]
+    module HtmlElements =
+        type internal IMarker = interface end
+        let private f name (attrs: HtmlAttribute list) (content: IHtmlContent list) =
```
colombod, line 34: could we call ```f``` something like ```htmlElement``` or anything that could make it easier for contributors to identify?

src/Microsoft.DotNet.Interactive.FSharp/FSharpKernelExtensions.fs
```diff
@@ -17,35 +17,53 @@ open XPlot.Plotly
...
-        let code = @"
-open System
-open System.Text
-open System.Threading.Tasks
-open System.Linq"
+        // F# has its own views on what namespaces are open by default in its scripting model
+        kernel
+
+    [<Extension>]
+    static member UseExtraNamespacesForTesting(kernel: FSharpKernelBase) =
```
jonsequitur, line 48: It looks like this should go in the test project.

src/Microsoft.DotNet.Interactive.FSharp/FSharpHtml.fs
dsyme: Will do.  I will also remove the tags "html" and "body" which we never use in this context to reduce namespace pollution

src/Microsoft.DotNet.Interactive.FSharp/FSharpKernelExtensions.fs
dsyme: Thanks, done

src/Microsoft.DotNet.Interactive.FSharp/FSharpHtml.fs
dsyme: Thanks, done

dsyme:
I've adjusted for code review (would love to get this one in as I'm needing it in documentation etc.)

Issue #690, opened 2020-08-07
Questions on Formatters 


dsyme:
Current Formatters are registered by type + mime-type.  I'm trying to determine the exact rule that is used to select the relevant formatter.

### Question 1 - Sometimes formatters seem to apply to subtypes but not always, e.g. `obj` formatters 

For example consider
```fsharp
Formatter<IComparable>.Register((fun (s:IComparable) writer -> writer.Write(s.ToString() + "!!!! (IComparable, html)")), "text/html");
Formatter<IComparable>.Register((fun (s:IComparable) writer -> writer.Write(s.ToString() + "!!!! (IComparable, plain)")), "text/plain");
```

Then values like `1`, `System.DateTime.Now` and `DayOfWeek.Monday` do invoke the IComparable printer.:
```
> 1
1!!!! (IComparable, html)
```
However, if I do this
```
Formatter<obj>.Register((fun s writer -> writer.Write(s.ToString() + "!!!! (obj, html)")), "text/html");
Formatter<obj>.Register((fun s writer -> writer.Write(s.ToString() + "!!!! (obj, plain)")), "text/plain");
```
then the `obj` formatter is not invoked at all unless the exact runtime type of the value is `System.Object`, e.g.

* `1` and `box 1` do not invoke the printer (surprise!)
* `DateTime.Now` does not invoke the printer (surprise!)
* `new System.Object()`  does invoke the printer ( as expected) 

I'm just a little confused what the rule is here. Is `obj` special?  Also, what's the priority if multiple conflicting formatters are registered, e.g. one form System.DateTime and one for IComparable?

### Question 2 - Formatters can't apply to selective values

There is no way for a formatter to apply only to specific values except via System.Type. The conext here is that in F# Interactive (AddPrinter, returning string) and older F# notebooks (AddHtmlPrinter, returning HTML), formatters can return `null` to indicate that they are not interested in formatting a specific value.  

This means you could, for example, register a formatter for type `System.Object` and pick specific classes of values to your heart's content, e.g. 

```fsharp
fsi.AddPrintTransformer(fun (x:obj) ->
    match x with
    | :? string as s when s = "beep" -> box ["quack"; "quack"; "quack"]
    | _ -> null)

let y = "beep"
```
This outputs:
```
val y : string = ["quack"; "quack"; "quack"]
```

For example, you can register a formatter for "PythonObject" but only decide to format Python tensors.   And you can have multiple such formatters active.

[...]

PR #694, opened 2020-08-08
State-free formatter selection

### LINKED ISSUE #692: Evaluation leads to a default formatter registration, then user-defined formatters are ignored
Evaluation implicitly leads to a default formatter registration which then leads to user-defined formatters being ignored.  THis is very confusing.

0. Restart kernel

1. Evaluate `DayOfWeek.Monday`

2. Register a formatter for IComparable

3. Evaluate `DayOfWeek.Monday` again

Expected: The formatter for IComparable is invoked

Actual: the same formatting as step 1 is used

Compare with 

0. Restart kernel

1. Register a formatter for IComparable

2. Evaluate `DayOfWeek.Monday` again

In this case, the formatter for IComparable is invoked.  The fact that I've previously formatted a DayOfWeek value should be irrelevant.

The formatter architecture is very sniffy, it seems full of hidden state like this.  Somehow evaluation causes default formatters to be populated and then those take precedence of user-specified formatters.  

There should be a simple declarative spec like 

* "we run all registered user-defined formatters in sequence finding the first that matches, then run default formatters", or

* "we look at all user-defined formatters and select the one that is closest match by type hierarchy.  If none match we do the same with the default formatters"

I took a  look at the implementation and TBH I saw some code fragments that indicate these type-indexed tables get automatically populated by defaults which I assume that leads to this kind of problem.

dsyme:
This is a implements the speclet at https://github.com/dotnet/interactive/issues/692#issuecomment-670981993 , now included in the docs

It implements "state free" formatter selection, so the relevant formatter is chosen by a simple and intuitive set of rules, and doesn't depend on what evaluations have been performed apart from formatter registrations

This fixes #692 and the problem with registering "object" formatters in #690.   Apart from that the behaviour of `display` is effectively unchanged.

The key rules are [`InferBestFormatter`](https://github.com/dotnet/interactive/pull/694/files#diff-e6c2e9c0248d36096dc2d295af93ba2cR384) and [`IsRelevantFormatterFor(Type, mimeType)`](https://github.com/dotnet/interactive/pull/694/files#diff-fbe198d9d68a26b6c3ff134d92eaf1ddR100)

This PR also aligns string-emits (e.g.  [here](https://github.com/dotnet/interactive/blob/master/src/Microsoft.DotNet.Interactive.Formatting/PocketView.cs#L194) and [here](https://github.com/dotnet/interactive/blob/master/src/Microsoft.DotNet.Interactive.Formatting/Formatter.cs#L265)) to 

1. always encode and 
2. never emit span. 

These need to be consistent.  Also the default string formatter for HTML was trying to emit a "span" but for practical purposes this wasn't being used because of the two links above.

Some cleanup happens along the way

* The lists of default formatters are now clean and simple, collecting up the various fragments of default formatter behaviour into declarative lists in [DefaultHtmlFormatterSet](https://github.com/dotnet/interactive/blob/50966710d66140cff5752c99d617430cd69e79f9/src/Microsoft.DotNet.Interactive.Formatting/DefaultHtmlFormatterSet.cs),  [DefaultPlainTextFormatterSet](https://github.com/dotnet/interactive/blob/50966710d66140cff5752c99d617430cd69e79f9/src/Microsoft.DotNet.Interactive.Formatting/DefaultPlainTextFormatterSet.cs) and [DefaultJsonFormatterSet](https://github.com/dotnet/interactive/blob/50966710d66140cff5752c99d617430cd69e79f9/src/Microsoft.DotNet.Interactive.Formatting/DefaultJsonFormatterSet.cs).

* Remove the formatter factory from the default sets (in favour of selecting a formatter using `IsRelevantFormatterFor`)

* The DefaultFormatterSets are now unchanging arrays (given to the selection rules implementing the spec)

* IFormatterSet, FormatterSetBase, TryGetFormatterForType, AddFormatterFactory and the associated state are removed because the selection rules are sufficient (e.g. allow selecting an 'Enumerable' formatter for a collection - there is no need to *create* a formatter with the precise type, we must just *select* one of the right type)

* The "Create" methods were awkward because they factored in some but not all default behaviour in a somewhat adhoc way.  These shouldn't be made public because of this and instead I've removed them and we have `GetBestFormatterFor(Type)` with `GetDefaultFormatterForAnyObject(Type, includeInternals)` and `GetDefaultFormatterForAnyEnumerable(Type)` internal.

* Many special cases where formatters with bespoke logic were being created in order to format multiple subtypes have been simplified (Type, RuntimeType, JToken).  We can now just create a formatter for System.Type and the formatter selection rules do the rest.

  For example 

                [typeof(Type)] = _formatterForSystemType,
                [typeof(Type).GetType()] = _formatterForSystemType,
                private static readonly HtmlFormatter<Type> _formatterForSystemType  = new HtmlFormatter<Type>((type, writer) =>

  becomes one default formatter, selected for all subtypes of Type.

                new HtmlFormatter<Type>((type, writer) => ...)

The same performance profile is kept (formatter selections are computed only once, until a state change occurs with another Register, and default formatters are not recomputed).


src/Microsoft.DotNet.Interactive.Formatting.Tests/HtmlFormatterTests.cs
```diff
@@ -61,7 +61,7 @@ public void Formatter_expands_properties_of_ExpandoObjects()
                 expando.Name = "socks";
                 expando.Count = 2;
 
-                var formatter = HtmlFormatter<ExpandoObject>.Create();
+                var formatter = HtmlFormatter<ExpandoObject>.GetBestFormatterFor();
```
jonsequitur, line 64: For a more intuitive signature and to align to the recent addition of `Formatter.Register<T>` as well as the non-generic signature, I'd suggest we do: `HtmlFormatter.GetBestFormatterFor<T>`.

src/Microsoft.DotNet.Interactive.Formatting.Tests/FormatterTests.cs
```diff
@@ -245,6 +245,76 @@ public void Formatters_can_be_registered_for_concrete_types(string mimeType, boo
...
+
+            // The more specific formatter is chosen
+            new object().ToDisplayString(mimeType)
+                                            .Should()
+                                            .Be("world");
+
+            // The object formatter is chosen
+            DateTime.Now.ToDisplayString(mimeType)
+                                            .Should()
+                                            .Be("world");
```
jonsequitur, line 315: Splitting the assertions into separate theory items or even separate tests will give clearer feedback during refactoring when not all cases pass or fail together.

An additional case that would help clarify the scenario: which formatter should `FileSystemInfo` use?

src/Microsoft.DotNet.Interactive.Formatting/DefaultHtmlFormatterSet.cs
```diff
@@ -74,38 +44,70 @@ protected override bool TryInferFormatter(Type type, out ITypeFormatter formatte
...
                 }),
 
-                [typeof(HtmlString)] = new HtmlFormatter<HtmlString>((view, writer) => view.WriteTo(writer, HtmlEncoder.Default)),
+                new HtmlFormatter<HtmlString>((view, writer) => view.WriteTo(writer, HtmlEncoder.Default)),
 
-                [typeof(JsonString)] = new HtmlFormatter<JsonString>((view, writer) => view.WriteTo(writer, HtmlEncoder.Default)),
+                new HtmlFormatter<JsonString>((view, writer) => view.WriteTo(writer, HtmlEncoder.Default)),
 
-                [typeof(PocketView)] = new HtmlFormatter<PocketView>((view, writer) => view.WriteTo(writer, HtmlEncoder.Default)),
+                new HtmlFormatter<PocketView>((view, writer) => view.WriteTo(writer, HtmlEncoder.Default)),
```
jonsequitur, line 51: Even prior to these changes, I believe all three of these cases should be replaceable by a single formatter for `IHtmlString`.

src/Microsoft.DotNet.Interactive.Formatting.Tests/HtmlFormatterTests.cs
dsyme: cool will do

src/Microsoft.DotNet.Interactive.Formatting.Tests/FormatterTests.cs
dsyme: Agreed thanks will do

src/Microsoft.DotNet.Interactive.Formatting/DefaultHtmlFormatterSet.cs
dsyme: that would be great

src/Microsoft.DotNet.Interactive.Formatting.Tests/HtmlFormatterTests.cs
dsyme: The following feel like they should not be public, though some tests use them

Would you prefer

1. they are public

2. they are internal and the tests use an IVT

3. I adjust the tests so there are not used directly (since they are really an implementation detail I think)

I think (3) is best?

dsyme:
@jonsequitur I'm feeling a bit quesy that this test fails (in master too)

```
            [Fact]
            public void It_formats_string_without_a_span_or_encoding()
            {
                var text = "hello<b>world</b>";

                var html = text.ToDisplayString("text/html");

                html.Should().BeEquivalentHtmlTo(text);
            }

```

The problem is that, while the text is valid HTML, we have invalid XML `hello<b>world</b>` created and the `BeEquivalentHtmlTo` chokes on this.

Similarly consider

                var text = "hello<b>world<b> WRONG HTML";

It feels to be rendering strings as HTML should render them as valid HTML, i.e. HTMLEncode.  Objects nested in PocketView do get an HtmlEncode:

```
                    case string s:
                        writer.Write(s.HtmlEncode());
                        break;
```

but for HTML rendering we use this default HTML formatter:

                new HtmlFormatter<string>((s, writer) => writer.Write(span(s))),

src/Microsoft.DotNet.Interactive.Formatting.Tests/HtmlFormatterTests.cs
```diff
@@ -45,13 +45,13 @@ public void Null_references_are_indicated()
...
             [Fact]
-            public void Formatter_does_not_expand_string()
+            public void Formatter_does_not_put_span_around_string()
             {
-                var formatter = HtmlFormatter<string>.GetBestFormatterFor();
+                var formatter = HtmlFormatter.GetBestFormatterFor<string>();
 
                 var s = "hello".ToDisplayString(formatter);
 
-                s.Should().Be("<span>hello</span>");
```
jonsequitur, line 54: This looks like a behavior-breaking change. The `span` element is present to avoid dropping text directly into the DOM.

dsyme:
@jonsequitur I went through and made the adjustments to make sure strings are processed by the default formatters, always producing valid HTML and JSON  when those mime types are requested.  Also added lots of testing around this.  ScriptContent had to be adjusted to make sure it is not HTMLEncoded, but that's as expected.

I think we could now also make the default for strings be HTML rather than plain text, there's no need to make them special.

src/Microsoft.DotNet.Interactive.Formatting.Tests/HtmlFormatterTests.cs
```diff
@@ -326,6 +326,70 @@ public void It_formats_non_generic_dictionaries_that_arent_generic_as_tables_wit
...
+                var date1 = DateTimeOffset.Now;
+
+                var html = date1.ToDisplayString("text/html");
+
+                html.Should().BeEquivalentHtmlTo(
+                    $"<span>{date1.ToString("u")}</span>");
+            }
+
+            [Fact]
+            public void It_formats_Enum_correctly()
```
jonsequitur, line 352: I would suggest `It_formats_enum_value_using_the_name` or something similarly more explicit about the intended outcome.

src/Microsoft.DotNet.Interactive.Formatting/PocketViewTags.cs
```diff
@@ -85,5 +87,11 @@ public static class PocketViewTags
         public static dynamic u => _.u;
         public static dynamic ul => _.ul;
         public static dynamic video => _.video;
+
+        // This can be used to indicates points
+        // where we are embedding an arbitrary value into the HTML content
+        // and we are explicitly indicating that the value can be either IHtmlContent
+        // or will otherwise be formatted as plaintext.
+        public static object arbitrary(object value) => value;
```
jonsequitur, line 95: If I understand this correctly, it's intended to defer formatting? 

We also have the `HTML` method which wraps the object in an `IHtmlContent`, which is more explicit about the intent in the HTML case.

src/Microsoft.DotNet.Interactive.Formatting.Tests/HtmlFormatterTests.cs
dsyme: So this is when grabbing the default formater directly.  

However the default "string" formatters were never actually being used (so `span` never actually emitted) because the actual behaviour of "display" was indeed to dump the string straight into the DOM (unencoded), because of this  https://github.com/dotnet/interactive/blob/master/src/Microsoft.DotNet.Interactive.Formatting/Formatter.cs#L265 

So we have three decisions

1. Do we dump strings encoded or not?  I think they have to be encoded or else invalid HTML is produced.  

2. Do we dump with `<span>` or not?  I think not - currently the defulat formatter says "span" but it's never actually emitted today.  It doesn't feel right to start emitting it - if the user wants a span they should make a span element?

3. the separate question of preserving whitespace or not (which we can del with separately)

src/dotnet-interactive.Tests/FormatterConfigurationTests.cs
```diff
@@ -79,6 +79,21 @@ public void ScriptContent_type_is_formatted()
             formattedValue.Value.Should().Be(@"<script type=""text/javascript"">alert('hello');</script>");
         }
 
+        [Fact]
+        public void ScriptContent_type_with_possible_html_characters_is_formatted_unchanged()
```
jonsequitur, line 83: Perhaps `ScriptContent_type_with_possible_html_characters_is_not_HTML_encoded`.

src/Microsoft.DotNet.Interactive.Formatting/PocketViewTags.cs
dsyme: Well, the intent is to mark the places we end up coming back around to either IHtmlContent or plain text formatting here https://github.com/dotnet/interactive/blob/master/src/Microsoft.DotNet.Interactive.Formatting/PocketView.cs#L213

Initially I chose the name `arbitraryIHtmlContentOrElseUsePlainTextFormatting` since it at least explains what's actually going to happen to this content :))

src/dotnet-interactive.Tests/FormatterConfigurationTests.cs
dsyme: thanks done

src/Microsoft.DotNet.Interactive.Formatting.Tests/HtmlFormatterTests.cs
dsyme: Thanks done

src/Microsoft.DotNet.Interactive.Formatting.Tests/HtmlFormatterTests.cs
jonsequitur: 1. We should encode.

2. A user can do `display("hello", "text/html")` and in the absence of this behavior, I believe they'll see no output. My memory is that this bug was the reason for the current behavior.

3. We should preserve whitespace. Perhaps a `pre` tag would be a more informative way to display strings.

src/Microsoft.DotNet.Interactive.Formatting/PocketViewTags.cs
dsyme: Happy to change this however you see fit.  The use of `dynamic` and `object` makes it hard to mark up these recursive injection places (= deferred formatting I guess) except via a discpline, and I find them quite important for understanding what's going on.

jonsequitur: By deferred I guess I mean that by calling `arbitrary`, we're deferring the decision back to the formatter, but yes, it's not deferred by very long. 

Would `FormatToHtml` capture the intent?

src/Microsoft.DotNet.Interactive.Formatting.Tests/HtmlFormatterTests.cs
dsyme: @jonsequitur I updated the comment above to explain how all three HTML string emits are being aligned to be consistent.

src/Microsoft.DotNet.Interactive.Formatting.Tests/HtmlFormatterTests.cs
jonsequitur: Agreed.

src/Microsoft.DotNet.Interactive.Formatting/PocketViewTags.cs
dsyme: > Would FormatToHtml capture the intent?

Well, the current situation is to very often format as plaintext :) So it's kind of the opposite to `FormatToHtml` at the moment

src/Microsoft.DotNet.Interactive.Formatting.Tests/HtmlFormatterTests.cs
dsyme: > I believe they'll see no output. My memory is that this bug was the reason for the current behavior.

They'll see the output (and note at the moment we're not actually emitting spans - as far as I know that `<span>` is never coming out around strings).  Even in a full document the text "hello" displays as "hello".  I don't know of any situation that where HTML will actually drop encoded text where a span would be valid.

dsyme: > We should preserve whitespace. Perhaps a pre tag would be a more informative way to display strings.

See https://stackoverflow.com/questions/45326130/how-can-i-display-the-text-in-pre-tag-in-the-default-fonts-that-html-uses/45326306 for a possibility (at least for `ToString()` outputs).  This _would_ require a wrapping `<span>` or perhaps even a `<p>`

src/Microsoft.DotNet.Interactive.Formatting/PocketViewTags.cs
dsyme: I'll leave it as `arbitrary` with tthe comment for now and we can resolve later

dsyme:
@jonsequitur This is now ready.   I'll continue to look at additional issues in other PRs, would be great to get this one in.

src/Microsoft.DotNet.Interactive.Formatting/PocketViewTags.cs
jonsequitur: To clarify, my understanding here is that we want `IHtmlContent` back (including if the value was generated by the `PlainTextFormatter` and then HTML-encoded). The current code doesn't know how to do that so it defers the decision back to the infrastructure. The result is plain text today but might eventually include nestable HTML.

Is this accurate? If so, is it useful to change `arbitrary` to something like:

```csharp
public IHtmlContent YouFigureItOut() => /*...*/
```

src/Microsoft.DotNet.Interactive.Formatting/HtmlFormatter.cs
```diff
@@ -11,30 +11,20 @@ namespace Microsoft.DotNet.Interactive.Formatting
...
-                                             {
-                                                 typeof(bool)
-                                             });
-
-            return (ITypeFormatter) genericCreateForAllMembers.Invoke(null, new object[]
-            {
-                includeInternals
-            });
-        }
+        public static ITypeFormatter GetBestFormatterFor(Type type) =>
```
jonsequitur, line 14: Can we find a name for these methods that's clearer about the intent? What makes a formatter "best"?

dsyme: Hmmm.... Maybe `GetCurrentFormatterFor` or `GetFormatterFor` or `GetPreferredFormatterFor` (since we already use that word)

dsyme: The API would then be

define:

Register
SetPreferredMimeTypeFor

query:

GetPreferredMimeTypeFor
GetPreferredFormatterFor
GetRegisteredFormatters
GetRegisteredMimeTypesFor

jonsequitur: That looks clearer.

PR #699, opened 2020-08-09
Update nuget-overview.md

### LINKED ISSUE #698: Documentation: How do I load nuget references from alternative package sources?
Trying to use dotnet interactive to create c# notebooks. I need to reference a package on a private Nuget feed. Looking at the documentation I do not see how to do this.

* I tried including a `nuget.config` in the notebook folder. 
* I tried various variations on adding source to the `#r` command.

```
#r "nuget:My.Private.Nuget -Source https://www.myget.org/F/{feed}/api/v3/index.json"
```

Result:
> No packages exist with this id in source(s): Microsoft Visual Studio Offline Packages, nuget.org

Update [dotnet interactive nuget docs](https://github.com/dotnet/interactive/blob/master/docs/nuget-overview.md) to communicate how to do more than just reference a nuget by id and version.

I also looked for details on how the `#r` command is implemented but I do not see enough in the architecture docs to get there. Sorry if I missed something.

KevM:
Fixes #698 

I noticed the documentation for magic commands after creating #698. 

What is a good demo alternative source? I am not sure my example is very good.

### DIFF
--- a/docs/nuget-overview.md
+++ b/docs/nuget-overview.md
@@ -19,5 +19,5 @@ If you'd like to use a specific version, you can specify it like this:
 If your nuget package is not hosted on the main Nuget feed you can specify an alternative nuget source using `#i`.
 
 ```csharp
-#i "https://www.myget.org/F/{feedurl}/api/v3/index.json"
+#i "nuget:https://www.myget.org/F/{feedurl}/api/v3/index.json"
 ```
--- a/docs/nuget-overview.md
+++ b/docs/nuget-overview.md
@@ -16,9 +16,21 @@ If you'd like to use a specific version, you can specify it like this:
 
 ## Adding a Nuget Source
 
-If your nuget package is not hosted on the main Nuget feed you can specify an alternative nuget source using `#i`. In the following example we are adding the dotnet pre-release nuget feed.
+If your nuget package is not hosted on the main Nuget feed you can specify an alternative nuget source using `#i`. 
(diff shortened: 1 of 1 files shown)


docs/nuget-overview.md
```diff
@@ -14,8 +14,10 @@ If you'd like to use a specific version, you can specify it like this:
...
 
+## Adding a Nuget Source
 
+If your nuget package is not hosted on the main Nuget feed you can specify an alternative nuget source using `#i`.
 
-
-
-
+```csharp
+#i "https://www.myget.org/F/{feedurl}/api/v3/index.json"
```
jonsequitur, line 22: A few suggestions: 

* Using a real feed URL would be useful, to avoid people using this example literally.
* An example pointing out that this also supports file paths.
* The official syntax is `#i "nuget:...". While your example will work, there's some discussion over making it stricter.

KevM: I addressed your feedback.

> An example pointing out that this also supports file paths.

I know you suggested to use a real example for the remote nuget source. Is it OK to use a fake example here for a local folder. It feels like this might be the most typical usage.

KevM:
I added a local folder example as suggested. For the remote source you suggested using a real example. Is it OK to use a fake example here for a local folder. It feels like this might be the more typical usage.

jonsequitur:
Thanks, @KevM!

PR #700, opened 2020-08-10
enable extension to expose static content

colombod:
kernel extensions can provide content if they implement the ```IStaticContentSource``` interface

### DIFF
--- a/src/Microsoft.DotNet.Interactive/Events/KernelExtensionLoaded.cs
+++ b/src/Microsoft.DotNet.Interactive/Events/KernelExtensionLoaded.cs
@@ -15,18 +15,15 @@ namespace Microsoft.DotNet.Interactive.Events
         [JsonIgnore]
         public IKernelExtension KernelExtension { get; }
 
-        public string KernelName { get; }
 
         [JsonConstructor]
-        public KernelExtensionLoaded(string extensionType, string kernelName, bool isStaticContentSource ,KernelCommand command = null) : base(command)
+        public KernelExtensionLoaded(string extensionType,KernelCommand command = null) : base(command)
         {
             ExtensionType = extensionType;
-            KernelName = kernelName;
         }
-        public KernelExtensionLoaded(IKernelExtension kernelExtension, string kernelName, KernelCommand command = null) : base(command)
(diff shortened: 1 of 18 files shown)


src/Microsoft.DotNet.Interactive/Events/KernelExtensionLoaded.cs
```diff
@@ -0,0 +1,44 @@
...
+            ExtensionType = extensionType;
+            ContentSourceName = contentSourceName;
+            IsStaticContentSource = isStaticContentSource;
+        }
+        public KernelExtensionLoaded(IKernelExtension kernelExtension, KernelCommand command = null) : base(command)
+        {
+            KernelExtension = kernelExtension;
+            ExtensionType = kernelExtension.GetType().Name;
+
+            if (kernelExtension is IStaticContentSource contentSource)
```
jonsequitur, line 30: This seems like extra bookkeeping.

src/dotnet-interactive.Tests/FileProviderTests.cs
```diff
@@ -0,0 +1,120 @@
...
+using FluentAssertions;
+
+using Microsoft.DotNet.Interactive.App.Http;
+using Microsoft.DotNet.Interactive.Tests;
+using Microsoft.DotNet.Interactive.Tests.Utility;
+
+using Xunit;
+using Xunit.Abstractions;
+
+namespace Microsoft.DotNet.Interactive.App.Tests
```
jonsequitur, line 17: To the extent you're testing interfaces directly and not over HTTP, it looks like these belong in `Microsoft.DotNet.Interactive.Tests`.

src/Microsoft.DotNet.Interactive/Events/KernelExtensionLoaded.cs
colombod: removed

src/dotnet-interactive.Tests/FileProviderTests.cs
colombod: that could bring too many asp.net dependencies

Issue #642, opened 2020-07-14
F# quotations display badly in default formatting


dsyme:
1. Evaluate "<@ 1 @>" in a notebook.

Expect: Some reasonable of the quotation.  
Actual, a big mess, see image at bottom 

For reference FSI shows this, which is not great but not awful.  
```
val it : Quotations.Expr<int> =
  Value (1)
    {CustomAttributes = [NewTuple (Value ("DebugRange"),
          NewTuple (Value ("stdin"), Value (1), Value (3), Value (1), Value (4)))];
     Raw = ...;
     Type = System.Int32;}
```
Better would really be this:
```
val it : Quotations.Expr<int> = Value (1)
```

![image](https://user-images.githubusercontent.com/7204669/87437732-99b2c000-c5e6-11ea-92d9-643215ba43f1.png)

dsyme:
@brettfo @KevinRansom I'd like to iterate on the design here.

I know we have extensible type-direct formatting through `Formatter<_>.Register(f)` but we need to get the basic default printing of a whole swathe of F# data values to be decent without use customization.   

Here's a full list of values we need to consider, I'll update these as I think of more, and start to cross them off as I check current status

* [x] Constants (integers, floating point, strings, decimal)
* [ ] Constants from .NET libraries (bigint, DateTime, TimeSpan)
* [ ] Tuples
* [ ] Struct tuples
* [ ] Records
* [ ] Single case unions without data
* [ ] Single case unions with data
* [ ] Multi case unions all without data
* [ ] Multi case unions some with data
* [ ] Enums
* [ ] .NET Object values from core libraries (e.g. Type, MethodInfo, Assembly)
* [ ] Lists of all of the above
* [ ] Arrays of all of the above
* [ ] Sequences of all of the above
* [ ] Sequences that are slow to compute
* [ ] Nested cases of the above (where applicable)
* [ ] Quotations (raw and typed)
* [ ] Function values (unapplied and partially applied)
* [ ] Async values
* [ ] Task values
* [ ] Lazy values
* [ ] IDictionary values
* [ ] ICollection values
* [ ] Map values
* [ ] Set values
* [ ] `dict` values
* [ ] values returned by `List.groupBy`, `Seq.groupBy` and so on

Then a set of things from important libraries which will need customization 

* [ ] Data Frames
* [ ] Charting
* [ ] Tensors
* many other examples

dsyme:
[...]

PR #705, opened 2020-08-11
[WIP] Make '%A' structured formatting be the default for F# plaintext 

dsyme:
This extends over #689 to make '%A' formatting by the standard for plaintext rendering for F#

Addresses much of #642 

It's work in progress


src/Microsoft.DotNet.Interactive.FSharp.Tests/Microsoft.DotNet.Interactive.FSharp.Tests.fsproj
```diff
@@ -14,6 +14,8 @@
     <Compile Include="ApiTests.fs" />
   </ItemGroup>
 
+  <ItemGroup />
```
jonsequitur, line 17: Is this a VS artifact?

src/Microsoft.DotNet.Interactive.FSharp/FSharpKernelExtensions.fs
```diff
@@ -17,35 +17,76 @@ open XPlot.Plotly
...
-        let code = @"
-open System
-open System.Text
-open System.Threading.Tasks
-open System.Linq"
+        // F# has its own views on what namespaces are open by default in its scripting model
+        kernel
+
+    [<Extension>]
+    static member UseExtraNamespacesForTesting(kernel: FSharpKernelBase) =
```
jonsequitur, line 71: This looks like it should be in the test project.

dsyme:
@jonsequitur asked this

> I haven't noticed in your PRs any explicit separation of F# from C# plain text formatting. Is it something you've run into yet? The tricky piece here is that since they're in process together and currently formatters are not language-specific, which is by design. But I think we need to decide when and where it's appropriate to differ.

Yes, this is actively on my mind, and I'm trying to make as many improvements as possible which are language-independent. However F# box-layout plaintext formatting is a good mechanism we need to integrate, and there is no equivalent for C# as yet.

We could possible look at incorporating an adapted version of that mechanism into .NET Interactive for C# as well.  But that's further down the line.  

I think the order is

1. improve the common basis (what I'm doing now)

2. incorporate F#-specific box layout plaintext formatting (basically https://github.com/dotnet/interactive/pull/705)

3. make some improvements to  F#-specific box layout plaintext formatting in dotnet/fsharp

4. think about whether to duplicate/generalise the mechanism or something similar for C#

dsyme:
So if we need F# plaintext formatting to be different (e.g. "1L" not "1"), then I guess that means we're going to have to start scoping the formatting settings to a particular kernel.

Currently everything is, well, very global....

In general how do we feel about this kind of thing?  e.g. a polyglot programmer can today register a formatter in C# and have it implicitly picked up by F#?  It seems likely better to avoid building in "global leakage" across kernels simply because C# and F# happen to be implemented using the same DLLs...

caroletchepare (approved):

dsyme:
I'm going to close this out for now

When I looked at this I didn't appreciate that the defaults weren't specific to each language.  I may salvage parts of this when I next take notebooks for a long spin from the F# perspective

dsyme:
I'm going to close this out for now

When I looked at this I didn't appreciate that the defaults weren't specific to each language.  I may salvage parts of this when I next take notebooks for a long spin from the F# perspective

PR #707, opened 2020-08-11
Allow nested HTML in tables but not nested tables

dsyme:
This is improves HTML formatting after discussion with @jonsequitur 

1. Allow formatters to return `bool` indicating if they are interested in formatting a value or not.  

2. Compose the chain of applicable formatters in preference order looking for the first that succeeds

3. Add an IFormatContext parameter propagated throughout formatting.  This currently holds only a boolean indicating if we are emitting inside a table or not, to allow formatters to refuse to emit nested tables

4. Use this to allow the columns of tables to be formatted as HTML, but nested tables will not be produced.

This will allow, for example, tables with images in the columns, or whatever HTML displays we like, just not nested tables (at least our formatters refuse to produce nested tables - user ones can).

TODO:

* [x] get it green
* [x] trial it
* [x] add testing


src/Microsoft.DotNet.Interactive.Formatting/HtmlFormatter.cs
```diff
@@ -25,6 +25,35 @@ public static class HtmlFormatter
...
+        public static bool PreformatEmbeddedPlainText { get; set; } = false;
+        public static bool LeftJustifyEmbeddedPlainText { get; set; } = false;
+
+        static HtmlFormatter()
+        {
+            Formatter.Clearing += (obj, sender) => PreformatEmbeddedPlainText = false;
+            Formatter.Clearing += (obj, sender) => LeftJustifyEmbeddedPlainText = false;
+        }
+
+        internal static IHtmlContent DisplayEmbeddedObjectAsPlainText(IFormatContext context, object value)
```
jonsequitur, line 37: I'd suggest `FormatEmbeddedObjectAsPlainText`.

src/Microsoft.DotNet.Interactive.Formatting/IFormatContext.cs
```diff
@@ -0,0 +1,29 @@
+﻿using System.IO;
+
+namespace Microsoft.DotNet.Interactive.Formatting
+{
+    public interface IFormatContext
+    {
+        /// <summary>Represents the approximate proportion of output width remaining to this format operation.</summary>
+        bool IsNestedInTable { get; }
```
jonsequitur, line 8: This seems very HTML-specific. I suspect there are other scenarios where the nesting of formatters might be of interest, e.g. is this potentially a generalization of recursion rules for plain text formatting or a way of thinking about creating drill-down points for visualizations that need to call back to the kernel?

dsyme: Yes. It's basically impossible not to either over-architect or under-architect this.  There are an infinite number of scenarios of interest, and each of them have an infinite number of possible refinements.  

One approach is make the context an immutable string --> object property bag and leave it up to different formatters to use whatever keys they like.   Another is to strongly type is according to our specific needs and then iterate/refine.  Switching to the immutable property bag (with a couple of predefined keys) at least ensures there's space for whatever information any formatter wants to register.

dsyme:
There is a spec question here about 

1. formatting of nested HTML objects with no visible members
2. default formatting of plaintext objects

For (1), HTML Formatting currently formats such an object using `ToString()` [because of this code](https://github.com/dotnet/interactive/blob/master/src/Microsoft.DotNet.Interactive.Formatting/HtmlFormatter%7BT%7D.cs#L44).  However in nested position it reverts to plain text formatting which formats this object as `{ TypeName:  }` (note two spaces after the colon).

Questions

1. These should align. 

2. The plain text formatting looks wrong to me (because of the `:`, the two spaces after the `:` and the braces too)

However in general I don't understand why plain text doesn't call `ToString()` on the object?  e.g. look at this:

image

src/Microsoft.DotNet.Interactive.Formatting/IFormatContext.cs
jonsequitur: I think that even the simple step of allowing the context to understand the current recursion depth and list expansion index could work. My assumption is that `IsNestedInTable` is currently functionally equivalent to `RecursionDepth > 1`.

jonsequitur:
Objects with no visible members should probably just be formatted using `ToString` (or display something indicating they're `null`) in all cases. The prefix of `"TypeName"` is an artifact from the logging library origins of the plain text formatter and isn't useful here.

dsyme:
> Objects with no visible members should probably just be formatted using ToString (or display something indicating they're null) in all cases. The prefix of "TypeName" is an artifact from the logging library origins of the plain text formatter and isn't useful here.


Yes.  To be honest for C# plain text I think even ones with members should use `{ <ToString()>:  <extra-members-we-decide-to-display> }`. We just have to trust and report the results of `ToString()` really.  It exists for exactly this scenario.

I'd also suggest we have a `PlainTextFormatter.MaxProperties` setting, initially set to 4 or 10 or something, at least in the case where the plaintext formatting is being invoked invoked in a nested context, e.g. inside a table.  Or indeed just omit property printing in nested table position.

jonsequitur:
> I'd also suggest we have a PlainTextFormatter.MaxProperties setting, initially set to 4 or 10 or something, at least in the case where the plaintext formatting is being invoked invoked in a nested context, e.g. inside a table. Or indeed just omit property printing in nested table position.

I think this would be quite useful, as would keeping property printing inside tables. Ideally these could become expandable-on-demand in the "explore object graph" scenario.

src/Microsoft.DotNet.Interactive.Formatting/IFormatContext.cs
dsyme: Random other ideas 

- If we want to keep this strongly typed, non-extensible and simple, then a more declarative name might be along the lines `SimpleDisplayRequested` indicating that some outer formatting has requested that items use a simpler display. A simple flag like this is easy enough to understand and is enough to allow all sorts of restrictions to kick in (reduce number of properties being displayed by default, no nested tables etc.)

- A slight variation in this iso to make that a number `ContentThreshold` starting at 1.0 for "full display" and reducing down in inner table recursions.

I don't really mind - maybe we could start with this version and move to a property bag later if needed.  Making IFormatContext a sealed immutable type would allow us to do whatever we liked in the long run (e.g. add more properties or a proprety bag).

I think I'll change to ContentThreshold for now.

dsyme:
> I think this would be quite useful, as would keeping property printing inside tables. Ideally these could become expandable-on-demand in the "explore object graph" scenario.

Yes.

For now I'll change the IFormatContext to be a sealed type currently holding just a ContentThreshold and use this to drive down the properties being displayed in inner positions.

dsyme:
Running tests locally I get strange failures in 

      ....Connected_kernel_help_description_provides_context_about_the_connection

that don't occur when I run the test either in isolation nor in CI but only when I run the tests en-masse using Visual Studio or `dotnet test`.  It's a bit strange - it feels like there must be some interaction between the tests

dsyme:
@jonsequitur This is ready now - I will spend a while trialling it on some of the examples where large outputs were showing in the HTML tables.

src/Microsoft.DotNet.Interactive.Formatting.Tests/HtmlFormatterTests.cs
```diff
@@ -192,28 +184,42 @@ public void It_displays_exceptions_thrown_by_properties_in_the_property_value_ce
...
+
+                formatter.Format(typeof(Dummy.DummyNotInSystemNamespace), writer);
+
+                writer.ToString()
+                      .Should()
+                      .Be("Dummy.DummyNotInSystemNamespace");
+            }
+
+            [Fact]
+            public void Properies_of_System_Type_instances_are_not_expanded_and_link_is_added()
```
jonsequitur, line 205: I'd suggest including a "when ... " in the names of this test and the previous one in order to embed more specification into the tests. Why are there these two different outcomes?

dsyme: Thanks, made this clearer now

dsyme:
@jonsequitur As you can see from the tests added in https://github.com/dotnet/interactive/pull/707/commits/35cdb023e597cbb244db881e75ade199d51c0630, this PR contains the necessary settings to enable preformatting, but none of them are yet activated

All of these are initially false.

    HtmlFormatter.PlainTextPreformat = false;
    HtmlFormatter.PlainTextPreformatDefaultFont = false;
    HtmlFormatter.PlainTextPreformatNoLeftJustify = false;

There is also this new setting with default value 20

    PlainTextFormatter.MaxProperties = 20

I'll add a test for this last one now

dsyme:
@jonsequitur 

Added a sane ordering for HTML tables containing keys from various objects.  The property columns were previously in a random order drawn from the dictionary keys - they are now sorted as follows:

  - first by which type they come from (e.g. the properties of the first type encountered occupy the first columns) 

  - second by the declaration order of the properties (as given by the first destructurer that presented the properties)

dsyme:
Just to show this working in practice with multi-line data (here generated by `sprintf %A` )


image

dsyme:
Here is an example showing that HTML is now generated for columns:

image

dsyme:
Here is an example showing nested images in HTML tables:

image

dsyme:
(BTW There are lots of other groovy options to auto-generate interesting variations on images at https://picsum.photos/)

dsyme:
This is ready

src/Microsoft.DotNet.Interactive.Formatting.Tests/HtmlFormatterTests.cs
```diff
@@ -560,44 +811,55 @@ IEnumerable<object> GetCollection()
...
-      <td>99</td>
-    </tr>
-    <tr>
-      <td>2</td>
-      <td>System.String</td>
-      <td>Hello, World</td>
-    </tr>
-  </tbody>
-</table>");
+                      .BeEquivalentHtmlTo($@"<table>
```
dsyme, line 821: These cahnges are because column entries are now showing HTML

src/Microsoft.DotNet.Interactive.Formatting.Tests/PlainTextFormatterTests.cs
```diff
@@ -475,26 +509,11 @@ public void ReadOnlyMemory_of_char_is_formatted_like_a_string()
                 writer.ToString().Should().Be("Hi!");
             }
 
-            [Fact]
```
dsyme, line 478: These tests are removed because the returned formatter is now often a composition of possible formatters (using AnonymousTypeFormatter) and its not useful to check the runtime type of the formatter on return.

To be honest I think we should remove `HtmlFormatter<T>` and `PlainTextFormatter<T>`, ITypeFormatter<T> etc. as actual public implementation types of ITypeFormatter, just use them as part of the API, and have all formatters be either delegates or ITypeFormatter (non-generic). THis will greatly reduce the API surface area.

src/Microsoft.DotNet.Interactive.Formatting.Tests/PlainTextFormatterTests.cs
```diff
@@ -15,22 +15,6 @@ namespace Microsoft.DotNet.Interactive.Formatting.Tests
 {
     public class PlainTextFormatterTests : FormatterTestBase
     {
-        [Fact]
```
dsyme, line 18: These tests are removed because the returned formatter is now often a composition of possible formatters (using AnonymousTypeFormatter) and its not useful to check the runtime type of the formatter on return.

src/Microsoft.DotNet.Interactive.Formatting/DefaultHtmlFormatterSet.cs
```diff
@@ -52,72 +55,121 @@ internal class DefaultHtmlFormatterSet
...
+                new HtmlFormatter<Type>((context, type, writer) =>
                 {
-                    PocketView view = span(
-                        a[href: $"https://docs.microsoft.com/dotnet/api/{type.FullName}?view=netcore-3.0"](
-                            type.ToDisplayString(PlainTextFormatter.MimeType)));
+                    string text = type.ToDisplayString(PlainTextFormatter.MimeType);
+                    
+                    // This is approximate
+                    bool isKnownDocType =
+                      type.Namespace != null &&
```
dsyme, line 102: If anyone knows how to improve this approximation beyond this please chime in

src/Microsoft.DotNet.Interactive.Formatting/Formatter.cs
```diff
@@ -106,10 +107,14 @@ public static void ResetToDefault()
...
-
+            
+            // In the lists of default formatters, the highest priority ones come first,
+            // so register those last.
+            _defaultTypeFormatters.PushRange(HtmlFormatter.DefaultFormatters.Reverse().ToArray());
+            _defaultTypeFormatters.PushRange(JsonFormatter.DefaultFormatters.Reverse().ToArray());
+            _defaultTypeFormatters.PushRange(PlainTextFormatter.DefaultFormatters.Reverse().ToArray());
+
+            // It is unclear if we need this default:
             _defaultPreferredMimeTypes.Push((typeof(string), PlainTextFormatter.MimeType));
```
dsyme, line 118: I'm not sure we need this default mime type for `string` anymore. Maybe there's a reason why

dsyme, line 118: I'm not sure we need this default mime type for `string` anymore. Maybe there's a reason why

src/Microsoft.DotNet.Interactive.Formatting/DefaultHtmlFormatterSet.cs
jonsequitur: The docs.microsoft.com APIs are quite good, and there's also a NuGet index allowing search by type, so we can eventually improve the link generation quite a lot. I think the rough heuristic here is fine for now.

src/Microsoft.DotNet.Interactive.Formatting/HtmlFormatter.cs
```diff
@@ -25,6 +68,32 @@ public static class HtmlFormatter
...
+            Formatter.FormatTo(value, context, writer, PlainTextFormatter.MimeType);
+            var text = writer.ToString();
+            return FormatString(text);
+        }
+
+        internal static IHtmlContent FormatString(string text)
+        {
+            var html = text.HtmlEncode();
+
+            if (PlainTextPreformat)
```
jonsequitur, line 83: I initially misunderstood this name to mean "something we do before formatting". 

Looking more closely at this and the associated properties, I think it might be more flexible and easier to understand by removing them and doing something like:

```csharp
public Func<Tag> PlainTextWrapperTag> { get; set; } = () => new Tag("div");
```

This lets people control all of these style elements and add others.

dsyme: I do see the problem.

Another approach would be to emit consistent and documented CSS classes, e.g. emit any and all of `plaintext`, `dotnet-plaintex`, `fsharp-plaintext`, `csharp-plaintext`, ... plus tags for other outputs - e.g. `dotnet-object` etc.

Then have a helper function `CSS`:

    CSS ".fsharp-plaintext { text-align: left; white-space: pre; }"

which applies the formatting selectively.  Some advantages being it is general and requires almost no documentation (beyond documenting the tags emitted).   

So in this case we would always emit `<div class="dotnet-plaintext">...</div>`....  And perhaps also a general wrapper `<div class="dotnet-fsharp">...</div>` at the outer of all F# kernel outputs (likewise C#, powershell) to allow them to be formatted selectively.

src/Microsoft.DotNet.Interactive.Formatting/DefaultPlainTextFormatterSet.cs
```diff
@@ -79,39 +97,53 @@ internal class DefaultPlainTextFormatterSet
...
                 // Newtonsoft.Json types -- these implement IEnumerable and their default output is not useful, so use their default ToString
-                new PlainTextFormatter<Newtonsoft.Json.Linq.JArray>((obj, writer) => writer.Write(obj)),
+                new PlainTextFormatter<Newtonsoft.Json.Linq.JArray>((context, obj, writer) =>
+                    {
+                        writer.Write(obj);
+                        return true;
+                    }),
 
-                new PlainTextFormatter<Newtonsoft.Json.Linq.JObject>((obj, writer) => writer.Write(obj)),
+                new PlainTextFormatter<Newtonsoft.Json.Linq.JObject>((context, obj, writer) =>
```
jonsequitur, line 117: `JObject` and `JArray` both derive from `JToken`, so this and the previous formatter can be combined.

src/Microsoft.DotNet.Interactive.Formatting.Tests/PlainTextFormatterTests.cs
jonsequitur: I think this simplification makes sense. I wonder whether `HtmlFormatter<T>` and `PlainTextFormatter<T>` have useful, distinct APIs that people will want to use, but most likely if people just implement their own `ITypeFormatter` the capabilities become much more open-ended anyway.

src/Microsoft.DotNet.Interactive.Formatting/Formatter.cs
jonsequitur: If it's working well without it, let's remove it. It occurs to me that a `"text/html"` representation of `string` has some nice possibilities in terms of debug views, encoding visualizations, etc., which are unavailable if we hard-code the `"text/plain"` mapping.

src/Microsoft.DotNet.Interactive.Formatting/HtmlFormatter.cs
jonsequitur: I think ultimately we'll want to emit documented, stable CSS classes like this. I'd like to design it independent of this PR. In the meantime, I'm concerned that creating fine-grained abstractions for this in .NET (a) is necessarily limited and (b) will later clash with the more natural CSS-class approach.

jonsequitur:
>     {

The name of this class is has likely become misleading at this point.

---
Refers to: src/Microsoft.DotNet.Interactive.Formatting/SingleLinePlainTextFormatter.cs:9 in 7b8d5c1. [](commit_id = 7b8d5c184821ed342e90772d84ffc98dbf18ac32, deletion_comment = False)

src/Microsoft.DotNet.Interactive.Formatting/PocketViewTags.cs
```diff
@@ -88,10 +88,11 @@ public static class PocketViewTags
...
         public static dynamic video => _.video;
 
-        // This can be used to indicates points
-        // where we are embedding an arbitrary value into the HTML content
-        // and we are explicitly indicating that the value can be either IHtmlContent
-        // or will otherwise be formatted as plaintext.
-        public static object arbitrary(object value) => value;
+        /// <summary>Create an object suitable for delayed expansion to HTML</summary>
+        public static object embed(object obj, FormatContext context)
+        {
```
jonsequitur, line 93: This method probably belongs in a different class and with a Pascal-cased name, as it seems to sit outside the DSL.

dsyme: I think it's part of the DSL as things stand, because it's used in DSL expressions and creates the special EmbeddedContent object later processed by the DSL writer.

But yes, this thing is somewhat needlessly delayed in its execution and could be taken outside the DSL.

dsyme: From an education point of view we would teach it as part of the DSL - "to embed an arbitrary object with a context use `embed`"

That reminds me, I'll need to update the C# Formatter.Register samples in the notebooks to propagate the context

src/Microsoft.DotNet.Interactive.Formatting/DefaultPlainTextFormatterSet.cs
dsyme: got it thanks, will be in next update

dsyme:
> The name of this class is has likely become misleading at this point.

Yes it's a bit strange at the moment.  I'll leave it as is because its design intention is to format single line text, even if it can generate multi- line (and has always been able to)

dsyme:
> I think ultimately we'll want to emit documented, stable CSS classes like this. I'd like to design it independent of this PR. In the meantime, I'm concerned that creating fine-grained abstractions for this in .NET (a) is necessarily limited and (b) will later clash with the more natural CSS-class approach.

I went through and did the CSS approach because it's just simpler to do that then to try to think of proper names for all those .NET identifiers, and document their use.

It's simple enough (just an existing use of the Javascript helper) and I've added test for it - I followed https://stackoverflow.com/questions/524696/how-to-create-a-style-tag-with-javascript

dsyme:
I've been through the C# docs

BTW I notice that we're not emitting any warning diagnostics from either C# or F#, e.g. for obsolete APIs.

dsyme:
One of the new integration tests is failing, trying to work it out as its disabled loacally

dsyme:
@colombod This makes the integration test pass locally - note the difference between Powershell and C#/F# - any idea what's going on here?  And why this test is enabled by default?

```
        [IntegrationTheory]
        [InlineData(Language.CSharp, "<div class=\"dni-plaintext\">2</div>")]
        [InlineData(Language.FSharp, "<div class=\"dni-plaintext\">2</div>")]
        [InlineData(Language.PowerShell, "2")]
        public async Task dotnet_kernels_can_execute_code(Language language, string expected)
        {
            using var client = await CreateClient();

            var events = client.Events.ToSubscribedList();

            client.SubmitCommand(new SubmitCode("1+1", targetKernelName: language.LanguageName()));

            events
                .Should()
                .EventuallyContainSingle<DisplayEvent>(
                    where: d => d.FormattedValues.Any(fv => fv.Value.Trim() == expected),
                    timeout: 10_000);
        }
```

dsyme:
OK I see:

> The unit tests in this project are run as a special leg of CI that assumes the `dotnet interactive` global tool has been installed.

dsyme:
This is ready though please check this https://github.com/dotnet/interactive/pull/707#issuecomment-673703393

jonsequitur:
> are we expecting HTML output from Powershell?

Yes.

jonsequitur:
@TylerLeonhardt @daxian-dbw Please take a look at this PR. It represents a fairly thorough update to our formatting APIs and I'd like to make sure there are no surprises.

dsyme:
> > are we expecting HTML output from Powershell?

> Yes.

From the code it looks to me like powershell is going to output to stdout here.  e.g. C# and F# kernels use `DIsplayValue` and `FormattedValue` but these aren't use in Powershall kernel

I looked at master https://mybinder.org/v2/gh/dotnet/interactive/master?urlpath=lab and Powershell isn't returning HTML for things like `(1,2)`.  Instead I just get pre output:
```
<pre>1
2
</pre>
```

In short there's been no change to master in this PR, all good

jonsequitur:
Ah, right. PowerShell does emit `DisplayedValueProduced` events (when using the `Out-Display` cmdlet) but not on eval, so this is fine.

Issue #691, opened 2020-08-07
missing diagnostics for some F#, C# code


dsyme:
Put this into a notebook and execute:

    open Microsof  t.DotN et

Expected: an error 
Actual: no error is shown 

Tested with https://mybinder.org/v2/gh/dotnet/interactive/master?urlpath=lab

dsyme:
There are other examples where errors aren't shown, e.g. evaluate

```
1+
```

or

```
1+!
```

dsyme:
Also noticed missing warnings diagnostics from C# code, e.g. System.Obsolete messages

dsyme:
@brettfo The problem here is thar FsiScript.Eval is succeeding with `Ok` even when there are error diagnostics from parsing and type checking:
```
            let result, diagnostics =
                try
                    script.Value.Eval(codeSubmission.Code, tokenSource.Token)
                with
                | ex -> Error(ex), [||]
```

I'll fix this up in .NET Interacive to double check for an Error severity diagnostic.  However I think we should also fix this in the scripting DLL shim in dotnet/fsharp.

dsyme:
A repro for missing errors in F#

```
1+!
```

A repro for missing warnings in both F# and C# (using an Obsolete API)

```
System.AppDomain.GetCurrentThreadId();
```

dsyme:
BTW the difference between these is very subtle in C# :-)
```
System.AppDomain.GetCurrentThreadId()
```
and
```
System.AppDomain.GetCurrentThreadId();
```
The first reveals an integer answer, the second the result is implicitly thrown away because of C#'s crazy "ignore this expression" rule.

dsyme:
Fixed by #713

PR #713, opened 2020-08-14
Fix missing silent diagnostics

dsyme:
This fixes issue #691 for F# and C#

* [x] F# errors noe always being reported 

* [x] F# warnings are not always being shown

* [x] C# warnings are now always being shown

Basically two problems

* Some F# evaluations were failing (error diangostics produced) but this wasn't being noticed

* The C# and F# Kernels were pushing diagnostics through DiagnosticsProduced (which is for logging) but they should also be producing StandardErrorValueProduced (which is what Jupyter and others want for diagnostics).

I've checked things in Jupyter notebooks and the results are correct.


dsyme:
@jonsequitur This is an important fix for usability, as without it we can't tell when F# code is failing to parse/type-check correctly.

jonsequitur:
I don't think publishing diagnostics as `StandardErrorProduced` events is the right approach here.  Semantically, they're not standard error, and they are showing in VS Code. 

Since Jupyter doesn't have a diagnostic concept, it sounds like within the `Microsoft.DotNet.Interactive.Jupyter` layer, we should consume `DiagnosticsProduced` events and publish them into one of Jupyter's channels.

dsyme:
> Since Jupyter doesn't have a diagnostic concept, it sounds like within the Microsoft.DotNet.Interactive.Jupyter layer, we should consume DiagnosticsProduced events and publish them into one of Jupyter's channels.

Yes, ok, I can change it to do this

dsyme:
@jonsequitur I adjusted the fix to this, I believe more along the lines you wanted.

The DiagnosticsProduced event is now a DisplayEvent - it seems to fit well as the formatted errors are somewhat subtly different and belong well in `FormattedValues`.  This will also allow us to format the diagnotics as HTML sometime in the future.

I adjusted the Powershell kernel to populate this as well for the one place it produced DiagnosticsProduced.

src/Microsoft.DotNet.Interactive.CSharp/CSharpKernel.cs
```diff
@@ -221,27 +221,35 @@ public async Task HandleAsync(SubmitCode submitCode, KernelInvocationContext con
 
             if (!context.CancellationToken.IsCancellationRequested)
             {
+                var diagnostics = ImmutableArray<CodeAnalysis.Diagnostic>.Empty;
```
dsyme, line 224: This grabs the diagnostics (whether they be errors or warnings) and reports them.

The formatting of error text is the same as before, however that text always goes to `DiagnosticsProduced` instead of being attached to the exceptions in `context.Fail(...)`

dsyme:
@jonsequitur 

The revised fix makes `DiagnosticsProduced` carry extra information for the formatted errors.  This is necessary because the formatted diagnostics we get from Roslyn/FCS vary somewhat from the simple ToString on the "Diagnostics" array in that event, e.g. they contain the error numbers and so on.  So we need to format the errors.  Previously we did that be slamming text into the failure message.   That's no good for warnings.

However adding this information changes the JSON contract for DiagnosticsProduced because extra information is now present. I think this is the right change because it's OK for extra values to be present in the JSON, even if it is ignored by the consumer (VSCode in this case)?

I've adjusted the JSON contract - thanks for the helpful messages about how to do that!

jonsequitur:
> The DiagnosticsProduced event is now a DisplayEvent 

I think this makes sense semantically. One thing worth discussing is a more explicit separation between display events that get stored in the `.ipynb` versus those that don't. Most language service events shouldn't, and while that mapping can live in the Jupter- and VS Code-specific layers, it does feel like an intrinsic difference that, though not very clearly modeled, has been erased by this change. 

To sum up, prior to this change:

* `DisplayEvent` types (which do get saved in the `.ipynb` file):
  * `StandardErrorValueProduced`
  * `StandardOutputValueProduced`
  * `DisplayedValueProduced`
  * `ReturnValueProduced`
* Language services events (which don't get saved in the `.ipynb` file):
  * `HoverTextProduced`
  * `CompletionsProduced`
  * `DiagnosticsProduced`

Also, could you add a screen shot of the new output in Jupyter?

brettfo:
Can you post screenshots of what this experience looks like in both Jupyter and VS Code?  Otherwise I like it.

dsyme:
> Can you post screenshots of what this experience looks like in both Jupyter and VS Code? Otherwise I like it.

#### C# warning in Jupyter

image

#### C# compilation error in Jupyter

image


#### C# Exception in VS Code

image


### C# Warning in VS Code

image

Note, hover tip is correct

### C# compilation error in VS Code

image

The F# ones are all the same

Note that in VS Code the compilation diagnostics show as red/yellow-squigglies and "Problems" but not in the status line after the attempted execution of the cell, which just shows `Compilation error`.  This could perhaps say "Compilation error - see diagnostics in 'Problems' tab" but that would need a fix in VS Code I think, it might be better if one of you made that.  I suppose the fix would be to add the extra message text here:
```
export function toVsCodeDiagnostic(diagnostic: Diagnostic): vscode.Diagnostic {
    return {
        range: convertToRange(diagnostic.linePositionSpan)!,
        message: diagnostic.message,
        severity: toDiagnosticSeverity(diagnostic.severity)
    };
}
```

dsyme:
OK I added that text on the status line

image

dsyme:
From chat with @jonsequitur 

@dsyme says:

> One query - the current PR suppresses DiagnosticsReported event when there are no diagnostics (Count = 0).  Is that right, or should we still be reporting that event ?

@jonsequitur says:

> It's probably fine to not emit the event at that point. Question would be does the UI rely on an empty event to clear the prior diagnostics.

@dsyme says

> And if not delivering a zero would force the IDE to flicker the errors (remove them all when a request is submitted)

dsyme:
Besides [this issue](https://github.com/dotnet/interactive/pull/713#issuecomment-674553700) this PR is ready

dsyme:
@jonsequitur @brettfo  I fixed this, so we now send empty diagnostics as before

> One query - the current PR suppresses DiagnosticsReported event when there are no diagnostics (Count = 0). Is that right, or should we still be reporting that event ?

jonsequitur:
> Note that in VS Code the compilation diagnostics show as red/yellow-squigglies and "Problems" but not in the status line after the attempted execution of the cell, which just shows Compilation error.

The current behavior shows the details of the compiler error in the output and I think we should likely keep that, since it prevents cell execution.

Similarly, I think the warning output in Jupyter is a bit alarming, because warnings should not prevent cell execution. We might consider using `Display` and some custom HTML styling (maybe with a ⚠️ ) rather than the Jupyter error channel.

dsyme:
@jonsequitur OK I fixed the tests and adjusted the error text on the cell status error back in VS code 

image

dsyme:
@jonsequitur this is ready I believe

jonsequitur:
I think this bears further discussion: https://github.com/dotnet/interactive/pull/713#issuecomment-674297962

If `DiagnosticsProduced` is going to be a `DisplayEvent`, which not e.g. `HoverTextProduced`?

Related to that, did you verify that these don't get saved into the `.ipynb` file?

dsyme:
> If DiagnosticsProduced is going to be a DisplayEvent, which not e.g. HoverTextProduced?

To me such hierarchical classifications are meaningless and I don't particularly see their role.  What's the point of them?  My position is that imposing and maintaining hierarchical classifications is 98% a waste of time for the software industry :-)

Anyway I adjusted it to use DisplayEvent because, well, its Jupyter processing most resembles other DisplayEvents.  But we already have other cases where display events do and don't carry data and do/don't vary slightly.

I can take that away with a little extra code duplication, I don't think it matter, we should just get rid of DisplayEvent and friends altogether.

> Related to that, did you verify that these don't get saved into the .ipynb file?

The warning diagnostics get saved in the ipynb as stream output, which I believe is what we'd expect:
```
    {
     "name": "stderr",
     "output_type": "stream",
     "text": [
      "\n",
      "input.fsx (1,1)-(1,36) typecheck warning This construct is deprecated. AppDomain.GetCurrentThreadId has been deprecated because it does not provide a stable Id when managed threads are running on fibers (aka lightweight threads). To get a stable identifier for a managed thread, use the ManagedThreadId property on Thread.  https://go.microsoft.com/fwlink/?linkid=14202\n",
      "\n"
     ]
    },
```

dsyme:
> Anyway I adjusted it to use DisplayEvent because, well, its Jupyter processing most resembles other DisplayEvents. But we already have other cases where display events do and don't carry data and do/don't vary slightly.

(So please just tell me which way you'd like it - subclass of DisplayEvent or not - I don't think it matters too much either way)

jonsequitur:
> The warning diagnostics get saved in the ipynb as stream output, which I believe is what we'd expect:

I think we would expect this when diagnostics arise from a code submission but not when they arise transiently while just typing.

While I agree that the `DisplayEvent` supertype is a bit of extra OOP goop, it currently represents a pragmatic behavioral distinction between what I think most people will generally regard as the "program output" (including console output, visualizations, cell return values), and "tooling" (editor features and language services). The host does ultimately own the handling of this distinction so removing `DisplayEvent` altogether would probably be the cleanest approach, but a good deal of host code for both Jupyter and VS Code will need to be reverified.

 So in sum I would say let's not inherit `DisplayEvent` for now, with the plan to remove it in a future PR.

dsyme:
> So in sum I would say let's not inherit DisplayEvent for now, with the plan to remove it in a future PR.

Okey dokey I'll change that

> > The warning diagnostics get saved in the ipynb as stream output, which I believe is what we'd expect:
> 
> I think we would expect this when diagnostics arise from a code submission but not when they arise transiently while just typing.

In Jupyter the waning diagnostics do only appear when the code is submitted for execution.  

As an imperfect comparison point, when Python execution emits (dynamic) warnings then they are in the output streams, so:

image

gives:

```
   {
     "name": "stderr",
     "output_type": "stream",
     "text": [
      "c:\\users\\dsyme\\appdata\\local\\programs\\python\\python37-32\\lib\\site-packages\\ipykernel_launcher.py:4: UserWarning: This is a warning message\n",
      "  after removing the cwd from sys.path.\n"
     ]
    }
```

I mean, it's a dynamic warning not a compilation warning, but I think it's ok to regard compilation warnings in this way all in all

dsyme:
>  brettfo changed the base branch from master to main 1 hour ago

👍

dsyme:
@jonsequitur I updated the base class, please check the latest changes (and we need to check they go green)

dsyme:
> it currently represents a pragmatic behavioral distinction between what I think most people will generally regard as the "program output" (including console output, visualizations, cell return values), and "tooling" (editor features and language services).

Right, diagnostics fall in this gap (outputs on stderr for Jupyter and command-line tools, outputs to tooling windows in IDE)

Anyway, updates are done, I'll be on vacation tomorrow so hopefully you can approve this tonight once it is green?

dsyme:
@jonsequitur Right, all greeen and ready

dsyme:
Yay, thanks!

jonsequitur:
Thank _you_, good sir.

PR #714, opened 2020-08-15
F# HoverText

krauthaufen:
Hey interactive people, I added support for `RequestHoverText` in `FSharpKernel` via adding (a very reduced variant) of [FsAutoComplete](https://github.com/fsharp/FsAutoComplete).

image

I'm aware that this is not really pretty, since changes in `FsAutoComplete` won't be easy to integrate (some files are modified slightly) but nonetheless hover-texts are really useful (escpecially for F#).

Maybe you have a better idea how to integrate FsAutoComplete's hover-text.
Cheers


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/interactive?pullRequest=714) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/interactive?pullRequest=714) before we can accept your contribution.<br/><br/>:x: krauthaufen [sign now](https://cla.dotnetfoundation.org/dotnet/interactive?pullRequest=714)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/interactive?pullRequest=714) it.</sub>

brettfo:
This looks wonderful!  Can you add the appropriate tests to the `src/Microsoft.DotNet.Interactive.Tests/LanguageServices/HoverTextTests.cs` file?

Long term we'd likely want this code migrated to the `dotnet/fsharp` repo, but for now I think we're ok with it existing here, especially if we have tests that we can use to confirm if/when the code is moved.

src/FsAutoComplete/Debug.fs
```diff
@@ -0,0 +1,39 @@
+// --------------------------------------------------------------------------------------
```
jonsequitur, line 1: It looks like this file can be removed.

jonsequitur:
An interesting enhancement here to traditional hover text is that we can also inspect the current value of the variable using `FSharpKernel.TryGetVariable`. Might be worth including?

krauthaufen:
I'll add some basic tests and remove more unnecessary code. @jonsequitur awesome idea, I will definetly look into that...

krauthaufen:
Hey, I just added a few tests and improved the layouting a little.

Btw. I also included the current-value when available:
image

jonsequitur:
That is very cool!

dsyme:
This is great work and I think we will be able to take this after an iteration or too

> I'm aware that this is not really pretty, since changes in FsAutoComplete won't be easy to integrate (some files are modified slightly) but nonetheless hover-texts are really useful (escpecially for F#).

We do need to be consolidating here.

Referencing https://github.com/fsharp/FsAutoComplete/issues/645 which covers the ground

dsyme:
@krauthaufen  Did you consider just callling the FCS function `GetStructuredToolTipText` directly?  (We still need to get a correctly configured FSharpChecker and so forth)

src/Microsoft.DotNet.Interactive.FSharp/FSharpKernel.fs
```diff
@@ -138,6 +140,122 @@ type FSharpKernelBase () as this =
...
+                                    sprintf "```fsharp\n%s\n```" code
+
+                                if not (String.IsNullOrWhiteSpace comment) then
+                                    comment
+
+                                if not (String.IsNullOrWhiteSpace newFooter) then
+                                    newFooter
+                            ]
+
+                        FormattedValue("text/markdown", stdinRx.Replace(fsiModuleRx.Replace(markdown, ""), ""))
```
dsyme, line 240: Why are we producing text/markdown?  Does HoverTextProduced require it?

dsyme, line 240: Why are we producing text/markdown?  Does HoverTextProduced require it?

src/Microsoft.DotNet.Interactive.FSharp/FSharpKernel.fs
```diff
@@ -138,6 +140,122 @@ type FSharpKernelBase () as this =
...
+                            let expr = if name.StartsWith "Stdin." then name.Substring 6 else name
+                            try return script.Value.Fsi.EvalExpression(expr) |> Some
+                            with _ -> return None
+                        | None -> return None
+
+                    | Error _ ->
+                        return None
+                }
+            
+            match! res.TryGetToolTipEnhanced (FSharp.Compiler.Range.mkPos line col) lineContent with
```
dsyme, line 184: Yes I'm wondering why we need FsAutoComplete tooltips here, why not the tooltips provided by FCS.  That might reduce the amount of code copied across (and leave us to incorporate the FsAutoComplete tooling into FCS).

src/Microsoft.DotNet.Interactive.FSharp/FSharpKernel.fs
```diff
@@ -138,6 +140,122 @@ type FSharpKernelBase () as this =
...
+                                    | Some _ -> Some sym.FullName
+                                    | None -> None
+                                | _ -> None
+                            | _ ->
+                                None
+
+                        match fullName with
+                        | Some name ->
+                            let expr = if name.StartsWith "Stdin." then name.Substring 6 else name
+                            try return script.Value.Fsi.EvalExpression(expr) |> Some
```
dsyme, line 176: I'd like to understand more about how "safe" this evaluation is (i.e. not have side-effects)

dsyme:
Note to self: the static checking in this PR (from which the tooltips are driven)  appears to be based on  the current type-check state of FSI, plus the checking of the cell contents of the cell being examined (but not of prior cell contents).   This differs from iFSharp etc. where the static checking concatenates all cell contents and then splits them out again.

This means executing cells (especially definitions and package loads) will light up more accurate tooling in subsequent cells

jonsequitur:
> the static checking in this PR (from which the tooltips are driven) appears to be based on the current type-check state of FSI, plus the checking of the cell contents of the cell being examined (but not of prior cell contents)

We've considered the possibility of doing a compile for language services evaluation when the notebook starts up. This would make the assumption that the cells should be evaluated in order, which I think is intuitive. It would present some challenges for polyglot scenarios where a submission in one languages influences the state of another, e.g. variable sharing.

src/Microsoft.DotNet.Interactive.FSharp/FSharpKernel.fs
krauthaufen: It's basically what ionide does and the tooltip-text contains markdown for e.g. generic arguments. Furthermore the F# syntax highlighting works with markdown code-blocks...

src/Microsoft.DotNet.Interactive.FSharp/FSharpKernel.fs
krauthaufen: Since it just runs for module-bindings I figured they were already evaluated in the FSI-semantics?

src/Microsoft.DotNet.Interactive.FSharp/FSharpKernel.fs
krauthaufen: I first tried to use the plain tooltips from FCS but they don't incorporate types/modules etc. (at least when used with overlapping names).
When hovering the type `int` i got the tooltip for the `int : ^T -> int` function, so i figured FsAutoComplete could be a way to go.

src/Microsoft.DotNet.Interactive.FSharp/FSharpKernel.fs
dsyme: > Furthermore the F# syntax highlighting works with markdown code-blocks...

Right, though that's a non-standard interpretation of F# code comments only really supported by F# tooling based off the ionide toolchain.

We're talking about standardizing it but it's a long way off, I don't think we should adopt it in F# notebooks

src/Microsoft.DotNet.Interactive.FSharp/FSharpKernel.fs
dsyme: > I first tried to use the plain tooltips from FCS but they don't incorporate types/modules etc. (at least when used with overlapping names).

You're right that the FsAutoComplete tooltips make improvements (though it's not yet clear to me if they also regress other things).

But in any case, for .NET Interactive we would take the approach that we build on FSharp.Compiler.Service and just have to accept the tooling possibilities that gives.  If we need to improve those tooltips, let's do it in FCS (e.g. by incorporating and aligning the tooltips code with Ionide)

brettfo:
@krauthaufen Can you update each test in the file `src/Microsoft.DotNet.Interactive.Tests/LanguageServices/HoverTextTests.cs` by adding the appropriate `[InlineData(Language.FSharp...)]` attributes and removing the `Skip = "not implemented in fsharp"` parameter?  There are some oddities in how the hover text request gets forwarded to the underlying kernel that these tests can enforce.

Once that file's updated (and a merge or rebase to fix conflicts) I'm ready to pull it.

krauthaufen:
Hey, sorry for the delay, I addressed all your requests however I have no idea how useful these tests are since they're very strict in the Hover-Texts. They all succeed atm. but minor format changes would immediately break them.

brettfo:
Thanks for adding those F# tests.  Yes, they're very strict, but they test that the F# kernel is setup to receive those commands, which is subtly different than directly testing the kernel itself.

The CI failure appears to be a refactoring that's causing problems with your merge from `main`.

brettfo (approved):

brettfo:
@krauthaufen Thank you again for porting over F# hover text.  I was eager to use it so I've published a new version of the VS Code extension.  Version 1.0.146603 has F# hover text and it's beautiful.

krauthaufen:
Cool, thanks

PR #723, opened 2020-08-24
tabular data formatters

colombod:


src/Microsoft.DotNet.Interactive.Formatting/DefaultTabularDataFormatterSet.cs
```diff
@@ -0,0 +1,33 @@
...
+{
+    internal static class DefaultTabularDataFormatterSet
+    {
+        internal static readonly ITypeFormatter[] DefaultFormatters =
+            new ITypeFormatter[]
+            {
+                new TabularDataFormatter<IEnumerable>((context, source, writer) =>
+                {
+
+                    if (source.GetType() != typeof(string))
```
jonsequitur, line 16:
```suggestion
                    if (!(source is string))
```

src/Microsoft.DotNet.Interactive.Formatting/TabularDataFieldList.cs
```diff
@@ -0,0 +1,11 @@
+﻿// Copyright(c).NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Collections.Generic;
+
+namespace Microsoft.DotNet.Interactive.Formatting
+{
+    public class TabularDataFieldList : List<TabularDataSchemaField>
```
jonsequitur, line 8: These types should not be public.

colombod: i made them public to be used in the data frame extensions, so we can reuse such schema objects

src/Microsoft.DotNet.Interactive.Formatting/TabularDataFormatter.cs
```diff
@@ -0,0 +1,155 @@
...
+{
+    public static class TabularDataFormatter
+    {
+        // https://specs.frictionlessdata.io/table-schema/#language
+        public const string MimeType = "application/table-schema+json";
+
+        internal static ITypeFormatter[] DefaultFormatters { get; } = DefaultTabularDataFormatterSet.DefaultFormatters;
+
+
+        public static TabularDataSet ToTabularDataSet(this IEnumerable source)
```
jonsequitur, line 21: This method and its return type should probably be internal.

The type of `source` should probably be `IEnumerable<T>` .This doesn't exactly enforce homogeneous contents (which are necessary for valid tabular formatting) but it helps.

src/Microsoft.DotNet.Interactive.Formatting/TabularDataSchema.cs
```diff
@@ -0,0 +1,17 @@
+﻿// Copyright(c).NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Collections.Generic;
+using Newtonsoft.Json;
+
+namespace Microsoft.DotNet.Interactive.Formatting
+{
+    public class TabularDataSchema
```
jonsequitur, line 9: This type should be internal.

src/Microsoft.DotNet.Interactive.Formatting/TabularDataSchemaField.cs
```diff
@@ -0,0 +1,22 @@
+﻿// Copyright(c).NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Newtonsoft.Json;
+
+namespace Microsoft.DotNet.Interactive.Formatting
+{
+    public class TabularDataSchemaField
```
jonsequitur, line 8: This type should be internal.

src/Microsoft.DotNet.Interactive.Formatting/TabularDataSet.cs
```diff
@@ -0,0 +1,35 @@
+﻿// Copyright(c).NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Collections;
+using Newtonsoft.Json.Linq;
+
+namespace Microsoft.DotNet.Interactive.Formatting
+{
+    public class TabularDataSet
```
jonsequitur, line 9: This type should be internal.

src/Microsoft.DotNet.Interactive.Formatting/TabularDataFieldList.cs
jonsequitur: I'd have to see the use case but I suspect we should not support this here.

src/Microsoft.DotNet.Interactive.Formatting/TabularDataSchemaField.cs
colombod: done for the moment

src/Microsoft.DotNet.Interactive.Formatting/TabularDataSet.cs
colombod: done for the moment

src/Microsoft.DotNet.Interactive.Formatting/TabularDataFieldList.cs
colombod: making those internal until we need them

PR #732, opened 2020-08-27
prepare cell output and metadata updates

brettfo:
According to microsoft/vscode#105283, notebook cell update apis are getting an ... update.

The existing way of setting the cell's metadata no longer works, but the new apis aren't yet fleshed out.  This PR does the appropriate metadata/output refactor so that when the real changes are in we won't have to do any real work to switch over.

There is also a slight change in that we now set the entire cell metadata instead of just the individual fields.  This is to fix a recent issue where a cell's metadata might be `null`.

### DIFF
--- a/src/dotnet-interactive-vscode/src/vscode/notebookProvider.ts
+++ b/src/dotnet-interactive-vscode/src/vscode/notebookProvider.ts
@@ -88,15 +88,15 @@ export class DotNetInteractiveNotebookContentProvider implements vscode.Notebook
 
     async executeCell(document: vscode.NotebookDocument, cell: vscode.NotebookCell): Promise<void> {
         const startTime = Date.now();
-        cell.metadata.runStartTime = startTime;
-        cell.metadata.runState = vscode.NotebookCellRunState.Running;
-        cell.outputs = [];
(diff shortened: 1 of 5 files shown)


colombod (approved):

src/dotnet-interactive-vscode/src/vscode/notebookProvider.ts
```diff
@@ -88,30 +88,55 @@ export class DotNetInteractiveNotebookContentProvider implements vscode.Notebook
...
+            });
         });
     }
 
+    static updateCellMetadata(document: vscode.NotebookDocument, cell: vscode.NotebookCell, metadata: vscode.NotebookCellMetadata) {
+        cell.metadata = metadata;
+        // const index = document.cells.findIndex(c => c === cell);
+        // if (index >= 0) {
+        //     vscode.notebook.activeNotebookEditor?.edit(editBuilder => {
+        //         editBuilder.replaceMetadata(index, metadata);
```
jrieken, line 124: Instead of relying on the active editor you can/should use the workspace edit, e.g 

```ts
const edit = new vscode.WorkspaceEdit();
edit.replaceCellMetadata(document.uri, cellIndex, metadata)
await vscode.workspace.applyEdit(edit)
```

PR #734, opened 2020-08-28
move notebook parsing and serialization into the kernel

brettfo:
This is a big one.

All notebook parsing and serialization has moved to the kernel.  This will give us a more accurate parse since the kernel knows the names of all available language kernels; the previous method was a hard-coded list in the VS Code extension.  A minor result was that the name of the PowerShell kernel is `pwsh`, not `powershell` that was previously assumed.  I updated the existing `.dib` sample notebooks with this so they round-trip without any changes.

VS code is the only component that touches the file system during read/write.  This is to make any future scenarios where the kernel might be remote much easier to handle.  Because of this and because we may want to support binary formats in the future, VS Code takes a `Uint8Array` that it read from disk and `base64` encodes it to send to the kernel where it's then parsed.  Similarly, serialized files are sent back as a `base64` encoded string and VS Code simply spits the bytes to disk.  There are explicit tests on each end that verify that a byte array is always `base64` encoded.  In the kernel, this is the Newtonsoft.Json default, but in VS Code I had to create custom `parse` and `stringify` functions that try their best to determine what to do.  When parsing, if the name of the field is `rawData` and the value is a `string` it's assumed to be a `base64` encoded byte array.  Similarly when serializing, if the name of the field is `rawData` and the value looks like either an array (has a `.length` property) or looks like a `Buffer` (has a `.data` property) then the same assumption is made.

I've added pretty extensive tests in the kernel for both parsing and serialization, particularly around `.ipynb` files since that file format is pretty loose; e.g., a cell's source can either be a single string or an array of string.  Both of those possibilities I handle and test for and I (arbitrarily?) chose the string array as the format to use when serializing.  This can be easily changed.

All parsing and serialization is based on the file extension; `.ipynb` will obviously write a Jupyter Notebook file, `.dib`/`.dotnet-interactive` will write our simple flat-file format, and everything else is an error.

I've also manually tested opening `.dib` and `.ipynb` files in VS Code and saving them as the opposite format and verifying the contents.  When saving as a `.ipynb` I then opened that file in Jupyter Lab and confirmed it had the same outputs as before and could also execute.


colombod (approved):

samples/vs-code/covid-19.dib
```diff
@@ -10,7 +10,7 @@ Also, due to travel restrictions, you should run this at home on isolated comput
 
 Since Johns Hopkins has put COVID-19 time series data on [GitHub](https://github.com/CSSEGISandData/COVID-19), let's take a look at it. We can download it using PowerShell:
 
-#!powershell
```
jonsequitur, line 13: Either should be valid. Was there a reason this needed to change?

brettfo: Correct that both are valid.  I was testing round-tripping some notebooks to make sure I didn't break anything and I figured our sample .dib notebook should round-trip with no edit.

PR #735, opened 2020-08-28
Adopt edit submission api

colombod:
The edit submission api is able to perform in scenarios where the notebook doesn't exist locally (like in a live share session) and routes the edits correctly.

As @jrieken helped me understand earlier:
> The notebook editor and the workspace edit API. The former is OK to use when running commands or other UI-like opertion, but when executing a cell we don't guarantee that a notebook editor exists (e.g your notebook document might be driven via a liveshare session) and then it's better to use the edit-api

### DIFF
--- a/src/dotnet-interactive-vscode/package.json
+++ b/src/dotnet-interactive-vscode/package.json
@@ -98,7 +98,7 @@
         },
         "dotnet-interactive.minimumInteractiveToolVersion": {
           "type": "string",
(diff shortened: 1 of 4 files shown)


brettfo:
Also update [this](https://github.com/dotnet/interactive/blob/9dc164ac7c4c0f94e31fa06b45c0323764671d42/src/dotnet-interactive-vscode/src/vscode/commands.ts#L96-L103).

PR #739, opened 2020-09-01
add command for new blank notebook

brettfo:
The new command's display name is ".NET Interactive: Create new blank notebook" for two reasons:

1. All of our other command names start with ".NET Interactive:" to remove any ambiguity and I thought that should continue.
2. The Python command to create a new Jupyter notebook is ["Create New Blank Jupyter Notebook"](https://github.com/microsoft/vscode-python/blob/de1a7bcd3d1eb337db27f60e5953cc5d325585a3/package.nls.json#L114).

I've added a keybinding of `Ctrl+Alt+Shift+N` to invoke this.  It's extremely verbose, but I don't want to steal any existing keybinding, and unfortunately that means we can't use the chord `Ctrl+N, D` because that causes the default `Ctrl+N` to be unbound because it's waiting for a chord to continue, and that's bad.  An alternative is to not have a default keybinding and let users add their own that simply invokes the command `dotnet-interactive.newNotebook`.  I'm open to either approach.

As for changes to code, there were really only 3 of note:

1. To create a new untitled notebook, I avoided calling `workbench.action.files.newUntitledFile { viewType: 'dotnet-interactive' }` because that creates a new notebook with the name `Untitled-1` (note, no file extension) and when the user hits `Ctrl+S`, the save dialog defaults to the `.txt` extension, and that can cause a whole bunch of problems because the file format is chosen depending on the extension, and if the user chooses `.txt` (or `.my-notebook`, etc.) we have no idea how to serialize the notebook.  Instead, I manually create a URI with the `.dib` file extension pre-applied.  This means when the user finally saves their notebook, it'll have the correct extension.  It's still possible for them to manually override this, but that's on them; if they're mangling the file extensions we can't help them.  To enable multiple unsaved notebooks, I look through the existing notebooks until I can find a numeric suffix that's unused, e.g., `Untitled-9.dib`.

2. Add the method `ClientMapper.reassociateClient`.  This is necessary because the notebook's runtime is associated with the notebook's URI.  In normal scenarios, this will be something like `C:\Users\brettfo\Desktop\my-notebook.dib`.  In the case of a new unsaved notebook, the URI is `Untitled-1.dib`.  Initially this is fine, but as soon as the user saves the notebook, the URI will suddenly be a file path like above and since we don't want them to loose their current runtime values and leave around a stale kernel, the reassociate method does exactly that: reassociates the kernel process with the new URI.

3. VS Code will call `openNotebook` when creating a new notebook viewer, but if it's an unsaved notebook created through the new command, nothing exists on disk.  The fix for this was simple; if it's a new unsaved file, short-circuit everything and create an empty notebook.

### DIFF
--- a/src/dotnet-interactive-vscode/src/clientMapper.ts
+++ b/src/dotnet-interactive-vscode/src/clientMapper.ts
@@ -27,6 +27,24 @@ export class ClientMapper {
         return client;
     }
 
+    reassociateClient(oldUri: Uri, newUri: Uri) {
+        const oldKey = ClientMapper.keyFromUri(oldUri);
+        const newKey = ClientMapper.keyFromUri(newUri);
+        if (oldKey === newKey) {
+            // no change
+            return;
+        }
+
+        const client = this.clientMap.get(oldKey);
+        if (!client) {
+            // no old client found, nothing to do
+            return;
+        }
+
+        this.clientMap.set(newKey, client);
+        this.clientMap.delete(oldKey);
+    }
+
     closeClient(uri: Uri) {
         let key = ClientMapper.keyFromUri(uri);
         let client = this.clientMap.get(key);
--- a/src/dotnet-interactive-vscode/src/vscode/commands.ts
+++ b/src/dotnet-interactive-vscode/src/vscode/commands.ts
@@ -3,10 +3,11 @@
 
 import * as vscode from 'vscode';
 import * as cp from 'child_process';
+import * as path from 'path';
 import { acquireDotnetInteractive } from '../acquisition';
 import { InstallInteractiveArgs, InteractiveLaunchOptions } from '../interfaces';
 import { ClientMapper } from '../clientMapper';
-import { getEol } from './vscodeUtilities';
+import { getEol, isUnsavedNotebook } from './vscodeUtilities';
 import { toNotebookDocument, DotNetInteractiveNotebookContentProvider } from './notebookProvider';
 
 export function registerAcquisitionCommands(context: vscode.ExtensionContext, dotnetPath: string) {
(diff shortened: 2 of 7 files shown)


colombod (approved):

src/dotnet-interactive-vscode/src/clientMapper.ts
```diff
@@ -27,6 +27,24 @@ export class ClientMapper {
         return client;
     }
 
+    reassociateClient(oldUri: Uri, newUri: Uri) {
```
colombod, line 30: can we test any of this?

src/dotnet-interactive-vscode/src/vscode/commands.ts
```diff
@@ -121,6 +122,32 @@ export function registerFileFormatCommands(context: vscode.ExtensionContext, cli
...
+                return isUnsavedNotebook(notebookUri) && path.basename(notebookUri.fsPath) === fileName;
+            }
+
+            return false;
+        }) >= 0;
+    }
+
+    function getNewNotebookName(): string {
+        let suffix = 1;
+        while (workspaceHasUnsavedNotebookWithName(`Untitled-${suffix}.dib`)) {
```
colombod, line 138: do we need to add the .dib ? Untitled-notebook-1 looks good enough and we leave the choice to the user at save time

brettfo: Yes, we need to specify `.dib`.  A few lines down that's the part that makes it the default extension when the user finally tries to save.

src/dotnet-interactive-vscode/src/clientMapper.ts
brettfo: Added test.

PR #742, opened 2020-09-02
SharpLab-like extension experiment

vzarytovskii:
A proof-of-concept of SharpLab-like extension for interactive.
It shows what actual C#/IL/JitAsm code was generated.

PoC version works only with C# code and uses a custom (prism) renderer.

What can/should be improved: 
* I can't CSS properly, so there are a bunch of hacks in there.
* Monaco styling can be (probably) used instead of a PrismJS. This will allow reusing user styles?
* Code can be generalized (i.e. have a common InspectKernel, which then will be invoking different implementations, based on the original kernel used, for example - C#, F#, VB, SQL etc...).
* RequireJS can (and probably should) be used to load prism resources.
* SharpLab part (JitDisassembler, which has just been copied over for the sake of making this PoC working) may be separated into a library, which can be published and consumed as a NuGet. This is to be discussed with SharpLab maintainers.
* Some tooltips can/should be added probably.
* Some additional styling - greying out some generated code (such as async state machine, in case if you compiling as a script).

How does it look in the notebook:

notebook

Edit: there are small problems with rendering right now, since it's being rendered, and then on dom ready, prism kicks in. It will be fixed shortly.


vzarytovskii:
Huh, interesting that Linux tests are failing, probably because of some Windows-specific is being used. I'll take a look at it.

src/Microsoft.DotNet.Interactive.ExtensionLab/InspectExtension.cs
```diff
@@ -0,0 +1,201 @@
...
+        }
+
+        private void Inspect(OptimizationLevel configuration, SourceCodeKind kind, Platform platform, KernelInvocationContext context)
+        {
+            if (!(context.Command is SubmitCode))
+                return;
+
+            var command = context.Command as SubmitCode;
+
+            // TODO: Is there a proper way of cleaning up code from the magic commands?
```
jonsequitur, line 51: The `SubmitCode` command has an internal `LanguageNode` that might be useful.  Maybe we should make it public?

vzarytovskii: Yeah, that will be useful, I think.

vzarytovskii:
Okay, it was much simpler than Windows-specific APIs, it was a regex + line endings 🤦‍♂️

vzarytovskii:
@jonsequitur I've added the original license from SharpLab to jit decompiler module.

jonsequitur (approved):

PR #745, opened 2020-09-03
vscode http tunnel

colombod:


src/dotnet-interactive-vscode/src/stdioKernelTransport.ts
```diff
@@ -18,41 +19,100 @@ import { LineReader } from './lineReader';
...
+
+    private async configureHttpArgs(args: string[]): Promise<string[]> {
+        let newArgs = [...args];
+        let index = newArgs.indexOf("--http-port");
+        if (index < 0) {
+            index = newArgs.indexOf("--http-port-range");
+            this.httpPort = await this.findFreePort();
+
+            if (index > 0) {
+                this.diagnosticChannel.appendLine("the --http-port-range command option is not supported in vscode extension, remove if from settings. The kernel will start with --http-port option isntead.");
```
jonsequitur, line 79:
```suggestion
                this.diagnosticChannel.appendLine("The --http-port-range option is not supported in the VS Code extension. Please use --http-port instead.");
```

jonsequitur: Should we error out instead? This seems like it will work nondeterministically, and if the user isn't looking at the console, they won't see the warning.

src/dotnet-interactive/CommandLine/CommandLineParser.cs
```diff
@@ -329,6 +330,39 @@ Command StdIO()
                     parseArgument: result => result.Tokens.Count == 0 ? HttpPortRange.Default : ParsePortRangeOption(result),
                     description: "Specifies the range of ports to use to enable HTTP services");
 
+                var httpPortOption = new Option<HttpPort>(
+                    "--http-port",
+                    description: "Specifies the port on which to enable HTTP services",
+                    parseArgument: result =>
+                    {
+                        if (result.Parent.Parent.Children.FirstOrDefault(c => c.Symbol == httpPortRangeOption) is OptionResult conflictingOption)
```
jonsequitur, line 338: You can do this now instead (possible requiring a System.CommandLine update):

```csharp
if (result.FindResultFor(httpPortRangeOption) is { } conflictingOption)
```

src/dotnet-interactive-vscode/src/stdioKernelTransport.ts
colombod: this will be  removed later

src/dotnet-interactive-vscode/src/stdioKernelTransport.ts
```diff
@@ -13,48 +14,129 @@ import {
...
-                : '';
-            this.diagnosticChannel.appendLine(message + messageCodeSuffix + messageSignalSuffix);
+    private async configureHttpArgs(args: string[]): Promise<string[]> {
+        let newArgs = [...args];
+        let index = newArgs.indexOf("--http-port");
+        if (index < 0) {
+            index = newArgs.indexOf("--http-port-range");
+            this.httpPort = await this.findFreePort();
+
+            //todo: this is a temporary work during transition
```
jonsequitur, line 101: Comments like this almost guarantee that the code will remain forever.

jonsequitur (approved):

Issue #769, opened 2020-09-10
Out-Display not working as expected in VS Code. 


jhoneill:
### Describe the bug
I've been experimenting outputting SVG, which worked in Jupyter with an older version of .Net Interactive, but not in VS Code, eventually I simplified this down to _any_ HTML markup not working, with this simple example
```
"<body><b>Test<b><body>" | out-display -mimetype "text/html"
```
On  1.0.115407 / Jupyter  outputs **Test**    -this is preserved if I save the file and re-open in vscode.
On 1.0.145905 / VS Code outputs `<body><b>Test<b><body>`

I can't tell if there is an intentional change in behavior (does out-display need to be called differently), or if Out-display or something downstream from it is failing. 

### Please complete the following:

**Which version of .NET Interactive are you using?**   1.0.115407 (works) and 1.0.145905 (fails)

- OS
    - [X] Windows 10
    - [ ] macOS
    - [ ] Linux (Please specify distro)
    - [ ] iOS
    - [ ] Android
 - Browser 
    - [ ] Chrome
    - [X] Edge  (86.0.622.10)
    - [ ] Firefox
    - [ ] Safari
 - Frontend
    - [X] Jupyter Notebook
    - [ ] Jupyter Lab
    - [ ] nteract
    - [X] Visual Studio Code  (1.49.0-insider)
    - [ ] Other (please specify)

### Screenshots
![image](https://user-images.githubusercontent.com/16209593/92727578-3f747800-f367-11ea-9be8-772cac458822.png)
Above, vs-code failing, below Jupyter working
![image](https://user-images.githubusercontent.com/16209593/92727635-59ae5600-f367-11ea-82c2-3cff294a7085.png)

jonsequitur:
This was somewhat by design, as a result of a recent formatter overhaul, but it's worth discussing this scenario again.

@dsyme We changed this behavior to support formatting all strings in `<div>` elements, rather than rendering them with `text/plain`. Among other things, this allows columnar whitespace layouts to be preserved. 

It also made things a bit more consistent. Two different behaviors were conflated in the `display` API previously, like this:

* `display("<b>hello</b>", "text/html")` -> The input is HTML, please don't encode it.
* `display(someObject, "text/html")` -> Please render a representation of the object into an HTML table.

The first case now encodes the HTML, so the meaning is consistent in both cases: Please transform the input into HTML.

If you want to use a string literal to emit HTML, you can still do this:

[...]

PR #770, opened 2020-09-10
check for double encoding

colombod:
displaying html and json literals produce double encoded values,

As Reported in #769

### DIFF
--- a/src/Microsoft.DotNet.Interactive.Formatting.Tests/HtmlFormatterTests.cs
+++ b/src/Microsoft.DotNet.Interactive.Formatting.Tests/HtmlFormatterTests.cs
@@ -26,7 +26,7 @@ namespace Microsoft.DotNet.Interactive.Formatting.Tests
         {
 
             [Fact]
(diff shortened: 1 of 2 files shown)


src/Microsoft.DotNet.Interactive.Formatting.Tests/HtmlFormatterTests.cs
```diff
@@ -24,6 +24,22 @@ public class HtmlFormatterTests : FormatterTestBase
     {
         public class Objects : FormatterTestBase
         {
+
+            [Fact]
+            public void does_not_double_encode_json_string()
```
jonsequitur, line 29:
```suggestion
            public void does_not_double_encode_HTML_string()
```

jonsequitur:
Thoughts on https://github.com/dotnet/interactive/issues/769#issuecomment-690500704?

I think we may want to keep the new behavior, and add some more .NET-ish APIs to make things more explicit, rather than restore this behavior.

jonsequitur:
Do we still need this?

PR #773, opened 2020-09-10
make vscode source format on save and make an initial pass at formatting

brettfo:


src/dotnet-interactive-vscode/.vscode/settings.json
```diff
@@ -6,8 +6,9 @@
     "search.exclude": {
         "out": true // set this to false to include "out" folder in search results
     },
+    "debug.allowBreakpointsEverywhere": true,
+    "editor.formatOnSave": true,
```
brettfo, line 10: This was the only line I added to this PR, the rest was Code applying the auto-formats.

jonsequitur (approved):

PR #792, opened 2020-09-17
pre load 'https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-…

KevinRansom:
People using the ExtensionLab package have seen package resolution errors.

We fixed the underlying issue by removing from it the dependency on the fsharp.compiler.private.scripting package.  However, this change, ensures that the notebook will be able to find it, if script authors choose to use earlier versions of the package.

Update the samples so that they can select the latest preview versions of the package.

### DIFF
--- a/src/Microsoft.DotNet.Interactive/PackageRestoreContext.cs
+++ b/src/Microsoft.DotNet.Interactive/PackageRestoreContext.cs
@@ -25,6 +25,8 @@ namespace Microsoft.DotNet.Interactive
 
         public PackageRestoreContext()
         {
+            // By default look in to the package source  "https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-tools/nuget/v3/index.json"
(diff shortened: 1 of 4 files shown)


src/Microsoft.DotNet.Interactive/PackageRestoreContext.cs
```diff
@@ -25,6 +25,8 @@ public class PackageRestoreContext : IDisposable
 
         public PackageRestoreContext()
         {
+            // By default look in to the package source  "https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-tools/nuget/v3/index.json"
+            AddRestoreSource("https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-tools/nuget/v3/index.json");
```
jonsequitur, line 29: Let's also add our own feed for now: https://dotnet.myget.org/F/dotnet-try/api/v3/index.json

colombod:
tests are not up to date with the code change

KevinRansom:
@colombod , thanks mate.

jonsequitur (approved):

Issue #724, opened 2020-08-25
Raspberry Pi install fails on cffi wheel due to missing dependencies


gnarlygeek:
### Describe the bug
When following the instructions here: 
        **https://github.com/dotnet/interactive/blob/main/docs/small-factor-devices.md** 
**The install on Raspberry PI 4 fails on the cffi wheel install due to a missing dependency.** 

The resolution is described in this StackOverflow item: https://stackoverflow.com/questions/22073516/failed-to-install-python-cryptography-package-with-pip-and-setup-py

The tl;dr is to run the following command and rerun the curl command: 
        **sudo apt-get install build-essential libssl-dev libffi-dev python-dev**

### Please complete the following:

- OS
    - [ ] Windows 10
    - [ ] macOS
    - [x ] Linux (Please specify distro) **"Raspbian GNU/Linux 10 (buster)"**
    - [ ] iOS
    - [ ] Android
 - Browser - N/A
 - Frontend N/A

colombod:
@gnarlygeek is this issue resolved for you?

gnarlygeek:
Hello Diego,

Yes it is. Sorry, I thought I had already replied. The script now works and the web page reference is corrected.

Thanks.

colombod:
Closing this one then

PR #793, opened 2020-09-18
Update setup-rapbian.sh

colombod:
Addressing #724, @gnarlygeek can you test if this addresses the issue?


tools/setup-rapbian.sh
```diff
@@ -4,7 +4,7 @@
 ### Install System.Drawing dependencies and virtualenv ###
 ##########################################################
 echo "Installing dependencies..."
-sudo apt install -y libgdiplus virtualenv
+sudo apt install -y libgdiplus virtualenv build-essential libssl-dev libffi-dev python-dev
```
jonsequitur, line 7: The file name should be `raspbian.sh`

gnarlygeek (changes requested):
This change resolved the issue and the installation completed successfully. There was a warning about a missing dependency for zlib, so you may want to inject that before finalizing.  Lack of zlib did not generate any other warnings or errors during the install.

The correction of the script name is also good, however the documentation page here: small-factor-devices.md also has the misspelling.

PR #798, opened 2020-09-23
update tool and proposed api

colombod:


src/XPlot.DotNet.Interactive.KernelExtensions/KernelExtension.cs
```diff
@@ -0,0 +1,69 @@
+﻿// // Copyright (c) .NET Foundation and contributors. All rights reserved.
+// // Licensed under the MIT license. See LICENSE file in the project root for full license information.
```
jonsequitur, line 2: Nit: Double double slashes.

jonsequitur (approved):

src/XPlot.DotNet.Interactive.KernelExtensions/KernelExtension.cs
colombod: Done

PR #802, opened 2020-09-29
extract api to configure aspnet hosting to expose .NET Interactive api

colombod:


src/Microsoft.DotNet.Interactive/JavascriptUtilities.cs
```diff
@@ -0,0 +1,46 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+
+namespace Microsoft.DotNet.Interactive.Http
+{
+    public static class JavascriptUtilities
+    {
+        public static string EnsureRequireJs(Uri requireJsUri = null, string callBackName = null)
```
jonsequitur, line 10: This name and return type could be clearer. `Ensure*` usually means ensure some thing has been done, and signals idempotency. This returns a `string`.

What's it really doing?

colombod: that returns the javascript code snippet that checks and pull in if needed the requirejs library

src/Microsoft.DotNet.Interactive.Http/FileProvider.cs
```diff
@@ -5,24 +5,24 @@
...
+namespace Microsoft.DotNet.Interactive.Http
 {
     public class FileProvider : IFileProvider, IDisposable
     {
         private readonly EmbeddedFileProvider _root;
         private readonly IDisposable _eventSubscription;
         private readonly ConcurrentDictionary<string, EmbeddedFileProvider> _providers = new ConcurrentDictionary<string, EmbeddedFileProvider>();
 
-        public FileProvider(Kernel kernel)
+        public FileProvider(Kernel kernel, Assembly rootProvider)
```
jonsequitur, line 21:
```suggestion
        public FileProvider(Kernel kernel, Assembly rootProviderAssembly)
```

colombod: sounds good

src/Microsoft.DotNet.Interactive.Http/AspNetExtensions.cs
```diff
@@ -0,0 +1,111 @@
...
+                    {
+                        foreach (var serviceDescriptor in services)
+                        {
+                            c.Add(serviceDescriptor);
+                        }
+                    }
+                });
+            return builder;
+        }
+        public static IServiceCollection AddDotnetInteractive(this IServiceCollection services)
```
jonsequitur, line 51: Maybe the naming should also include the fact that this is adding HTTP-specific functionality?

src/Microsoft.DotNet.Interactive.Http/AspNetExtensions.cs
```diff
@@ -0,0 +1,111 @@
...
+                        .AllowCredentials()
+                        .SetIsOriginAllowed((host) => true);
+
+                });
+            });
+            services.AddSignalR();
+
+            return services;
+        }
+        public static IApplicationBuilder UseDotNetInteractive<T>(this IApplicationBuilder app, T kernel, Assembly staticResourceRoot, HttpProbingSettings httpProbingSettings, HttpPort httpPort) where T : Kernel
```
jonsequitur, line 71: Ditto.

src/Microsoft.DotNet.Interactive.Http/AspNetExtensions.cs
```diff
@@ -0,0 +1,111 @@
...
+            app.UseWebSockets();
+            app.UseCors("default");
+            app.UseRouting();
+            app.UseRouter(r =>
+            {
+                r.Routes.Add(new VariableRouter(kernel));
+                r.Routes.Add(new KernelsRouter(kernel));
+                var htmlNotebookFrontedEnvironment = kernel.FrontendEnvironment as HtmlNotebookFrontedEnvironment;
+
+                if (htmlNotebookFrontedEnvironment != null)
```
jonsequitur, line 87:
```suggestion
                if (htmlNotebookFrontedEnvironment is { } )
```

src/Microsoft.DotNet.Interactive.Http/AspNetExtensions.cs
```diff
@@ -0,0 +1,111 @@
...
+                {
+                    r.Routes.Add(new DiscoveryRouter(htmlNotebookFrontedEnvironment));
+                    r.Routes.Add(new HttpApiTunnelingRouter(htmlNotebookFrontedEnvironment));
+                }
+
+                if (htmlNotebookFrontedEnvironment == null || htmlNotebookFrontedEnvironment.RequiresAutomaticBootstrapping)
```
jonsequitur, line 93:
```suggestion
                if (htmlNotebookFrontedEnvironment is null || htmlNotebookFrontedEnvironment.RequiresAutomaticBootstrapping)
```

src/dotnet-interactive/CommandLine/CommandLineParser.cs
```diff
@@ -305,14 +300,8 @@ Command HttpServer()
                 httpCommand.Handler = CommandHandler.Create<StartupOptions, KernelHttpOptions, IConsole, InvocationContext>(
```
jonsequitur, line 300: I just noticed that `KernelHttpOptions` has nothing to do with HTTP. Does it need to exist and why would it be different from the already-existing `DefaultKernel` concept?

docs/javascript-overview.md
```diff
@@ -47,4 +47,13 @@ The `interactive` object contains the following properties, corresponding to the
...
-* `interactive.value`
\ No newline at end of file
+* `interactive.value`
+
+## Loading external JavaScript modules at runtime
+
+Sometimes you might need to import JavaScript modules in your notebook. You can use the `interactive` object to do so.
+
+Here is an example.
+![image](https://user-images.githubusercontent.com/375556/95768858-e6be3500-0cae-11eb-925c-53d46670bd09.png)
```
jonsequitur, line 57: Let's use code rather than images to document APIs. This will be hard to search for otherwise.

docs/javascript-overview.md
```diff
@@ -47,4 +47,13 @@ The `interactive` object contains the following properties, corresponding to the
...
+The `interactive.configureRequire` takes as input a [requirejs configuration object](https://www.tutorialspoint.com/requirejs/requirejs_configuration.htm) and returns a function that you can now use to load the dependecies.
```
jonsequitur, line 58: Let's add a simple, working code sample.

src/dotnet-interactive/CommandLine/CommandLineParser.cs
```diff
@@ -529,7 +515,7 @@ static HttpPortRange ParsePortRangeOption(ArgumentResult result)
             SetUpFormatters(frontendEnvironment, startupOptions, TimeSpan.FromSeconds(15));
 
             kernel.DefaultKernelName = defaultKernelName;
-
+            services.AddKernel(kernel);
```
jonsequitur, line 518: Doing this here seems to blur some boundaries. I would expect the caller to this method to handle registration, and for this method to be non-side-affecting.

docs/javascript-overview.md
colombod: removed image and made test example that users can actually use and try

docs/javascript-overview.md
colombod: done

src/dotnet-interactive/CommandLine/CommandLineParser.cs
colombod: that part of  code has always been like that as the kernel is created as side effect while executing the command line parser. the caller is passing in the ServiceCollection and merging it with the one used during asp.net configuration

docs/javascript-overview.md
```diff
@@ -53,7 +53,39 @@ The `interactive` object contains the following properties, corresponding to the
 
 Sometimes you might need to import JavaScript modules in your notebook. You can use the `interactive` object to do so.
 
-Here is an example.
-![image](https://user-images.githubusercontent.com/375556/95768858-e6be3500-0cae-11eb-925c-53d46670bd09.png)
+Here is an example that configures requirejs to load d3.js from cdn.
```
jonsequitur, line 56:
```suggestion
Here is an example that configures RequireJS to load D3.js from a CDN.
```

docs/javascript-overview.md
```diff
@@ -53,7 +53,39 @@ The `interactive` object contains the following properties, corresponding to the
...
+```js
+#!js
+dtreeRequire = interactive.configureRequire({
+    paths: {
+        d3: "https://d3js.org/d3.v5.min"
+    }
+});
+```
+
+Now you can require the module and use it. Using `#!html` to inject an `svg` element and then `#!js` to load the `d3.js` library and append a circle to the dom
```
jonsequitur, line 67:
```suggestion
Now you can require the module and use it. You can use the `#!html` magic command to inject an SVG element and then the `#!js` magic command to load the D3.js library and append a circle.
```

docs/javascript-overview.md
```diff
@@ -53,7 +53,39 @@ The `interactive` object contains the following properties, corresponding to the
...
+
+
+```
+#!html
+<sgv id"renderTarget"></svg>
+
+#!js
+#!html
+<svg id="renderTarget" width=300 height=300></svg>
```
jonsequitur, line 76: You have an extra `#!js` here and the `renderTarget` is being added twice.

docs/javascript-overview.md
```diff
@@ -53,7 +53,39 @@ The `interactive` object contains the following properties, corresponding to the
...
+    }
+});
+```
+
+Now you can require the module and use it. Using `#!html` to inject an `svg` element and then `#!js` to load the `d3.js` library and append a circle to the dom
+
+
+```
+#!html
+<sgv id"renderTarget"></svg>
```
jonsequitur, line 72:
```suggestion
<svg id"renderTarget"></svg>
```

docs/javascript-overview.md
colombod: fixed

docs/javascript-overview.md
colombod: fixed

jonsequitur (approved):

PR #835, opened 2020-10-14
WIP: Add SQL Kernel to dotnet-interactive

corivera:
This is a draft PR for adding a SQL kernel to dotnet interactive. This is still very much work-in-progress, with only basic scenarios working.

Things that currently work:
1. Executing queries that return a single result set.
2. Very rudimentary intellisense. The Completion and Hover callbacks are set up, but I'm still ironing out some issues with the service calls.

Things that still need to be completed:
- [x] Fully enable intellisense for SQL kernel code
- [x] Enable queries that return multiple result sets
- [x] Figure out a way to load SqlToolsService dynamically, either through a nuget package or some kind of command line configuration argument
- [x] Fix a bug where the SqlToolsService process isn't being properly terminated when dotnet-interactive exits.
- [x] Only spawn a single instance of the SqlToolsService process, rather than having each new SQL kernel create its own instance.

CC @jonsequitur @colombod


src/Microsoft.DotNet.Interactive.ExtensionLab/MsSqlKernel.cs
```diff
@@ -3,21 +3,155 @@
...
-            return new SqlConnection(_connectionString);
+            if (!_connected)
+            {
+                await _serviceClient.ConnectAsync(_tempFileUri, _connectionString);
+                await _connectionCompleted.Task;
+                _connected = true;
+            }
+
+            var queryResult = await _serviceClient.ExecuteSimpleQueryAsync(_tempFileUri, command.Code);
+            var tableString = GetTableStringForSimpleResult(queryResult);
```
jonsequitur, line 99: In general we want kernels to focus on returning data and not worry about MIME types or formatting. This will make them more flexible. The `SqliteKernel` approach might be a good shape, or we can come up with something different, but ideally it will be consistent across data kernels.

src/Microsoft.DotNet.Interactive.ExtensionLab/MsSqlServiceClient.cs
```diff
@@ -84,58 +84,69 @@ public async Task<bool> DisconnectAsync(string ownerUri)
...
         public async Task<Hover> ProvideHoverAsync(string filePath, RequestHoverText command)
         {
-            var tempFileStream = File.Open(filePath, FileMode.OpenOrCreate);
-            try
+            string oldFileContents;
+            using (var tempFileStream = File.Open(filePath, FileMode.OpenOrCreate))
+            using (var reader = new StreamReader(tempFileStream))
+            {
+                oldFileContents = await reader.ReadToEndAsync();
+            }
```
jonsequitur, line 131: You can potentially simplify using `File.ReadAllText`.

src/Microsoft.DotNet.Interactive.ExtensionLab/MsSqlServiceClient.cs
```diff
@@ -84,58 +84,69 @@ public async Task<bool> DisconnectAsync(string ownerUri)
...
+
+            var fileUri = GetUriForFilePath(filePath);
+            if (!oldFileContents.Equals(command.Code))
             {
+                using (var tempFileStream = File.Open(filePath, FileMode.OpenOrCreate))
                 using (var writer = new StreamWriter(tempFileStream))
                 {
                     await writer.WriteLineAsync(command.Code);
                 }
```
jonsequitur, line 140: You can potentially simplify using `File.AppendAllText` or `File.WriteAllText`.

src/Microsoft.DotNet.Interactive.ExtensionLab/MsSqlServiceClient.cs
```diff
@@ -156,6 +167,24 @@ public async Task<QueryExecuteSubsetResult> ExecuteQueryExecuteSubsetAsync(strin
             return await rpc.InvokeWithParameterObjectAsync<QueryExecuteSubsetResult>("query/subset", queryExecuteSubsetParams);
         }
 
+        public async Task SendTextChangeNotification(string ownerUri, string newText, string oldText)
+        {
+            var oldLines = oldText.Split('\n');
```
jonsequitur, line 172: This should probably also anticipate `\r\n`.

jonsequitur: Oh, I see you're just using this to count the lines. That can be made a little more explicit given the use case by doing something like this:

```csharp
var numberOfLines = oldText.Split(`\n`).Length;
```

src/Microsoft.DotNet.Interactive.ExtensionLab/MsSqlServiceClient.cs
corivera: This part also needs to create the temp file if it doesn't exist.

corivera: Disregard that, looks like GetTempFileName also creates the file.

corivera:
@jonsequitur I've updated the PR with your comments. Could you take another look? It's mostly complete, apart from a disconnect bug with SQL Tools Service I just started seeing today. I'll also try adding some basic unit tests after that.

corivera:
Looking into build failures.

src/Microsoft.DotNet.Interactive.ExtensionLab/MsSqlKernel.cs
```diff
@@ -3,21 +3,219 @@
...
+
+            _serviceClient.OnConnectionComplete += HandleConnectionComplete;
+            _serviceClient.OnQueryComplete += HandleQueryCompleteAsync;
+            _serviceClient.OnIntellisenseReady += HandleIntellisenseReady;
+
+            RegisterForDisposal(() =>
+            {
+                if (_connected)
+                {
+                    _serviceClient.DisconnectAsync(_tempFileUri).Wait();
```
jonsequitur, line 58: This could probably stand to have a timeout.

To avoid the possibility of a deadlock in the presence of a `SynchronizationContext` (which you have no control over here), you can do this:

```csharp
Task.Run(() => _serviceClient.DisconnectAsync(_tempFileUri)).Wait();
```

src/Microsoft.DotNet.Interactive.ExtensionLab/MsSqlKernel.cs
```diff
@@ -3,21 +3,219 @@
...
-    public class MsSqlKernel : SqlKernel
+    public class MsSqlKernel :
+        Kernel,
+        IKernelCommandHandler<SubmitCode>,
+        IKernelCommandHandler<RequestCompletions>,
+        IKernelCommandHandler<RequestHoverText>
     {
+        private bool _connected = false;
+        private bool _intellisenseReady = false;
+        private readonly string _tempFilePath;
```
jonsequitur, line 32: Any reason not to use `FileInfo`?

src/Microsoft.DotNet.Interactive.ExtensionLab/MsSqlKernel.cs
```diff
@@ -3,21 +3,219 @@
...
+        private readonly string _tempFileUri;
```
jonsequitur, line 33: Any reason not to use `Uri`?

src/Microsoft.DotNet.Interactive.ExtensionLab/MsSqlKernel.cs
```diff
@@ -3,21 +3,219 @@
...
+            RegisterForDisposal(() => File.Delete(_tempFilePath));
+        }
+
+        private void HandleConnectionComplete(object sender, ConnectionCompleteParams connParams)
+        {
+            if (connParams.OwnerUri.Equals(_tempFileUri))
+            {
+                if (connParams.ErrorMessage != null)
+                {
+                    _connectionCompleted.SetException(new Exception(connParams.ErrorMessage));
```
jonsequitur, line 70: It's generally a good idea to avoid using `Exception`. Is there a more specific exception we could use here? What are some examples of why an error will occur?

src/Microsoft.DotNet.Interactive.ExtensionLab/MsSqlServiceClient.cs
```diff
@@ -0,0 +1,206 @@
...
+using Microsoft.DotNet.Interactive.Events;
+using StreamJsonRpc;
+
+namespace Microsoft.DotNet.Interactive.ExtensionLab
+{
+    public class MsSqlServiceClient : IDisposable
+    {
+        public static readonly MsSqlServiceClient Instance = new MsSqlServiceClient();
+
+        private Process process;
```
jonsequitur, line 21:
```suggestion
        private Process _process;
```

src/Microsoft.DotNet.Interactive.ExtensionLab/MsSqlServiceClient.cs
```diff
@@ -0,0 +1,206 @@
...
+        private JsonRpc rpc;
```
jonsequitur, line 22:
```suggestion
        private JsonRpc _rpc;
```

src/Microsoft.DotNet.Interactive.ExtensionLab/MsSqlKernel.cs
```diff
@@ -3,21 +3,219 @@
...
+                            var subsetResult = await _serviceClient.ExecuteQueryExecuteSubsetAsync(subsetParams);
+
+                            if (subsetResult.Message != null)
+                            {
+                                context.Display(subsetResult.Message);
+                            }
+                            else
+                            {
+                                var tableString = GetTableStringForResult(resultSummary.ColumnInfo, subsetResult.ResultSubset.Rows);
+                                context.Display(tableString);
```
jonsequitur, line 144: This should return a data structure rather than assuming a specific MIME type.

Thoughts about `IEnumerable<IEnumerable<IEnumerable<(string name, object value)>>>` or similar?

src/Microsoft.DotNet.Interactive.ExtensionLab/MsSqlKernel.cs
corivera: This ErrorMessage can come from any type of Exception. We just catch the base Exception class on the service side for reporting the error message: https://github.com/microsoft/sqltoolsservice/blob/main/src/Microsoft.SqlTools.ServiceLayer/Connection/ConnectionService.cs#L565

src/Microsoft.DotNet.Interactive.ExtensionLab/MsSqlKernel.cs
corivera: What's the difference here versus just waiting on the disconnect task itself?

src/Microsoft.DotNet.Interactive.ExtensionLab/MsSqlKernel.cs
jonsequitur: Would `SqlException` be appropriate? A custom exception type?

src/Microsoft.DotNet.Interactive.ExtensionLab/MsSqlKernel.cs
jonsequitur: If there's a `SynchronizationContext` present, then the task will be scheduled on that context, at which point calling a blocking method such as `Task.Wait` or `Task.Result` can result in a deadlock. 

Calling `Task.Run` will ensure that the task is not scheduled on the current `SynchronizationContext` if there is one.

colombod:
Build is still broken, these are are the messages

DataFrameKernelExtensionTests.cs(32,27): error VSTHRD200: Use "Async" suffix in names of methods that return an awaitable type.

jonsequitur (approved):

PR #836, opened 2020-10-15
simplified loading csv from string

colombod:


src/Microsoft.DotNet.Interactive.ExtensionLab.Tests/DataFrameKernelExtensionTests.cs
```diff
@@ -37,13 +37,10 @@ public async Task it_registers_formatters()
...
-            using var stream = new MemoryStream();
-            using var writer = new StreamWriter(stream);
-            writer.AutoFlush = true;
-            writer.Write(@"id,name,color,deliciousness
+            using var stream = @"id,name,color,deliciousness
 1,apple,green,10
 2,banana,yellow,11
-3,cherry,red,9000");
+3,cherry,red,9000".ToStream();
             stream.Position = 0;
```
jonsequitur, line 44: Shouldn't have to set `stream.Position` here.

jonsequitur (approved):

Issue #842, opened 2020-10-19
Copy "FSharp.Core.xml" from "FSharp.Core" nuget package to build/publish output


TIHan:
When we build "dotnet-interactive", the file "FSharp.Core.xml" is not included in the build/publish output. It needs be copied from the "FSharp.Core" nuget package.

This is necessary in order to get documentation in VSCode notebooks for the "FSharp.Core" library.

Issue #626, opened 2020-07-08
F# completion item documentation is incorrect.


brettfo:
F# completion documentation shows the type name

```
Microsoft.FSharp.Control.FSharpAsync`1[FSharp.Compiler.SourceCodeServices.FSharpToolTipText`1[System.String]]
```

instead of the actual content.

![image](https://user-images.githubusercontent.com/926281/86952910-898f7200-c108-11ea-9e4d-dc4087b6bd8f.png)

Issue stems from [here](https://github.com/dotnet/interactive/blob/86d3f806832b53425bb1c95d2d88437cd467f05d/src/Microsoft.DotNet.Interactive.FSharp/FSharpKernel.fs#L74-L76).

PR #843, opened 2020-10-19
F# documentation completion item fix

TIHan:
Resolves this: https://github.com/dotnet/interactive/issues/626

It does not add any semantic/syntax highlights on the text for the tooltip. We should do that in a separate PR.

Another note: "FSharp.Core.xml" must be in the published output of "dotnet-interactive.csproj" in order to see "FSharp.Core" library documentation in notebooks. See here: https://github.com/dotnet/interactive/issues/842


TIHan:
This is ready

jonsequitur:
Got tests?

KevinRansom (approved):

src/Microsoft.DotNet.Interactive.FSharp.Tests/KernelTests.fs
```diff
@@ -44,6 +44,24 @@ type KernelTests() =
             
             texts
 
+    let getCompletions (line : int) (column : int) (code : #seq<string>)  =
+        let code = String.concat "\r\n" code
```
jonsequitur, line 48: We probably won't want the `\r` on non-Windows OSs, yeah?

TIHan: Probably not, good catch

jonsequitur (approved):

PR #854, opened 2020-10-27
Remove the hardcoded plotly extension

cartermp:
yeet it outa here in favor of xplot.plotly supporting interactive extensions


colombod:
the project is still reference by the solution file I think

PR #863, opened 2020-11-02
move to net standard 2.1

colombod:

### DIFF
--- a/src/Microsoft.DotNet.Interactive/KernelSupportsNugetExtensions.cs
+++ b/src/Microsoft.DotNet.Interactive/KernelSupportsNugetExtensions.cs
@@ -75,7 +75,7 @@ namespace Microsoft.DotNet.Interactive
 
                         if (token != null &&
                             !token.StartsWith("nuget:") &&
-                            !Path.EndsInDirectorySeparator(token))
+                            !EndsInDirectorySeparator(token))
                         {
                             return new FileInfo(token);
                         }
@@ -143,6 +143,11 @@ namespace Microsoft.DotNet.Interactive
 
                 return Task.CompletedTask;
             }
+        }
+
+        private static bool EndsInDirectorySeparator(string path)
+        {
+            return path.Length > 0 && path.EndsWith(Path.DirectorySeparatorChar);
         }
 
         private class PackageReferenceComparer : IEqualityComparer<PackageReference>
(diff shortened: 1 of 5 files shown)


src/Microsoft.DotNet.Interactive/KernelSupportsNugetExtensions.cs
```diff
@@ -145,6 +145,11 @@ private static Command r()
             }
         }
 
+        private static bool EndsInDirectorySeparator(string path)
```
jonsequitur, line 148: This can be an extension method I suppose.

src/Microsoft.DotNet.Interactive/Microsoft.DotNet.Interactive.csproj
```diff
@@ -1,10 +1,10 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>netcoreapp3.1</TargetFramework>
+    <TargetFramework>netstandard2.1</TargetFramework>
+    <LangVersion>8.0</LangVersion>
```
jonsequitur, line 5: This can be latest.

src/Microsoft.DotNet.Interactive/Microsoft.DotNet.Interactive.csproj
```diff
@@ -39,6 +39,7 @@
     <PackageReference Include="FSharp.Compiler.Private.Scripting">
       <PrivateAssets>all</PrivateAssets>
     </PackageReference>
+    <PackageReference Include="System.Text.Json" Version="4.7.2" />
```
jonsequitur, line 42: Is this needed?

jonsequitur (approved):

src/Microsoft.DotNet.Interactive/Microsoft.DotNet.Interactive.csproj
colombod: without that build fails due to conflict resolution on the package

src/Microsoft.DotNet.Interactive/KernelSupportsNugetExtensions.cs
colombod: is not used anywhere else

jonsequitur: It's a minor readability nitpick, not important.

PR #870, opened 2020-11-10
[WIP] SQL kernel updates

jonsequitur:


RiaanVR:
@jonsequitur I am currently experimenting with some the stuff you demo'd at .NET conf, I'm running into a failure when running the #!connect, about can't find SQL Tools Service, I'm guessing this PR is a long term solution for that, do you have a short term solution (workaround) until this is merged? I'm using version 1.0.0-beta.20574.9 of `Microsoft.DotNet.Interactive.ExtensionLab`

jonsequitur:
@RiaanVR We're working out how to publish the SQL Tools as NuGet packages. This is mainly a licensing task at this point. You can download and install it directly in the meantime and provide the path via an environment variable. But you might run into package version issues, especially as we've moved from .NET Core 3.1 to .NET 5 in the interim.

PR #872, opened 2020-11-10
debounce diagnostics request on execute

colombod:


brettfo (approved):

src/dotnet-interactive-vscode/src/interactiveClient.ts
```diff
@@ -87,7 +88,8 @@ export class InteractiveClient {
         return serializedNotebook.rawData;
     }
 
-    async execute(source: string, language: string, outputObserver: { (outputs: Array<CellOutput>): void }, diagnosticObserver: (diags: Array<Diagnostic>) => void, token?: string | undefined): Promise<void> {
+    async execute(id: string, source: string, language: string, outputObserver: { (outputs: Array<CellOutput>): void }, diagnosticObserver: (diags: Array<Diagnostic>) => void, token?: string | undefined): Promise<void> {
```
colombod, line 91: @brettfo do you have a better suggestion about this id, we use is a key parameter for the `debounce` operator, I am wondering if we can have a better parameter story for this function, maybe use a executeConfiguration at the end?

```
async execute(source: string, language: string, outputObserver: { (outputs: Array<CellOutput>): void }, diagnosticObserver: (diags: Array<Diagnostic>) => void, settings : {token?: string | undefined, id?: string | undefined} |undefined): Promise<void>
```

Issue #884, opened 2020-11-17
Can't install ClockExtension  sample


PawelStadnicki:
### Describe the bug
I run the ClockExtension.ipynb notebook .
The Build and Pack step were fine
but when I want to load the extension I got following error.
It is same on my both computers, after fresh dotnet-interactive installation with vscode extension

```
Installed package clockextension version 1.4.
Loading kernel extension "ClockKernelExtension" from assembly C:\Users\...\.nuget\packages\clockextension\1.4.0\interactive-extensions\dotnet\ClockExtension.dl
Error: Microsoft.DotNet.Interactive.KernelExtensionLoadException: Failure loading Kernel Extension
---> System.MissingMethodException: Method not found: 'System.Threading.Tasks.Task`1<Microsoft.DotNet.Interactive.Events.DisplayedValue> Microsoft.DotNet.Interactive.KernelInvocationContextExtensions.DisplayAsync(Microsoft.DotNet.Interactive.KernelInvocationContext, System.Object, System.String)'.
at ClockExtension.ClockKernelExtension.OnLoadAsync(Kernel kernel)
at System.Runtime.CompilerServices.AsyncMethodBuilderCore.Start[TStateMachine](TStateMachine& stateMachine)
at ClockExtension.ClockKernelExtension.OnLoadAsync(Kernel kernel)
at Microsoft.DotNet.Interactive.Extensions.AssemblyBasedExtensionLoader.LoadFromAssembly(FileInfo assemblyFile, Kernel kernel, KernelInvocationContext context) in F:\workspace\_work\1\s\src\Microsoft.DotNet.Interactive\Extensions\AssemblyBasedExtensionLoader.cs:line 130
--- End of inner exception stack trace ---
```

Note that **I have no  idea** what the path in error is F:\workspace etc

### Please complete the following:

**Which version of .NET Interactive are you using?** 
Version: 1.0.156101+537f46e89a1690a55315d805eb90a5877aff9de2

Build date: 2020-11-13T08:54:27.7781028Z

- OS
    - Windows 10
    - [ ] macOS
    - [ ] Linux (Please specify distro)
    - [ ] iOS
    - [ ] Android
 - Browser 
    -  Chrome
    - [ ] Edge
    - [ ] Firefox
    - [ ] Safari
 - Frontend
    - [ ] Jupyter Notebook
    - [ ] Jupyter Lab
    - [ ] nteract
    - Visual Studio Code
    - [ ] Other (please specify)

.

[...]

PR #886, opened 2020-11-17
Fix clock extension example

colombod:
Addressing issue #884


samples/extensions/ClockExtension/ClockKernelExtension.cs
```diff
@@ -38,10 +38,10 @@ public async Task OnLoadAsync(Kernel kernel)
                 };
 
             clockCommand.Handler = CommandHandler.Create(
-                async (int hour, int minute, int second, KernelInvocationContext context) =>
-            {
```
jonsequitur, line 42: There are `void` returning overloads so you don't need to return a `Task`.

jonsequitur (approved):

Issue #869, opened 2020-11-09
Consume .NET 5.0 SDK when RTM is available.


brettfo:
When the .NET 5.0 SDK RTM is publicly available, this repo should move to that.  This will likely be a 2 step process:

1. Update global.json with SDK version and get CI and signed builds to pass.
2. [optional] update `<TargetFramework>` values in appropriate projects to `net5.0`.
   - If this step is taken, the dotnet requirement checks for the VS Code extension should also be updated [here](https://github.com/dotnet/interactive/blob/e8365bba80d734dc3f9ec6bb0fd16890a3f151a4/src/dotnet-interactive-vscode/package.json#L113).

akdor1154:
Maybe related to this.. is there any method for a user to specify framework version? I.e. currently it's 3.1.0 and I can't see any method to force it to 5.0.0, which I have installed. If this issue gets fixed, will it be set to 5.0.0, with no way to force 3.1.0?

jonsequitur:
Currently there isn't but we'd like to enable this eventually.

dhirschfeld:
I'd love to try this out with .NET5 which has just been released! :tada:
```
> .\dotnet-interactive.exe --help
It was not possible to find any compatible framework version
The framework 'Microsoft.AspNetCore.App', version '3.1.0' was not found.
  - The following frameworks were found:
      5.0.0 at [C:\Miniconda\dotnet\shared\Microsoft.AspNetCore.App]

You can resolve the problem by installing the specified framework and/or SDK.

The specified framework can be found at:
  - https://aka.ms/dotnet-core-applaunch?framework=Microsoft.AspNetCore.App&framework_version=3.1.0&arch=x64&rid=win10-x64
```

kurt-mueller-osumc:
Awesome, this is just awesome. I'm so excited y'all :).

PR #893, opened 2020-11-18
migrate to net5.0

brettfo:
Retarget to `net5.0`.  This will enable `#r` to other `net5.0` assemblies.

During testing an issue was found where F# doesn't follow ref->impl type forwards.  An issue has been created at dotnet/fsharp#10496 and a workaround applied to this codebase.  I also included a testcase with this scenario.

Creating as draft until all CI issues have been worked out.

~~An issue with ILSpy which powers the `#!inspect` magic command has been found and filed at icsharpcode/ILSpy#2228 and a local workaround has been added.~~

The 5.0 SDK requires MSBuild 16.8 which requires VS 16.8.  The internal VMs have been updated.

### DIFF
--- a/src/Microsoft.DotNet.Interactive.ExtensionLab/Inspector/CSharpDecompiler/CSharpDecompiler.cs
+++ b/src/Microsoft.DotNet.Interactive.ExtensionLab/Inspector/CSharpDecompiler/CSharpDecompiler.cs
@@ -15,7 +15,9 @@ namespace Microsoft.DotNet.Interactive.ExtensionLab.Inspector.CSharpDecompiler
         {
             var decompilationLanguageVersion = Defaults.GetDecompilationLanguageVersion(inspectionOptions.DecompilationLanguage);
 
-            var decompiler = new ICSharpCode.Decompiler.CSharp.CSharpDecompiler(module: assembly, assemblyResolver: new UniversalAssemblyResolver($"{Defaults.InternalAssemblyName}.dll", true, "net5.0"), new DecompilerSettings(decompilationLanguageVersion))
+            var assemblyResolver = new UniversalAssemblyResolver($"{Defaults.InternalAssemblyName}.dll", true, ".NETCoreApp,Version=v5.0");
+            var settings = new DecompilerSettings(decompilationLanguageVersion);
+            var decompiler = new ICSharpCode.Decompiler.CSharp.CSharpDecompiler(module: assembly, assemblyResolver: assemblyResolver, settings: settings)
             {
                 DebugInfoProvider = debugInfoProvider
             };
--- a/src/Microsoft.DotNet.Interactive.ExtensionLab/Inspector/CSharpDecompiler/CSharpDecompiler.cs
+++ b/src/Microsoft.DotNet.Interactive.ExtensionLab/Inspector/CSharpDecompiler/CSharpDecompiler.cs
@@ -15,7 +15,11 @@ namespace Microsoft.DotNet.Interactive.ExtensionLab.Inspector.CSharpDecompiler
         {
             var decompilationLanguageVersion = Defaults.GetDecompilationLanguageVersion(inspectionOptions.DecompilationLanguage);
 
-            var assemblyResolver = new UniversalAssemblyResolver($"{Defaults.InternalAssemblyName}.dll", true, ".NETCoreApp,Version=v5.0");
+            var systemPrivateCoreLibLocation = typeof(object).Assembly.Location;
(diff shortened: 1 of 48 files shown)


src/Microsoft.DotNet.Interactive.Tests/VariableSharingTests.cs
```diff
@@ -117,6 +117,92 @@ public class VariableSharingTests
                   .Be("123:System.Int32");
         }
 
+        [Theory]
```
jonsequitur, line 120: It's not clear how these are different from the existing tests.

brettfo: Instead of simply repeating the value, an operation is performed on it, e.g., `x + 1`.  This ensures that the type of `x` and the type of `1` refer to the same `System.Int32`.

jonsequitur: Why not change the existing tests rather than duplicate? Or make it clear that these tests are also verifying that the type is correct?

brettfo: made test names more explicit

src/Microsoft.DotNet.Interactive.ExtensionLab/Inspector/CSharpDecompiler/CSharpDecompiler.cs
```diff
@@ -15,7 +15,10 @@ internal static string Decompile(in PEFile assembly, in InspectionOptions inspec
         {
             var decompilationLanguageVersion = Defaults.GetDecompilationLanguageVersion(inspectionOptions.DecompilationLanguage);
 
-            var decompiler = new ICSharpCode.Decompiler.CSharp.CSharpDecompiler(module: assembly, assemblyResolver: new UniversalAssemblyResolver($"{Defaults.InternalAssemblyName}.dll", true, "netcoreapp3.1"), new DecompilerSettings(decompilationLanguageVersion))
```
siegfriedpammer, line 18: See my comments at https://github.com/icsharpcode/ILSpy/issues/2228#issuecomment-731844279... UniversalAssemblyResolver does not support the TFM format as used by the project system. It uses the format used by `System.Runtime.Versioning.TargetFrameworkAttribute`. You should use `.NETCoreApp,Version=3.1` instead of `".netcoreapp3.1"`.

src/Microsoft.DotNet.Interactive.ExtensionLab/Inspector/CSharpDecompiler/RuntimeAssemblyResolver.cs
```diff
@@ -0,0 +1,45 @@
...
+using System.IO;
+using ICSharpCode.Decompiler.Metadata;
+
+namespace Microsoft.DotNet.Interactive.ExtensionLab.Inspector.CSharpDecompiler
+{
+    /// <summary>
+    /// This assembly resolver is to work around issue https://github.com/icsharpcode/ILSpy/issues/2228 where assemblies
+    /// from the 5.0.0 runtime can't be resolved.
+    /// </summary>
+    internal class RuntimeAssemblyResolver : IAssemblyResolver
```
siegfriedpammer, line 13: This should not be necessary.

src/Microsoft.DotNet.Interactive.ExtensionLab/Inspector/CSharpDecompiler/CSharpDecompiler.cs
brettfo: Thank you for following up.  I added a reply to the original issue and using `.NETCoreApp,Version=v5.0` was the proper solution.

src/Microsoft.DotNet.Interactive.ExtensionLab/Inspector/CSharpDecompiler/RuntimeAssemblyResolver.cs
brettfo: Removed.

KevinRansom (approved):

jonsequitur (approved):

colombod (approved):

dhirschfeld:
It looks like this closes #869?

brettfo:
@dhirschfeld Mostly, yes.  I have a few more checks to do then I'll push an update to the VS Code extension and this can all be considered done.

brettfo:
Publish is now done, closing other issue.

PR #900, opened 2020-11-21
Add ASP.NET support

halter73:
There's a lot left to do before undrafting, but I said I'd open a draft PR today!

I want to have `#!aspnet` enable the ASP.NET environment I set up in UseAspNet(). I also want to move this out of the C# Kernel so it can be used with other languages.

I'm still experimenting heavily with what the inner loop should look like, but you can get a feel for the current experience by looking at https://github.com/halter73/AspNetNotebooks/blob/master/AspNetMin.ipynb.

image

In my opinion, the really cool thing about this is being able to modify an ASP.NET web app, make a request, and see the response along with any diagnostic output in a fraction of a second without leaving the notebook. This is possible because we don't have to restart the process, host or server to deploy changes. I might do a quick video demo once this is a little more polished.


src/Microsoft.DotNet.Interactive.CSharp/AspNetKernelExtensions.cs
```diff
@@ -0,0 +1,169 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Microsoft.DotNet.Interactive.Commands;
+
+namespace Microsoft.DotNet.Interactive.CSharp
```
jonsequitur, line 6: I suspect that this code will be more at home in the `Microsoft.DotNet.Interactive.Http` project.

src/Microsoft.DotNet.Interactive.CSharp/AspNetKernelExtensions.cs
```diff
@@ -0,0 +1,169 @@
...
+{
+    public static class AspNetKernelExtensions
+    {
+        public static CSharpKernel UseAspNet(
+            this CSharpKernel kernel)
+        {
+            var command = new SubmitCode(@"
+Environment.SetEnvironmentVariable($""ASPNETCORE_{WebHostDefaults.PreventHostingStartupKey}"", ""true"");
+
+class LogLevelMonitor : IOptionsMonitor<LoggerFilterOptions>
```
jonsequitur, line 16: We can likely implement these things in library code, possibly as an `IKernelExtension`, rather than using a code submission.

src/Microsoft.DotNet.Interactive.CSharp/AspNetKernelExtensions.cs
```diff
@@ -0,0 +1,169 @@
...
+                   App.Build()(httpContext));
+        });
+    })
+    .ConfigureLogging(loggingBuilder =>
+    {
+        loggingBuilder.ClearProviders();
+        loggingBuilder.AddSimpleConsole(options => options.ColorBehavior = LoggerColorBehavior.Disabled);
+    });
+
+var __AspNet_HostRunAsyncTask = __AspNet_HostBuilder.Build().RunAsync();
```
colombod, line 161: Is there a way to modify a host that is already running?

src/Microsoft.DotNet.Interactive.Http/AspNetKernelExtensions.cs
```diff
@@ -0,0 +1,323 @@
...
+                    };
+
+                    await kernel.SetVariableAsync("App", capturedApp, typeof(IApplicationBuilder));
+                    await kernel.SetVariableAsync("Endpoints", capturedEndpoints, typeof(IEndpointRouteBuilder));
+                    await kernel.SetVariableAsync("HttpClient", httpClient);
+
+                    // We would do an "is" check on the kernel type, but we don't have a reference to the containing project
+                    if (kernel.Name == "csharp")
+                    {
+                        await kernel.SendAsync(new SubmitCode($"using {typeof(EndpointRouteBuilderExtensions).Namespace};"), CancellationToken.None);
```
halter73, line 68: Add all usings here instead of to the CSharpKernel by default.

src/dotnet-interactive/CommandLine/CommandLineParser.cs
```diff
@@ -479,15 +480,17 @@ static HttpPortRange ParsePortRangeOption(ArgumentResult result)
                     .UseDefaultNamespaces()
                     .UseXplot()
                     .UseMathAndLaTeX()
-                    .UseDotNetVariableSharing(),
+                    .UseDotNetVariableSharing()
+                    .UseAspNet(),
```
halter73, line 484: Only use on C# kernel for now.

src/Microsoft.DotNet.Interactive.Http/AspNetKernelExtensions.cs
```diff
@@ -0,0 +1,323 @@
...
+using static Microsoft.DotNet.Interactive.Formatting.PocketViewTags;
+
+namespace Microsoft.DotNet.Interactive.Http
+{
+    public static class AspNetKernelExtensions
+    {
+        private const string _aspnetLogsKey = "aspnet-logs";
+
+        public static T UseAspNet<T>(this T kernel)
+            where T : DotNetKernel
```
halter73, line 37: For first round, limit to CSharpKernel.

src/Microsoft.DotNet.Interactive.Http/AspNetKernelExtensions.cs
```diff
@@ -0,0 +1,323 @@
...
+                    FormatHttpResponseMessage(responseMessage, textWriter).Wait();
+                }
+                finally
+                {
+                    ExecutionContext.RestoreFlow();
+                }
+            }, HtmlFormatter.MimeType);
+
+            //KernelInvocationContext.Current.Publish(...)
+            //kernel.AddMiddleware()
```
halter73, line 92: Remove dead code.

src/Microsoft.DotNet.Interactive.Http/AspNetKernelExtensions.cs
```diff
@@ -0,0 +1,323 @@
...
+            where T : DotNetKernel
+        {
+            // TODO: Manage lifetime. Allow stopping, restarting, selecting port/in-memory etc...
+            var isActive = false;
+
+            var directive = new Command("#!aspnet", "Activate ASP.NET")
+            {
+                Handler = CommandHandler.Create<string, string, KernelInvocationContext>(async (from, name, context) =>
+                {
+                    // REVIEW: Commands cannot run concurrently, right?
```
halter73, line 46: right

src/Microsoft.DotNet.Interactive.FSharp/Microsoft.DotNet.Interactive.FSharp.fsproj
```diff
@@ -50,4 +50,8 @@
     <PackageReference Include="XPlot.Plotly" Version="3.0.1" />
   </ItemGroup>
 
+  <ItemGroup>
+    <FrameworkReference Include="Microsoft.AspNetCore.App" />
+  </ItemGroup>
```
halter73, line 55: Revert all kernel csproj changes

src/Microsoft.DotNet.Interactive.FSharp/FSharpKernel.fs
```diff
@@ -373,7 +373,7 @@ type FSharpKernel () as this =
         | _ ->
             false
 
-    override _.SetVariableAsync(name: string, value: Object) : Task = 
+    override _.SetVariableAsync(name: string, value: Object, declaredType: Type) : Task =
```
jonsequitur, line 376: @TIHan This is an interesting and useful change to the `SetVariableAsync` that might should work its way down to FSI. Thoughts?

src/Microsoft.DotNet.Interactive.CSharp/AspNetKernelExtensions.cs
halter73: Not yet, but I would love to add support for that.

halter73:
image

I think this is finally ready to merge! I've updated the PR to minimize the impact on the kernels themselves. This does add an optional declared Type parameter to DotNetKernel.SetVariableAsync(), but it shouldn't cause any problems for people taking dependencies on the Kernels like before when I changed the C# kernel's TFM from netstandard2.1 to net5.0.

src/Microsoft.DotNet.Interactive.AspNetCore/AspNetCoreKernelExtensions.cs
```diff
@@ -0,0 +1,104 @@
...
+                {
+                    ExecutionContext.RestoreFlow();
+                }
+            }, HtmlFormatter.MimeType);
+
+            return kernel;
+        }
+
+        // REVIEW: Should we make this a public CSharp/DotNetKernel extension method?
+        private static Task SetVariableAsync<T>(this CSharpKernel kernel, string name, T value)
```
halter73, line 99: Should we make this public?

src/Microsoft.DotNet.Interactive.AspNetCore/HttpClientFormatter.cs
```diff
@@ -0,0 +1,129 @@
...
+
+            var responseHeaders = details[open: true](summary("Headers"), HeaderTable(responseMessage.Headers, responseMessage.Content.Headers));
+            var responseBody = details[open: true](summary("Body"), pre(responseBodyString));
+
+            var output = div[@class: _containerClass](
+                style[type: "text/css"](_flexCss),
+                div(h2("Request"), hr(), requestLine, requestHeaders, requestBody),
+                div(h2("Response"), hr(), responseLine, responseHeaders, responseBody));
+
+            //Pocket.LoggerExtensions.Info(Pocket.Logger.Log, output.ToString());
```
jonsequitur, line 84: Want to remove this comment?

jonsequitur (approved):

src/Microsoft.DotNet.Interactive.AspNetCore/AspNetCoreKernelExtensions.cs
jonsequitur: I think a public version of this is useful but should be on `DotNetKernel` (or ` T where T : DotNetKernel`) rather than on `CSharpKernel`.

halter73:
⬆️📅 I addressed the PR feedback and added more tests.

jonsequitur (approved):

halter73:
@jonsequitur The PR builds are consistently hanging after printing `Running tests: D:\a\1\s\artifacts\bin\Microsoft.DotNet.Interactive.Tests\Release\net5.0\Microsoft.DotNet.Interactive.Tests.dll [net5.0|x64]` This is the same on Windows and Linux. Locally, when I run `dotnet test -c Release` in `src\dotnet-interactive.Tests` which appears to be the test project behind Microsoft.DotNet.Interactive.Tests.dll, all 120 tests pass in 24 seconds. Any idea what's going on?

jonsequitur:
We've been seeing this issue in our CI builds for a few weeks. It's not related to your PR.

Have you rebased recently?

halter73:
@jonsequitur In my last commit, I added `#!aspnet-stop` so I could call it from AspNetCoreTests.DisposeAsync() and that looks to have fixed the issues with the test project hanging on AzDO. Should I merge this now?

PR #910, opened 2020-11-25
implement interfaces directly in FSharpKernel

brettfo:
dotnet/fsharp#2867 was completed and wound up in .NET 5 so we can now implement the same generic interface multiple times on the same type with different generic arguments so we can remove a wrapper project that's no longer needed.


KevinRansom (approved):
Noice

jonsequitur (approved):

PR #926, opened 2020-12-04
fix #r immediately after language selector

### LINKED ISSUE #921: #r import not working after #!language commands
### Describe the bug
Creating a new notebook via the VSCode Interactive Notebook extension generates a notebook using the C# kernel.
Switching to F# code blocks does not change the kernel specified in metadata, but instead adds a `#!fsharp` command at the top of each block that's not visible in VSCode.

Using `#r` directive to import a NuGet package in a F# code block works in VSCode. When run in a jupyter notebook, such as when sharing with Binder, the import seems to have no effect after the `#!language` commands, resulting in undefined type/namespace errors.

I've created a repository to reproduce the issue:
https://github.com/bolte-17/Dotnet-Interactive-Nuget-Reference-Issue
It's runnable via Binder:
https://mybinder.org/v2/gh/bolte-17/Dotnet-Interactive-Nuget-Reference-Issue/d71f31562d0553f13917267f1e2a8c066bd5ae34

1.0.157409+69b86441f75e5663aa2c1568ae971127de9cd02e

- OS
    - [x] Windows 10
    - [ ] macOS
    - [ ] iOS
    - [ ] Android
 - Browser 
    - [x] Chrome
    - [ ] Edge
    - [ ] Firefox
    - [ ] Safari
 - Frontend
    - [x] Jupyter Notebook
    - [ ] Jupyter Lab
    - [ ] nteract
    - [x] Visual Studio Code
    - [ ] Other (please specify)

brettfo:
When a submission is split into syntax nodes, a language switcher directive, e.g., `#!csharp`, was setting the kernel name to `#!` instead of `csharp`.  The result of this was that when the following was submitted:

```
#!csharp
#r "nuget: SomePackage"
```

The `#r` was inappropriately associated with the kernel `#!` which doesn't exist so it was never flagged as a directive that needed to be handled.  The fix is to set the kernel name appropriately, but then to also not let code submissions get appended to a language switcher node.

~~To enable testing the `DirectiveCommand` type needed to be exposed, but granting `InternalsVisibleTo` the test project opened up a whole other can of worms.  There likely needs to be a discussion around if exposing this type is appropriate.~~

This was also verified manually.

Fixes #921.

### DIFF
--- a/src/Microsoft.DotNet.Interactive.Tests/LanguageKernelPackageTests.cs
+++ b/src/Microsoft.DotNet.Interactive.Tests/LanguageKernelPackageTests.cs
@@ -948,5 +948,31 @@ typeof(System.Device.Gpio.GpioController).Assembly.Location
             }
             // (OSPlatform.OSX is not supported by this library
         }
+
+        [Fact]
+        public async Task Pound_r_nuget_works_immediately_after_a_language_selector()
+        {
+            var kernel = CreateCompositeKernel(defaultKernelLanguage: Language.CSharp);
+
+            var code = @"
+#!csharp
+#r ""nuget: System.Text.Json, 4.6.0""
+";
+
+            var command = new SubmitCode(code);
+
+            var result = await kernel.SendAsync(command);
+
+            using var events = result.KernelEvents.ToSubscribedList();
+
+            events
+                .Should()
+                .ContainSingle<PackageAdded>()
+                .Which
+                .PackageReference
+                .PackageName
+                .Should()
+                .Be("System.Text.Json");
+        }
     }
 }
--- a/src/Microsoft.DotNet.Interactive/Parsing/DirectiveNode.cs
+++ b/src/Microsoft.DotNet.Interactive/Parsing/DirectiveNode.cs
@@ -18,7 +18,7 @@ namespace Microsoft.DotNet.Interactive.Parsing
         internal DirectiveNode(
             DirectiveToken directiveToken,
             SourceText sourceText,
-            PolyglotSyntaxTree? syntaxTree) : base("#!", sourceText, syntaxTree)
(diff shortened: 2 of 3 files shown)


src/Microsoft.DotNet.Interactive.Tests/SubmissionSplittingTests.cs
```diff
@@ -76,5 +76,30 @@ public void RequestDiagnostics_can_be_split_into_separate_commands()
...
+            };
+
+            kernel.DefaultKernelName = "csharp";
+
+            var command = new SubmitCode(code);
+            var commands = kernel.SubmissionParser.SplitSubmission(command);
+
+            commands
+                .Should()
+                .ContainSingle<DirectiveCommand>(d => d.ParseResult?.CommandResult?.Command?.Name == "#!nuget-restore");
```
jonsequitur, line 102: This is really an implementation detail. I wonder if there's a cleaner way to test this.

jonsequitur, line 102: An assertion like this will give better feedback:

```csharp
commands
    .Should()
    .ContainSingle<DirectiveCommand>()
    .Which
    .ParseResult
    .CommandResult
    .Command
    .Name
    .Should()
    .Be("#!nuget-restore");
```

brettfo: I wrote it this way because there are multiple `DirectiveCommand`s returned, but only one should have the specified command name, not just a single `DirectiveCommand`.  Is there a different overload that will give the nice error message?

brettfo: Test was rewritten to be a regular package add.

jonsequitur (approved):

colombod (approved):

PR #929, opened 2020-12-04
disable .NET interactive first run experience

colombod:


azure-pipelines.yml
```diff
@@ -84,8 +84,8 @@ stages:
           inputs:
             versionSpec: $(NodeJSVersion)        
 
-        - task: UseDotNet@2
-          displayName: Add dotnet 3.0
+        - task: UseDotNet@5
```
jonsequitur, line 87: Why not call this something like `AddDotNetSDK` so the naming is more durable?

jonsequitur (approved):

azure-pipelines.yml
colombod: sounds ok

brettfo: The task `UseDotNet@2` should stay the same.  `UseDotNet` is the name of an Azure DevOps thing that we don't control and `@2` is their plugin/action version that we also don't control.  The only thing you can and should change is the `displayName` field.

colombod: done

azure-pipelines.yml
```diff
@@ -196,7 +196,7 @@ stages:
             versionSpec: $(NodeJSVersion)
 
         - task: UseDotNet@2
-          displayName: Add dotnet
+          displayName: Use .NET SDK
```
brettfo, line 199: Some say `Use .NET SDK` and others `Use .NET SDK 5.0`.  I'd prefer simply `Use .NET SDK` since the version will eventually change again, but ultimately I don't care too much.

brettfo (approved):

PR #930, opened 2020-12-04
Re enable some tests 2

jonsequitur:


colombod (approved):

azure-pipelines.yml
```diff
@@ -42,6 +42,7 @@ stages:
           ${{ if and(ne(variables['System.TeamProject'], 'public'), notin(variables['Build.Reason'], 'PullRequest')) }}:
             name: NetCoreInternal-Pool
             queue: buildpool.windows.10.amd64.vs2019
+        timeoutInMinutes: 300
```
brettfo, line 45: I'm going to be really sad if this is what we need.

jonsequitur: It's just a test to distinguish between a deadlock and extreme slowness.

These tests run in 15 minutes or less locally.

PR #949, opened 2020-12-18
SQL kernel updates

pensivebrian:
Various updates to SQL kernel in the ExtensionLabs project:
* Updated the logic to load the SqlToolsService native binary based on the resolved Microsoft.SqlToolsService nuget package.
* Bumped the EF nuget packages to 3.1.8.
* We now explicitly declare the nuget dependencies in the Microsoft.Dotnet.Interactive.ExtensionLabs.csproj.
* Updated the tests that connect to SQL Server to load the connection string from environment variable TEST_MSSQL_CONNECTION_STRING. This will keep connection strings out of the source code, and should prevent accidental credential checkins.
* Misc .cs code changes from Jon's https://github.com/dotnet/interactive/pull/870 branch.

We are not moving to the latest EF 5.X version due to assembly version conflicts with the Humanizer and Microsoft.Extensions.DependencyInjection when the latest versions are pulled in by EF 5.

### DIFF
--- a/src/Microsoft.DotNet.Interactive.ExtensionLab.Tests/MsSqlConnectionTests.cs
+++ b/src/Microsoft.DotNet.Interactive.ExtensionLab.Tests/MsSqlConnectionTests.cs
@@ -82,7 +82,7 @@ SELECT TOP 100 * FROM Person.Person
 
             events.Should().NotContainErrors();
 
-            result = await kernel.SubmitCodeAsync("adventureworks.AddressTypes.Count()");
+            result = await kernel.SubmitCodeAsync("adventureworks.AddressType.Count()");
 
             events = result.KernelEvents.ToSubscribedList();
 
--- a/src/Microsoft.DotNet.Interactive.ExtensionLab.Tests/MsSqlConnectionTests.cs
+++ b/src/Microsoft.DotNet.Interactive.ExtensionLab.Tests/MsSqlConnectionTests.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System;
 using System.Threading.Tasks;
 
 using FluentAssertions;
@@ -19,6 +20,14 @@ namespace Microsoft.DotNet.Interactive.ExtensionLab.Tests
         public MsSqlConnectionTests(ITestOutputHelper output)
         {
             _output = output;
+            
+            // FIX: (MsSqlConnectionTests) 
+            if (!string.IsNullOrWhiteSpace(Environment.GetEnvironmentVariable(MsSqlServiceClient.SqlToolsServiceEnvironmentVariableName)))
+            {
+                Environment.SetEnvironmentVariable(
+                    MsSqlServiceClient.SqlToolsServiceEnvironmentVariableName,
+                    @"/Users/bro/code/sqltoolsservice/src/Microsoft.SqlTools.ServiceLayer/bin/Debug/netcoreapp3.1/osx.10.11-x64/publish/MicrosoftSqlToolsServiceLayer");
+            }
         }
 
         [MsSqlFact("Persist Security Info=False; Integrated Security=true; Initial Catalog=AdventureWorks2019; Server=localhost", Skip = "not ready")]
--- a/src/Microsoft.DotNet.Interactive.ExtensionLab.Tests/MsSqlConnectionTests.cs
+++ b/src/Microsoft.DotNet.Interactive.ExtensionLab.Tests/MsSqlConnectionTests.cs
@@ -26,7 +26,7 @@ namespace Microsoft.DotNet.Interactive.ExtensionLab.Tests
         {
             var csharpKernel = new CSharpKernel().UseNugetDirective();
             await csharpKernel.SubmitCodeAsync(@$"
-#r ""nuget:runtime.osx-x64.native.microsoft.sqltoolsservice,3.0.0-release.52""
+#r ""nuget:microsoft.sqltoolsservice,3.0.0-release.52""
 ");
             
             var kernel = new CompositeKernel
--- a/src/Microsoft.DotNet.Interactive.ExtensionLab.Tests/MsSqlConnectionTests.cs
+++ b/src/Microsoft.DotNet.Interactive.ExtensionLab.Tests/MsSqlConnectionTests.cs
@@ -82,7 +82,7 @@ SELECT TOP 100 * FROM Person.Person
 
             events.Should().NotContainErrors();
 
-            result = await kernel.SubmitCodeAsync("adventureworks.AddressType.Count()");
+            result = await kernel.SubmitCodeAsync("adventureworks.AddressTypes.Count()");
 
             events = result.KernelEvents.ToSubscribedList();
 
--- a/src/Microsoft.DotNet.Interactive.ExtensionLab.Tests/MsSqlConnectionTests.cs
+++ b/src/Microsoft.DotNet.Interactive.ExtensionLab.Tests/MsSqlConnectionTests.cs
@@ -20,29 +20,30 @@ namespace Microsoft.DotNet.Interactive.ExtensionLab.Tests
         public MsSqlConnectionTests(ITestOutputHelper output)
         {
             _output = output;
-            
-            // FIX: (MsSqlConnectionTests) 
-            if (!string.IsNullOrWhiteSpace(Environment.GetEnvironmentVariable(MsSqlServiceClient.SqlToolsServiceEnvironmentVariableName)))
-            {
-                Environment.SetEnvironmentVariable(
-                    MsSqlServiceClient.SqlToolsServiceEnvironmentVariableName,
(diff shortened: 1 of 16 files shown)


src/Microsoft.DotNet.Interactive.ExtensionLab/MsSqlKernelConnection.cs
```diff
@@ -55,8 +90,8 @@ private async Task InitializeDbContextAsync(MsSqlConnectionOptions options, Kern
             context.Display($"Scaffolding a `DbContext` and initializing an instance of it called `{options.KernelName}` in the C# kernel.", "text/markdown");
 
             var submission1 = @$"
-#r ""nuget:Microsoft.EntityFrameworkCore.Design""
-#r ""nuget:Microsoft.EntityFrameworkCore.SqlServer""
+#r ""nuget:Microsoft.EntityFrameworkCore.Design,3.1.8""
+#r ""nuget:Microsoft.EntityFrameworkCore.SqlServer,3.1.8""
```
jonsequitur, line 94: This can move to a .NET 5 version now.

src/Microsoft.DotNet.Interactive.ExtensionLab/MsSqlServiceClient.cs
```diff
@@ -15,11 +15,22 @@ namespace Microsoft.DotNet.Interactive.ExtensionLab
...
+
+        public const string SqlToolsServiceEnvironmentVariableName = "DOTNET_SQLTOOLSSERVICE";
+
+        public MsSqlServiceClient(string serviceExePath = null)
+        {
+            _serviceExePath = serviceExePath;
+
+            if (string.IsNullOrWhiteSpace(_serviceExePath))
+            {
+                _serviceExePath = Environment.GetEnvironmentVariable(SqlToolsServiceEnvironmentVariableName);
```
jonsequitur, line 31: If we're unable to find a value for this, it probably deserves an exception at this point.

src/Microsoft.DotNet.Interactive.ExtensionLab.Tests/MsSqlConnectionTests.cs
```diff
@@ -21,19 +22,28 @@ public MsSqlConnectionTests(ITestOutputHelper output)
...
         }
 
-        [MsSqlFact("Persist Security Info=False; Integrated Security=true; Initial Catalog=AdventureWorks2019; Server=localhost", Skip = "not ready")]
-        public async Task It_can_connect_and_query_data()
+        private async Task<CompositeKernel> CreateKernel()
         {
-            using var kernel = new CompositeKernel
+            var csharpKernel = new CSharpKernel().UseNugetDirective();
+            await csharpKernel.SubmitCodeAsync(@$"
+#r ""nuget:runtime.osx-x64.native.microsoft.sqltoolsservice,3.0.0-release.52""
```
pensivebrian, line 29: @jonsequitur - I've hard coded the SqlToolsService nuget reference for the time being. Have any ideas what is the best way to handle this? I was thinking, based on the running platform, I could try and resolve the platform reference.

src/Microsoft.DotNet.Interactive.ExtensionLab/MsSqlKernelConnection.cs
pensivebrian: Fixed and unit test passes

src/Microsoft.DotNet.Interactive.ExtensionLab.Tests/MsSqlConnectionTests.cs
jonsequitur: You don't need to reference the platform-specific packages directly. Reference Microsoft.SqlToolsService and the appropriate platform-specific package will be installed transitively.

pensivebrian: I see. Thanks for the tip!

src/Microsoft.DotNet.Interactive.ExtensionLab/MsSqlServiceClient.cs
pensivebrian: Fixed by moving the check from when we start the process to the constructor.

src/Microsoft.DotNet.Interactive.ExtensionLab/MsSqlKernelConnection.cs
pensivebrian: I take that back... the scaffold unit test is failing so I'm investigating this.

jonsequitur: Try version 5.0.1.

pensivebrian: 5.0.1 didn't help. Here's the exception I'm seeing from submitting the scaffolding code. 3.1.8 works, but 3.1.9 and above fail.

 ```
System.MissingMethodException: Method not found: 'Microsoft.Extensions.DependencyInjection.ServiceProvider Microsoft.Extensions.DependencyInjection.ServiceCollectionContainerBuilderExtensions.BuildServiceProvider(Microsoft.Extensions.DependencyInjection.IServiceCollection)
```

Comment (pensivebrian) [reply src/Microsoft.DotNet.Interactive.ExtensionLab/MsSqlKernelConnection.cs]: Narrowed this down a bit. The unit tests are loading assemblies from `dotnet-interactive/artifacts/bin/Microsoft.DotNet.Interactive.ExtensionLab.Tests/Debug/net5.0` which contain the wrong Microsoft.Extensions.DependencyInjection.dll 3.1.9 assembly rather than using the nuget store where the other EF 5.0 binaries are loaded.

Comment (pensivebrian) [reply src/Microsoft.DotNet.Interactive.ExtensionLab/MsSqlKernelConnection.cs]: Two package dependencies are causing this (so far):
1. Microsoft.ApplicationInsights.AspNetCore is pulling in all kinds of Microsoft.Extension.* dependencies, including the Microsoft.Extensions.DependencyInjection.
2. ICSharpCode.Decompiler is pulling in an outdated version of Humanizer, which conflicts with the version used by EF 5.0, resulting in the same Missing Method exception.

For 2, splitting the SQL kernel into its own assembly would resolve this long term. I need read up a bit more about how to address 1 in the .net core world. Let me know if you have ideas.

Comment (jonsequitur) [reply src/Microsoft.DotNet.Interactive.ExtensionLab/MsSqlKernelConnection.cs]: Would moving some of these dependencies to newer version help? If we're not yet on Microsoft.Extensions.DependencyInjection 5.0.1, we probably should be.

Comment (pensivebrian) [reply src/Microsoft.DotNet.Interactive.ExtensionLab/MsSqlKernelConnection.cs]: I fixed this by declaring explicit nuget dependencies on EF and the Humanizer.Core packages for the ExtensionLab csproj. The existing code had runtime dependencies on EF. Since that dependency was not declared, the build tooling would binplace the minimum package version detected for other projects. Be making them explicit, the newer package versions are binplaced. Unit EF scaffolding tests now pass after some minor changes to the code injected into the kernel.

File: src/dotnet-interactive/dotnet-interactive.csproj [hunk 4]
```diff
@@ -37,6 +37,12 @@
   </ItemGroup>
 
   <ItemGroup>
+    <PackageReference Include="Humanizer.Core" Version="2.8.26" />
+    <PackageReference Include="Microsoft.EntityFrameworkCore.Design" Version="5.0.1">
+      <PrivateAssets>all</PrivateAssets>
+      <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
+    </PackageReference>
+    <PackageReference Include="Microsoft.EntityFrameworkCore.SqlServer" Version="5.0.1" />
```
jonsequitur, line 45: We should not add these dependencies here. Let's see if we can figure out another approach.

src/Microsoft.DotNet.Interactive.ExtensionLab/MsSqlKernelConnection.cs
```diff
@@ -55,8 +90,8 @@ private async Task InitializeDbContextAsync(MsSqlConnectionOptions options, Kern
             context.Display($"Scaffolding a `DbContext` and initializing an instance of it called `{options.KernelName}` in the C# kernel.", "text/markdown");
 
             var submission1 = @$"
-#r ""nuget:Microsoft.EntityFrameworkCore.Design""
-#r ""nuget:Microsoft.EntityFrameworkCore.SqlServer""
+#r ""nuget:Microsoft.EntityFrameworkCore.Design,3.1.8""
```
colombod, line 93: could we use 3.1.8 as default and have this version part of the #!connect option?

pensivebrian: That's a good idea. There will likely be other EF options to consider such as an option to pluralize table names. 

Currently, any EF version greater than 3.1.8 won't work due to assumptions on the generated .cs scaffolding code and assembly version conflicts, so we hard code this for now and address the EF issues in another PR.

src/dotnet-interactive/dotnet-interactive.csproj
pensivebrian: I removed them from this project.

jonsequitur (approved):

Issue #856, opened 2020-10-29
Bidirectional communication between kernel and notebook UI


idg10:
We want to be able to create 'live' UI elements in a notebook that can be updated by ongoing activity elsewhere in the notebook. Today, we can produce HTML output and client-side JavaScript to enable interactivity, but there is currently no built-in mechanism to connect this in any way to code running in the kernel.

To support this, we would like it to be possible to open channels of communication by which code running inside the kernel (either code cells, or extensions or other components loaded by those cells) is able to send messages to and receive messages from JavaScript (or TypeScript) code running in client hosting the notebook.

It's not completely impossible to achieve this today, but it's messy and limited. We have built a prototype that provides limited support for such connectivity. It works by running a simple web server inside the kernel process, and then connecting to this from JavaScript code on the client side. UI to kernel communication can be achieved with `fetch` operations from the client, and we used SignalR to enable the kernel to send messages to the client code. We have been able to demonstrate the kind of 'live' notebook cells we have in mind. However, there are a number of issues with this technique.

For example, currently it only works when the kernel process is local; we have not yet made it work for remote scenarios. We have only attempted this in Visual Studio Code, and not yet other hosts. Nor have we attempted to get it working in CodeSpaces yet. In principle, these are all just challenges which could be overcome with sufficient engineering effort. However, it is likely to be fragile, because there may be scenarios in which our chosen communication mechanism might not be open to us. In any case, it duplicates work already being done by .NET Interactive.

.NET Interactive already maintains a channel enabling communication between the UI process and the kernel process. (As it happens, it even uses SignalR today.) Instead of building a parallel mechanism to this, it would be far more robust and reliable to be able to piggy back on the channel that .NET Interactive has necessarily already established.

We are requesting that a message passing service be layered on top of this.

In the kernel, the existing `Microsoft.DotNet.Interactive` API's `Kernel` type would offer a method (say, `GetChannel`) enabling an object representing a named channel to be acquired. It might look something like this:

```csharp
public interface IKernelChannel
{
    IObservable<KernelChannelMessage> Messages { get; }
    // TODO: should this be async, enabling callers to know the message has been delivered?
    void SendMessage(KernelChannelMessage message);
}
```

A similar facility would exist for client-side code. JavaScript in the notebook would be able to obtain a very similar sort of service. The `KernelClient` interface would offer a method (say, `getChannel`) which would return an object providing the same two services: an observable source of messages, and a method by which a message can be sent.

[...]

PR #977, opened 2021-01-12
Client-side custom commands

idg10:
This arises from the discussion at #856 which it addresses by way of #895

This change introduces:
 * a client-side kernel that can run inside a notebook
 * a backchannel for the existing SignalR transport, enabling commands and events to flow in both directions
 * support for defining custom commands, and handling them either in .NET or in client-side JS


src/Microsoft.DotNet.Interactive/Connection/IndiscriminateProxyKernel.cs
```diff
@@ -0,0 +1,44 @@
...
+using Microsoft.DotNet.Interactive.Commands;
+using Microsoft.DotNet.Interactive.Events;
+
+namespace Microsoft.DotNet.Interactive.Connection
+{
+    /// <summary>
+    /// Passes all commands on without any preconceived idea of whether it will be handled.
+    /// </summary>
+    /// <remarks>
+    /// <see cref="ProxyKernel"/> only supports SubmitCode, RequestCompletions, RequestDiagnostics,
```
jonsequitur, line 16: I'm wondering if this separate class is needed. `ProxyKernel` should ideally proxy all command types.

src/Microsoft.DotNet.Interactive/Kernel.cs
```diff
@@ -215,6 +218,19 @@ public FrontendEnvironment FrontendEnvironment
...
         public void AddDirective(Command command) => SubmissionParser.AddDirective(command);
 
+        public virtual void RegisterCommandHandler<TCommand>(string commandType, Func<TCommand, KernelInvocationContext, Task> handler)
+            where TCommand : KernelCommand
+        {
+            RegisterCommandType<TCommand>(commandType);
+            _dynamicHandlers[typeof(TCommand)] = (command, context) => handler((TCommand)command, context);
+        }
+
+        public void RegisterCommandType<TCommand>(string commandType)
```
jonsequitur, line 228: Should we derive the command name from the type for consistency?

src/Microsoft.DotNet.Interactive/Kernel.cs
```diff
@@ -215,6 +218,19 @@ public FrontendEnvironment FrontendEnvironment
 
         public void AddDirective(Command command) => SubmissionParser.AddDirective(command);
 
+        public virtual void RegisterCommandHandler<TCommand>(string commandType, Func<TCommand, KernelInvocationContext, Task> handler)
```
jonsequitur, line 221: Does this need to be `virtual`?

src/Microsoft.DotNet.Interactive/Connection/IndiscriminateProxyKernel.cs
idg10: It wasn't obvious to me whether anything depended on its current behaviour of deciding not to try to handle commands it didn't have prior knowledge of.

If that selectivity is a feature other users of `ProxyKernel` depend on, would it be better to implement this as a filtering mechanism? E.g., `ProxyKernel` could take a delegate that inspects a command and returns a `bool` indicating whether the command should be sent, at which point we could just use a delegate that always returns `true` to get the "proxy everything" behaviour.

So, do we need that selectivity elsewhere?

(Also, in the back of my mind I wasn't sure if we might end up wanting a proxy to have a clear idea of which commands the remote would be able to handle, in which case this "indiscriminate" proxy might be a temporary feature, eventually replace by something that has some knowledge of which command types the client kernel has handlers for.)

src/Microsoft.DotNet.Interactive/Kernel.cs
idg10: Good point. Done.

src/Microsoft.DotNet.Interactive/Kernel.cs
idg10: No. At the time, I was thinking that different kernels might want to handle this differently, but it was a vague thought. Since I don't have a clear idea of what an `override` of this might actually do, I've pushed a commit making it non-`virtual`.

src/Microsoft.DotNet.Interactive/Connection/IndiscriminateProxyKernel.cs
jonsequitur: > eventually replace by something that has some knowledge of which command types the client kernel has handlers for

I think it makes perfect sense to allow the remote kernel to decide which messages it can receive. So I think the existing proxy kernel should be indiscriminate at this point. The next step would be to add support for a command that performs capability discovery, including which commands, magic commands, and languages are supported by a kernel.

src/Microsoft.DotNet.Interactive.Js/src/dotnet-interactive/client-side-kernel.ts
```diff
@@ -0,0 +1,99 @@
...
+                if (contextEventsSubscription) {
+                    contextEventsSubscription.dispose();
+                }
+
+                resolvePromise();
+            })();
+            return promise;
+
+        } else {
+            // TODO: what's the right behaviour if there is no handler?
```
colombod, line 51: I think for the moment is fine to reqject the promise. In the .NET counterpart a handler is always set

src/Microsoft.DotNet.Interactive.Js/src/dotnet-interactive/client-side-kernel.ts
```diff
@@ -0,0 +1,99 @@
...
+export class ClientSideKernel implements Kernel {
+    private _commandHandlers: { [commandType: string]: (envelope: KernelCommandEnvelope, context: KernelInvocationContext) => Promise<void> } = {};
+    private readonly _eventSubscribers: { [token: string]: KernelEventEnvelopeObserver} = {};
+    private readonly _tokenGenerator: TokenGenerator = new TokenGenerator();
+
+    // Is it worth us going to efforts to ensure that the Promise returned here accurately reflects
+    // the command's progress? The only thing that actually calls this is the kernel transport, through
+    // the callback set up by attachKernelToTransport, and the callback is expected to return void, so
+    // nothing is ever going to look at the promise we return here.
+    send(command: KernelCommandEnvelope): Promise<void> {
```
colombod, line 18: why do we use KernelEnvelopes at this level? in the .NET counterpart we use the envelopes only we we get close to the signalr, stdio and namedpipe transport, the envelopes are useful at the boundaries to be used as filters using the tokens

src/Microsoft.DotNet.Interactive.Js/src/dotnet-interactive/dotnet-interactive-interfaces.ts
```diff
@@ -20,8 +20,24 @@ export interface KernelClient {
     submitCommand(commandType: string, command?: any): Promise<string>;
 }
 
+export interface KernelInvocationContext extends Disposable {
+    subscribeToKernelEvents(observer: KernelEventEnvelopeObserver): DisposableSubscription;
+    complete(commandEnvelope: KernelCommandEnvelope): void;
```
colombod, line 25: this interface should handle KernelCommands and KernelEvents without exposing the envelope types

src/Microsoft.DotNet.Interactive.Js/tests/client-side-kernel-invocation-context.spec.ts
```diff
@@ -0,0 +1,305 @@
...
+
+        let toDispose: Disposable[] = [];
+        function use<T extends Disposable>(disposable: T): T {
+            toDispose.push(disposable);
+            return disposable;
+        }
+        afterEach(() => toDispose.forEach(d => d.dispose()));
+
+        it("publishes CommandHandled when Complete is called", async () => {
+            let context = use(ClientSideKernelInvocationContext.establish(commandEnvelope));
```
colombod, line 39: Same observation here, the public api deals in KernelEvents and KernelCommands, the envelopes should come into place near the transport.

src/Microsoft.DotNet.Interactive.Js/src/dotnet-interactive/client-side-kernel.ts
```diff
@@ -0,0 +1,99 @@
+// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+import { ClientSideKernelInvocationContext } from "./client-side-kernel-invocation-context";
+import { DisposableSubscription, KernelEventEnvelopeObserver, KernelCommandEnvelope, KernelTransport, Disposable, KernelEventEnvelope } from "./contracts";
+import { Kernel, KernelInvocationContext } from "./dotnet-interactive-interfaces";
+import { TokenGenerator } from "./tokenGenerator";
+
+export class ClientSideKernel implements Kernel {
+    private _commandHandlers: { [commandType: string]: (envelope: KernelCommandEnvelope, context: KernelInvocationContext) => Promise<void> } = {};
```
colombod, line 10: a command handler signature could should just deal with kernelcommands

jonsequitur: Are you referring to the envelope versus command, @colombod?

src/Microsoft.DotNet.Interactive.Js/src/dotnet-interactive/client-side-kernel.ts
jonsequitur: If no handler is set in the .NET implementation, I believe it will always coincide with a `NoSuitableKernelException`.

src/Microsoft.DotNet.Interactive/Connection/IndiscriminateProxyKernel.cs
idg10: OK, I've pushed an update that modifies the existing `ProxyKernel` to work in exactly the same way that this new `IndiscriminateProxyKernel` did, and I've now removed `IndiscriminateProxyKernel`.

src/Microsoft.DotNet.Interactive.Js/src/dotnet-interactive/client-side-kernel.ts
idg10: Because we need to know what particular command type is being sent, and JavaScript's runtime typing doesn't work like .NET's.

In .NET, you can have a reference with a static type of `KernelCommand` where the object referred to is of some more specific type. But in JS, these things are just essentially dictionaries. I might have declared a variable in TS as being of type `RequestHoverText`, but that type is essentially erased by the compiler, and any object I assign into that variable will contain no clues to tell me whether it's a `RequestHoverText` or a `RequestCompletions`, or any other command type that is structurally identical.

So if you want to know what sort of command you have, you need to carry some sort of tag, and since `KernelCommandEnvelope` already does that, I just used that instead of inventing a new type.

I did wonder about using TypeScript's ability to discriminate type based on properties having certain known values, but at that point I would have invented something that was almost indistinguishable from `KernelCommandEnvelope`.

At one point, in all the places like this where I needed to pass a command indicating its type, I had pairs of arguments: `command: KernelCommand, commandType: KernelCommandType`, but `KernelCommandEnvelope` is a more convenient way of doing that.

However, I can see an argument for introducing some new type that is a command-type-tagged kernel command that is not a `KernelCommandEnvelope` because I understand that the original intent behind `KernelCommandEnvelope` is that it's a serialization representation. But it seemed to me that unless I took the radical step of changing how commands are represented throughout the TypeScript part of this codebase so that everything uses this proposed tagged representation, that I'd just be introducing inconsistency.

Perhaps this more radical approach is the right one, though. (Or possibly there's a simpler approach I haven't thought of.)

colombod (approved):

PR #982, opened 2021-01-14
use cell metadata to track language in ipynb metadata in vscode

brettfo:
To make us work better with the Jupyter extension to VS Code, this enables our VS Code kernel to sniff the cell metadata for a language.  To do this, then, our `.ipynb` handler needs to emit this metadata when saving from our end.  The metadata looks like this:

``` json
...
"metadata": {
  "dotnet_interactive": {
    "language": "fsharp"
  }
}
...
```

I've also updated our own Jupyter protocol handler to sniff out this cell metadata and apply it whenever possible.

There was a refactor of our VS Code extension to always activate whenever a notebook is loaded so we can register our kernel.  This meant that our auto-install of .NET and our global tool was moved to when we initialize a notebook.

When a VS Code notebook is switched to our kernel, we iterate through all cells and check for metadata that will help us assign a language.

A follow-up PR will remove our VS Code `.ipynb` handling entirely, but that first requires microsoft/vscode-jupyter#4363 to be addressed, and part of that included adding some metadata to our kernel spec files so they can identify our VS Code version of the kernel:

``` json
...
"metadata": {
  "vscode": {
    "extension_id": "ms-dotnettools.dotnet-interactive-vscode",
    "kernel_id": "dotnet-interactive"
  }
}
...
```

This does complicate the issue of Jupyter Lab.  When we're executing a cell there, the cell metadata is passed back to us and we can use it, but the user currently has no way of viewing/updating the cell metadata.  This will likely require us building/publishing a Jupyter Lab extension specifically for us.  The good news is that we'll likely only need the single `"language"` field, so there should be very little churn on this front.

### DIFF
--- a/src/Microsoft.DotNet.Interactive.Jupyter.Tests/CompleteRequestHandlerTests.cs
+++ b/src/Microsoft.DotNet.Interactive.Jupyter.Tests/CompleteRequestHandlerTests.cs
@@ -1,10 +1,13 @@
 ﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System.Collections.Generic;
 using System.Threading.Tasks;
 using FluentAssertions;
 using FluentAssertions.Extensions;
 using Microsoft.DotNet.Interactive.Jupyter.Protocol;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
 using Recipes;
 using Xunit;
 using Xunit.Abstractions;
@@ -32,6 +35,32 @@ namespace Microsoft.DotNet.Interactive.Jupyter.Tests
             JupyterMessageSender.ReplyMessages
                 .Should()
                 .ContainSingle(r => r is CompleteReply);
+        }
+
+        [Fact]
+        public void cell_language_can_be_pulled_from_metadata_when_present()
+        {
+            var metaData = new Dictionary<string, object>()
+            {
+                { "dotnet_interactive", JObject.Parse(JsonConvert.SerializeObject(new { language = "fsharp" })) }
+            };
+            var request = ZeroMQMessage.Create(new CompleteRequest("1+1"), metaData: metaData);
+            var context = new JupyterRequestContext(JupyterMessageSender, request);
+            var language = context.GetLanguage();
+            language
+                .Should()
+                .Be("fsharp");
+        }
+
+        [Fact]
+        public void cell_language_defaults_to_null_when_it_cant_be_found()
+        {
+            var request = ZeroMQMessage.Create(new CompleteRequest("1+1"));
+            var context = new JupyterRequestContext(JupyterMessageSender, request);
+            var language = context.GetLanguage();
+            language
+                .Should()
+                .BeNull();
         }
     }
 }
--- a/src/Microsoft.DotNet.Interactive.Jupyter.Tests/CompleteRequestHandlerTests.cs
+++ b/src/Microsoft.DotNet.Interactive.Jupyter.Tests/CompleteRequestHandlerTests.cs
@@ -6,8 +6,7 @@ using System.Threading.Tasks;
 using FluentAssertions;
 using FluentAssertions.Extensions;
 using Microsoft.DotNet.Interactive.Jupyter.Protocol;
-using Newtonsoft.Json;
-using Newtonsoft.Json.Linq;
+using Microsoft.DotNet.Interactive.Notebook;
 using Recipes;
 using Xunit;
 using Xunit.Abstractions;
@@ -42,7 +41,7 @@ namespace Microsoft.DotNet.Interactive.Jupyter.Tests
         {
             var metaData = new Dictionary<string, object>()
             {
-                { "dotnet_interactive", JObject.Parse(JsonConvert.SerializeObject(new { language = "fsharp" })) }
+                { "dotnet_interactive", new InputCellMetadata { Language = "fsharp" } }
             };
             var request = ZeroMQMessage.Create(new CompleteRequest("1+1"), metaData: metaData);
             var context = new JupyterRequestContext(JupyterMessageSender, request);
--- a/src/Microsoft.DotNet.Interactive.Tests/Notebook/JupyterNotebookDocumentFileFormatTests.cs
+++ b/src/Microsoft.DotNet.Interactive.Tests/Notebook/JupyterNotebookDocumentFileFormatTests.cs
@@ -113,6 +113,89 @@ namespace Microsoft.DotNet.Interactive.Tests.Notebook
         }
 
         [Fact]
+        public void cell_metadata_can_specify_language_that_overrides_notebook()
+        {
+            var jupyter = new
+            {
+                cells = new object[]
+                {
+                    new
+                    {
+                        cell_type = "code",
+                        execution_count = 1,
+                        metadata = new
+                        {
+                            dotnet_interactive = new
+                            {
+                                language = "fsharp"
+                            }
+                        },
+                        source = "// this should be F#"
+                    }
+                },
+                metadata = new
+                {
+                    kernelspec = new
+                    {
+                        display_name = $".NET (C#)",
+                        language = "C#",
+                        name = $".net-csharp"
+                    },
+                    language_info = new
+                    {
+                        file_extension = ".cs",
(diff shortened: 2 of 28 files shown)


src/Microsoft.DotNet.Interactive.Jupyter.Tests/CompleteRequestHandlerTests.cs
```diff
@@ -33,5 +36,31 @@ public async Task send_completeReply_on_CompleteRequest()
                 .Should()
                 .ContainSingle(r => r is CompleteReply);
         }
+
+        [Fact]
+        public void cell_language_can_be_pulled_from_metadata_when_present()
+        {
+            var metaData = new Dictionary<string, object>()
+            {
+                { "dotnet_interactive", JObject.Parse(JsonConvert.SerializeObject(new { language = "fsharp" })) }
```
jonsequitur, line 45: Can we do this with the appropriate message type rather than a `JObject`?

src/Microsoft.DotNet.Interactive.Tests/Notebook/JupyterNotebookDocumentFileFormatTests.cs
```diff
@@ -112,6 +112,89 @@ public void missing_metadata_defaults_to_csharp_kernel()
                 });
         }
 
+        [Fact]
+        public void cell_metadata_can_specify_language_that_overrides_notebook()
+        {
+            var jupyter = new
```
jonsequitur, line 118: Why are we using anonymous types for this sort of thing?

src/dotnet-interactive-vscode/src/tests/unit/ipynbMetadata.test.ts
```diff
@@ -0,0 +1,363 @@
+// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+import { expect } from 'chai';
+import { DotNetCellMetadata, getCellLanguage, getDotNetMetadata, getLanguageInfoMetadata, LanguageInfoMetadata, withDotNetMetadata } from '../../ipynbUtilities';
+
+describe('ipynb metadata tests', () => {
+
+    //------------------------------------------------------- document metadata
```
jonsequitur, line 9: We can do nested `describe`s instead of comments for categorization, which will help make test output clearer.

src/Microsoft.DotNet.Interactive.Jupyter.Tests/CompleteRequestHandlerTests.cs
brettfo: It comes back from the parser as `JObject` by default, and given that we may change this in the future, we'll be the most flexible if we stay this way.  @colombod, what's your take on it?

src/Microsoft.DotNet.Interactive.Tests/Notebook/JupyterNotebookDocumentFileFormatTests.cs
brettfo: This structure is really json, but I didn't want to deal with all of the quote escaping so I did it this way when I first added the notebook parser a few months ago.

src/dotnet-interactive-vscode/src/tests/unit/ipynbMetadata.test.ts
brettfo: done.

src/Microsoft.DotNet.Interactive.Jupyter.Tests/CompleteRequestHandlerTests.cs
jonsequitur: This is part of why it stood out for me. Since `JObject` is an implementation detail and we plan to move to `System.Text.Json`, all the more reason to keep serializer types out of the tests.

colombod: then metadata is definitely a Dictionary<string,object> and the values can be as deep as necessary.

src/Microsoft.DotNet.Interactive.Tests/Notebook/JupyterNotebookDocumentFileFormatTests.cs
```diff
@@ -549,7 +687,13 @@ public void serialized_code_cells_have_appropriate_shape()
                     {
                         cell_type = "code",
                         execution_count = 1,
-                        metadata = new { },
+                        metadata = new
+                        {
+                            dotnet_interactive = new
```
colombod, line 692: We should represent this metadata with an object, we own this object so we can be more strict

src/dotnet-interactive-vscode/src/ipynbUtilities.ts
```diff
@@ -0,0 +1,104 @@
...
+    return arg
+        && typeof arg.language === 'string';
+}
+
+export function getDotNetMetadata(metadata: CellMetadata | undefined): DotNetCellMetadata {
+    if (metadata &&
+        metadata.custom &&
+        metadata.custom.metadata &&
+        metadata.custom.metadata.dotnet_interactive &&
+        isDotNetCellMetadata(metadata.custom.metadata.dotnet_interactive)) {
```
colombod, line 25: can't we use `isDotNetCellMetadata(metadata?.custom?.metadata?.dotnet_interactive)`

src/Microsoft.DotNet.Interactive.Jupyter.Tests/CompleteRequestHandlerTests.cs
colombod: something like
image
image

src/Microsoft.DotNet.Interactive.Tests/Notebook/JupyterNotebookDocumentFileFormatTests.cs
brettfo: I'd argue that this file _shouldn't_ use strict types, because we're testing that we can deserialize JSON of a known shape, and I don't want the test to be "serialize-then-deserialize" because that doesn't properly cover the scenario of "deserialize-from-well-known-shape".  I just went with anonymous types because I didn't want to deal with strings with JSON and all the escaping.

src/dotnet-interactive-vscode/src/ipynbUtilities.ts
brettfo: That works for the check, but it makes the assignment inside the `if`-block more difficult.  By doing it this way the TypeScript compiler lets me simply say `return metadata.custom.metadata.dotnet_interactive;` instead of `return <DotNetCellMetadata>metadata?.custom?.metadata?.dotnet_interactive;`.

src/Microsoft.DotNet.Interactive.Tests/Notebook/JupyterNotebookDocumentFileFormatTests.cs
jonsequitur: I'm not suggesting using a string. Using a defined type instead of an anonymous type has advantages for refactoring, discoverability, test readability, and overall maintenance.

src/Microsoft.DotNet.Interactive.Jupyter.Tests/CompleteRequestHandlerTests.cs
brettfo: done

src/Microsoft.DotNet.Interactive.Tests/Notebook/JupyterNotebookDocumentFileFormatTests.cs
brettfo: Discussed this privately.

src/Microsoft.DotNet.Interactive.Tests/Notebook/JupyterNotebookDocumentFileFormatTests.cs
brettfo: I discussed this with @colombod and he's in agreement that anonymous types are fine here because we really don't want to be using strong typing because that misses the point of the input to the function we're actually testing being able to have many different shapes (e.g., a cell's source can either be a single string or an array of strings, and we can't model that with strong types, but we have to be able to test both.)

colombod (approved):

PR #1001, opened 2021-01-22
use cell's first line to set language if possible

brettfo:
When loading an `.ipynb` notebook, peek at the first line to see if a language is set, e.g., `#!pwsh`.  If so, use that as the cell's language, otherwise fall back to the previous methods.

### DIFF
--- a/src/dotnet-interactive-vscode/src/tests/unit/ipynbMetadata.test.ts
+++ b/src/dotnet-interactive-vscode/src/tests/unit/ipynbMetadata.test.ts
@@ -329,7 +329,7 @@ describe('ipynb metadata tests', () => {
     });
 
     describe('cell language selection', () => {
-        it(`cell language is first determined from cell text`, () => {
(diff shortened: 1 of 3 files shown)


src/dotnet-interactive-vscode/src/tests/unit/ipynbMetadata.test.ts
```diff
@@ -329,25 +329,51 @@ describe('ipynb metadata tests', () => {
     });
 
     describe('cell language selection', () => {
-        it(`cell language is first determined from cell metadata`, () => {
+        it(`cell language is first determined from cell text`, () => {
```
jonsequitur, line 332: The passive voice is a little funny alongside the "it..." style of the mocha tests. Alternatively...

```suggestion
        it(`sets the cell language based on the cell text`, () => {
```

jonsequitur (approved):

PR #1027, opened 2021-02-01
add signature help

### LINKED ISSUE #728: Add signature help to C# kernel
This will be implemented as `CSharpKernel : IKernelCommandHandler<SignatureHelp>`.

brettfo:
Add commands and events for signature help and implement it for the C# kernel (#728).  I borrowed the OmniSharp implementation for C# signature help and I kept all of the implementation in a single directory so it's easy to move around later if need be.

image

Follow up items:

- Generate better markdown for signature help instead of just plain text.  I don't want to block this PR on markdown formatting discussions, because once this is in, the markdown can be easily changed and it'll automatically flow through.
- Signature help doesn't display doc comments for symbols from metadata (as opposed to symbols from source), and I suspect it's due to how we add metadata references, because we're calling the same method from Roslyn to get this information.  Hover text has a similar problem.  My guess is that once we fix this for one feature, it'll work for the other.

Fixes #728.


colombod (approved):

colombod (commented):
ci is now broken with docker build due to a dependency 
```
Step 18/33 : RUN dotnet interactive jupyter install --http-port-range ${HTTP_PORT_RANGE}
 ---> Running in 56c27beb519b
It was not possible to find any compatible framework version
The framework 'Microsoft.AspNetCore.App', version '3.1.0' was not found.
  - The following frameworks were found:
      5.0.2 at [/usr/share/dotnet/shared/Microsoft.AspNetCore.App]

You can resolve the problem by installing the specified framework and/or SDK.
```

colombod (approved):

src/Microsoft.DotNet.Interactive.CSharp/CSharpKernel.cs
```diff
@@ -172,6 +174,16 @@ public async Task HandleAsync(RequestHoverText command, KernelInvocationContext
             
         }
 
+        public async Task HandleAsync(RequestSignatureHelp command, KernelInvocationContext context)
+        {
+            var document = _workspace.UpdateWorkingDocument(command.Code);
+            var signatureHelp = await SignatureHelpGenerator.GenerateSignatureInformation(document, command);
+            if (signatureHelp != null)
```
jonsequitur, line 181:
```suggestion
            if (signatureHelp is { })
```

src/Microsoft.DotNet.Interactive.CSharp/SignatureHelp/SignatureHelpGenerator.cs
```diff
@@ -0,0 +1,163 @@
...
+using Microsoft.DotNet.Interactive.Events;
+
+namespace Microsoft.DotNet.Interactive.CSharp.SignatureHelp
+{
+    internal static class SignatureHelpGenerator
+    {
+        public static async Task<SignatureHelpProduced> GenerateSignatureInformation(Document document, RequestSignatureHelp command)
+        {
+            var invocation = await GetInvocation(document, command.LinePosition);
+            if (invocation == null)
```
jonsequitur, line 20:
```suggestion
            if (invocation is null)
```

src/Microsoft.DotNet.Interactive.CSharp/SignatureHelp/SignatureHelpGenerator.cs
```diff
@@ -0,0 +1,163 @@
...
+            return score;
+        }
+
+        private static SignatureInformation BuildSignature(IMethodSymbol symbol)
+        {
+            var parameters = symbol.Parameters
+                .Select(parameter =>
+                    new ParameterInformation(
+                        label: parameter.Name,
+                        documentation: new FormattedValue("text/markdown", DocumentationConverter.ConvertDocumentation(parameter.GetDocumentationCommentXml(expandIncludes: true)))));
```
jonsequitur, line 153: We talked about deferring this to the `Formatter` so it can be customized more easily and consistently.

src/Microsoft.DotNet.Interactive.CSharp/SignatureHelp/SignatureHelpGenerator.cs
```diff
@@ -0,0 +1,163 @@
...
+
+            var signatureInformation = new SignatureInformation(
+                label: symbol.ToDisplayString(SymbolDisplayFormat.MinimallyQualifiedFormat),
+                documentation: new FormattedValue("text/markdown", DocumentationConverter.ConvertDocumentation(symbol.GetDocumentationCommentXml(expandIncludes: true))),
```
jonsequitur, line 157: Ditto.

src/Microsoft.DotNet.Interactive/Events/SignatureHelpProduced.cs
```diff
@@ -0,0 +1,54 @@
...
+            Signatures = signatures;
+            ActiveSignature = activeSignature;
+            ActiveParameter = activeParameter;
+        }
+
+        public static SignatureHelpProduced Empty(RequestSignatureHelp command) => new SignatureHelpProduced(command, null, 0, 0);
+
+        public IReadOnlyList<SignatureInformation> Signatures { get; }
+
+        public int ActiveSignature { get; }
```
jonsequitur, line 50: Is this an index into the `Signatures` property? Can we give it a clearer name?

src/Microsoft.DotNet.Interactive/Events/SignatureHelpProduced.cs
```diff
@@ -0,0 +1,54 @@
...
+
+        public int ActiveParameter { get; }
```
jonsequitur, line 52: This one is also a little unclear.

src/Microsoft.DotNet.Interactive.CSharp/SignatureHelp/SignatureHelpGenerator.cs
brettfo: I think this is a future work item.  The type returned from the Roslyn API is `string` which _might_ be valid XML, and the type eventually returned by F# will be something different.  Eventually we'll need to add `Formatter.Register<SignatureHelpDocumentation>(...)`, but we don't yet have a common type for `SignatureHelpDocumentation` that encompasses everything that might be present in all kinds of sighelp.

jonsequitur: Got it.

Issue #983, opened 2021-01-14
SQL extension fails executing query "select 1; select 1, 2"


pensivebrian:
A SQL query with multiple batches returning different column counts fails with the below exception against SQL Server. The following query reproduces the issue: `select 1; select 1, 2`

```
Error: System.ArgumentException: Can not add property (No column name) to Newtonsoft.Json.Linq.JObject. Property with the same name already exists on object.
at Newtonsoft.Json.Linq.JObject.ValidateToken(JToken o, JToken existing)
at Newtonsoft.Json.Linq.JContainer.InsertItem(Int32 index, JToken item, Boolean skipParentCheck)
at Microsoft.DotNet.Interactive.Formatting.TabularDataFormatter.Generate[T](IEnumerable`1 source) in /Users/bro/code/dotnet-interactive/src/Microsoft.DotNet.Interactive.Formatting/TabularDataFormatter.cs:line 46
at Microsoft.DotNet.Interactive.Formatting.TabularDataFormatter.ToTabularJsonString[T](IEnumerable`1 source) in /Users/bro/code/dotnet-interactive/src/Microsoft.DotNet.Interactive.Formatting/TabularDataFormatter.cs:line 21
at Microsoft.DotNet.Interactive.ExtensionLab.SqlKernelsExtension.<>c.<OnLoadAsync>b__0_0(IEnumerable`1 source, TextWriter writer)
at Microsoft.DotNet.Interactive.Formatting.Formatter.<>c__DisplayClass49_0`1.<Register>b__0(FormatContext context, Object value, TextWriter writer) in /Users/bro/code/dotnet-interactive/src/Microsoft.DotNet.Interactive.Formatting/Formatter.cs:line 448
at Microsoft.DotNet.Interactive.Formatting.AnonymousTypeFormatter`1.Format(FormatContext context, T instance, TextWriter writer) in /Users/bro/code/dotnet-interactive/src/Microsoft.DotNet.Interactive.Formatting/AnonymousTypeFormatter{T}.cs:line 31
at Microsoft.DotNet.Interactive.Formatting.TypeFormatter`1.Microsoft.DotNet.Interactive.Formatting.ITypeFormatter.Format(FormatContext context, Object instance, TextWriter writer) in /Users/bro/code/dotnet-interactive/src/Microsoft.DotNet.Interactive.Formatting/TypeFormatter{T}.cs:line 24
[...]

PR #1036, opened 2021-02-04
Handle duplicate column names in MSSQL and SQLite query results.

corivera:
Fixes https://github.com/dotnet/interactive/issues/983

Table Schemas require column names to be unique, but SQL query results can have duplicate columns names due to aliasing or being unnamed. This change solves that problem by appending a number to the end of a duplicate column name, similar to what would happen if you download duplicate files in a web browser.

### DIFF
--- a/src/Microsoft.DotNet.Interactive.ExtensionLab.Tests/SqlKernelsExtensionTests.cs
+++ b/src/Microsoft.DotNet.Interactive.ExtensionLab.Tests/SqlKernelsExtensionTests.cs
@@ -90,17 +90,6 @@ SELECT * FROM fruit
             } while (reader.NextResult());
         }
 
-        [Fact]
-        public void can_alias_duplicate_column_names_in_sql_results()
-        {
-            var testNames = new string[] { "apple", "banana", "apple", "orange", "orange", "BANANA", "orange", "BaNaNa" };
-            var expectedAliases = new string[] { "apple", "banana", "apple (1)", "orange", "orange (1)", "BANANA", "orange (2)", "BaNaNa" };
-
-            SqlKernelUtils.AliasDuplicateColumnNames(testNames);
-
-            Assert.Equal(expectedAliases, testNames);
-        }
-
         public void Dispose()
         {
             Formatter.ResetToDefault();
--- a/src/Microsoft.DotNet.Interactive.ExtensionLab.Tests/SqlKernelsExtensionTests.cs
+++ b/src/Microsoft.DotNet.Interactive.ExtensionLab.Tests/SqlKernelsExtensionTests.cs
@@ -29,7 +29,6 @@ namespace Microsoft.DotNet.Interactive.ExtensionLab.Tests
                 .UsingExtension("json");
         }
 
-
         [Fact]
         public async Task can_generate_tabular_json_from_database_table_result()
         {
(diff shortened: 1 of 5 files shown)


src/Microsoft.DotNet.Interactive.ExtensionLab.Tests/SqlKernelsExtensionTests.cs
```diff
@@ -90,6 +90,17 @@ private static IEnumerable<IEnumerable<IEnumerable<(string name, object value)>>
...
 
+        [Fact]
+        public void can_alias_duplicate_column_names_in_sql_results()
+        {
+            var testNames = new string[] { "apple", "banana", "apple", "orange", "orange", "BANANA", "orange", "BaNaNa" };
+            var expectedAliases = new string[] { "apple", "banana", "apple (1)", "orange", "orange (1)", "BANANA", "orange (2)", "BaNaNa" };
+
+            SqlKernelUtils.AliasDuplicateColumnNames(testNames);
+
+            Assert.Equal(expectedAliases, testNames);
```
jonsequitur, line 101: Please use FluentAssertions, e.g. `expectedAliases.Should().BeEquivalentTo(testNames);`

src/Microsoft.DotNet.Interactive.ExtensionLab/SqlKernelUtils.cs
```diff
@@ -0,0 +1,40 @@
+// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Collections.Generic;
+
+namespace Microsoft.DotNet.Interactive.ExtensionLab
+{
+    public static class SqlKernelUtils
```
jonsequitur, line 8: This should be kept `internal`.

src/Microsoft.DotNet.Interactive.ExtensionLab.Tests/SqlKernelsExtensionTests.cs
```diff
@@ -90,6 +90,17 @@ private static IEnumerable<IEnumerable<IEnumerable<(string name, object value)>>
             } while (reader.NextResult());
         }
 
+        [Fact]
+        public void can_alias_duplicate_column_names_in_sql_results()
+        {
+            var testNames = new string[] { "apple", "banana", "apple", "orange", "orange", "BANANA", "orange", "BaNaNa" };
+            var expectedAliases = new string[] { "apple", "banana", "apple (1)", "orange", "orange (1)", "BANANA", "orange (2)", "BaNaNa" };
```
jonsequitur, line 97: It seems like we should do one of the following:

* Also number the first occurrence of "apple", or
* Start the numbering at 2.

jonsequitur: By testing this method in isolation, we're not actually guaranteeing that the end-to-end works. A more robust approach would be to verify against the actual serialized output.

src/Microsoft.DotNet.Interactive.ExtensionLab/SqlKernelUtils.cs
corivera: It wouldn't be visible to the test project then. Is that alright with you?

src/Microsoft.DotNet.Interactive.ExtensionLab.Tests/SqlKernelsExtensionTests.cs
corivera: I'll do the second option. Doing the first would require two passes through the names array.

src/Microsoft.DotNet.Interactive.ExtensionLab/SqlKernelUtils.cs
jonsequitur: That's ideal. It's an implementation detail and shouldn't be tested directly.

src/Microsoft.DotNet.Interactive.ExtensionLab.Tests/SqlKernelsExtensionTests.cs
jonsequitur: If one approach is clearer or more common than the other, I wouldn't sweat an extra loop through the array. The perf hit there is negligible especially on top of network I/O and serialization costs.

jonsequitur (approved):

PR #1037, opened 2021-02-04
fix struct deserialization

brettfo:
This fixes an issue where System.Text.Json can't deserialize a struct by replacing those types with our own.  This also uncovered an issue where the deep object inspection was defaulting to compile time types instead of runtime.  A helper was added to do the correct deep inspection and all existing calls to `BeEquivalentTo()` were audited and updated if necessary.

Without this change, the `LinePosition` and `LinePositionSpan` structs are always deserialized to `default`, e.g., `null`/`0` which means all language service commands were reporting for line 0, character 0, regardless of where the actual request took place.

### DIFF
--- a/src/Microsoft.DotNet.Interactive.CSharp/CSharpKernel.cs
+++ b/src/Microsoft.DotNet.Interactive.CSharp/CSharpKernel.cs
@@ -149,7 +149,7 @@ namespace Microsoft.DotNet.Interactive.CSharp
             
             var document = _workspace.UpdateWorkingDocument(command.Code);
             var text = await document.GetTextAsync();
-            var cursorPosition = text.Lines.GetPosition(command.LinePosition.AsCodeAnalysisLinePosition());
+            var cursorPosition = text.Lines.GetPosition(command.LinePosition.ToCodeAnalysisLinePosition());
             var service = QuickInfoService.GetService(document);
             var info = await service.GetQuickInfoAsync(document, cursorPosition);
             
--- a/src/Microsoft.DotNet.Interactive.CSharp/CSharpKernel.cs
+++ b/src/Microsoft.DotNet.Interactive.CSharp/CSharpKernel.cs
@@ -149,7 +149,7 @@ namespace Microsoft.DotNet.Interactive.CSharp
             
             var document = _workspace.UpdateWorkingDocument(command.Code);
             var text = await document.GetTextAsync();
-            var cursorPosition = text.Lines.GetPosition(command.LinePosition);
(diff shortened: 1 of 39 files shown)


src/Microsoft.DotNet.Interactive.CSharp/CSharpKernel.cs
```diff
@@ -149,7 +149,7 @@ public async Task HandleAsync(RequestHoverText command, KernelInvocationContext
             
             var document = _workspace.UpdateWorkingDocument(command.Code);
             var text = await document.GetTextAsync();
-            var cursorPosition = text.Lines.GetPosition(command.LinePosition);
+            var cursorPosition = text.Lines.GetPosition(command.LinePosition.AsCodeAnalysisLinePosition());
```
jonsequitur, line 152:
```suggestion
            var cursorPosition = text.Lines.GetPosition(command.LinePosition.ToCodeAnalysisLinePosition());
```

jonsequitur: This is more symmetrical with `FromCodeAnalysisLinePosition`.

src/Microsoft.DotNet.Interactive.Tests/Server/SerializationTests.cs
```diff
@@ -65,10 +67,20 @@ public void All_event_types_are_round_trip_serializable(KernelEvent @event)
 
             var deserializedEnvelope = KernelEventEnvelope.Deserialize(json);
 
+            // ignore these specific properties because they're not serialized
+            var ignoredProperties = new HashSet<string>()
+            {
+                $"{nameof(CommandFailed)}.{nameof(CommandFailed.Exception)}",
+                $"{nameof(DisplayEvent)}.{nameof(DisplayEvent.Value)}"
```
jonsequitur, line 74: `DisplayEvent` is abstract so what about e.g. `ReturnValueProduced`?

jonsequitur (approved):

src/Microsoft.DotNet.Interactive.Tests/Server/SerializationTests.cs
brettfo: I initially tried that, but since `DisplayEvent` is the declaring type, this is the only thing that matches.

brettfo: Also, due to how that method is implemented, I have access either to the type information through `SelectedMemberInfo` like I currently do, _or_ I can use the `EventType` property, but not both, and since the member name (`SelectedMemberInfo.Name`) is the really important part, we have to fall back to this.

PR #1041, opened 2021-02-05
retarget extension for stable vscode

brettfo:
Purely by accident, our extension works on stable VS Code instead of Insiders like we advertise.  This PR simply makes that official by pulling in the appropriate *.d.ts files from VS Code.

In the future we'll split out a separate Insiders-only version of the extension for rapid development, but that will require figuring out some infrastructure trickery, so in the meantime I've made it so that by default our extension provides support for our own .ipynb files, _but_ if (1) the extension is running an insiders build, (2) the Jupyter extension is installed, and (3) a new option is checked (default `false`), then we'll allow the Jupyter extension to handle the .ipynb file format for us.  This buys us some time until we figure out how to properly split the extension.

There are also some additional changes that recently popped up, including:

1. When our kernel is selected for a given notebook, always set the document metadata to contain our Jupyter kernelspec settings to allow that notebook to be directly opened in Jupyter Lab.
2. When a .ipynb is opened with the Jupyter extension (following the strict options set above), we sniff out the kernelspec data and if it looks like it's probably one of ours, we say that our kernel is the preferred one.  This will make the scenario where a user is given a notebook from a Jupyter Lab user and they open it on a machine with only VS Code installed, then we will be appropriately selected as the kernel.  If we don't see the appropriate metadata, we'll still return a non-preferred version of our kernel so it can always be manually selected.

Future work to be done includes:
1. When we're using the Jupyter extension to handle the file and a new notebook is created, when the notebook is eventually saved we have no way of reassociating `Untitled-1.ipynb` to `C:\MyNotebook.ipynb`.

### DIFF
--- a/src/dotnet-interactive-vscode/package.json
+++ b/src/dotnet-interactive-vscode/package.json
@@ -12,7 +12,7 @@
   "//version": "The version '42.42.42' is auto-set during CI package creation.",
   "version": "42.42.42",
   "engines": {
-    "vscode": "^1.46.0"
+    "vscode": "^1.53.0"
   },
   "bugs": {
     "url": "https://github.com/dotnet/interactive/issues"
@@ -41,6 +41,7 @@
   ],
   "activationEvents": [
     "onNotebook:dotnet-interactive",
+    "onNotebook:dotnet-interactive-jupyter",
     "onNotebook:*",
     "onCommand:dotnet-interactive.acquire",
     "onCommand:dotnet-interactive.newNotebook",
@@ -49,8 +50,7 @@
   ],
   "main": "./out/vscode/extension.js",
   "extensionDependencies": [
-    "ms-dotnettools.vscode-dotnet-runtime",
-    "ms-toolsai.jupyter"
+    "ms-dotnettools.vscode-dotnet-runtime"
   ],
   "contributes": {
     "notebookProvider": [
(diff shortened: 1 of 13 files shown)


colombod (approved):

src/dotnet-interactive-vscode/package.json
```diff
@@ -62,6 +62,16 @@
             "filenamePattern": "*.{dib,dotnet-interactive}"
           }
         ]
+      },
+      {
+        "viewType": "dotnet-interactive-jupyter",
+        "displayName": ".NET Interactive for Jupyter Notebooks",
```
jonsequitur, line 68:
```suggestion
        "displayName": ".NET Interactive for Jupyter",

```

src/dotnet-interactive-vscode/package.json
```diff
@@ -104,8 +114,13 @@
...
         "dotnet-interactive.minimumInteractiveToolVersion": {
           "type": "string",
-          "default": "1.0.207101",
+          "default": "1.0.210401",
           "description": "The minimum required version of the .NET Interactive tool."
+        },
+        "dotnet-interactive.useJupyterExtensionForIpynbFiles": {
+          "type": "boolean",
+          "default": false,
+          "description": "(EXPERIMENTAL) Use the Jupyter extension to handle .ipynb notebook files if possible (requries restart)."
```
jonsequitur, line 123:
```suggestion
          "description": "(EXPERIMENTAL) Use the Jupyter extension to handle .ipynb files if possible (requires restart)."

```

src/dotnet-interactive-vscode/package.json
brettfo: done

src/dotnet-interactive-vscode/package.json
brettfo: done

PR #1042, opened 2021-02-05
don't crash when required ipynb fields are missing

### LINKED ISSUE #1039: Audit calls to `JsonElement.GetProperty(string)` in NotebookFileFormatHandler.cs
The method `JsonElement.GetProperty(string)` throws if the specified property was not found which can cause some notebooks to not open.

All calls to `GetProperty(string)` in that file need to be audited.  E.g., [this line](https://github.com/dotnet/interactive/blob/fc09b100054dd827b2e02a27920f98e240d0b1c5/src/Microsoft.DotNet.Interactive/Notebook/NotebookFileFormatHandler.cs#L156) will throw if the `language` key is not present.

brettfo:
When parsing a Jupyter notebook (.ipynb) if there are missing fields, System.Text.Json will throw where Newtonsoft.Json would return `null`.  This adds tests for those missing cases and removes all calls to `.GetProperty()` with a safe variant that we already had.  Some convenience extension methods were added for `Nullable<JsonElement>`.

Fixes #1039.

### DIFF
--- a/src/Microsoft.DotNet.Interactive.Tests/Notebook/JupyterNotebookDocumentFileFormatTests.cs
+++ b/src/Microsoft.DotNet.Interactive.Tests/Notebook/JupyterNotebookDocumentFileFormatTests.cs
@@ -633,6 +633,278 @@ namespace Microsoft.DotNet.Interactive.Tests.Notebook
         }
 
         [Fact]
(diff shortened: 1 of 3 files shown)


colombod (approved):
LGTM(Y)

PR #1057, opened 2021-02-10
Adding cancel execution command

colombod:
Add tests for quit command


src/Microsoft.DotNet.Interactive.Tests/CompositeKernelTests.cs
```diff
@@ -425,6 +426,189 @@ public async Task Deferred_commands_on_composite_kernel_are_execute_on_first_sub
                     typeof(CommandSucceeded));
         }
 
+        [Fact]
+        public async Task quit_command_cancels_all_deferred_commands_on_composite_kernel()
```
jonsequitur, line 430: These aren't really just tests of the `CompositeKernel`. It might be useful to split them into a new test class and move toward a more vertical test organization.

src/Microsoft.DotNet.Interactive.Tests/LanguageKernelTests.cs
```diff
@@ -1264,5 +1267,208 @@ public async Task SetVariableAsync_can_redeclare_an_existing_variable_and_change
...
+            Task.WhenAll(
+                    Task.Run(async () =>
+                    {
+                        await kernel.SendAsync(submitCodeCommand);
+                    }),
+                    Task.Run(async () =>
+                    {
+                        await Task.Delay(100);
+                        await kernel.SendAsync(cancelCommand);
+                    })).Wait();
```
jonsequitur, line 1416: Make sure to enforce a timeout here.

src/Microsoft.DotNet.Interactive.Tests/LanguageKernelTests.cs
```diff
@@ -1264,5 +1267,208 @@ public async Task SetVariableAsync_can_redeclare_an_existing_variable_and_change
...
+                .Which
+                .Exception
+                .Should()
+                .BeOfType<OperationCanceledException>();
+        }
+
+        [Theory]
+        [InlineData(Language.CSharp)]
+        [InlineData(Language.FSharp)]
+        public async Task cancel_command_cancels_all_deferred_commands(Language language)
```
jonsequitur, line 1437: This looks like a duplicate. Consolidating will help make it clearer what we're testing.

src/Microsoft.DotNet.Interactive/Commands/Cancel.cs
```diff
@@ -0,0 +1,15 @@
...
+
+using System.Threading.Tasks;
+
+namespace Microsoft.DotNet.Interactive.Commands
+{
+    public class Cancel : KernelCommand
+    {
+        public Cancel(string targetKernelName = null): base(targetKernelName)
+        {
+            Handler = (command, context) => Task.CompletedTask;
```
jonsequitur, line 12: It's preferable to override `InvokeAsync`. We should at some point limit the settable `Handler` property to only those commands that vary behavior by kernel.

src/Microsoft.DotNet.Interactive/CompositeKernel.cs
```diff
@@ -213,6 +213,22 @@ protected override void SetHandlingKernel(KernelCommand command, KernelInvocatio
...
+                    CancelInflightCommands();
+                    ClearPendingCommands();
+                    kernel.ClearPendingCommands();
+                    break;
+                case Quit _:
+                    CancelInflightCommands();
+                    ClearPendingCommands();
+                    kernel.CancelInflightCommands();
+                    kernel.ClearPendingCommands();
+                    break;
```
jonsequitur, line 229: Do these two cases need to vary in terms of behavior?

src/Microsoft.DotNet.Interactive/CompositeKernel.cs
```diff
@@ -213,6 +213,22 @@ protected override void SetHandlingKernel(KernelCommand command, KernelInvocatio
                 throw new NoSuitableKernelException(command);
             }
 
+            switch (command)
+            {
+                case Cancel _:
+
+                    CancelInflightCommands();
+                    ClearPendingCommands();
```
jonsequitur, line 221: Is "pending" different from "deferred"?

src/Microsoft.DotNet.Interactive/CompositeKernel.cs
```diff
@@ -213,6 +213,22 @@ protected override void SetHandlingKernel(KernelCommand command, KernelInvocatio
                 throw new NoSuitableKernelException(command);
             }
 
+            switch (command)
```
jonsequitur, line 216: Rather than duplicating all of this, why not move it to the base and put the actual cancellation logic into a virtual method?

src/Microsoft.DotNet.Interactive/Commands/Quit.cs
```diff
@@ -0,0 +1,47 @@
...
+using System.Text.Json.Serialization;
+using System.Threading.Tasks;
+
+namespace Microsoft.DotNet.Interactive.Commands
+{
+    public class Quit : KernelCommand
+    {
+        private static readonly CompositeDisposable DisposeOnQuit = new();
+
+        public static void RegisterForDisposalOnQuit(IDisposable disposable)
```
jonsequitur, line 16: Why have potentially different disposal semantics than `Kernel.Dispose`? This looks like something that doesn't belong in the library. It should probably just be a host concern.

src/Microsoft.DotNet.Interactive/Commands/Quit.cs
```diff
@@ -0,0 +1,47 @@
...
+        public Quit(string targetKernelName = null): this(() =>
+        {
+            Environment.Exit(0);
+        }, targetKernelName)
+        {
+        }
+
+        public Quit(Action onQuit, string targetKernelName = null) : base(targetKernelName)
+        {
+            if (onQuit == null)
```
jonsequitur, line 34:
```suggestion
            if (onQuit is null)
```

src/Microsoft.DotNet.Interactive/KernelInvocationContext.cs
```diff
@@ -70,14 +75,27 @@ public void Complete(KernelCommand command)
...
         {
-            Publish(new CommandFailed(exception, Command, message));
+            if (!IsComplete)
+            {
+                Publish(new CommandFailed(exception, Command, message));
 
-            _events.OnCompleted();
-            IsComplete = true;
+                _events.OnCompleted();
+                _cancellationTokenSource.Cancel(false);
```
jonsequitur, line 96: It looks like this gets called twice when someone is calling `Cancel`.

src/Microsoft.DotNet.Interactive.Tests/CompositeKernelTests.cs
colombod: moved to QuiCommandTests class

src/Microsoft.DotNet.Interactive.Tests/LanguageKernelTests.cs
colombod: ok

src/Microsoft.DotNet.Interactive/CompositeKernel.cs
colombod: might not be the case

src/Microsoft.DotNet.Interactive/CompositeKernel.cs
colombod: pending includes all things in the trampoline queue and the deferred

src/Microsoft.DotNet.Interactive/Commands/Quit.cs
colombod: changed

src/Microsoft.DotNet.Interactive/Commands/Quit.cs
colombod: we can have the host to register the OnQuit handler, and throw if it is not set

src/Microsoft.DotNet.Interactive/CompositeKernel.cs
jonsequitur: So it might be clearer to say `ClearQueuedAndDeferredCommand` since both of those are terms that already exist in the code.

src/Microsoft.DotNet.Interactive.Tests/QuitCommandTests.cs
```diff
@@ -0,0 +1,215 @@
...
+using FluentAssertions.Execution;
+using Microsoft.DotNet.Interactive.Commands;
+using Microsoft.DotNet.Interactive.Events;
+using Microsoft.DotNet.Interactive.Tests.Utility;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace Microsoft.DotNet.Interactive.Tests
+{
+    [Collection("Do not parallelize")]
```
jonsequitur, line 16: Why should these not be parallelizable? That sounds like a smell.

src/Microsoft.DotNet.Interactive/KernelScheduler.cs
```diff
@@ -0,0 +1,274 @@
...
+            public DeferredOperation(OnExecuteDelegate onExecute, GetDeferredOperationsDelegate getDeferredOperations)
+            {
+                OnExecute = onExecute;
+                GetDeferredOperations = getDeferredOperations;
+            }
+        }
+
+    }
+
+    internal sealed class ClockwiseSynchronizationContext : SynchronizationContext, IDisposable
```
jonsequitur, line 201: Let's rename this and move it to a different file.

src/Microsoft.DotNet.Interactive/KernelScheduler.cs
```diff
@@ -0,0 +1,274 @@
...
+            }
+
+            async Task DoWork(ScheduledOperation scheduleOperation)
+            {
+                using var _ = Logger.OnEnterAndExit("DoWork");
+                if (!scheduleOperation.CompletionSource.Task.IsCanceled)
+                {
+                    try
+                    {
+                        await scheduleOperation.OnExecuteAsync(scheduleOperation.Value);
```
jonsequitur, line 122: No need to pass the operation's value to itself.

src/Microsoft.DotNet.Interactive/KernelScheduler.cs
```diff
@@ -0,0 +1,274 @@
...
+            }
+
+            public TaskCompletionSource<U> CompletionSource { get; }
+        }
+
+        private class DeferredOperation
+        {
+            public GetDeferredOperationsDelegate GetDeferredOperations { get; }
+            public OnExecuteDelegate OnExecute { get; }
+            public DeferredOperation(OnExecuteDelegate onExecute, GetDeferredOperationsDelegate getDeferredOperations)
```
jonsequitur, line 192:
```suggestion
            public DeferredOperation(OnExecuteDelegate onExecuteAsync, GetDeferredOperationsDelegate getDeferredOperations)
```

Issue #1048, opened 2021-02-09
MSSQL: Can't get intellisense results before running query cell


corivera:
When typing out a query in a new MSSQL kernel cell, intellisense for things like column names doesn't show up until after the cell is run.

corivera:
Looks like this might actually be related to the number of lines in the cell. In the following query, I can't get column names after the "d" table alias below unless there's a blank line immediately after the query. Even without running the query first, I can get intellisense back by just adding that blank line after the query text.

#!sqlkerneltest
select d.___
from sys.databases d

PR #1062, opened 2021-02-12
Improve intellisense behavior for MSSQL kernel.

corivera:
Fixes https://github.com/dotnet/interactive/issues/1048

I was noticing some issues in the MSSQL intellisense where things wouldn't autocomplete unless there were extra spaces or blank lines in the cell. My fixes here were to handle newlines correctly using Environment.NewLine so that there wouldn't be offset spacing in the text, as well as incrementing the last character column number by 1 to address off-by-1 issues. That last one doesn't really make sense, since we don't have to increment the line number by 1, so that may be a bug in the SqlToolsService somewhere. Trying lots of different text to autocomplete showed no issues when incrementing the column number in this way, and there were definitely lots of autocomplete issues without doing it. I also cleaned up an unused return type, and fixed a potential array out-of-bounds error when parsing the text to autocomplete.


src/Microsoft.DotNet.Interactive.SqlServer/MsSqlServiceClient.cs
```diff
@@ -158,10 +156,10 @@ public async Task<QueryExecuteSubsetResult> ExecuteQueryExecuteSubsetAsync(Query
 
         public async Task SendTextChangeNotificationAsync(Uri ownerUri, string newText, string oldText)
         {
-            var oldTextLines = oldText.Split('\n');
+            var oldTextLines = oldText.Split(Environment.NewLine);
```
jonsequitur, line 159: It's a common scenario that the compute and the editor are on different machines with different newlines, or that a notebook file was written with `\r\n` newlines but is being evaluated with `\n` newlines. So the split should handle both.

colombod: in ntoebook parsing we use this logic 
`s.Split('\n').Select(l => l.EndsWith('\r') ? l[0..^1] : l);`

corivera: Fixed with Diego's suggestion.

src/Microsoft.DotNet.Interactive.SqlServer/MsSqlServiceClient.cs
```diff
@@ -158,10 +156,10 @@ public async Task<QueryExecuteSubsetResult> ExecuteQueryExecuteSubsetAsync(Query
 
         public async Task SendTextChangeNotificationAsync(Uri ownerUri, string newText, string oldText)
         {
-            var oldTextLines = oldText.Split('\n');
+            var oldTextLines = oldText.Split('\n').Select(text => text.EndsWith('\r') ? text[0..^1] : text).ToArray();
```
jonsequitur, line 159: You can simplify like this:

```csharp
text.Split(new []{ '\n', '\r' }, StringSplitOptions.RemoveEmptyEntries)
```

corivera: Does that still preserve blank lines? We need the original number of lines in the text in order to get the right document change ranges.

jonsequitur: It does if you don't specify `StringSplitOptions.RemoveEmptyEntries`.

corivera: Either way doesn't seem to work. When using this test string, "abc\r\ndef\r\n\r\nabc\r\nabcdef", the parsing returns 4 lines when specifying RemoveEmptyEntries, while it returns 9 when leaving out that option. We expect 5 lines given the Windows line endings in that string.

jonsequitur: Ah, righto.

src/Microsoft.DotNet.Interactive.SqlServer.Tests/MsSqlServiceClientTests.cs
```diff
@@ -0,0 +1,61 @@
...
+            int oldTextLineCount = 5;
+            int oldTextLastCharacterNum = 6;
+            string newText = string.Join(lineEnding, "abc", "def");
+            var testUri = new Uri("untitled://test");
+
+            var docChange = MsSqlServiceClient.GetDocumentChangeForText(testUri, newText, oldText);
+
+            docChange.ContentChanges.Length
+                .Should()
+                .Equals(1);
```
jonsequitur, line 37: This is a common `FluentAssertions` gotcha. You should make this change throughout these. The `Equals` call here is `Object.Equals` which returns false but doesn't throw, so you'll get false positives.

```suggestion
                .Be(1);
```

jonsequitur (approved):

Issue #1063, opened 2021-02-12
Use notebook error mime type for display


brettfo:
VS Code Stable doesn't support the output mime type `application/x.notebook.error-traceback`, so we're temporarily using `text/plain` to ensure it appears in the UI.  Once Stable supports this mime type, the `text/plain` fallback will be removed.

PR #1064, opened 2021-02-12
split vscode stable/insiders APIs into separate projects

brettfo:
To allow the same extension to work on both stable and insiders VS Code, this adds separate sub-projects so the different APIs can be consumed at the same time.

Note that due to #1063 error display on insiders will revert to `text/plain`.


colombod (approved):

colombod (approved):
Should this also update the .NET Interactive tool? we have merged the json error fix right?

brettfo:
No need to update the tool; when the CI builds it inserts the version of the tool that was built 10 minutes prior via [this](https://github.com/dotnet/interactive/blob/411b53ced326a1a32c1f3cb1aa1fd8b5c8a8a74f/eng/package/PackVSCodeExtension.ps1#L23).

colombod:
Is it good for merging then?

azure-pipelines.yml
```diff
@@ -114,7 +114,7 @@ stages:
           displayName: NPM tests
           workingDirectory: "$(Build.SourcesDirectory)/src/Microsoft.DotNet.Interactive.Js"
 
-        - script: npm i
+        - script: npm install
           displayName: Install NPM packages for VS Code extension
```
colombod, line 118: Since we have the need for powershell is there a way to ensure powershell presence and also for node to latest lts for our builds too?

Issue #1058, opened 2021-02-11
MSSQL: Query that returns 0 rows Throws Exception


chlafreniere:
### Describe the bug
Craft a query that won't return any rows. e.g.:
`SELECT * from sys.tables where name = 'foo'`

### Please complete the following:

- OS
    - [ ] Windows 10
    - [x] macOS
    - [ ] Linux (Please specify distro)
    - [ ] iOS
    - [ ] Android
 - Browser 
    - [ ] Chrome
    - [ ] Edge
    - [ ] Firefox
    - [ ] Safari
 - Frontend
    - [ ] Jupyter Notebook
    - [ ] Jupyter Lab
    - [ ] nteract
    - [x] Visual Studio Code
    - [ ] Other (please specify)

### Screenshots
![image](https://user-images.githubusercontent.com/40371649/107596537-de2e2a80-6bcc-11eb-933d-fc4e42a44245.png)

### Relevant stack trace

```
Error: StreamJsonRpc.RemoteInvocationException: Start row cannot be less than 0 or greater than the number of rows in the result set (Parameter 'startRow')
at StreamJsonRpc.JsonRpc.InvokeCoreAsync[TResult](RequestId id, String targetName, IReadOnlyList`1 arguments, IReadOnlyList`1 positionalArgumentDeclaredTypes, IReadOnlyDictionary`2 namedArgumentDeclaredTypes, CancellationToken cancellationToken, Boolean isParameterObject)
at Microsoft.DotNet.Interactive.SqlServer.MsSqlServiceClient.ExecuteQueryExecuteSubsetAsync(QueryExecuteSubsetParams subsetParams)
at Microsoft.DotNet.Interactive.SqlServer.MsSqlKernel.<>c__DisplayClass14_0.<<HandleAsync>b__0>d.MoveNext()
--- End of stack trace from previous location ---
at Microsoft.DotNet.Interactive.SqlServer.MsSqlKernel.HandleAsync(SubmitCode command, KernelInvocationContext context)
at Microsoft.DotNet.Interactive.Kernel.HandleAsync(KernelCommand command, KernelInvocationContext context) in D:\workspace\_work\1\s\src\Microsoft.DotNet.Interactive\Kernel.cs:line 296
[...]

PR #1077, opened 2021-02-17
Skip retrieving MSSQL result sets that contain no rows.

corivera:
Fixes https://github.com/dotnet/interactive/issues/1058

We throw an error from SqlToolsService when trying to retrieve a result set that doesn't have any rows. This change skips that retrieval step when the result set has zero rows, and instead displays a single empty row so that we still get column names in the Nteract data explorer.


src/Microsoft.DotNet.Interactive.SqlServer/MsSqlKernel.cs
```diff
@@ -129,25 +129,31 @@ public async Task HandleAsync(SubmitCode command, KernelInvocationContext contex
...
+                                    OwnerUri = _tempFileUri.ToString(),
+                                    BatchIndex = batchSummary.Id,
+                                    ResultSetIndex = resultSummary.Id,
+                                    RowsStartIndex = 0,
+                                    RowsCount = Convert.ToInt32(resultSummary.RowCount)
+                                };
+                                var subsetResult = await _serviceClient.ExecuteQueryExecuteSubsetAsync(subsetParams);
+                                if (subsetResult.Message != null)
+                                {
+                                    context.Fail(message: subsetResult.Message);
```
jonsequitur, line 146: Is a message always an error?

src/Microsoft.DotNet.Interactive.SqlServer.Tests/MsSqlConnectionTests.cs
```diff
@@ -126,5 +126,44 @@ public async Task Field_types_are_deserialized_correctly()
...
+
+            var value = events.Should()
+                              .ContainSingle<DisplayedValueProduced>()
+                              .Which;
+
+            var tables = (IEnumerable<IEnumerable<IEnumerable<(string, object)>>>) value.Value;
+            var table = tables.Single().ToTable();
+
+            table.Count.Should().Be(1);
+            table[0].Count.Should().Be(3);
```
jonsequitur, line 163: I don't think we should be creating fake data in order to display the Data Explorer. This data might be displayed or shared in other ways and the record count should be accurate.

src/Microsoft.DotNet.Interactive.SqlServer/MsSqlKernel.cs
corivera: It used to be, but now it looks like the property was removed on the SqlToolsService side: https://github.com/microsoft/sqltoolsservice/blob/main/src/Microsoft.SqlTools.ServiceLayer/QueryExecution/Contracts/SubsetRequest.cs#L47

I've removed it from our subset result class.

src/Microsoft.DotNet.Interactive.SqlServer.Tests/MsSqlConnectionTests.cs
corivera: Are we able to display only columns (without data) in the data explorer? That's the behavior in other SQL query editors like SSMS and ADS. Otherwise, I can just display a message that no rows were returned.

jonsequitur: I don't know. A simple message would be fine for now. A nicer display for schema info might be nice to have but we can discuss later.

jonsequitur (approved):

corivera:
@jonsequitur Pushed up my latest changes, so feel free to merge whenever.

jonsequitur (approved):

PR #1083, opened 2021-02-19
default to using the jupyter extension

### LINKED ISSUE #1060: Python notebooks cannot be opened if .NET interactive is installed
Once the jupyter extension ships, if .NET extension is installed, jupyter will use the new editor and register itself as the notebook provider. 

To allow python notebooks to still be supported when .NET interactive is installed, .NET interactive can't be the content provider.

If it is, then only .NET interactive languages can be used.

This PR here is the submission on the Jupyter side that enables jupyter and .NET in stable:
https://github.com/microsoft/vscode-jupyter/pull/4772

brettfo:
This switches to using the Jupyter extension by default in both stable and insiders and allows an explicit override to force our own handling via an option.

I have not yet added a direct dependency on the Jupyter extension because there are currently some broken APIs when using that against insiders that prevent our extension from properly registering, so I'm leaving open the fallback of disabling the Jupyter extension and allowing our own ipynb handling to be used.  Once the Jupyter extension has been updated to work on insiders again, I'll come back and directly add our dependency on it via `package.json`.

Fixes #1060.

### DIFF
--- a/src/dotnet-interactive-vscode/package.json
+++ b/src/dotnet-interactive-vscode/package.json
@@ -117,10 +117,10 @@
           "default": "1.0.211604",
           "description": "The minimum required version of the .NET Interactive tool."
         },
-        "dotnet-interactive.useJupyterExtensionForIpynbFiles": {
+        "dotnet-interactive.useDotNetInteractiveExtensionForIpynbFiles": {
           "type": "boolean",
           "default": false,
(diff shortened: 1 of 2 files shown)


brettfo:
cc @rchiodo

jonsequitur (approved):

PR #1084, opened 2021-02-19
Remove CSharpKernel.UseDefaultFormatting and default usings for PocketView & PocketViewTags

### LINKED ISSUE #938: PocketViewTags._ is in scope and overrides discard in C#
### Describe the bug

When assigning to a discard in C#, I get `error CS0200:  Property or indexer 'PocketViewTags._' cannot be assigned to -- it is read only`. It seems that the notebook is interpreting `_` as a reference to `PocketViewTags._` rather than a discard variable. To reproduce in the VSCode Insiders extension, I can simply open a new notebook and in the first cell type `_ = 1;`. Running the cell then generates the error.

I know that the C# specs say that the `_` discard is only valid if no other variables of that name are in scope, I was just surprised to find that `PocketViewTags._` was.

1.0.160204+9a96ad3eee7b38d6db907a97513009e04f967825

- OS
    - [x] Windows 10
    - [ ] macOS
    - [ ] iOS
    - [ ] Android
 - Browser 
    - [ ] Chrome
    - [ ] Edge
    - [ ] Firefox
    - [ ] Safari
 - Frontend
    - [ ] Jupyter Notebook
    - [ ] Jupyter Lab
    - [ ] nteract
    - [x] Visual Studio Code
    - [ ] Other (please specify)

### Screenshots
image

jonsequitur:
This will fix #938.

Changes include: 
- [x] Remove `CSharpKernel.UseDefaultFormatting` extension method (which only added usings for `PocketViewTags` anyway)
- [x] Improve information displayed when loading various extensions.
- [x] Update notebooks that use `PocketView` to add those usings directly.


src/Microsoft.DotNet.Interactive.Tests/LanguageKernelFormattingTests.cs
```diff
@@ -26,6 +27,18 @@ public LanguageKernelFormattingTests(ITestOutputHelper output) : base(output)
         {
         }
 
+        protected override CSharpKernel CreateCSharpKernel()
+        {
+            var cSharpKernel = base.CreateCSharpKernel();
+
+            cSharpKernel.DeferCommand(new SubmitCode(@$"
```
brettfo, line 34:
```suggestion
            cSharpKernel.DeferCommand(new SubmitCode($@"
```

Both types of string are technically allowed, but `$@` is the canonical way to do it.

src/Microsoft.DotNet.Interactive.Tests/LanguageKernelPackageTests.cs
```diff
@@ -315,8 +316,10 @@ public async Task Dependency_version_conflicts_are_resolved_correctly()
 #r ""nuget:Microsoft.Data.Analysis,0.4.0""
 ");
 
-            await kernel.SubmitCodeAsync(@"
-using Microsoft.Data.Analysis;");
+            await kernel.SubmitCodeAsync(@$"
```
brettfo, line 319:
```suggestion
            await kernel.SubmitCodeAsync($@"
```

colombod (approved):

Issue #1090, opened 2021-02-19
MSSQL: Display messages from SQL Server


lucyzhang929:
Messages displayed in ADS:
![Screen Shot 2021-02-19 at 3 39 11 PM](https://user-images.githubusercontent.com/23462877/108573407-4fef2e00-72c9-11eb-8169-a611a156cd0c.png)

Interactive:
![Screen Shot 2021-02-19 at 3 42 07 PM](https://user-images.githubusercontent.com/23462877/108573273-0b639280-72c9-11eb-8469-388c1db22f26.png)

PR #1093, opened 2021-02-22
Display messages from SQL Server

lucyzhang929:
Fixes https://github.com/dotnet/interactive/issues/1090
Screen Shot 2021-02-22 at 9 53 16 AM

### DIFF
--- a/src/Microsoft.DotNet.Interactive.SqlServer/MsSqlKernel.cs
+++ b/src/Microsoft.DotNet.Interactive.SqlServer/MsSqlKernel.cs
@@ -183,7 +183,7 @@ namespace Microsoft.DotNet.Interactive.SqlServer
                     }
                     else
                     {
-                        context.Display(messageParams.Message.Message.ToString());
+                        context.Display(messageParams.Message.Message);
                     }
                 }
                 catch (Exception e)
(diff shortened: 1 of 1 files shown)


src/Microsoft.DotNet.Interactive.SqlServer/MsSqlKernel.cs
```diff
@@ -181,6 +181,10 @@ public async Task HandleAsync(SubmitCode command, KernelInvocationContext contex
                         context.Fail(message: messageParams.Message.Message);
                         completion.SetResult(true);
                     }
+                    else
+                    {
+                        context.Display(messageParams.Message.Message.ToString());
```
jonsequitur, line 186: Does this branch also need a `completion.SetResult(true);`?

corivera: Completion in this case would be handled at the end of processing the query results. We set the completion above because it's a failure case.

corivera:
@lucyzhang929 Have you looked into whether there are any race conditions here when running queries with multiple result sets? Would any messages show up out of order relative to the results themselves?

src/Microsoft.DotNet.Interactive.SqlServer/MsSqlKernel.cs
corivera, line 186: The Message field is already a string, so we shouldn't need ToString() here.

corivera (approved):

lucyzhang929:
@corivera Tested a bunch of different queries and didn't see any issues with ordering
Screen Shot 2021-02-22 at 11 31 55 AM

jonsequitur (approved):

PR #1101, opened 2021-02-25
ensure a non-existant dotnet always returns something actionable

brettfo:
While testing something else I discovered that if `dotnet.exe` is _completely_ missing from the machine, then our up-to-date check never returns and we'll spin forever.  The fix was to also listen to the spawned process's `'error'` event.  I also moved the function to a common location so I could add a test.


src/dotnet-interactive-vscode/src/tests/unit/misc.test.ts
```diff
@@ -296,4 +296,13 @@ describe('Miscellaneous tests', () => {
             expect(reshaped).to.equal('some error message');
         });
     });
+
+    it('executing a non-existant process still returns', async () => {
```
jonsequitur, line 300:
```suggestion
    it('executing a non-existent process still returns', async () => {
```

jonsequitur (approved):

colombod (approved):

PR #1102, opened 2021-02-25
Cancel Command execution

colombod:
New scheduler logic
New kernel routing
Implement cancel logic


src/Microsoft.DotNet.Interactive.Tests/Utility/CancellableCommand.cs
```diff
@@ -0,0 +1,33 @@
+// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Threading.Tasks;
+using Microsoft.DotNet.Interactive.Commands;
+
+namespace Microsoft.DotNet.Interactive.Tests.Utility
+{
+    public class CancellableCommand : KernelCommand
```
jonsequitur, line 9: Is this being used?

src/Microsoft.DotNet.Interactive/Commands/KernelCommand.cs
```diff
@@ -40,6 +40,9 @@ protected KernelCommand(string targetKernelName = null, KernelCommand parent = n
 
         internal static KernelCommand None { get; } = new NoCommand();
 
+        [JsonIgnore]
```
jonsequitur, line 43: The `[JsonIgnore]` shouldn't be needed for `internal` properties.

src/Microsoft.DotNet.Interactive/CompositeKernel.cs
```diff
@@ -202,22 +203,42 @@ protected override void SetHandlingKernel(KernelCommand command, KernelInvocatio
...
+            var targetKernelName = command switch
+            {
+                { } kcb => kcb.TargetKernelName ?? DefaultKernelName,
+                _ => DefaultKernelName
+            };
+
+            Kernel kernel;
 
-            if (kernel is null)
+            if (targetKernelName != null)
```
jonsequitur, line 217:
```suggestion
            if (targetKernelName is not null)
```

colombod: done

PR #1105, opened 2021-02-26
Add do-nothing SQL kernel to the builtin kernels in order to enable SQL language coloring

corivera:
This change adds a do-nothing SQL kernel that can be used to enable language coloring for SQL code. This is required because we can only define language grammars for fixed kernel names, but MSSQL kernel names are user-defined via the **connect** magic command, and thus aren't known ahead of time. The default SQL kernel displays a help message on execute if no actual SQL kernel exists, and does nothing otherwise. I also added a Language field to the base Kernel class so that we can better determine a kernel's capabilities without relying on its name.

Additionally, I incorporated the language grammar changes from @brettfo's PR here, for sake of completeness: https://github.com/dotnet/interactive/pull/1099


src/Microsoft.DotNet.Interactive.SqlServer/MsSqlKernel.cs
```diff
@@ -36,7 +36,7 @@ public class MsSqlKernel :
         public MsSqlKernel(
             string name,
             string connectionString,
-            MsSqlServiceClient client) : base(name)
+            MsSqlServiceClient client) : base(name, "SQL")
```
jonsequitur, line 39: This should probably be `"T-SQL"` or `"MSSQL"` or something more specific.

src/Microsoft.DotNet.Interactive/KernelExtensions.cs
```diff
@@ -45,6 +45,29 @@ public static Kernel FindKernel(this Kernel kernel, string name)
             };
         }
 
+        public static Kernel FindKernel(this Kernel kernel, Func<Kernel, bool> selector)
```
jonsequitur, line 48: This should return `IEnumerable<Kernel>` since there can be more than one that matches.

It would be more straightforward to use `VisitSubkernels` and avoid looking at the `ChooseKernelDirective`s.

src/Microsoft.DotNet.Interactive/SqlKernel.cs
```diff
@@ -0,0 +1,42 @@
...
+            {
+                context.Display(@"
+A SQL kernel is not currently defined.
+
+SQL kernels are provided as part of the Microsoft.DotNet.Interactive.SqlServer package, which can be installed by using the following nuget command:
+#r Microsoft.DotNet.Interactive.SqlServer
+
+Once installed, you can find more info about creating a SQL kernel and running queries by running the following help command:
+#!connect mssql -h
+                ");
```
jonsequitur, line 37: If the message is emitted as `"text/html"` we can include a link to the package on NuGet.org, and people can find additional information that way.

src/Microsoft.DotNet.Interactive/SqlKernel.cs
```diff
@@ -0,0 +1,42 @@
...
+        public Task HandleAsync(SubmitCode command, KernelInvocationContext context)
+        {
+            var sqlKernel = this.FindKernel(kernel => kernel.Language?.ToUpper() == "SQL");
+            if (sqlKernel == null)
+            {
+                context.Display(@"
+A SQL kernel is not currently defined.
+
+SQL kernels are provided as part of the Microsoft.DotNet.Interactive.SqlServer package, which can be installed by using the following nuget command:
+#r Microsoft.DotNet.Interactive.SqlServer
```
jonsequitur, line 33:
```suggestion
````
```console
#r "nuget:Microsoft.DotNet.Interactive.SqlServer"
```
````
```

jonsequitur, line 33: The `#r` directive should be updated to `#r "nuget:Microsoft.DotNet.Interactive.SqlServer,*-*"`.

Formatted as Markdown, you'll want:

image

corivera:
Closing this PR since @colombod is working on a different solution for this.

PR #1108, opened 2021-02-26
Fix MsSqlConnectionTests

lucyzhang929:
Fix tests that were broken by displaying SQL Server messages. (Now there are two DisplayedValueProduced events)

### DIFF
--- a/src/Microsoft.DotNet.Interactive.SqlServer.Tests/MsSqlConnectionTests.cs
+++ b/src/Microsoft.DotNet.Interactive.SqlServer.Tests/MsSqlConnectionTests.cs
@@ -21,7 +21,7 @@ namespace Microsoft.DotNet.Interactive.SqlServer.Tests
             await csharpKernel.SubmitCodeAsync(@$"
 #r ""nuget:microsoft.sqltoolsservice,3.0.0-release.53""
 ");
-            
+
             var kernel = new CompositeKernel
             {
                 csharpKernel,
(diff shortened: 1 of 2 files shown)


src/Microsoft.DotNet.Interactive.SqlServer.Tests/MsSqlConnectionTests.cs
```diff
@@ -56,11 +56,18 @@ public async Task It_can_connect_and_query_data()
...
-                  .ContainSingle<DisplayedValueProduced>()
-                  .Which
-                  .FormattedValues
-                  .Should()
-                  .ContainSingle(f => f.MimeType == HtmlFormatter.MimeType);
+                .Contain(e => e is DisplayedValueProduced && e.As<DisplayedValueProduced>()
+                .FormattedValues
+                .First()
+                .MimeType
+                .Equals(PlainTextFormatter.MimeType));
```
jonsequitur, line 63:
```suggestion
                .Be(PlainTextFormatter.MimeType));
```

This is a big gotcha when using FluentAssertions. The `Equals` call here is `object.Equals` and if it returns false, no exception will be thrown and the test will pass.

src/Microsoft.DotNet.Interactive.SqlServer.Tests/MsSqlConnectionTests.cs
```diff
@@ -56,11 +56,18 @@ public async Task It_can_connect_and_query_data()
             events.Should().NotContainErrors();
 
             events.Should()
-                  .ContainSingle<DisplayedValueProduced>()
-                  .Which
-                  .FormattedValues
-                  .Should()
-                  .ContainSingle(f => f.MimeType == HtmlFormatter.MimeType);
+                .Contain(e => e is DisplayedValueProduced && e.As<DisplayedValueProduced>()
```
jonsequitur, line 59: We have an overload extension or two for `ContainSingle` that can be helpful here. Here's an example:

```cs
events
  .Should()  
  .NotContainErrors()
  .And.ContainSingle<DisplayedValueProduced>(e => e.FormattedValues.Any(f => f.MimeType == "text/html")) // <-- this is a filter   
  .Which.FormattedValues.Should().ContainSingle(f => f.MimeType == "text/html")
  .Which.Value.Should().Contain("...")
```

jonsequitur: (That might be able to be more concise but just wanted to call attention to those APIs.)

src/Microsoft.DotNet.Interactive.SqlServer.Tests/MsSqlConnectionTests.cs
lucyzhang929: Oh wait, would it be fine if Equals returns false here since it is inside Contains()?

jonsequitur: Returning false doesn't provide any signal to the test runner that there was a problem. The `Be` method will cause an exception to be thrown, which the test runner can see.

PR #1109, opened 2021-02-26
[Draft] Theme nteract data explorer

lucyzhang929:


src/Microsoft.DotNet.Interactive/Commands/UpdateDisplayedValue.cs
```diff
@@ -9,7 +9,7 @@ namespace Microsoft.DotNet.Interactive.Commands
 {
     public class UpdateDisplayedValue : KernelCommand
     {
-        public UpdateDisplayedValue(FormattedValue formattedValue, string valueId)
+        public UpdateDisplayedValue(FormattedValue formattedValue, string valueId, string targetKernelName = null, string theme = null)
```
jonsequitur, line 12: We shouldn't make any changes to events or commands as this will have an impact on our public API contract.

src/Microsoft.DotNet.Interactive/Kernel.cs
```diff
@@ -24,6 +24,7 @@ namespace Microsoft.DotNet.Interactive
 {
     public abstract partial class Kernel : IDisposable
     {
+        public string theme = "light";
```
jonsequitur, line 27: The kernel shouldn't know anything about frontend concerns such as theming.

PR #1112, opened 2021-02-27
Cancel SQL queries after the KernelInvocationContext's CancellationToken is set as cancelled

corivera:
This is a preemptive change for when cancellation support is enabled for cells in general. I tested with my own CancellationToken that would get cancelled shortly after starting a query, and everything seemed to work fine.


src/Microsoft.DotNet.Interactive.SqlServer/MsSqlKernel.cs
```diff
@@ -121,6 +121,8 @@ public async Task HandleAsync(SubmitCode command, KernelInvocationContext contex
                 return;
             }
 
+            // If a query handler is already defined, then it means another query is already running in parallel.
```
corivera, line 124: This comment isn't relevant to this change, but there seemed to be some confusion over what this check was for, so I added a comment for it.

jonsequitur (approved):

PR #1119, opened 2021-03-02
fix message from sql kernel

colombod:


LadyNaggaga (commented):
This explanation is a lot clearer

src/Microsoft.DotNet.Interactive/SqlKernel.cs
```diff
@@ -48,21 +48,21 @@ public Task HandleAsync(SubmitCode command, KernelInvocationContext context)
             if (mssqlKernelNames.Count == 0)
             {
                 context.Display(@"
-A SQL kernel is not currently defined.
+A SQL connection has not been established.
```
chlafreniere, line 51: What's the loc story out of curiosity?

colombod: We do not have one at the moment

colombod:
> This explanation is a lot clearer

All thanks to @jonsequitur

chlafreniere (approved):

jonsequitur (approved):

PR #1120, opened 2021-03-02
split into two separate extensions

brettfo:
Splits the VS Code extension into two separate ones for stable and insiders.  The root of the stable extension is `src/dotnet-interactive-vscode/stable` and the root of the insiders extension is `src/dotnet-interactive-vscode/insiders`.

The vast majority of the code between the extensions is shared and that is stored in the `src/dotnet-interactive-vscode/common` directory.  To enable the build to work as expected, symlinks were created to place that directory under `src/dotnet-interactive-vscode/(stable|insiders)/src/common`.  This brings in one huge caveat: **you either need to install/fix git to support symlinks on Windows (non-default) (preferred) or you'll need to run the included `src/dotnet-interactive-vscode/ensure-symlinks.ps1` script when switching from a branch that previously didn't have the symlinks.**  By doing this there's a ton of code that we don't have to duplicate and worry about getting out-of-sync.  All of this information is in `src/dotnet-interactive-vscode/DEVGUIDE.md`.

An official build will now produce the following output:

- `artifacts/vscode`
  - `stable`
    - `dotnet-interactive-vscode-1.0.12345600.vsix`
  - `insiders`
    - `dotnet-interactive-vscode-1.0.12345601.vsix`

The version numbers are similar to what we do now, except we also append "00" or "01" to the build ID.  This is because the build ID is computed from the date by Arcade and I don't want to mess with this, but we need an insiders build to always be larger than the stable build, hence the extensions look like "xxxxxx00" for stable and "xxxxxx01" for insiders.

With that setup and with each extension targeting a different version of VS Code (`^1.53.0` for stable, `1.54.0-insider` for insiders), we can publish either the stable or insiders version and the appropriate one will get acquired by the end user, **with one caveat**: Every time we publish an extension on stable, we _must_ immediately publish the corresponding insiders extension, but that's fully automatiable.  The inverse, however, is not true; we can publish a new insiders build any time we want and users on stable won't get it because of the vs engine mismatch.

The plan with this is to release the stable version rarely, e.g., when a new version of VS Code is released, and to release the insiders version frequently (weekly?  daily?).

A new insiders-only release pipeline was created and once we verify that publish mechanism works (after merge) I'll update the stable-only release pipeline with the same changes.

[Internal build](https://dev.azure.com/dnceng/internal/_build/results?buildId=1019837&view=results) is green with the expected artifact and package content.


brettfo:
With this setup, the only duplicated code is the TextMate grammars and JS bootstrapper API, and that's only because the VS Extension packager doesn't support them, but shared unit tests should ensure that nothing deviates between the two.

jonsequitur (approved):

colombod (approved):

PR #1123, opened 2021-03-03
text fixes during extension loading

jonsequitur:
This tweaks a few of the messages from extensions.

I'm trying to make it a little clearer which extensions came from which assemblies:

image
image


brettfo (approved):

brettfo:
https://github.com/dotnet/interactive/blob/974732778592b69c526648c54ef4c48219058e97/src/Microsoft.DotNet.Interactive.ExtensionLab.Tests/SQLiteConnectionTests.cs#L55

``` diff
-.Contain("- `#!sql-mydb`");
+.Contain("    #!sql-mydb\n    SELECT TOP * FROM ...");
```

PR #1126, opened 2021-03-04
initialise signalr connection on extension load

colombod:
In a notebook without any js cell the signalr hub would not initialise.
This pr ensure that when the vscode extension is loaded and fully configured a client connection is established. This makes the  clientside kernel work in all scenarios


colombod:
@idg10 this is addressing a nasty issue when the notebook doesn't do any js cell

PR #1129, opened 2021-03-04
Upgrade to System.Reactive 5.0.0

HowardvanRooijen:
Upgrade to use the latest System.Reactive package to avoid conflicts with external libraries who also use the latest version.


colombod:
We have another one here #1131 because I file is missing in yours. Do you want to update your one to include that change or should we merge the one @brettfo create?

HowardvanRooijen:
#1131 looks better than mine!

PR #1153, opened 2021-03-10
data explorer api

colombod:


src/Microsoft.DotNet.Interactive.ExtensionLab/DataExplorer.cs
```diff
@@ -0,0 +1,74 @@
...
+using System.Threading.Tasks;
+using Microsoft.DotNet.Interactive.Commands;
+using Microsoft.DotNet.Interactive.Events;
+using Microsoft.DotNet.Interactive.Formatting;
+
+namespace Microsoft.DotNet.Interactive.ExtensionLab
+{
+    public class DataExplorer
+    {
+        public DataExplorer(TabularJsonString source)
```
jonsequitur, line 15: We should probably rename this to `TableSchemaJsonString`.

src/Microsoft.DotNet.Interactive.ExtensionLab/DataExplorer.cs
```diff
@@ -0,0 +1,74 @@
...
+        {
+            throw new NotImplementedException();
+        }
+
+        public string Id { get; }
+
+        public TabularJsonString TabularData { get; }
+    }
+
+    public class SetDataExplorerData : KernelCommand
```
jonsequitur, line 36: Why is this different from the existing display command?

src/Microsoft.DotNet.Interactive.ExtensionLab/DataExplorer.cs
```diff
@@ -0,0 +1,74 @@
...
+    {
+        [JsonPropertyName("dataExplorerId")]
+        public string DataExplorerId { get; }
+        public GetDataExplorerData(string dataExplorerId, string targetKernelName = null, KernelCommand parent = null) : base(targetKernelName, parent)
+        {
+            DataExplorerId = dataExplorerId;
+        }
+    }
+
+    public class FilteredDataProduced : KernelEvent
```
jonsequitur, line 61: Why not use the existing display and update commands?

src/Microsoft.DotNet.Interactive.ExtensionLab/SandDanceKernelExtension.cs
```diff
@@ -53,16 +53,16 @@ public static class SandDanceExplorerExtensions
 
         public static void RegisterFormatters()
         {
-            Formatter.Register<SandDanceExplorer>((explorer, writer) =>
+            Formatter.Register<DataExplorer>((explorer, writer) =>
```
jonsequitur, line 56: What if I want to use more than one different data explorer at the same time?

src/Microsoft.DotNet.Interactive.ExtensionLab/DataExplorer.cs
colombod: the goal here is to handle this command in the clientside kernel to just replace the data via api instead of re-display

colombod:
@WalternativE keep an eye on this PR as will change few things for you integration with sanddance

src/Microsoft.DotNet.Interactive.ExtensionLab/System.Collections.Generic/EnumerableExtensions.cs
```diff
@@ -2,22 +2,37 @@
...
 {
     public static class EnumerableExtensions
     {
-        public static SandDanceExplorer ExploreWithSandDance<T>(this IEnumerable<T> source)
+        public static SandDanceDataExplorer ExploreWithSandDance<T>(this IEnumerable<T> source, bool immediateDisplay = true, KernelInvocationContext invocationContext = null)
         {
-            var explorer = new SandDanceExplorer();
-            explorer.LoadData(source);
+            var explorer = new SandDanceDataExplorer(source.ToTabularDataResource());
+            if (immediateDisplay)
```
jonsequitur, line 16: This case looks like it should be a different method, or even a simple constructor.

src/Microsoft.DotNet.Interactive.ExtensionLab/TabularDataResourceExtensions.cs
```diff
@@ -0,0 +1,31 @@
...
+            {
+                (invocationContext ?? KernelInvocationContext.Current)?.Display(explorer, HtmlFormatter.MimeType);
+            }
+            return explorer;
+        }
+
+        public static NteractDataExplorer ExploreWithNteract(this TabularDataResource source, bool immediateDisplay = true, KernelInvocationContext invocationContext = null)
+        {
+            var explorer = new NteractDataExplorer(source);
+            if (immediateDisplay)
```
jonsequitur, line 24: Ditto.

src/Microsoft.DotNet.Interactive.ExtensionLab.Tests/NteractKernelExtensionTests.cs
```diff
@@ -77,7 +78,7 @@ public async Task it_checks_and_load_require()
             using var kernel = new CompositeKernel();
 
             var kernelExtension = new NteractKernelExtension();
-            DataExplorerExtensions.Settings.UseUri("https://a.cdn.url/script.js");
+            NteractDataExplorerExtensions.Settings.UseUri("https://a.cdn.url/script.js");
```
jonsequitur, line 81: Any reason to use a method instead of a setter here?

Also, thoughts on how not to make this state global? Can we close over this in the middleware instead, for example? Is changing it once it's set a use case? Is multiple explorers from different CDNs a use case?

src/Microsoft.DotNet.Interactive.ExtensionLab/System.Collections.Generic/EnumerableExtensions.cs
colombod: Thank you! I missed the point here!

src/Microsoft.DotNet.Interactive.ExtensionLab/TabularDataResourceExtensions.cs
colombod: same here

src/Microsoft.DotNet.Interactive.ExtensionLab.Tests/NteractKernelExtensionTests.cs
colombod: the Settings class is gone and configuration is done via the Use* api

jonsequitur (approved):

PR #1157, opened 2021-03-11
allow js execution to return events to be published

brettfo:
This allows the capture of `console.log`, et al., as well as wrapping submitted code in `try/catch` and properly failing the command and proper execution time tracking.

Execution time and `console.log()`:
image

Properly fail on exception:
image


src/Microsoft.DotNet.Interactive.Http/AspNetExtensions.cs
```diff
@@ -82,12 +82,14 @@ public static IServiceCollection AddDotnetInteractiveHttpApi(this IServiceCollec
             {
                 r.Routes.Add(new VariableRouter(kernel));
                 r.Routes.Add(new KernelsRouter(kernel));
-                var htmlNotebookFrontedEnvironment = kernel.FrontendEnvironment as HtmlNotebookFrontedEnvironment;
```
colombod, line 85: when renaming types because of spelling errors you should condier the feelings onf the person that did tath typo

src/Microsoft.DotNet.Interactive.Js/package.json
```diff
@@ -11,7 +11,8 @@
     "test": "mocha --opts mocha.opts",
     "ciTest": "npm test -- --reporter mocha-multi-reporters --reporter-options configFile=testConfig.json",
     "rollup-dotnet-interactive": "npm run build-ci -- -i src/dotnet-interactive.ts -o dist/dotnet-interactive.js",
-    "rollup-dotnet-interactive-code-mirror": "npm run build-ci -- -i src/dotnet-interactive-code-mirror.ts -o dist/dotnet-interactive-code-mirror.js"
+    "rollup-dotnet-interactive-code-mirror": "npm run build-ci -- -i src/dotnet-interactive-code-mirror.ts -o dist/dotnet-interactive-code-mirror.js",
```
colombod, line 14: we should remove all the code mirror stuff we left behind :)

colombod (approved):

src/Microsoft.DotNet.Interactive/FrontendEnvironment.cs
```diff
@@ -1,11 +1,17 @@
...
+using System.Threading.Tasks;
 
 namespace Microsoft.DotNet.Interactive
 {
-    public abstract class FrontendEnvironment        
+    public abstract class FrontendEnvironment
     {
         public bool AllowStandardInput { get; set; }
+
+        public virtual Task ExecuteClientScript(string code, KernelInvocationContext context)
```
brettfo, line 12: This method might not be needed here; I think it's only ever consumed directly off of `HtmlNotebookFrontendEnvironment`.

brettfo: Just kidding, it's used in `JavascriptKernel`.  We can either leave the method here, or check for a strong type in the JS kernel.

PR #1163, opened 2021-03-15
pass reference assemblies to interactive workspace to enable bcl doc comments

### LINKED ISSUE #1072: Enable doc comments from BCL/Framework types

brettfo:
The `CSharpScript` object only tracks runtime assemblies, so passing those to the interactive workspace meant that any types that get forwarded to `System.Private.CoreLib.dll` (e.g., `System.Environment` defined in `System.Runtime.dll`/`System.Runtime.xml` is implemented in `System.Private.CoreLib.dll` which doesn't have a corresponding `System.Private.CoreLib.xml` file) won't have proper doc comments, because the defining ref assembly and the xml doc comment file names don't match.

The fix was to:

1. **Find and enumerate all reference assemblies and use those for the workspace.**  The reference assembly location is computed from the runtime location.  This same path calculation is used internally in F# and if at any point the path calculation fails, the runtime assemblies will be used.  In this scenario, doc comments won't be loaded, but the end user will still get parameter names and basic signature help, so they're not left completely in the cold.
2. **Combine the assemblies from (1) with anything referenced via explicit `#r "path/to/assembly.dll"`**, and
3. **all references from NuGet packages.**

Fixes #1072.

### DIFF
--- a/src/Microsoft.DotNet.Interactive.CSharp/InteractiveWorkspace.cs
+++ b/src/Microsoft.DotNet.Interactive.CSharp/InteractiveWorkspace.cs
@@ -119,7 +119,7 @@ namespace Microsoft.DotNet.Interactive.CSharp
             return solution;
         }
 
-        private IEnumerable<MetadataReference> GetReferenceSet(Compilation compilation)
+        private IReadOnlyCollection<MetadataReference> GetReferenceSet(Compilation compilation)
         {
             var references =
                 _referenceAssemblies
--- a/src/Microsoft.DotNet.Interactive.CSharp/InteractiveWorkspace.cs
+++ b/src/Microsoft.DotNet.Interactive.CSharp/InteractiveWorkspace.cs
@@ -1,7 +1,10 @@
 ﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System;
 using System.Collections.Generic;
+using System.IO;
+using System.Linq;
 using System.Reactive.Disposables;
 
 using Microsoft.CodeAnalysis;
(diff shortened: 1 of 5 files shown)


colombod (approved):

src/Microsoft.DotNet.Interactive.CSharp/InteractiveWorkspace.cs
```diff
@@ -59,6 +119,16 @@ private Solution RemoveWorkingProjectFromSolution(Solution solution)
             return solution;
         }
 
+        private IEnumerable<MetadataReference> GetReferenceSet(Compilation compilation)
```
jonsequitur, line 122: Since you're `ToArray`ing this you might as well return `IReadOnlyList<MetadataReference>`.

brettfo: Done.

src/Microsoft.DotNet.Interactive.Tests/LanguageServices/SignatureHelpTests.cs
```diff
@@ -81,6 +82,35 @@ public async Task signature_help_can_handle_language_switching_and_offsets()
...
+                .Subject
+                .LastOrDefault()
+                .Should()
+                .NotBeNull()
+                .And
+                .Subject
+                .As<SignatureInformation>()
+                .Documentation.Value
+                .Should()
+                .Contain(expectedDocumentationSubstring);
```
jonsequitur, line 111: `assertion.Should().BeReallyLong()`

jonsequitur (approved):

src/Microsoft.DotNet.Interactive.Tests/LanguageServices/SignatureHelpTests.cs
brettfo: I'll try to add a few GUIDs to make it conform to corporate standards.

PR #1164, opened 2021-03-15
hang individual tests after 5 minutes

brettfo:
Also remove unnecessary build script redirection.

Created as a draft to run repeatedly.


eng/Versions.props
```diff
@@ -5,7 +5,7 @@
     <!-- opt-in properties -->
     <UsingToolNuGetRepack>true</UsingToolNuGetRepack>
     <UsingToolSymbolUploader>true</UsingToolSymbolUploader>
-    <TestRunnerAdditionalArguments>-parallel none</TestRunnerAdditionalArguments>
+    <TestRunnerAdditionalArguments>-parallel none --blame-hang-timeout 20m</TestRunnerAdditionalArguments>
```
jonsequitur, line 8: If this applies to individual tests then I'd go down to `5m` or less.

PR #1167, opened 2021-03-17
allow tests to be skipped in signed build

brettfo:
To make testing build changes faster, I added a variable `SkipTests` to the signed build definition that defaults to `false`.  If set to `true` the obvious will happen.  This PR is to ensure I didn't break PR builds.

The behavior of `SkipTests=false` being the default instead of `RunTests=true` is because the PR build doesn't define that variable and I want the skipping of tests to be very explicit.  If instead the condition was (pseudo) `if $(RunTests) == 'true' ...` then if the variable was ever removed or renamed, tests would never be run.  I'd rather err on the side of running tests.

Marked as draft until internal builds (one with `SkipTests` as default `false`, one with `true`) finish with expected results.

I also dropped the Windows leg timeout down to 90 minutes; 5 hours was too much given the hanging tests we're dealing with, but 90 minutes should give us plenty of time to finish, even if it's a slow run.

Edit:
Internal builds behaved as expected.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -45,7 +45,7 @@ stages:
           ${{ if and(ne(variables['System.TeamProject'], 'public'), notin(variables['Build.Reason'], 'PullRequest')) }}:
             name: NetCoreInternal-Pool
             queue: buildpool.windows.10.amd64.vs2019
-        timeoutInMinutes: 300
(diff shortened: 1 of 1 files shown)


jonsequitur (approved):

brettfo:
PR build behaved as expected, will merge.

Issue #1146, opened 2021-03-08
If tool feed isn't available through VPN/proxy, extension fails to create/open notebook.


brettfo:
As the investigation in #1137 confirmed, if a VPN/proxy/firewall is preventing access to the default tool feed `https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-tools/nuget/v3/index.json` then notebooks cannot be opened or created because the required tool couldn't be installed.  If by chance a newer version of the tool exists on a feed that _is_ available, however, e.g., NuGet.org, then the user can proceed.

We'll need a way to better detect and communicate to the user that network settings may be preventing the tool from being acquired.

PR #1171, opened 2021-03-19
Fix explorer rendering

colombod:
Addressing #1146.

The goal is to have immediate rendering on ExploreWith*() and then retain the Explorer object to use its api (data update and more) 
 
cc @WalternativE


src/Microsoft.DotNet.Interactive.ExtensionLab/NteractDataExplorerExtensions.cs
```diff
@@ -0,0 +1,91 @@
...
+using System.Text;
+using Microsoft.AspNetCore.Html;
+using Microsoft.DotNet.Interactive.Formatting;
+using Microsoft.DotNet.Interactive.Http;
+
+namespace Microsoft.DotNet.Interactive.ExtensionLab
+{
+    public static class NteractDataExplorerExtensions
+    {
+        public static TabularDataResourceExplorerView<NteractDataExplorer> ToView(this NteractDataExplorer explorer)
```
jonsequitur, line 14: Let's try to get rid of this method. It looks like it should just be an implementation detail of the formatter registrations.

src/Microsoft.DotNet.Interactive.ExtensionLab/NteractDataExplorerExtensions.cs
```diff
@@ -0,0 +1,91 @@
...
+            code.AppendLine($"<div id=\"{divId}\" style=\"height: 100ch ;margin: 2px;\">");
+            code.AppendLine("</div>");
+            code.AppendLine(@"<script type=""text/javascript"">");
+            GenerateCode(data, code, divId, "https://cdnjs.cloudflare.com/ajax/libs/require.js/2.3.6/require.min.js", uri, context, cacheBuster);
+            code.AppendLine(" </script>");
+            code.AppendLine("</div>");
+            return new HtmlString(code.ToString());
+        }
+
+        private static void GenerateCode(TabularDataResourceJsonString data, StringBuilder code, string divId, string requireUri, Uri uri, string context, string cacheBuster)
```
jonsequitur, line 49: `Generate*` implies that these methods perform some side effect. The intention of the code can be made clearer by refactoring them to extension methods on the `StringBuilder`, e.g. `AppendRequireLoader`, etc.

src/Microsoft.DotNet.Interactive.ExtensionLab/TabularDataResourceExplorerView.cs
```diff
@@ -0,0 +1,21 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using Microsoft.DotNet.Interactive.Formatting;
+
+namespace Microsoft.DotNet.Interactive.ExtensionLab
+{
+    public class TabularDataResourceExplorerView<T> : ITabularDataResourceExplorer
```
jonsequitur, line 9: I suspect we can do away with this type.

PR #1176, opened 2021-03-21
Load a csx file from relative path into Jupyter Notebook, fixes #478

### LINKED ISSUE #478: Load a csx file from relative path into Jupyter Notebook doesn't work
I try to load a csx file into Jupyter Notebook and it throws the next error:
error CS1504: Source file 'Models.cs' could not be opened -- Could not find file.
The relative path doesn't work for #load directive, but the absolute path is ok.

E.g. Having a Jupyter notebook like this: C:/path/to/myNotebook.ipynb:
``#load "C:/path/to/file.csx" works``
``#load "file.csx" doesn't work``
``#load "./file.csx" doesn't work ``
``#load "../to/file.csx" doesn't work``

.NET Interactive 1.0.115407

- OS
    - [x] Windows 10
    - [ ] macOS
    - [ ] iOS
    - [ ] Android
 - Browser 
    - [x] Chrome
    - [ ] Edge
    - [x] Firefox
    - [ ] Safari
 - Frontend
    - [x] Jupyter Notebook
    - [x] Jupyter Lab
    - [ ] nteract
    - [ ] Visual Studio Code
    - [ ] Other (please specify)

strabu:
Makes this possible: https://github.com/strabu/dotnet-interactive-issue-repro

Would be great if we could have relative path-imports in VSCode-Notebooks!

greetings,
Alex

### DIFF
--- a/src/Microsoft.DotNet.Interactive.CSharp/CSharpKernel.cs
+++ b/src/Microsoft.DotNet.Interactive.CSharp/CSharpKernel.cs
@@ -53,6 +53,7 @@ namespace Microsoft.DotNet.Interactive.CSharp
         internal ScriptOptions ScriptOptions =
             ScriptOptions.Default
                          .WithMetadataResolver(CachingMetadataResolver.Default.WithBaseDirectory(Directory.GetCurrentDirectory()))
+                         .WithSourceResolver(new SourceFileResolver(ImmutableArray<string>.Empty, Directory.GetCurrentDirectory()))
                          .WithLanguageVersion(LanguageVersion.Latest)
                          .AddImports(
                              "System",
@@ -315,9 +316,9 @@ namespace Microsoft.DotNet.Interactive.CSharp
             if (_currentDirectory != currentDirectory)
             {
                 _currentDirectory = currentDirectory;
-                ScriptOptions = ScriptOptions.WithMetadataResolver(
-                    CachingMetadataResolver.Default.WithBaseDirectory(
-                        _currentDirectory));
+                ScriptOptions = ScriptOptions
+                    .WithMetadataResolver(CachingMetadataResolver.Default.WithBaseDirectory(_currentDirectory))
+                    .WithSourceResolver(new SourceFileResolver(ImmutableArray<string>.Empty, _currentDirectory));
             }
 
             if (ScriptState is null)
--- a/src/Microsoft.DotNet.Interactive.CSharp/CSharpKernel.cs
+++ b/src/Microsoft.DotNet.Interactive.CSharp/CSharpKernel.cs
@@ -36,7 +36,8 @@ namespace Microsoft.DotNet.Interactive.CSharp
         IKernelCommandHandler<RequestDiagnostics>,
         IKernelCommandHandler<RequestHoverText>,
         IKernelCommandHandler<RequestSignatureHelp>,
-        IKernelCommandHandler<SubmitCode>
+        IKernelCommandHandler<SubmitCode>,
+        IKernelCommandHandler<ChangeWorkingDirectory>
     {
         internal const string DefaultKernelName = "csharp";
 
@@ -50,10 +51,20 @@ namespace Microsoft.DotNet.Interactive.CSharp
 
         private Lazy<PackageRestoreContext> _packageRestoreContext;
 
-        internal ScriptOptions ScriptOptions =
-            ScriptOptions.Default
-                         .WithMetadataResolver(CachingMetadataResolver.Default.WithBaseDirectory(Directory.GetCurrentDirectory()))
-                         .WithSourceResolver(new SourceFileResolver(ImmutableArray<string>.Empty, Directory.GetCurrentDirectory()))
+        internal ScriptOptions ScriptOptions;
+
+        private readonly AssemblyBasedExtensionLoader _extensionLoader = new AssemblyBasedExtensionLoader();
+
+        private string _workingDirectory;
+
+        public CSharpKernel() : base(DefaultKernelName)
+        {
+            _workspace = new InteractiveWorkspace();
+
+            //For the VSCode-Add-In Directory.GetCurrentDirectory() would here return something like: c:\Users\<username>\AppData\Roaming\Code\User\globalStorage\ms-dotnettools.dotnet-interactive-vscode
+            //...so we wait for RunAsync to read Directory.GetCurrentDirectory() the first time.
+
+            ScriptOptions = ScriptOptions.Default
                          .WithLanguageVersion(LanguageVersion.Latest)
                          .AddImports(
                              "System",
@@ -69,14 +80,6 @@ namespace Microsoft.DotNet.Interactive.CSharp
                              typeof(Kernel).Assembly,
                              typeof(CSharpKernel).Assembly,
                              typeof(PocketView).Assembly);
-
-        private readonly AssemblyBasedExtensionLoader _extensionLoader = new();
-        private string _currentDirectory;
-
-        public CSharpKernel() : base(DefaultKernelName)
-        {
-            _workspace = new InteractiveWorkspace();
-            _currentDirectory = Directory.GetCurrentDirectory();
 
             _packageRestoreContext = new Lazy<PackageRestoreContext>(() =>
             {
@@ -177,7 +180,7 @@ namespace Microsoft.DotNet.Interactive.CSharp
         {
             await EnsureWorkspaceIsInitializedAsync(context);
 
-            var document = _workspace.ForkDocumentForLanguageServices(command.Code); 
+            var document = _workspace.ForkDocumentForLanguageServices(command.Code);
             var signatureHelp = await SignatureHelpGenerator.GenerateSignatureInformation(document, command, context.CancellationToken);
             if (signatureHelp is { })
             {
@@ -258,7 +261,7 @@ namespace Microsoft.DotNet.Interactive.CSharp
                 if (exception is CodeSubmissionCompilationErrorException { InnerException: CompilationErrorException innerCompilationException })
                 {
                     diagnostics = innerCompilationException.Diagnostics;
-                    // In the case of an error the diagnostics get attached to both the 
+                    // In the case of an error the diagnostics get attached to both the
                     // DiagnosticsProduced and CommandFailed events.
                     message =
                         string.Join(Environment.NewLine,
@@ -307,19 +310,23 @@ namespace Microsoft.DotNet.Interactive.CSharp
             }
         }
 
+        public Task HandleAsync(ChangeWorkingDirectory command, KernelInvocationContext context)
+        {
+            _workingDirectory = command.WorkingDirectory;
+            return Task.CompletedTask;
+        }
+
         private async Task RunAsync(
             string code,
             CancellationToken cancellationToken = default,
             Func<Exception, bool> catchException = default)
         {
-            var currentDirectory = Directory.GetCurrentDirectory();
-            if (_currentDirectory != currentDirectory)
-            {
-                _currentDirectory = currentDirectory;
-                ScriptOptions = ScriptOptions
-                    .WithMetadataResolver(CachingMetadataResolver.Default.WithBaseDirectory(_currentDirectory))
-                    .WithSourceResolver(new SourceFileResolver(ImmutableArray<string>.Empty, _currentDirectory));
-            }
+            if (_workingDirectory == null)
+                _workingDirectory = Directory.GetCurrentDirectory();
+
+            ScriptOptions = ScriptOptions
+                .WithMetadataResolver(CachingMetadataResolver.Default.WithBaseDirectory(_workingDirectory))
+                .WithSourceResolver(new SourceFileResolver(ImmutableArray<string>.Empty, _workingDirectory));
 
             if (ScriptState is null)
             {
@@ -374,7 +381,7 @@ namespace Microsoft.DotNet.Interactive.CSharp
             var document = _workspace.ForkDocumentForLanguageServices(code);
             var service = CompletionService.GetService(document);
             var completionList = await service.GetCompletionsAsync(document, cursorPosition, cancellationToken: contextCancellationToken);
-           
+
             if (completionList is null)
             {
                 return Enumerable.Empty<CompletionItem>();
--- a/src/Microsoft.DotNet.Interactive.CSharp/CSharpKernel.cs
+++ b/src/Microsoft.DotNet.Interactive.CSharp/CSharpKernel.cs
@@ -37,7 +37,7 @@ namespace Microsoft.DotNet.Interactive.CSharp
         IKernelCommandHandler<RequestHoverText>,
         IKernelCommandHandler<RequestSignatureHelp>,
         IKernelCommandHandler<SubmitCode>,
-        IKernelCommandHandler<ChangeWorkingDirectory>
+        IKernelCommandHandler<ChangeWorkspaceDirectory>
     {
         internal const string DefaultKernelName = "csharp";
 
@@ -55,7 +55,7 @@ namespace Microsoft.DotNet.Interactive.CSharp
 
         private readonly AssemblyBasedExtensionLoader _extensionLoader = new AssemblyBasedExtensionLoader();
 
-        private string _workingDirectory;
+        private string _workspaceDirectory;
 
         public CSharpKernel() : base(DefaultKernelName)
         {
@@ -310,9 +310,9 @@ namespace Microsoft.DotNet.Interactive.CSharp
             }
         }
 
-        public Task HandleAsync(ChangeWorkingDirectory command, KernelInvocationContext context)
+        public Task HandleAsync(ChangeWorkspaceDirectory command, KernelInvocationContext context)
         {
-            _workingDirectory = command.WorkingDirectory;
+            _workspaceDirectory = command.WorkspaceDirectory;
             return Task.CompletedTask;
         }
 
@@ -321,12 +321,12 @@ namespace Microsoft.DotNet.Interactive.CSharp
             CancellationToken cancellationToken = default,
             Func<Exception, bool> catchException = default)
         {
-            if (_workingDirectory == null)
-                _workingDirectory = Directory.GetCurrentDirectory();
+            if (_workspaceDirectory == null)
+                _workspaceDirectory = Directory.GetCurrentDirectory();
 
             ScriptOptions = ScriptOptions
-                .WithMetadataResolver(CachingMetadataResolver.Default.WithBaseDirectory(_workingDirectory))
-                .WithSourceResolver(new SourceFileResolver(ImmutableArray<string>.Empty, _workingDirectory));
+                .WithMetadataResolver(CachingMetadataResolver.Default.WithBaseDirectory(_workspaceDirectory))
+                .WithSourceResolver(new SourceFileResolver(ImmutableArray<string>.Empty, _workspaceDirectory));
 
             if (ScriptState is null)
             {
--- a/src/Microsoft.DotNet.Interactive.CSharp/CSharpKernel.cs
+++ b/src/Microsoft.DotNet.Interactive.CSharp/CSharpKernel.cs
@@ -37,7 +37,7 @@ namespace Microsoft.DotNet.Interactive.CSharp
         IKernelCommandHandler<RequestHoverText>,
         IKernelCommandHandler<RequestSignatureHelp>,
         IKernelCommandHandler<SubmitCode>,
-        IKernelCommandHandler<ChangeWorkspaceDirectory>
+        IKernelCommandHandler<ChangeWorkingDirectory>
     {
         internal const string DefaultKernelName = "csharp";
 
@@ -55,7 +55,7 @@ namespace Microsoft.DotNet.Interactive.CSharp
 
         private readonly AssemblyBasedExtensionLoader _extensionLoader = new AssemblyBasedExtensionLoader();
 
-        private string _workspaceDirectory;
+        private string _workingDirectory;
 
         public CSharpKernel() : base(DefaultKernelName)
         {
@@ -310,9 +310,9 @@ namespace Microsoft.DotNet.Interactive.CSharp
             }
         }
 
-        public Task HandleAsync(ChangeWorkspaceDirectory command, KernelInvocationContext context)
+        public Task HandleAsync(ChangeWorkingDirectory command, KernelInvocationContext context)
         {
-            _workspaceDirectory = command.WorkspaceDirectory;
+            _workingDirectory = command.WorkingDirectory;
             return Task.CompletedTask;
         }
 
@@ -321,12 +321,12 @@ namespace Microsoft.DotNet.Interactive.CSharp
             CancellationToken cancellationToken = default,
             Func<Exception, bool> catchException = default)
         {
-            if (_workspaceDirectory == null)
-                _workspaceDirectory = Directory.GetCurrentDirectory();
+            if (_workingDirectory == null)
+                _workingDirectory = Directory.GetCurrentDirectory();
 
             ScriptOptions = ScriptOptions
-                .WithMetadataResolver(CachingMetadataResolver.Default.WithBaseDirectory(_workspaceDirectory))
-                .WithSourceResolver(new SourceFileResolver(ImmutableArray<string>.Empty, _workspaceDirectory));
+                .WithMetadataResolver(CachingMetadataResolver.Default.WithBaseDirectory(_workingDirectory))
+                .WithSourceResolver(new SourceFileResolver(ImmutableArray<string>.Empty, _workingDirectory));
 
             if (ScriptState is null)
             {
--- a/src/Microsoft.DotNet.Interactive.Tests/LanguageKernelPackageTests.cs
+++ b/src/Microsoft.DotNet.Interactive.Tests/LanguageKernelPackageTests.cs
@@ -190,7 +190,7 @@ json"
 
             var kernel = CreateKernel(language);
             var absolutePathOneLevelHigher = Directory.GetParent(currentDirectory).FullName;
-            await kernel.SendAsync(new ChangeWorkingDirectory(absolutePathOneLevelHigher));        
+            await kernel.SendAsync(new ChangeWorkspaceDirectory(absolutePathOneLevelHigher));        
 
             await kernel.SendAsync(new SubmitCode("Environment.CurrentDirectory"));
 
--- a/src/Microsoft.DotNet.Interactive.Tests/LanguageKernelPackageTests.cs
+++ b/src/Microsoft.DotNet.Interactive.Tests/LanguageKernelPackageTests.cs
@@ -158,6 +158,45 @@ json"
         }
 
         [Theory]
+        [InlineData(Language.CSharp, false)]
+        [InlineData(Language.FSharp, false)]
+        [InlineData(Language.CSharp, true)]
+        public async Task it_can_load_script_files_using_load_directive_with_relative_path(Language language, bool changeWorkingDirectory)
+        {
+            var srcDir = DirectoryUtility.GetPathToSrcDirectory();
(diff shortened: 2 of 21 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/interactive?pullRequest=1176) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/interactive?pullRequest=1176) before we can accept your contribution.<br/><br/>:x: strabu [sign now](https://cla.dotnetfoundation.org/dotnet/interactive?pullRequest=1176)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/interactive?pullRequest=1176) it.</sub>

strabu:
I broke the Linux-Build with my first commit and fixed/tested it on my MacBook later with https://github.com/dotnet/interactive/pull/1176/commits/5a5a652d6f67e1ed1e05eed28061c4216a188b0d

src/Microsoft.DotNet.Interactive.Tests/LanguageKernelPackageTests.cs
```diff
@@ -164,7 +164,7 @@ public async Task it_can_load_assembly_references_using_r_directive_with_relativ
         public async Task it_can_load_script_files_using_load_directive_with_relative_path(Language language, bool changeWorkingDirectory)
         {
             var srcDir = DirectoryUtility.GetPathToSrcDirectory();
-            var pathToScripts = Path.Combine(srcDir, @"Microsoft.DotNet.Interactive.Tests\");
+            var pathToScripts = Path.Combine(srcDir, @"Microsoft.DotNet.Interactive.Tests"+Path.DirectorySeparatorChar);
```
colombod, line 167: do we need the trailing `'`'` at all?

src/Microsoft.DotNet.Interactive.Tests/Utility/DirectoryUtility.cs
```diff
@@ -61,9 +61,10 @@ public static class DirectoryUtility
 
         public static string GetPathToSrcDirectory()
         {
-            string pathAssembly = System.Reflection.Assembly.GetExecutingAssembly().Location;            
-            string folderAssembly = Path.GetDirectoryName(pathAssembly);            
-            return Path.GetFullPath(folderAssembly + @"..\..\..\..\..\..\src");            
+            string pathAssembly = System.Reflection.Assembly.GetExecutingAssembly().Location;
+            string folderAssembly = Path.GetDirectoryName(pathAssembly);
+            var ps = Path.DirectorySeparatorChar;
+            return Path.GetFullPath(folderAssembly + $"..{ps}..{ps}..{ps}..{ps}..{ps}..{ps}src");
```
colombod, line 67: This is safer with
```csharp
return System.IO.Path.Combine(folderAssembly ,"..","..","..","..","..","..","src");
```

strabu: Yes, I agree. This looks much better than the string-interpolations.

jonsequitur (approved):

src/Microsoft.DotNet.Interactive.Tests/Utility/DirectoryUtility.cs
strabu: it had to put this to make the 3 related tests succeed:
count the dots :)

return Path.Combine(folderAssembly, "..", "..", "..", "..", "..", "src");

src/Microsoft.DotNet.Interactive.Tests/LanguageKernelPackageTests.cs
strabu: no, you are right.
Probably it would be better to put a forward-slash in these 2 lines before "RelativeLoadingSample":

https://github.com/dotnet/interactive/blob/5a5a652d6f67e1ed1e05eed28061c4216a188b0d/src/Microsoft.DotNet.Interactive.Tests/LanguageKernelPackageTests.cs#L188-L189

src/Microsoft.DotNet.Interactive.Tests/Utility/DirectoryUtility.cs
```diff
@@ -58,5 +58,12 @@ public static class DirectoryUtility
                     t.content);
             }
         }
+
+        public static string GetPathToSrcDirectory()
+        {
+            string pathAssembly = System.Reflection.Assembly.GetExecutingAssembly().Location;
+            string folderAssembly = Path.GetDirectoryName(pathAssembly);            
+            return Path.Combine(folderAssembly, "..", "..", "..", "..", "..", "src");
```
brettfo, line 66: I'd feel more comfortable if the .csx/.fsx files were added as `<Content>` in the .csproj and copied to the output directory, instead.  With this layout anybody that builds and tests with the env var `%DISABLEARCADE%=1` will be broken, aka, @colombod and @jonsequitur.

jonsequitur: Or better, just emit the file during the test run, so the total code for the test case is more cohesive.

strabu: please have a look at this: https://github.com/dotnet/interactive/pull/1176/commits/433e8f6991745e7a7ffe7ba4d388de27f4fc506c

src/Microsoft.DotNet.Interactive.Tests/Utility/DirectoryUtility.cs
strabu: We got rid of the whole method here: https://github.com/dotnet/interactive/pull/1176/commits/433e8f6991745e7a7ffe7ba4d388de27f4fc506c

colombod (approved):

strabu:
@colombod sorry, clicked the "re-request-button" by accident

src/Microsoft.DotNet.Interactive.CSharp/CSharpKernel.cs
```diff
@@ -298,9 +299,9 @@ public async Task HandleAsync(SubmitCode submitCode, KernelInvocationContext con
             if (_currentDirectory != currentDirectory)
             {
```
KevinRansom, line 300: I can't get my head around the fact that this code exists in the code base (knowing my luck I could even have added it :-)).  A different thread can call SetWorkingDirectory between 298 and 300, I am assuming this is intended to maintain some sort of invariant that the cache and resolver use the current working directory.

Anything that changes "the current working directory for the notebook needs to be independent of Directory.GetWorkingDirectory()/Directory.SetWorkingDirectory().  A command to change the working directory seems like the right way to go, and that command need not ... probably (definitely in my head) should not invoke Directory.SetWorkingDirectory( ... ).

I don't believe there is any invariant that we can maintain around Directory.GetWorkingDirectory() returning us anything "notebook usefull".  It should ideally be reserved for script authors and perhaps interoperability with third party stuff that relies on it.

jonsequitur: I think we need a separate concept which is the notebook directory, which is not mutable at runtime.

KevinRansom: I'm okay with it being mutable, so long as the mutability is controlled.  The problem with Directory.GetCurrentDirectory() is that there are a bunch of ways of changing it, and we only see the effects.  So a command to change the Project Root or whatever we decide to call it is fine, probably even necessary for example when an IDE implements the "save as" gesture.

jonsequitur: We have a command for this currently: [ChangeWorkingDirectory](https://github.com/dotnet/interactive/blob/8438829658e70de5ff20a172748cc10f8654d1da/src/Microsoft.DotNet.Interactive/Commands/ChangeWorkingDirectory.cs#L10-L10), which is used as you're describing by the VS Code extension. This might be the wrong name for it.

src/Microsoft.DotNet.Interactive.CSharp/CSharpKernel.cs
```diff
@@ -53,6 +53,7 @@ public class CSharpKernel :
         internal ScriptOptions ScriptOptions =
             ScriptOptions.Default
                          .WithMetadataResolver(CachingMetadataResolver.Default.WithBaseDirectory(Directory.GetCurrentDirectory()))
```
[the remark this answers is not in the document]
strabu: Yes, calling Directory.GetCurrentDirectory() twice could lead to different results.
We could initialize ScriptOptions in the Constructor and not worry about it any longer.
Shall we?

strabu:
To add one more thing to the WorkingDirectory-discussion, this (see screenshot) is what I can do with the change at the moment in VSCode.

Actually as a Notebook-User I would probably not want that behaviour. 
I'd prefer my "magic" #load-Directive to give me the same result relative to my .ipynb-File, no matter what the CurrentDirectory is.

When I open VSCode one folder-level higher my "#load"-directives don't work any more with the current implementation.
I would have to change it and put the missing folder after "#load".

What I want as a user is: open my notebook with VSCode on any level (maybe it's in a monorepo some folders below)
and I want my Notebook to work. It should not be necessary for me as a consumer of the notebook, to know on which level I have to "Open Folder..." to make it work.

In my opinion these relative Paths have to be based on the location where the current Notebook-file is.
Is there a way to make it relative to .ipynb?

image

src/Microsoft.DotNet.Interactive.CSharp/CSharpKernel.cs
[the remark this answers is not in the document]
jonsequitur: It's intentional that this (let's not call it "working directory") can be changed after the kernel is already up and running, though this should really be a host gesture (e.g. [ChangeWorkingDirectory](https://github.com/dotnet/interactive/blob/8438829658e70de5ff20a172748cc10f8654d1da/src/Microsoft.DotNet.Interactive/Commands/ChangeWorkingDirectory.cs#L10-L10) sent by the VS Code extension, and not a user gesture.

src/Microsoft.DotNet.Interactive.CSharp/CSharpKernel.cs
strabu: Maybe "ScriptRootDirectory" would be a good name for it. From the perspective of the VS Code extension you would understand what it will do.
And it should probably not change the current WorkingDirectory is that correct?

jonsequitur: The working directory should default to this location as it does today, but it would not change when the user changes the working directory.

strabu: I've pushed another change that handles the ChangeWorkingDirectory-Command.

In the VS Code extension it's now possible to do:  Directory.SetCurrentDirectory("..")
but the "#load"-directive is not affected by that.

src/Microsoft.DotNet.Interactive.Tests/LanguageKernelPackageTests.cs
```diff
@@ -116,41 +116,93 @@ public async Task it_can_load_assembly_references_using_r_directive_separate_sub
...
-            var code = language switch
+            KernelEvents.Should()
+                        .ContainSingle<CommandSucceeded>(c => c.Command == command);
+        }
+
+        [Theory]
+        [InlineData(Language.CSharp, false)]
+        [InlineData(Language.FSharp, false)]
+        [InlineData(Language.CSharp, true)]
+        public async Task it_can_load_script_files_using_load_directive_with_relative_path(Language language, bool changeCurrentDirectoryInUserCode)
```
jonsequitur, line 152: Can we add a `changeCurrentDirectoryInUserCode == true` case for F#?

strabu: I'll also put an IKernelCommandHandler for ChangeWorkingDirectory into FSharpKernel.fs because this also needs to be handled on that side.

I couldn't figure out how to change the working directory in FSharp-Script. That's why FSharp is commented for this test: "it_can_load_script_files_using_load_directive_with_relative_path_after_command_changeWorkingDirectory"

Is there somebody else who can have a look at the FSharp-Part?
Unfortunately I don't know a lot about FSharp yet.

strabu: I was hesitating when it came to renaming "WorkingDirectory" because I was worrying to break the Contract to the Typescript-world, I don't know how you are auto-generating the Typescript-Contract and what I would break on this side.
But finally I renamed "WorkingDirectory" to "WorkspaceDirectory" where appropriate.
The CSharpKernel uses an "InteractiveWorkspace" so this seemed like a good choice to me.

The VSCode-Add-In still works after that change. I believe the Command "ChangeWorkingDirectory" was not sent from there so far.

jonsequitur: Sorry, I was being unclear. We shouldn't change the JSON API contracts. As long as we don't rename the command, we avoid that. I was really trying to work out what a better name for this concept would be and we can plan a contract change for later if we come up with something we like. 

For F# help, we can summon @KevinRansom or @brettfo. For the TypeScript-y bits, we can summon @brettfo.

src/Microsoft.DotNet.Interactive.CSharp/CSharpKernel.cs
```diff
@@ -286,19 +290,23 @@ public async Task HandleAsync(SubmitCode submitCode, KernelInvocationContext con
...
-                ScriptOptions = ScriptOptions.WithMetadataResolver(
-                    CachingMetadataResolver.Default.WithBaseDirectory(
-                        _currentDirectory));
-            }
+            if (_workspaceDirectory == null)
+                _workspaceDirectory = Directory.GetCurrentDirectory();
+
+            ScriptOptions = ScriptOptions
+                .WithMetadataResolver(CachingMetadataResolver.Default.WithBaseDirectory(_workspaceDirectory))
+                .WithSourceResolver(new SourceFileResolver(ImmutableArray<string>.Empty, _workspaceDirectory));
```
brettfo, line 309: I'd like to avoid setting the `_workspaceDirectory` variable unless we explicitly enter the `HandleAsync()` method that you added.  I'd prefer something like this:

``` csharp
var workingDirectory = _workspaceDirectory ?? Directory.GetCurrentDirectory();

ScriptOptions = ScriptOptions
    .WithMetadataResolver(CachingMetadataResolver.Default.WithBaseDirectory(workingDirectory))
    .WithSourceResolver(new SourceFileResolver(ImmutableArray<string>.Emtpy, workingDirectory));
```

Same change applied to the F# kernel, too.

src/Microsoft.DotNet.Interactive.CSharp/CSharpKernel.cs
```diff
@@ -430,7 +438,7 @@ void ISupportNuget.RegisterResolvedPackageReferences(IReadOnlyList<ResolvedPacka
             ScriptOptions = ScriptOptions.AddReferences(references);
         }
 
-        Task<PackageRestoreResult> ISupportNuget.RestoreAsync() => _packageRestoreContext.Value.RestoreAsync();
+        Task<PackageRestoreResult> ISupportNuget.RestoreAsync() => _packageRestoreContext.Value.RestoreAsync();
```
brettfo, line 441: Nit: extra whitespace.

src/Microsoft.DotNet.Interactive.Tests/LanguageKernelPackageTests.cs
```diff
@@ -116,41 +116,94 @@ public async Task it_can_load_assembly_references_using_r_directive_separate_sub
...
 
-            if (changeWorkingDirectory)
+            if (changeCurrentDirectoryInUserCode)
             {
+                //Even when a user changes the current directory, loading from a relative path is not affected.
                 await kernel.SendAsync(new SubmitCode("System.IO.Directory.SetCurrentDirectory(\"..\")"));
             }
 
-            var fullName = new FileInfo(typeof(JsonConvert).Assembly.Location).FullName;
+            var dllName = new FileInfo(typeof(JsonConvert).Assembly.Location).Name;
```
brettfo, line 132: whitespace

src/Microsoft.DotNet.Interactive.Tests/LanguageKernelPackageTests.cs
```diff
@@ -116,41 +116,94 @@ public async Task it_can_load_assembly_references_using_r_directive_separate_sub
...
+
+            if (changeCurrentDirectoryInUserCode)
             {
-                Language.CSharp => $"#r \"{relativePath}\"",
-                Language.FSharp => $"#r \"{relativePath}\""
+                //Even when a user changes the current directory, loading from a relative path is not affected.
+                await kernel.SendAsync(new SubmitCode("System.IO.Directory.SetCurrentDirectory(\"..\")"));
+            }
+
+            await kernel.SendAsync(new SubmitCode("Environment.CurrentDirectory"));
```
brettfo, line 166: This doesn't look necessary for the test.

src/Microsoft.DotNet.Interactive.Tests/LanguageKernelPackageTests.cs
```diff
@@ -116,41 +116,94 @@ public async Task it_can_load_assembly_references_using_r_directive_separate_sub
...
+            {                
+                Language.CSharp => $"#load \"RelativeLoadingSample.csx\"",
+                Language.FSharp => $"#load \"RelativeLoadingSample.fsx\""
             };
 
             var command = new SubmitCode(code);
+            await kernel.SendAsync(command);
+
+            KernelEvents.Should()
+                        .ContainSingle<CommandSucceeded>(c => c.Command == command);
```
brettfo, line 180: The success condition should be that a standard outptu value of `"hello!"` was produced, since that's what's in the .csx/.fsx files.

src/Microsoft.DotNet.Interactive.Tests/LanguageKernelPackageTests.cs
```diff
@@ -116,41 +116,94 @@ public async Task it_can_load_assembly_references_using_r_directive_separate_sub
...
+            var code = language switch
+            {
+                Language.CSharp => $"#load \"{relativePath}/RelativeLoadingSample.csx\"",
+                Language.FSharp => $"#load \"{relativePath}/RelativeLoadingSample.fsx\""
+            };
+
+            var command = new SubmitCode(code);
             await kernel.SendAsync(command);
 
             KernelEvents.Should()
```
brettfo, line 209: Same success condition as above: stdout contained `"hello!"`

src/Microsoft.DotNet.Interactive/Commands/ChangeWorkspaceDirectory.cs
```diff
@@ -7,20 +7,20 @@
...
+            WorkspaceDirectory = workspaceDirectory;
         }
 
         public override Task InvokeAsync(KernelInvocationContext context)
         {
-            Directory.SetCurrentDirectory(WorkingDirectory);
-            context.Publish(new WorkingDirectoryChanged(WorkingDirectory, this));
-            return Task.CompletedTask;
+            Directory.SetCurrentDirectory(WorkspaceDirectory);
+            context.Publish(new WorkspaceDirectoryChanged(WorkspaceDirectory, this));
```
brettfo, line 22: whitespace

src/Microsoft.DotNet.Interactive.Tests/LanguageKernelPackageTests.cs
brettfo: I'm not sure what the F# part is to update it's probing paths for `#load` so I'll throw this one over to @KevinRansom.

KevinRansom: It's not just **#load** it's also **#r**, and I suppose **#i**.  @dsyme has been pushing a PR that addresses a **#r** relative load issue that's related to TypeProviders, and I have been maliciously sitting on it.  Because I wanted to think about it some.

Historically **#r** and **#load** in F# has been relative to the actual script in which it is situated.  This is actually the cleanest approach, it makes scripts with relative references more reusable.  The executing script is not always a file on disk, but an in memory file, often part of a notebook file, this is pretty much identical to standard FSI though, where the executing script is not on disk but treated as if it were, and in the current working directory when FSI starts.  There are circumstances when the executing script is a file generated by the scripting infrastructure, E.g. the list of references generated in a **#r "nuget: ..."**, that script is generated in the temp directory with a random number in the path name and, if I remember correctly a thread.id so relative to that is really not a thing.

I would really ... really like to consider all of the scenarios, before blindly updating fsi to change how #load happens.  Changing FSI now would break existing scripts, that have worked this way on released FSI for 13 years, you can probably guess that I would be reluctant to make that change.  However, the way, one should probably think about this issue is that:

**#r**, **#i** and **#load** are relative to the location for the file in which they are found, and the "project directory(working directory) or however you want to think about is the path for in-memory scripts executed within the notebook.

So for a notebook: **//user/kevinr/notebooks/kevinsfirst/kevinsfirst.dib**

#load "compilationscripts/somefile.fsx" will load from:
**//user/kevinr/notebooks/kevinsfirst/compilationscripts/somefile.fsx**

And if that file has a line:
#load "codegen/makeframework.fsx"  will load from:
**//user/kevinr/notebooks/kevinsfirst/compilationscripts/codegen/makeframework**

This does leave the issue of what to do about the nuget / typeprovider scenario, @dsyme has a proposal and a PR, I am not certain it is exactly right but his proposals are mostly spot on, so I expect it fine will be when I finally try to convince myself.

src/Microsoft.DotNet.Interactive.CSharp/CSharpKernel.cs
strabu: Thanks for your feedback. Just installed "Trailing Whitespace Visualizer by Mads Kristensen" to make myself more whitespace-aware ;)

src/Microsoft.DotNet.Interactive/Commands/ChangeWorkspaceDirectory.cs
```diff
@@ -7,20 +7,20 @@
 
 namespace Microsoft.DotNet.Interactive.Commands
 {
-    public class ChangeWorkingDirectory : KernelCommand
+    public class ChangeWorkspaceDirectory : KernelCommand
```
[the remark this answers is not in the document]
strabu: OK @brettfo, @jonsequitur, I'll revert this https://github.com/dotnet/interactive/pull/1176/commits/bb97a329efd9a8963b5ae65976deead977060ddc

src/Microsoft.DotNet.Interactive.Tests/LanguageKernelPackageTests.cs
strabu: sorry, probably I'm whitespace-blind.

src/Microsoft.DotNet.Interactive/Commands/ChangeWorkspaceDirectory.cs
strabu: reverted

src/Microsoft.DotNet.Interactive.CSharp/CSharpKernel.cs
strabu: reverted

src/Microsoft.DotNet.Interactive.CSharp/CSharpKernel.cs
strabu: Yes, I agree,
"_workspaceDirectory" should not mutate magically as a side-effect.

jonsequitur:
@strabu Could you rebase this onto `main` please?

strabu:
> @strabu Could you rebase this onto `main` please?

have done that, please check

jonsequitur:
Thanks. There are still a couple of outstanding comments. If you have any questions or want clarification on anything, please let us know. We'd love to get this merged.

src/Microsoft.DotNet.Interactive.Tests/LanguageKernelPackageTests.cs
strabu: fixed in https://github.com/dotnet/interactive/pull/1176/commits/0dd55f30e66ea43e619e2fa75ba4cccdbbae6f6e

src/Microsoft.DotNet.Interactive.Tests/LanguageKernelPackageTests.cs
strabu: you are right, probably a leftover from debugging...
fixed in https://github.com/dotnet/interactive/pull/1176/commits/0dd55f30e66ea43e619e2fa75ba4cccdbbae6f6e

src/Microsoft.DotNet.Interactive.Tests/LanguageKernelPackageTests.cs
strabu: done in https://github.com/dotnet/interactive/pull/1176/commits/0dd55f30e66ea43e619e2fa75ba4cccdbbae6f6e

src/Microsoft.DotNet.Interactive.Tests/LanguageKernelPackageTests.cs
strabu: done in https://github.com/dotnet/interactive/pull/1176/commits/0dd55f30e66ea43e619e2fa75ba4cccdbbae6f6e

strabu:
> Thanks. There are still a couple of outstanding comments. If you have any questions or want clarification on anything, please let us know. We'd love to get this merged.

@jonsequitur I've changed the small issues that were left from the tests and rebased onto `main` again.

In the 2 sample-Notebooks for C# and F# at https://github.com/strabu/dotnet-interactive-issue-repro 
there is now a line on top: 
`Directory.SetCurrentDirectory(".."); //this should not mess up the following imports`
...that demonstrates that a user cannot mess up the following `#load`-statements with relative paths.

In VSCode I can now `Open Folder` and point it to: `C:\Users\Alex\source\repos\dotnet-interactive-issue-repro`
and I can also do `Open Folder` and select: `C:\Users\Alex\source\repos\dotnet-interactive-issue-repro\Notebook`
in both cases the C# and F# Sample-Notebooks can import their scripts based on their relative paths.

So I'm now personally getting what I wanted initially as a VSCode-.NET-Interactive-Notebook-user.

To my understanding there are 2 things left from the feedback above that can be discussed further:

1) If we should make the change suggested by @brettfo related to:
`var workingDirectory = _workspaceDirectory ?? Directory.GetCurrentDirectory();`
If yes, the command `ChangeWorkingDirectory` needs to be called by the VSCode-Plugin which is not happening currently as far as I see. I did not wanted to make additional changes in Typescript-Code because I could accomplish my goal without doing so.

2) The discussion about changing the WorkingDirectory for F# and all it's implications and whether it's necessary now.

jonsequitur:
> 1. If we should make the change suggested by @brettfo related to:
> var workingDirectory = _workspaceDirectory ?? Directory.GetCurrentDirectory();
> If yes, the command ChangeWorkingDirectory needs to be called by the VSCode-Plugin which is not happening currently as far as I see. I did not wanted to make additional changes in Typescript-Code because I could accomplish my goal without doing so.

I'm not sure I understand the reasoning for one approach versus the other. @brettfo or @colombod?

> 2. The discussion about changing the WorkingDirectory for F# and all it's implications and whether it's necessary now.

I'd say this is out of scope for this change.

jonsequitur:
@strabu Sorry for the slow response here. Could you rebase this please?

strabu:
> @strabu Sorry for the slow response here. Could you rebase this please?

OK, have done that.

PR #1182, opened 2021-03-23
update insiders kernel execution model

brettfo:
VS Code Insiders just updated with breaking changes to notebook kernel execution.  This PR separates the two different execution models (stable vs. insiders) into the two `versionSpecificFunctions.ts` files and makes `notebookKernel.ts` a thin wrapper around these version-specific functions.  Stable is unchanged, just rearranged; Insiders is very different regarding starting and stopping execution as well as setting outputs.

Scenarios covered in both Stable and Insiders:

1. Execute a cell that produces regular output.
2. Execute a cell that updates output (e.g., NuGet package restore.)
3. Execute a cell that produces errors (e.g., compilation failure).
4. Execute a long-running cell then restart the kernel; verify the cell's state is no longer executing.

Notably broken in this PR is execution cancellation in Insiders.  That will be addressed next.


colombod (approved):
approvified

PR #1186, opened 2021-03-24
Automatic clientSideKernel connection

colombod:
with this  change it is possible to do
```
Kernel.SendViaRootAsync(new CustomCommand(targetKernelName: "clientSideKernel"));
```

The api will try to create a connected kernel automatically


src/Microsoft.DotNet.Interactive/Kernel.Static.cs
```diff
@@ -17,6 +17,8 @@ public partial class Kernel
     {
         public static Kernel Current => KernelInvocationContext.Current.HandlingKernel;
 
+        public static Kernel Root => Current.ParentKernel;
```
jonsequitur, line 20: This should walk all the way to the root in case of deeper hierarchies and return self in the case where the current kernel is the root.

PR #1200, opened 2021-03-29
use mermaid to explain types

colombod:
Add support for mermaid markdown and helpers to print diagram from types
Example with mermaid kernel language
image

And class diagram
image


.gitignore
```diff
@@ -387,3 +387,5 @@ src/Microsoft.DotNet.Interactive.nteract.js/lib/
 src/Microsoft.DotNet.Interactive.SandDance.js/lib/
 
 src/Microsoft.DotNet.Interactive.SandDance.js/dist/
+
+src/Microsoft.DotNet.Interactive.Mermaid.js/dist/
```
jonsequitur, line 391: I think we should rename these .js projects. Following the .NET namespaces convention is a little confusing.

src/Microsoft.DotNet.Interactive.ExtensionLab.Tests/MermaidKernelExtensionTests.cs
```diff
@@ -0,0 +1,106 @@
...
+
+            await extension.OnLoadAsync(kernel);
+
+            kernel.ChildKernels
+                .Should()
+                .ContainSingle(k => k is MermaidKernel);
+        }
+
+        [Fact]
+        public async Task register_formatters()
```
jonsequitur, line 47: Can we come up with a test name that explains the intention a bit better?

src/Microsoft.DotNet.Interactive.ExtensionLab/MermaidExtensions.cs
```diff
@@ -0,0 +1,94 @@
...
+        private static IHtmlContent GenerateHtml(MermaidMarkdown markdown, Uri libraryUri, string libraryVersion, string cacheBuster)
+        {
+            var requireUri = new Uri("https://cdnjs.cloudflare.com/ajax/libs/require.js/2.3.6/require.min.js");
+            var divId = Guid.NewGuid().ToString("N");
+            var code = new StringBuilder();
+            var functionName = $"loadMermaid_{divId}";
+            code.AppendLine("<div style=\"background-color:white;\">");
+           
+            code.AppendLine(@"<script type=""text/javascript"">");
+            AppendJsCode(code,  divId, functionName, libraryUri, libraryVersion, cacheBuster, markdown.ToString());
```
jonsequitur, line 49: Making this an extension method might make it more readable.

src/Microsoft.DotNet.Interactive.ExtensionLab/MermaidKernelExtension.cs
```diff
@@ -0,0 +1,42 @@
...
+        public string Name => "Mermaid";
+        public Task OnLoadAsync(Kernel kernel)
+        {
+            if (kernel is CompositeKernel compositeKernel)
+            {
+                compositeKernel.Add(new MermaidKernel());
+            }
+
+            kernel.UseMermaid();
+            KernelInvocationContext.Current?.Display(
```
jonsequitur, line 20: It should work, and be cleaner, to send a `DisplayValue` command directly to the kernel, no?

jonsequitur:
This PR needs more screen shots.

src/Microsoft.DotNet.Interactive.ExtensionLab.Tests/MermaidKernelExtensionTests.cs
```diff
@@ -0,0 +1,136 @@
...
+                .Single(fm => fm.MimeType == HtmlFormatter.MimeType)
+                .Value;
+
+            this.Assent(formattedData.FixedGuid(), _configuration);
+        }
+
+        [Fact]
+        public void can_generate_class_diagram_from_type()
+        {
+            var diagram = typeof(CompositeKernel).ToClassDiagram();
```
jonsequitur, line 108: It might be a good idea to generate this from something more stable.

That said, this also looks like it might be a very nice way to built an assertion that the public API hasn't changed, once a type is stable, e.g. https://github.com/dotnet/command-line-api/issues/1242#issuecomment-811531652.

src/Microsoft.DotNet.Interactive.ExtensionLab/MermaidKernelExtension.cs
colombod: makes sense

src/Microsoft.DotNet.Interactive.ExtensionLab.Tests/MermaidKernelExtensionTests.cs
colombod: sure

henrygab: We used to have a tool that did something like this, back when developing Windows Vista (or was it XP?) ... to ensure the managed APIs were unchanged between builds.  (In part it was to prevent cross-branch build breaks.)  So, I'm pretty sure a tool already exists for ensuring exposed APIs are unchanged.... might not need to write one is what I mean.

jonsequitur: We we don't need to write one, just write a few lines of code in a test plus [Assent](https://www.nuget.org/packages/Assent/) or [ApprovalTests](https://www.nuget.org/packages/approvaltests/).

jonsequitur (approved):

PR #1212, opened 2021-03-31
Fast path scheduling for language service commands

jonsequitur:
The goal here is to improve UI responsiveness by allowing language service (and other) commands that don't change the state of the kernel to be scheduled in parallel to commands, like `SubmitCode`, that do.

### DIFF
--- a/src/Microsoft.DotNet.Interactive/Kernel.cs
+++ b/src/Microsoft.DotNet.Interactive/Kernel.cs
@@ -27,6 +27,7 @@ namespace Microsoft.DotNet.Interactive
         private readonly CompositeDisposable _disposables;
 
         private readonly Dictionary<Type, KernelCommandInvocation> _dynamicHandlers = new();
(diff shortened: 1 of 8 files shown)


src/Microsoft.DotNet.Interactive/Kernel.cs
```diff
@@ -265,13 +265,29 @@ public void RegisterCommandType<TCommand>()
...
                             case Cancel cancel:
                                 cancel.KernelUri = Uri;
-                                cancel.TargetKernelName = Name;
-                                scheduler.CancelCurrentOperation();
+                                cancel.TargetKernelName = Name; 
+                                Scheduler.CancelCurrentOperation();
                                 await InvokePipelineAndCommandHandler(cancel);
                                 break;
 
+                            case RequestDiagnostics requestDiagnostics:
```
brettfo, line 273: Also `LanguageServiceCommand` which includes hover text, completion, etc.

PR #1219, opened 2021-04-01
better api for mermaid support

colombod:


src/Microsoft.DotNet.Interactive.ExtensionLab.Tests/MermaidKernelExtensionTests.cs
```diff
@@ -123,7 +172,24 @@ public void can_generate_class_diagram_from_generic_type()
         [Fact]
         public void can_generate_class_diagram_to_a_specified_depth()
         {
-            var diagram = typeof(List<Dictionary<string, object>>).ToClassDiagram(graphDepth:2);
+            var diagram = typeof(List<Dictionary<string, object>>).ToClassDiagram(new ClassDiagramConfiguration(2));
+
+            this.Assent(diagram.ToString());
```
jonsequitur, line 177: I would do this without `Assent`. Maybe verify that a specific 3rd-level item isn't present?

colombod: that means parsing lines :(

colombod: I made it

src/Microsoft.DotNet.Interactive.ExtensionLab.Tests/MermaidKernelExtensionTests.cs
```diff
@@ -70,6 +70,23 @@ graph TD
...
+                new CSharpKernel().UseNugetDirective(),
+            };
+
+            var extension = new MermaidKernelExtension();
+
+            await extension.OnLoadAsync(kernel);
+
+            string formatted = typeof(List<string>).ExploreWithUmlClassDiagram().ToDisplayString(HtmlFormatter.MimeType);
+
+            this.Assent(formatted.FixedGuid());
```
jonsequitur, line 87: This test also doesn't really verify what the name of the test is verifying. It's verifying that the output matches the approval, which isn't readable from here. It's not verifying the content. For example, any object can be formatted to HTML using our default formatters. We should probably verify something about the content itself.

colombod: Ok will fix it later

colombod: changed

jonsequitur (approved):

PR #1228, opened 2021-04-08
Improve language service performance

colombod:
Prevent RequestDiagnostics from producing events when other language services are running
Produce diagnostics only on last command handling

### DIFF
--- a/src/Microsoft.DotNet.Interactive.CSharp/InteractiveWorkspace.cs
+++ b/src/Microsoft.DotNet.Interactive.CSharp/InteractiveWorkspace.cs
@@ -25,7 +25,8 @@ namespace Microsoft.DotNet.Interactive.CSharp
         private Compilation _currentCompilation;
         private DocumentId _workingDocumentId;
         private readonly IReadOnlyCollection<MetadataReference> _referenceAssemblies;
-        private readonly List<MetadataReference> _packageManagerReferences = new List<MetadataReference>();
+        private readonly List<MetadataReference> _packageManagerReferences = new();
+        private readonly object _workspaceLock = new();
 
         public InteractiveWorkspace() : base(MefHostServices.DefaultHost, WorkspaceKind.Interactive)
         {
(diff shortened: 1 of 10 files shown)


src/Microsoft.DotNet.Interactive/Kernel.cs
```diff
@@ -275,6 +275,24 @@ public void RegisterCommandType<TCommand>()
...
+
+                                if (_languageServiceCommandInFlight > 0 && command is RequestDiagnostics)
+                                {
+                                    context.Complete(command);
+                                }
+
+                               
+
+                                _inFlightDiagnosticsCommandInvocationContext?.Complete(_inFlightDiagnosticsCommandInvocationContext.Command);
+                                if (command is RequestDiagnostics)
```
jonsequitur, line 287: I would refactor so that the command types occupy different cases in the switch, and remove the need for this `if` block and the one below.

src/Microsoft.DotNet.Interactive/Kernel.cs
```diff
@@ -390,6 +417,8 @@ protected internal void SetScheduler(KernelScheduler<KernelCommand, KernelComman
 
         private static readonly List<KernelCommand> EmptyCommandList = new(0);
         private readonly SemaphoreSlim _fastPathSchedulerLock = new(1);
+        private KernelInvocationContext _inFlightDiagnosticsCommandInvocationContext;
+        private int _languageServiceCommandInFlight = 0;
```
jonsequitur, line 421:
```suggestion
        private int _countOfLanguageServiceCommandsInFlight = 0;
```

src/Microsoft.DotNet.Interactive/Kernel.cs
colombod: done

src/Microsoft.DotNet.Interactive.CSharp/InteractiveWorkspace.cs
```diff
@@ -240,28 +245,59 @@ private ProjectId AddProjectWithPreviousSubmissionToSolution(Compilation compila
 
         public Document UpdateWorkingDocument(string code)
         {
-            var solution = CurrentSolution;
-           
-            solution = solution.RemoveDocument(_workingDocumentId);
+            Document workingDocument;
 
-            var documentDebugName = $"Working from #{_submissionCount}";
+            lock (_workspaceLock)
```
jonsequitur, line 250: Are we sure the lock is needed?

src/Microsoft.DotNet.Interactive.Tests/LanguageKernelTests.cs
```diff
@@ -633,6 +634,62 @@ public async Task requested_diagnostics_are_remapped_to_the_appropriate_span(Lan
...
+        {
+            var kernel = CreateKernel(language);
+            MarkupTestFile.GetLineAndColumn("Console.$$", out var output, out var line, out var column);
+
+            var requestDiagnosticsCommand = new RequestDiagnostics(output);
+
+            var results = Task.WhenAll(
+                kernel.SendAsync(new RequestCompletions(output, new LinePosition(line,column))),
+                kernel.SendAsync(requestDiagnosticsCommand)
+            ).Result;
```
jonsequitur, line 681: You should await this.

colombod: done

src/Microsoft.DotNet.Interactive.CSharp/InteractiveWorkspace.cs
colombod: just deleted the entire api

src/Microsoft.DotNet.Interactive/Kernel.cs
```diff
@@ -317,6 +338,24 @@ public void RegisterCommandType<TCommand>()
             return context.Result;
         }
 
+        private async Task RunOnFastPath(CancellationToken cancellationToken, KernelInvocationContext context,
```
jonsequitur, line 341: `CancellationToken` is typically the last parameter.

jonsequitur (approved):

jonsequitur:
A couple of minor comments that could be resolved in a future PR. If this is helping perf, let's merge it.

src/Microsoft.DotNet.Interactive/Kernel.cs
colombod: fixing now

PR #1235, opened 2021-04-10
Fixing tests

WhiteBlackGoose:
That's a draft PR. It's my first serious PR, so I hope to get feedback on what I'm doing wrong. Currently my plan is to fix tests which fail locally.

There are locale-related failures of tests. I suggest making them independent from the locales.

### DIFF
--- a/src/Microsoft.DotNet.Interactive.PowerShell.Tests/PowerShellKernelTests.cs
+++ b/src/Microsoft.DotNet.Interactive.PowerShell.Tests/PowerShellKernelTests.cs
@@ -163,10 +163,10 @@ for ($j = 0; $j -le 4; $j += 4 ) {
 
             outputs.Should().HaveCountGreaterThan(1);
             
-            outputs
+            string.Join("", 
+                outputs
                 .SelectMany(e => e.FormattedValues.Select(v => v.Value))
-                .First(s => s.Trim().Length > 0)
-                .ToLowerInvariant()
+                ).ToLowerInvariant()
                 .Should()
                 .ContainAll("build-server", "restore");
         }
--- a/src/Microsoft.DotNet.Interactive.PowerShell.Tests/PowerShellKernelTests.cs
+++ b/src/Microsoft.DotNet.Interactive.PowerShell.Tests/PowerShellKernelTests.cs
@@ -170,7 +170,7 @@ for ($j = 0; $j -le 4; $j += 4 ) {
                 .First(s => s.Trim().Length > 0)
                 .ToLowerInvariant()
                 .Should()
-                .Match("*ping*data*");
+                .ContainAll("[::1]", "32");
         }
 
         [Fact]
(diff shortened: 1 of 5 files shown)


src/Microsoft.DotNet.Interactive.Tests/LanguageKernelPackageTests.cs
```diff
@@ -813,17 +813,11 @@ public async Task Pound_r_nuget_does_not_repeat_notifications_for_previous_r_nug
 
         [Theory]
         [InlineData(Language.CSharp)]
-        //[InlineData(Language.FSharp)]   /// Reenable when --- https://github.com/dotnet/fsharp/issues/8775
+        [InlineData(Language.FSharp)]
```
WhiteBlackGoose, line 816: Here the issue was closed, so I reenabled it. 🦆

src/Microsoft.DotNet.Interactive.PowerShell.Tests/PowerShellKernelTests.cs
```diff
@@ -155,9 +155,7 @@ public async Task PowerShell_native_executable_output_is_collected()
         {
             var kernel = CreateKernel(Language.PowerShell);
 
-            var command = Platform.IsWindows
-                ? new SubmitCode("ping.exe -n 1 localhost")
-                : new SubmitCode("ping -c 1 localhost");
+            var command = new SubmitCode("dotnet --help");
```
WhiteBlackGoose, line 158: `ping` has literally nothing in common on my Russian Windows machine and English Linux machine. Here:

Mine: `обмен пакетами с desktop-1if9leb [::1] с 32 байтами данных:`
CI: `ping localhost (127.0.0.1) 56(84) bytes of data.\n`

So let it be something else

WhiteBlackGoose:
I *think* I fixed all such tests that I could run and that failed.

WhiteBlackGoose:
@jonsequitur it fails on test `server_listens_on_ephemeral_port`, which I didn't even change. The test expects an opened local port to be at least 10_000, whereas on CI it opened on 7645. Should we/I/you somehow fix it?

// also, when my PRs will be reviewed? :).

WhiteBlackGoose:
@vzarytovskii isn't it better to fix the test rather than wait until it opens under a >10000 port? 😄

vzarytovskii:
> @vzarytovskii isn't it better to fix the test rather than wait until it opens under a >10000 port? 😄 

I've restarted it not to fix the CI run, but to see whether it will reproduce or not, since on windows 10, the ephemeral port range should be in between 49152 and 65535.

It can be one of two things - either we hit the misconfigured agent, or default behaviour had changed.

Also question is whether we care if port is real "ephemeral" and has to fit in the "official" range, or we're fine with it being >= 1025.

WhiteBlackGoose:
Now it passes ¯\_(ツ)_/¯

WhiteBlackGoose:
Yes, @colombod.

colombod (approved):

PR #1237, opened 2021-04-12
package upgrading

colombod:


src/Microsoft.DotNet.Interactive.CSharp.Tests/Microsoft.DotNet.Interactive.CSharp.Tests.csproj
```diff
@@ -39,4 +39,8 @@
     </PackageReference>
   </ItemGroup>
 
+  <ItemGroup>
+    <PackageReference Update="Microsoft.NET.Test.Sdk" Version="16.9.4" />
```
brettfo, line 43: Remove all references to `Microsoft.NET.Test.Sdk` because Arcade handles this for us and it'll cause problems when Arcade isn't disabled.  The proper place to upgrade this package is [here](https://github.com/dotnet/interactive/blob/06ac1da7d95da4306e81624ddea927e64ffb98f4/Directory.Build.targets#L11).  N.b., however, that the version of Arcade we're using specifies a Test.SDK of 16.6.1, but since this only applies if DisableArcade=1, you can use the 16.9.4 if you like.

src/Microsoft.DotNet.Interactive.ExtensionLab.Tests/Microsoft.DotNet.Interactive.ExtensionLab.Tests.csproj
```diff
@@ -37,7 +37,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Update="Microsoft.NET.Test.Sdk" Version="16.6.1" />
+    <PackageReference Update="Microsoft.NET.Test.Sdk" Version="16.9.4" />
```
brettfo, line 40: All of these are what should be removed.

brettfo (changes requested):

src/Microsoft.DotNet.Interactive.ExtensionLab.Tests/Microsoft.DotNet.Interactive.ExtensionLab.Tests.csproj
colombod: doing it

src/Microsoft.DotNet.Interactive.CSharp.Tests/Microsoft.DotNet.Interactive.CSharp.Tests.csproj
colombod: doing it

PR #1243, opened 2021-04-13
Fix pound i

KevinRansom:
A better fix and couple more tests.

### DIFF
--- a/src/Microsoft.DotNet.Interactive.Tests/LanguageKernelPackageTests.cs
+++ b/src/Microsoft.DotNet.Interactive.Tests/LanguageKernelPackageTests.cs
@@ -432,58 +432,20 @@ Formatter.Register<DataFrame>((df, writer) =>
             events.OfType<DisplayEvent>().Select(e => e.GetType()).Should().ContainInOrder(typeof(DisplayedValueProduced), typeof(DisplayedValueUpdated));
         }
 
-
-        // For the restore sources formatted text, strip out the restore source lines and return them
-        private static System.Collections.Generic.IEnumerable<string> GetLines(string value)
-        {
-            var startToken = @"<li><span>";
-            var startLength = startToken.Length;
-            var endToken = @"</span></li>";
-            var endLength = endToken.Length;
-            var pos = 0;
-            while(pos <= value.Length)
-            {
-                var st = value.IndexOf(startToken, pos);
-                if (st >= 0)
-                {
-                    var p = st + startLength;
-                    var en = value.IndexOf(endToken, p);
-                    if (en >= 0)
-                    {
-                        yield return value.Substring(p, en - p);
-                        pos = en + 1;
-                    }
-                    else
-                    {
-                        yield return value.Substring(p);
-                        pos = value.Length + 1;
-                        break;
-                    }
-                }
-                else
-                {
-                    break;
-                }
-            }
-        }
-
         [Theory]
         [InlineData(Language.CSharp)]
         [InlineData(Language.FSharp)]
-        public async Task Pound_i_nuget_with_multi_submissions_combines_the_DisplayValueProduced_Text(Language language)
+        public async Task Pound_i_nuget_with_multi_submissions_combines_the_Text_Produced(Language language)
         {
             var kernel = CreateKernel(language);
-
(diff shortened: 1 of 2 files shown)


src/Microsoft.DotNet.Interactive.Tests/LanguageKernelPackageTests.cs
```diff
@@ -425,18 +425,68 @@ public async Task Pound_i_nuget_displays_list_of_added_sources(Language language
...
-                  .ContainSingle<DisplayedValueUpdated>()
-                  .Which
-                  .FormattedValues
-                  .Should()
-                  .ContainSingle(v => v.MimeType == "text/html")
-                  .Which
-                  .Value
-                  .Should()
-                  .ContainAll("Restore sources", "https://completelyFakerestoreSource");
+            Assert.Collection(events,
```
jonsequitur, line 433: Could you use FluentAssertions and make the specification a little more focused? Maybe try `events.Should().ContainInOrder(...)` and only specify the events we care about?

src/Microsoft.DotNet.Interactive.Tests/LanguageKernelPackageTests.cs
```diff
@@ -425,18 +425,68 @@ public async Task Pound_i_nuget_displays_list_of_added_sources(Language language
...
+
+            using var events = kernel.KernelEvents.ToSubscribedList();
+
+            await kernel.SubmitCodeAsync(@"
+#i ""nuget:https://completelyFakerestoreSourceCommand1.1""
+#i ""nuget:https://completelyFakerestoreSourceCommand1.2""
+");
+
+            var displayValueProducedEvents = KernelEvents.Where(e => e is DisplayedValueProduced);
+            Assert.True(displayValueProducedEvents.Count() == 1);
```
jonsequitur, line 462: Please remove the intermediate sanity check assertions.

src/Microsoft.DotNet.Interactive.Tests/LanguageKernelPackageTests.cs
```diff
@@ -425,18 +425,68 @@ public async Task Pound_i_nuget_displays_list_of_added_sources(Language language
...
+            var displayValueUpdatedEvents = KernelEvents.Where(e => e is DisplayedValueUpdated);
+            Assert.True(displayValueProducedEvents.Count() == 1);
+
+            await kernel.SubmitCodeAsync(@"
+#i ""nuget:https://completelyFakerestoreSourceCommand2.1""
+#i ""nuget:https://completelyFakerestoreSourceCommand2.2""
+");
+
+            // Only produces a single DisplayedValueProduced event
+            displayValueProducedEvents = KernelEvents.Where(e => e is DisplayedValueProduced);
```
jonsequitur, line 473: This test doesn't look like it's testing something different from the one above.

src/Microsoft.DotNet.Interactive.Tests/LanguageKernelPackageTests.cs
KevinRansom: @jonsequitur , lordy ... I am not fluent in fluentassertions, but yes, I can try.

src/Microsoft.DotNet.Interactive.Tests/LanguageKernelPackageTests.cs
KevinRansom: It verifies that 2 submissions accumulate the output.

jonsequitur: Ah, I see. I missed that.

jonsequitur: Can we make the test name more explicit in that case?

jonsequitur (approved):

src/Microsoft.DotNet.Interactive.Tests/LanguageKernelPackageTests.cs
KevinRansom: Yes ... naming and cache eviction the two hardest problems in computer science.  Not that what I do is science. :-)

PR #1244, opened 2021-04-13
Automate command forwarding to client side kernel

colombod:
Simplify sending commands to client side kernel leveraging the routing of commands via the JavascriptKernel
image

image

image

### DIFF
--- a/src/Microsoft.DotNet.Interactive/FrontendEnvironment.cs
+++ b/src/Microsoft.DotNet.Interactive/FrontendEnvironment.cs
@@ -1,7 +1,9 @@
 ﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System;
 using System.Threading.Tasks;
+using Microsoft.DotNet.Interactive.Commands;
 
 namespace Microsoft.DotNet.Interactive
 {
@@ -16,6 +18,11 @@ namespace Microsoft.DotNet.Interactive
             context.Display(scriptContent);
 
             return Task.CompletedTask;
+        }
+
+        public virtual Task ForwardCommand(KernelCommand command, KernelInvocationContext context)
+        {
+            throw new NotSupportedException();
         }
     }
 }
--- a/src/Microsoft.DotNet.Interactive/FrontendEnvironment.cs
+++ b/src/Microsoft.DotNet.Interactive/FrontendEnvironment.cs
@@ -1,9 +1,7 @@
 ﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
-using System;
 using System.Threading.Tasks;
-using Microsoft.DotNet.Interactive.Commands;
 
 namespace Microsoft.DotNet.Interactive
 {
@@ -18,11 +16,6 @@ namespace Microsoft.DotNet.Interactive
             context.Display(scriptContent);
 
             return Task.CompletedTask;
-        }
-
-        public virtual Task ForwardCommand(KernelCommand command, KernelInvocationContext context)
-        {
-            throw new NotSupportedException();
         }
     }
 }
--- a/src/Microsoft.DotNet.Interactive/JavaScriptKernel.cs
+++ b/src/Microsoft.DotNet.Interactive/JavaScriptKernel.cs
@@ -31,16 +31,14 @@ namespace Microsoft.DotNet.Interactive
             _client = kernelClient ?? throw new ArgumentNullException(nameof(kernelClient));
         }
 
-        internal override async Task HandleAsync(KernelCommand command, KernelInvocationContext context)
+        protected override Func<TCommand, KernelInvocationContext, Task> CreateDefaultHandlerForCommandType<TCommand>()
         {
-            TrySetHandler(command, context);
+            return (kernelCommand, _) => ForwardCommand(kernelCommand);
+        }
 
-            if (_client is not null && IsCommandTypeRegistered(command.GetType()) && command.Handler is null)
-            {
-                command.Handler = (kernelCommand, _) => _client.SendAsync(kernelCommand); 
-            }
-
-            await command.InvokeAsync(context);
+        private Task ForwardCommand(KernelCommand command)
+        {
+            return _client.SendAsync(command);
         }
     }
 }
--- a/src/Microsoft.DotNet.Interactive/JavaScriptKernel.cs
+++ b/src/Microsoft.DotNet.Interactive/JavaScriptKernel.cs
@@ -1,8 +1,10 @@
 ﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System;
 using System.Threading.Tasks;
 using Microsoft.DotNet.Interactive.Commands;
+using Microsoft.DotNet.Interactive.Connection;
 
 namespace Microsoft.DotNet.Interactive
 {
@@ -10,6 +12,7 @@ namespace Microsoft.DotNet.Interactive
         Kernel,
         IKernelCommandHandler<SubmitCode>
     {
+        private KernelClientBase _client;
         public const string DefaultKernelName = "javascript";
 
         public JavaScriptKernel() : base(DefaultKernelName)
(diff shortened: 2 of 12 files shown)


src/Microsoft.DotNet.Interactive/FrontendEnvironment.cs
```diff
@@ -17,5 +19,10 @@ public virtual Task ExecuteClientScript(string code, KernelInvocationContext con
 
             return Task.CompletedTask;
         }
+
+        public virtual Task ForwardCommand(KernelCommand command, KernelInvocationContext context)
```
jonsequitur, line 23: Why not put this directly on the `JavaScriptKernel` (and make it private)?

src/Microsoft.DotNet.Interactive/JavaScriptKernel.cs
```diff
@@ -22,5 +22,17 @@ public JavaScriptKernel() : base(DefaultKernelName)
...
             return FrontendEnvironment.ExecuteClientScript(command.Code, context);
         }
+
+        internal override async Task HandleAsync(KernelCommand command, KernelInvocationContext context)
+        {
+            TrySetHandler(command, context);
+
+            if (CommandTypeIsRegistered(command.GetType()) && command.Handler is null)
+            {
+                command.Handler = (kernelCommand, invocationContext) => FrontendEnvironment.ForwardCommand(kernelCommand, invocationContext);
```
jonsequitur, line 32: This means that the command will have its handler assigned twice. This is kind of confusing. This kernel should have the real handler and the implementation of that real handler should transparently forward the command.

src/Microsoft.DotNet.Interactive/Server/KernelCommandEnvelope.cs
```diff
@@ -25,6 +25,10 @@ static KernelCommandEnvelope()
             ResetToDefaults();
         }
 
+        internal static bool IsCommandTypeRegistered(Type commaType)
+        {
+            return _commandTypesByCommandTypeName.ContainsKey(commaType.Name);
```
jonsequitur, line 30: I didn't realize this was static. This should really all be kernel-level state.

colombod: that is a registration for serialization is static with the envelope not the kernel, do we need to track these kind of registration in more than one place?

src/Microsoft.DotNet.Interactive/JavaScriptKernel.cs
colombod: the handler will only be assigned once, if we end in the if statement is because this is a command we want to forward to the client side kernel

src/Microsoft.DotNet.Interactive/FrontendEnvironment.cs
colombod: ok, so we pass the KernelClient object to the jskernel?

src/Microsoft.DotNet.Interactive/JavaScriptKernel.cs
jonsequitur: I mean we have two different blocks of code responsible for it, one in `TrySetHandler` and another here. This can be more concise and explicit by letting the handlers in this kernel forward directly. Remember not all commands have settable handlers, so this code won't work for some commands.

src/Microsoft.DotNet.Interactive/Server/KernelCommandEnvelope.cs
jonsequitur: Or shift them to the kernels maybe? Just trying to reduce global state.

colombod: The envelope serializers will have to know that state

src/Microsoft.DotNet.Interactive/JavaScriptKernel.cs
colombod: I have changed that part so that at registration time the jskernel creates default handlers to forward commands to clietnside kernel.  This way we only have 1 logic for handler assignment

src/Microsoft.DotNet.Interactive/Server/KernelCommandEnvelope.cs
colombod: removed that

src/Microsoft.DotNet.Interactive/FrontendEnvironment.cs
colombod: The jskernel obtains a kernelClient for commadn forwarding, removed all additional logic from  FrontendEnvironment type

src/Microsoft.DotNet.Interactive/JavaScriptKernel.cs
```diff
@@ -22,5 +25,20 @@ public JavaScriptKernel() : base(DefaultKernelName)
         {
             return FrontendEnvironment.ExecuteClientScript(command.Code, context);
         }
+
+        public void SetKernelClient(KernelClientBase kernelClient)
```
jonsequitur, line 29: Any reason for this to be public? Any reason for it not to be a constructor parameter?

src/Microsoft.DotNet.Interactive/Kernel.cs
```diff
@@ -204,8 +205,25 @@ public void RegisterCommandHandler<TCommand>(Func<TCommand, KernelInvocationCont
...
         public void RegisterCommandType<TCommand>()
-            where TCommand : KernelCommand =>
-            KernelCommandEnvelope.RegisterCommandTypeReplacingIfNecessary<TCommand>();
+            where TCommand : KernelCommand
+        {
+            if (_registeredCommandTypes.Add(typeof(TCommand)))
+            {
+                KernelCommandEnvelope.RegisterCommandTypeReplacingIfNecessary<TCommand>();
+                var defaultHandler = CreateDefaultHandlerForCommandType<TCommand>();
+                if (defaultHandler is not null)
```
jonsequitur, line 214: It seems like this should not be allowed to return null.

src/Microsoft.DotNet.Interactive/JavaScriptKernel.cs
colombod: right now we have scenarios where (synapse) there can be a jsclient but not able (yet) to forward commands to the client side kernel

src/Microsoft.DotNet.Interactive/Kernel.cs
colombod: done

src/Microsoft.DotNet.Interactive/Kernel.cs
```diff
@@ -204,8 +208,34 @@ public void RegisterCommandHandler<TCommand>(Func<TCommand, KernelInvocationCont
...
+        {
+            return (_,_) => Task.CompletedTask;
+        }
+
+        public virtual IEnumerable<Type> SupportedCommands()
+        {
+            return  _registeredCommandTypes;
+        }
+
+        private IEnumerable<Type> GetCommandFromInterfaceImplementation()
```
jonsequitur, line 232:
```suggestion
        private IEnumerable<Type> GetSupportedCommandTypesFromInterfaceImplementation()
```

src/Microsoft.DotNet.Interactive/Kernel.cs
```diff
@@ -204,8 +208,34 @@ public void RegisterCommandHandler<TCommand>(Func<TCommand, KernelInvocationCont
...
 
         public void RegisterCommandType<TCommand>()
-            where TCommand : KernelCommand =>
-            KernelCommandEnvelope.RegisterCommandTypeReplacingIfNecessary<TCommand>();
+            where TCommand : KernelCommand
+        {
+            if (_registeredCommandTypes.Add(typeof(TCommand)))
+            {
+                KernelCommandEnvelope.RegisterCommandTypeForSerialization<TCommand>();
+                var defaultHandler = CreateDefaultHandlerForCommandType<TCommand>() ?? throw new InvalidOperationException("CreateDefaultHandlerForCommandType should not return null");
```
jonsequitur, line 216: This still bugs me. We should differentiate the command types in a future PR.

jonsequitur (approved):

src/Microsoft.DotNet.Interactive/Kernel.cs
colombod: we should, agreed

Issue #1096, opened 2021-02-23
Display BigInteger values as decimals, not objects


idg10:
The `BigInteger` type (in `System.Numerics`) built into .NET does not display its numeric value in a Notebook. So if you have, for example:

```cs
BigInteger v = 42;
display(v);
```

You will see this tabular output:

IsPowerOfTwo | IsZero | IsOne | IsEven | Sign
-- | -- | -- | -- | --
False | False | False | True | 1

In other words, the notebook has displayed all the properties, instead of the numeric value. This is surprising. Normally when I ask a notebook to display a number I expect its numeric value in some form.

Currently I need to call `ToString()` on a `BigInteger` to get its value. This feels like something .NET Interactive should be doing automatically.

WhiteBlackGoose:
@jonsequitur if you think it's not hard to fix it, assign me to this one, I'll try to work on it

jonsequitur:
All yours. Thanks @WhiteBlackGoose!

PR #1246, opened 2021-04-14
BigInteger should be displayed as plain text

WhiteBlackGoose:
Unlike what we see in https://github.com/dotnet/interactive/issues/1096

Before:
image

After:
image

### DIFF
--- a/src/Microsoft.DotNet.Interactive.Formatting.Tests/HtmlFormatterTests.cs
+++ b/src/Microsoft.DotNet.Interactive.Formatting.Tests/HtmlFormatterTests.cs
@@ -7,6 +7,7 @@ using System.Collections.Generic;
 using System.Dynamic;
 using System.IO;
 using System.Linq;
+using System.Numerics;
 using FluentAssertions;
 using FluentAssertions.Extensions;
 using Xunit;
@@ -305,7 +306,21 @@ string";
                           $"{PlainTextBegin}{instance.HtmlEncode()}{PlainTextEnd}");
             }
 
+            [Fact]
+            public void It_does_not_affect_BigInteger()
+            {
+                var formatter = HtmlFormatter.GetPreferredFormatterFor(typeof(BigInteger));
 
+                var writer = new StringWriter();
+
+                var instance = BigInteger.Parse("78923589327589332402359");
+
+                formatter.Format(instance, writer);
+
+                writer.ToString()
+                    .Should()
+                    .Be("<div class=\"dni-plaintext\">78923589327589332402359</div>");
+            }
         }
         public class Sequences : FormatterTestBase
         {
--- a/src/Microsoft.DotNet.Interactive.Formatting.Tests/HtmlFormatterTests.cs
+++ b/src/Microsoft.DotNet.Interactive.Formatting.Tests/HtmlFormatterTests.cs
@@ -15,11 +15,10 @@ using static Microsoft.DotNet.Interactive.Formatting.Tests.Tags;
 
 namespace Microsoft.DotNet.Interactive.Formatting.Tests
 {
-    public class Tags
+    public static class Tags
     {
         public const string PlainTextBegin = "<div class=\"dni-plaintext\">";
         public const string PlainTextEnd = "</div>";
-
     }
     public class HtmlFormatterTests : FormatterTestBase
     {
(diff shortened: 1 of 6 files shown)


src/Microsoft.DotNet.Interactive.Formatting/DefaultHtmlFormatterSet.cs
```diff
@@ -155,6 +155,13 @@ internal class DefaultHtmlFormatterSet
                     return formatter.Format(context, value, writer);
                 }),
 
+                // BigInteger should be displayed as plain text
+                new HtmlFormatter<System.Numerics.BigInteger>((context, value, writer) =>
+                {
+                    HtmlFormatter.FormatStringAsPlainText(value.ToString(), writer);
```
WhiteBlackGoose, line 161: I did it same way it's done for strings, but I'm not sure it should work like this.

src/Microsoft.DotNet.Interactive.Formatting.Tests/HtmlFormatterTests.cs
```diff
@@ -305,7 +306,21 @@ is a
...
+
+                var writer = new StringWriter();
+
+                var instance = BigInteger.Parse("78923589327589332402359");
 
+                formatter.Format(instance, writer);
+
+                writer.ToString()
+                    .Should()
+                    .Be("<div class=\"dni-plaintext\">78923589327589332402359</div>");
```
WhiteBlackGoose, line 322: Current `main` would output a table (like in the screenshot). We want to see a plain text instead

src/Microsoft.DotNet.Interactive.Formatting/DefaultHtmlFormatterSet.cs
jonsequitur: Calling `ToString` here will bypass the `PlainTextFormatter`, which might have its own customization in place. To make sure the behavior is consistent, you can call into the other formatter like this:

```cs
value.FormatTo(writer, PlainTextFormatter.MimeType);
```

With that in mind, a complete fix should include a default for that formatter as well, otherwise:

image

src/Microsoft.DotNet.Interactive.Formatting/DefaultHtmlFormatterSet.cs
```diff
@@ -155,6 +156,13 @@ internal class DefaultHtmlFormatterSet
                     return formatter.Format(context, value, writer);
                 }),
 
+                // BigInteger should be displayed as plain text
+                new HtmlFormatter<BigInteger>((context, value, writer) =>
+                {
+                    HtmlFormatter.FormatObjectAsPlainText(context, value, writer);
```
WhiteBlackGoose, line 162: @jonsequitur you said to write `value.FormatTo(writer, PlainTextFormatter.MimeType);`, but then it would return a number without html wrapping (without `<div class="dni-plaintext">`, `</div>`). So I did it same way as for `object`. What do you think?

src/Microsoft.DotNet.Interactive.Formatting.Tests/HtmlFormatterTests.cs
```diff
@@ -305,8 +305,23 @@ is a
...
+
+                var writer = new StringWriter();
+
+                var instance = BigInteger.Parse("78923589327589332402359");
 
+                formatter.Format(instance, writer);
+
+                writer.ToString()
+                    .Should()
+                    .Be($"{PlainTextBegin}78923589327589332402359{PlainTextEnd}");
```
WhiteBlackGoose, line 321: Adding to [comment](https://github.com/dotnet/interactive/pull/1246/files#r613817544). Here, if it was via `FormatTo`,  it would return the result without `PlainTextBegin`. Now it seems to be working:
image

WhiteBlackGoose:
It would be nice if someones left a review... 🦆

src/Microsoft.DotNet.Interactive.Formatting/DefaultHtmlFormatterSet.cs
jonsequitur: That works great.

jonsequitur (approved):

jonsequitur:
Thanks, @WhiteBlackGoose!

Issue #1241, opened 2021-04-13
#i "nuget:somepath"  does not validate that the path exists


KevinRansom:
#i some path in nuget fails if the path doesn't exist.

WhiteBlackGoose:
Assign me to this one, sounds doable

Issue #1181, opened 2021-03-23
Failing on #r "nuget" puts the kernel in unusable state


colombod:
### Describe the bug
If a cell performs `#r "nuget"` with errors no other package loading instruction will work;
![image](https://user-images.githubusercontent.com/375556/112088765-4ab32680-8b88-11eb-8929-7c1e9c8432cb.png)

## Version
1.0.216815

- OS
    - [X] Windows 10
    - [ ] macOS
    - [ ] Linux (Please specify distro)
    - [ ] iOS
    - [ ] Android
 - Browser 
    - [ ] Chrome
    - [ ] Edge
    - [ ] Firefox
    - [ ] Safari
 - Frontend
    - [ ] Jupyter Notebook
    - [ ] Jupyter Lab
    - [ ] nteract
    - [ ] Visual Studio Code 
    - [X] Visual Studio Code Insiders
    - [ ] Other (please specify)

KevinRansom:
Fixed

PR #1247, opened 2021-04-15
Fix #r for invalid packages

KevinRansom:
Fixes #https://github.com/dotnet/interactive/issues/1181

image

Becomes this:
image


src/Microsoft.DotNet.Interactive/PackageRestoreContext.cs
```diff
@@ -234,6 +233,11 @@ public async Task<PackageRestoreResult> RestoreAsync()
                     succeeded: false,
                     requestedPackages: newlyRequested,
                     errors: errors);
+
+                foreach(var r in newlyRequested)
+                {
+                    _requestedPackageReferences.TryRemove(r.PackageName.ToLower(CultureInfo.InvariantCulture), out var _);
+                }
```
jonsequitur, line 240: This looks like a breaking change. We don't want to remove the requested packages, we just want to not re-request them. The information is useful in other ways, e.g. we know someone requested `FluentAssertions,*` and received `FluentAssertions,5.7.0`.

KevinRansom: @jonsequitur, it's fine this is the failed code path, as soon as the developer re-submits the script code _requestedPackageReferences will be refreshed.

colombod: Why not marking the packages as 
- pending
- succeeded
- failed
So we avoid requesting again anything that worked or failed (maybe keep the failure reason too?)

KevinRansom: @colombod, failure may be due to a point in time effect such as a network latency problem or it may succeed by adding a #i "nuget:something" or a #i may fail, but succeed after the developer created the actual directory.

jonsequitur: Deliberately re-requesting a failed package is a great scenario that deserves its own test, e.g. failing, adding a `#i`, trying again, and succeeding.

src/Microsoft.DotNet.Interactive.Tests/Server/SerializationTests.cs
```diff
@@ -315,6 +315,7 @@ IEnumerable<KernelEvent> events()
                     new ResolvedPackageReference(
                         packageName: "ThePackage",
                         packageVersion: "1.2.3",
+                        false,
```
KevinRansom, line 318: Kevin, do you still need this?

WhiteBlackGoose:
Does it close #1241 ?

src/Microsoft.DotNet.Interactive.Tests/PackageRestoreContextTests.cs
```diff
@@ -181,17 +200,24 @@ public async Task Can_get_path_to_nuget_package_when_multiple_packages_are_added
             File.Exists(path).Should().BeTrue();
         }
 
+        [Fact]
+        public async Task Fail_if_RestoreSourceHasAnInvalidUri()
```
colombod, line 204: Nice!

src/Microsoft.DotNet.Interactive.Tests/Server/SerializationTests.cs
KevinRansom: I still haven't checked ...

KevinRansom: Yes I do:   here:  https://github.com/dotnet/interactive/pull/1247/files#diff-7c24de179534b7b7914d7b376592645b4fca966c1a4d87d2ea2bda1a6bcbbe84R123

KevinRansom:
> Does it close #1241 ?

@WhiteBlackGoose no, it does not.

Kevin

src/Microsoft.DotNet.Interactive.Tests/LanguageKernelPackageTests.cs
```diff
@@ -457,11 +450,8 @@ public async Task Pound_i_nuget_with_multi_submissions_combines_the_Text_Produce
 
             using var events = result.KernelEvents.ToSubscribedList();
 
-            // For the DisplayedValueUpdated events strip out the restore sources
-            // Verify that they match the expected values
-            events.Should()
-                  .ContainSingle<DisplayedValueProduced>()
-                  .Which.FormattedValues
+            events.OfType<DisplayedValueProduced>()
+                  .First().FormattedValues
```
jonsequitur, line 454: Assuming we're expecting exactly one `DisplayedValueProduced` event, this is better written like this to provide better output:

```cs
events.Should().ContainSingle<DisplayedValueProduced>(e => e.MimeType)
    .Which.Value
    .Should()
    .ContainAll(expectedList);
```

src/Microsoft.DotNet.Interactive.Tests/LanguageKernelPackageTests.cs
```diff
@@ -496,10 +487,8 @@ public async Task Pound_i_nuget_with_multi_submissions_combines_the_Text_Updates
 
             using var events = result.KernelEvents.ToSubscribedList();
 
-            // For the DisplayedValueUpdated events strip out the restore sources
-            // Verify that they match the expected values
-            events.OfType<DisplayedValueUpdated>()
-                  .Last().FormattedValues
+            events.OfType<DisplayedValueProduced>()
+                  .First().FormattedValues
```
jonsequitur, line 491: Same as above.

src/Microsoft.DotNet.Interactive.Tests/PackageRestoreContextTests.cs
```diff
@@ -110,6 +110,25 @@ public async Task Packages_from_previous_requests_are_not_returned_in_subsequent
...
+            restoreContext.GetOrAddPackageReference("NonExistentNugetPackage", "99.99.99-NoReallyIDontExist");
+            await restoreContext.RestoreAsync();
+
+            // Even though the previous restore failed, this one should succeed
+            restoreContext.GetOrAddPackageReference("FluentAssertions", "5.7.0");
+            var result = await restoreContext.RestoreAsync();
+
+            result.ResolvedReferences
+                       .Should()
+                       .Contain(r => r.PackageName == "FluentAssertions");
```
jonsequitur, line 128: This assertion isn't checking that the package isn't in the set.

src/Microsoft.DotNet.Interactive.Tests/PackageRestoreContextTests.cs
jonsequitur, line 204: Snake_case_please.

src/Microsoft.DotNet.Interactive.Tests/LanguageKernelPackageTests.cs
KevinRansom: I think there are two displayValueProduced.
1 for restore sources and 1 for packages loaded.

src/Microsoft.DotNet.Interactive/KernelSupportsNugetExtensions.cs
```diff
@@ -178,36 +166,58 @@ public static string GetDisplayValueId(PackageReference package)
...
+                    object InstallMessage(string message, IEnumerable<string> items, string progress = null)
+                    {
+                        if (items.Count() > 0)
+                        {
+                            return div(
+                                strong(message),
+                                ul(items.Select(s => li(span(s + progress)))));
+                        }
+                        else
+                            return div();
```
jonsequitur, line 198:
```suggestion
                       {
                            return div();
                        }
```

src/Microsoft.DotNet.Interactive/KernelSupportsNugetExtensions.cs
```diff
@@ -178,36 +166,58 @@ public static string GetDisplayValueId(PackageReference package)
...
-                        var message = InstallingPackageMessage(package) + "...";
-                        var displayedValue = context.Display(message);
-                        displayedValues[id] = displayedValue;
-                        messages.Add(package, message);
-                        requestedPackageIds.Add(id, package);
-                    }
+                    object install = InstallMessage("Restore sources", (kernel.RestoreSources.OrderBy(s => s)));
+                    CreateOrUpdateDisplayValue(context, restoreSourcesPropertyName, install, HtmlFormatter.MimeType);
+
+                    var requested = newlyRequestedPackages.Select(s => s.PackageName).OrderBy(s => s);
```
jonsequitur, line 219: This variable name isn't informative. Do we need `newlyRequestedPackages` or should we eagerly transform to the ordered list of sources?

src/Microsoft.DotNet.Interactive/KernelSupportsNugetExtensions.cs
```diff
@@ -178,36 +166,58 @@ public static string GetDisplayValueId(PackageReference package)
             public static IEqualityComparer<PackageReference> Instance { get; } = new PackageReferenceComparer();
         }
 
+        private static void CreateOrUpdateDisplayValue(KernelInvocationContext context, string name, object content, string mimeType)
```
jonsequitur, line 169: Why 'object`?

src/Microsoft.DotNet.Interactive/KernelSupportsNugetExtensions.cs
```diff
@@ -178,36 +166,58 @@ public static string GetDisplayValueId(PackageReference package)
...
-                    foreach (var package in newlyRequestedPackages)
-                    {
-                        var id = PackageReferenceComparer.GetDisplayValueId(package);
-                        var message = InstallingPackageMessage(package) + "...";
-                        var displayedValue = context.Display(message);
-                        displayedValues[id] = displayedValue;
-                        messages.Add(package, message);
-                        requestedPackageIds.Add(id, package);
-                    }
+                    object install = InstallMessage("Restore sources", (kernel.RestoreSources.OrderBy(s => s)));
```
jonsequitur, line 216: A `ToArray` to prevent multiple enumerations might be good here.

src/Microsoft.DotNet.Interactive/PackageRestoreContext.cs
```diff
@@ -54,7 +57,13 @@ private IEnumerable<string> NativeProbingRoots()
             }
         }
 
-        public void AddRestoreSource(string source) => _restoreSources.Add(source);
+        public void AddRestoreSource(string source)
+        {
+            if (!_resolvedRestoreSources.TryGetValue(source, out var _))
```
jonsequitur, line 62: You shouldn't need the `if` condition given you're using `TryAdd`.

src/Microsoft.DotNet.Interactive/PackageRestoreContext.cs
```diff
@@ -112,7 +120,7 @@ private IEnumerable<string> NativeProbingRoots()
             {
                 yield return Tuple.Create("i", rs);
             }
-            foreach (var pr in RequestedPackageReferences)
+            foreach (var pr in RequestedPackageReferences.Concat(ResolvedPackageReferences.Where(p => p.Requested).Select(p => new PackageReference(p.PackageName, p.PackageVersion))).OrderBy(r => r.PackageName))
```
jonsequitur, line 123: I don't understand what the `Requested` property means. If it's in this collection, wasn't it by definition "requested"?

src/dotnet-interactive-vscode/common/interfaces/contracts.ts
```diff
@@ -335,6 +335,7 @@ export interface ResolvedPackageReference extends PackageReference {
     assemblyPaths: Array<string>;
     probingPaths: Array<string>;
     packageRoot: string;
+    requested: boolean;
```
jonsequitur, line 338: We should avoid modifying this contract. If we need this property as an implementation detail, we can leave the contract alone by making the property internal.

src/Microsoft.DotNet.Interactive.Tests/LanguageKernelPackageTests.cs
jonsequitur: The test isn't loading any packages.

src/Microsoft.DotNet.Interactive/KernelSupportsNugetExtensions.cs
KevinRansom: So that we can use string or IHTMLContent.  I use a string to empty the message out.

src/Microsoft.DotNet.Interactive/PackageRestoreContext.cs
KevinRansom: The if allows me to add to requested, if it's not already in resolved.

src/dotnet-interactive-vscode/common/interfaces/contracts.ts
KevinRansom: Okay makes sense.  So long as t's never serialized right?

src/Microsoft.DotNet.Interactive/PackageRestoreContext.cs
KevinRansom: This is getpackagemanager lines, previously resolved nuget packages contribute to future resolutions also.

src/Microsoft.DotNet.Interactive.Tests/LanguageKernelPackageTests.cs
KevinRansom: Ahh but DoNugetRestore produces two messages always.
1 :  https://github.com/dotnet/interactive/blob/0dcf94198da95e8387295c39ab55a8ea941d9854/src/Microsoft.DotNet.Interactive/KernelSupportsNugetExtensions.cs#L216
2: https://github.com/dotnet/interactive/blob/0dcf94198da95e8387295c39ab55a8ea941d9854/src/Microsoft.DotNet.Interactive/KernelSupportsNugetExtensions.cs#L221

src/Microsoft.DotNet.Interactive.Tests/LanguageKernelPackageTests.cs
KevinRansom: I think there are two displayValueProduced.
1 for restore sources and 1 for packages loaded.

src/Microsoft.DotNet.Interactive/KernelSupportsNugetExtensions.cs
KevinRansom: done.

src/Microsoft.DotNet.Interactive/KernelSupportsNugetExtensions.cs
KevinRansom: I suppose I could have used null.

src/Microsoft.DotNet.Interactive.Tests/PackageRestoreContextTests.cs
KevinRansom: done

src/Microsoft.DotNet.Interactive/KernelSupportsNugetExtensions.cs
KevinRansom: done

src/Microsoft.DotNet.Interactive.Tests/PackageRestoreContextTests.cs
KevinRansom: What the NonExistentNugetPackage with the version # 99.99.99-NoReallyIDontExist?

Because the test is intended to verify, that a failing package restore does not cause future resolves to fail.  I suppose we could test that the first resolve failed.

src/Microsoft.DotNet.Interactive.ExtensionLab.Tests/DataFrameKernelExtensionTests.it_registers_formatters.received.json
```diff
@@ -0,0 +1,4 @@
+id           name         color        deliciousness
+1            apple        green        10           
+2            banana       yellow       11           
+3            cherry       red          9000
```
brettfo, line 4: This file shouldn't be part of the PR; it looks like an artifact from an earlier test run.

src/Microsoft.DotNet.Interactive/KernelSupportsNugetExtensions.cs
```diff
@@ -241,54 +243,24 @@ internal static KernelCommandInvocation DoNugetRestore()
...
-                                displayedValue.Update(
-                                    $"Installed package {package.PackageName} version {package.PackageVersion}");
-                            }
-                        }
                     }
                     else
                     {
                         var errors = $"{string.Join(Environment.NewLine, result.Errors)}";
-
+                        CreateOrUpdateDisplayValue(context, installingPackagesPropertyName, "", HtmlFormatter.MimeType);
```
jonsequitur: We probably shouldn't hard-code the MIME type at this layer. If we want more interesting formatting than just text, then pass in a rich object and set up a formatter for it.

src/Microsoft.DotNet.Interactive/KernelSupportsNugetExtensions.cs
jonsequitur: `null` which actually get written out as the string "null" by the default formatter.

src/Microsoft.DotNet.Interactive/PackageRestoreContext.cs
jonsequitur: But what does this property signify? I'm still not understanding it.

KevinRansom: REquested Package references are the collected set for this submission, resolve are the ones that were accumualated from successfull earlier submissions

src/Microsoft.DotNet.Interactive/PackageRestoreContext.cs
jonsequitur: Oh, I misread these as the same dictionary. In this case, we've introduced a potential concurrency issue here, though it's very unlikely given that the scheduler synchronizes command invocation. I wonder if consolidating these into a single collection would simplify things.

src/Microsoft.DotNet.Interactive.ExtensionLab.Tests/DataFrameKernelExtensionTests.it_registers_formatters.received.json
KevinRansom: good eyes.

src/Microsoft.DotNet.Interactive/PackageRestoreContext.cs
jonsequitur: I'm probably still missing the point but they're in separate collections that signify this exact distinction, so why the need for a property?

src/Microsoft.DotNet.Interactive.Tests/PackageRestoreContextTests.cs
jonsequitur: Ah, I see. A better test name might be "a failing package restore does not cause future resolves to fail". Based on the test name, I thought this was verifying that we don't re-attempt the invalid test.

src/Microsoft.DotNet.Interactive/ResolvedPackageReference.cs
```diff
@@ -32,15 +31,14 @@ public class ResolvedPackageReference : PackageReference
...
-            Requested = requested;
         }
 
         public IReadOnlyList<string> AssemblyPaths { get; }
 
         public IReadOnlyList<string> ProbingPaths { get; }
 
         public string PackageRoot { get; }
-        internal bool Requested { get; }
+        public bool Requested { get; }
```
colombod, line 41: you have to delete this property too, the contract tests are broken

colombod (changes requested):

src/Microsoft.DotNet.Interactive/ResolvedPackageReference.cs
KevinRansom: thanks

src/Microsoft.DotNet.Interactive.Tests/Server/SerializationTests.Event_contract_has_not_been_broken.approved.PackageAdded.json
```diff
@@ -1 +1 @@
-{"event":{"packageReference":{"assemblyPaths":["/path/to/a.dll"],"probingPaths":["/probing/path/1","/probing/path/2"],"packageRoot":"/the/package/root","packageName":"ThePackage","packageVersion":"1.2.3","isPackageVersionSpecified":true}},"eventType":"PackageAdded","command":{"token":"the-token","commandType":"SubmitCode","command":{"code":"#r \"nuget:ThePackage,1.2.3\"","submissionType":0,"targetKernelName":null}}}
```
jonsequitur, line 1: This looks like there's still a contract break.

KevinRansom:
Yes, and I think there may actually be a bug in the package manager ... too.

src/Microsoft.DotNet.Interactive/InstallPackagesMessage.cs
```diff
@@ -0,0 +1,105 @@
...
+        {
+            Formatter.Register<InstallPackagesMessage>(m => m.FormatAsPlainText(), PlainTextFormatter.MimeType);
+            Formatter.Register<InstallPackagesMessage>(m => m.FormatAsHtml(), HtmlFormatter.MimeType);
+            Formatter.SetPreferredMimeTypeFor(typeof(InstallPackagesMessage), PlainTextFormatter.MimeType);
+        }
+
+        public IEnumerable<string> RestoreSources;
+        public IEnumerable<string> InstalledPackages;
+        public IEnumerable<string> InstallingPackages;
+        public int Progress;
```
jonsequitur, line 28: Properties instead of fields, please.

src/Microsoft.DotNet.Interactive/InstallPackagesMessage.cs
```diff
@@ -0,0 +1,105 @@
...
+using Microsoft.AspNetCore.Html;
+using static Microsoft.DotNet.Interactive.Formatting.PocketViewTags;
+
+namespace Microsoft.DotNet.Interactive
+{
+    // Data bag containing the data to be displayed when reporting nuget resolve progress 
+    public class InstallPackagesMessage
+    {
+        static InstallPackagesMessage ()
+        {
```
jonsequitur, line 19: We shouldn't register formatters in a static constructor because they can be reset and there's no way to get back to this state. It could introduce some annoying test ordering dependencies, among other problems.

src/Microsoft.DotNet.Interactive/InstallPackagesMessage.cs
```diff
@@ -0,0 +1,105 @@
...
+                    strong(message),
+                    ul(items.Select(s => li(span(s + progress)))));
+            }
+            else
+            {
+                return div();
+            }
+        }
+
+        private string FormatAsHtml()
```
jonsequitur, line 56: Does this not get double-encoded?

In order to avoid that, this should return `IHtmlContent`.

src/Microsoft.DotNet.Interactive/InstallPackagesMessage.cs
KevinRansom: Oh okay :-)

src/Microsoft.DotNet.Interactive/InstallPackagesMessage.cs
KevinRansom: I will find somewhere to register it.

src/Microsoft.DotNet.Interactive/InstallPackagesMessage.cs
```diff
@@ -0,0 +1,105 @@
...
+        static InstallPackagesMessage ()
+        {
+            Formatter.Register<InstallPackagesMessage>(m => m.FormatAsPlainText(), PlainTextFormatter.MimeType);
+            Formatter.Register<InstallPackagesMessage>(m => m.FormatAsHtml(), HtmlFormatter.MimeType);
+            Formatter.SetPreferredMimeTypeFor(typeof(InstallPackagesMessage), PlainTextFormatter.MimeType);
+        }
+
+        public IEnumerable<string> RestoreSources;
+        public IEnumerable<string> InstalledPackages;
+        public IEnumerable<string> InstallingPackages;
```
jonsequitur, line 27: Any reason for these to be `IEnumerable<string>` rather than `IReadOnlyList<string>`?

KevinRansom: The only thing you can do with them is iterate them.  It's pretty much the smallest possible contract.

src/Microsoft.DotNet.Interactive/InstallPackagesMessage.cs
jonsequitur: I wonder if we should add a convention, so e.g. if `InstallPackagesMessage` implemented `ITypeFormatter<InstalPackagesMessage>`, we would just use that implementation.

The one issue there would be around MIME types, because it's intended to map to just one: 

https://github.com/dotnet/interactive/blob/54c98c456b7afeba45c599cba8aae4134a97c81e/src/Microsoft.DotNet.Interactive.Formatting/ITypeFormatter.cs#L8-L15

src/Microsoft.DotNet.Interactive/InstallPackagesMessage.cs
KevinRansom: If I understand this test correctly, I think it works as you would expect:   
https://github.com/dotnet/interactive/blob/54c98c456b7afeba45c599cba8aae4134a97c81e/src/Microsoft.DotNet.Interactive.Formatting.Tests/PocketViewWithFormatterTests.cs#L25

src/Microsoft.DotNet.Interactive/InstallPackagesMessage.cs
jonsequitur: If the underlying implementation is always expected to be non-dynamic / non-lazy, then we can avoid allocations (e.g. `Count` rather than `Count()`) using `IReadOnlyList<string>`. It's just a habit.

src/Microsoft.DotNet.Interactive/InstallPackagesMessage.cs
jonsequitur: Also just a habit: if you have a method that returns `string` but that `string` contains HTML, it's just a matter of time before it gets double encoded. But yeah this works like you expect:

image

src/Microsoft.DotNet.Interactive/InstallPackagesMessage.cs
KevinRansom: IReadOnlyList<string> is one of the ugliest type names in existance.

src/Microsoft.DotNet.Interactive/InstallPackagesMessage.cs
```diff
@@ -0,0 +1,105 @@
...
+                result.Append(System.Environment.NewLine);
+            }
+            return result.ToString();
+        }
+
+        static public void RegisterInstallPackagesMimeType(string mimeType)
+        {
+            Formatter.Register<InstallPackagesMessage>(m => m.FormatAsPlainText(), PlainTextFormatter.MimeType);
+            Formatter.Register<InstallPackagesMessage>(m => m.FormatAsHtml(), HtmlFormatter.MimeType);
+            Formatter.SetPreferredMimeTypeFor(typeof(InstallPackagesMessage), PlainTextFormatter.MimeType);
```
jonsequitur, line 102: The MIME type preference shouldn't be needed. This isn't the standard default and I would think HTML would be preferred.

src/Microsoft.DotNet.Interactive/KernelSupportsNugetExtensions.cs
```diff
@@ -189,25 +192,16 @@ internal static KernelCommandInvocation DoNugetRestore()
                         return;
                     }
 
-                    var messages = new Dictionary<PackageReference, string>(new PackageReferenceComparer());
-                    var displayedValues = new Dictionary<string, DisplayedValue>();
+                    InstallPackagesMessage.RegisterInstallPackagesMimeType(PlainTextFormatter.MimeType);
```
jonsequitur, line 195: We shouldn't do this on every call, because it might overwrite people's custom formatters.

You can add these registrations to `DefaultPlainTextFormatterSet` and `DefaultHtmlFormatterSet`.

src/Microsoft.DotNet.Interactive/InstallPackagesMessage.cs
KevinRansom: Actually, I like the plaintext better, it redraws without flashing.

src/Microsoft.DotNet.Interactive.Tests/LanguageKernelPackageTests.cs
KevinRansom: now we produce a single displayValueProduced for aeach restore

src/Microsoft.DotNet.Interactive.Tests/PackageRestoreContextTests.cs
KevinRansom: done

src/Microsoft.DotNet.Interactive/PackageRestoreContext.cs
KevinRansom: changed to use GetOrAdd()

src/Microsoft.DotNet.Interactive/PackageRestoreContext.cs
KevinRansom: I got rid of the property, it wasn't necessary to distinguish between them after all.

src/dotnet-interactive-vscode/common/interfaces/contracts.ts
KevinRansom: removed property

src/Microsoft.DotNet.Interactive/KernelSupportsNugetExtensions.cs
KevinRansom: using a data object + formatter now.

src/Microsoft.DotNet.Interactive/ResolvedPackageReference.cs
KevinRansom: removed property

src/Microsoft.DotNet.Interactive/InstallPackagesMessage.cs
KevinRansom: If I don't have to register formatters, then this can go.

src/Microsoft.DotNet.Interactive.PackageManagement/KernelSupportsNugetExtensions.cs
```diff
@@ -216,79 +208,43 @@ internal static KernelCommandInvocation DoNugetRestore()
...
                         totalWaitMs += delay;
-                        if (totalWaitMs > TimeSpan.FromMinutes(1.5).TotalMilliseconds)
-                        {
-                            throw new TimeoutException($"Package restore took longer than expected for packages: {string.Join(", ", newlyRequestedPackages.Select(p => p.PackageName))}.");
-                        }
+                        install.Progress = 1 + install.Progress;
+                        CreateOrUpdateDisplayValue(context, installPackagesPropertyName, install);
 
-                        foreach (var package in messages.Keys.ToArray())
+                        if (totalWaitMs > TimeSpan.FromMinutes(1.5).TotalMilliseconds)
```
jonsequitur, line 214: We removed this timeout a couple of weeks ago.

src/Microsoft.DotNet.Interactive.PackageManagement/PackageRestoreContext.cs
```diff
@@ -51,7 +54,10 @@ private IEnumerable<string> NativeProbingRoots()
             }
         }
 
-        public void AddRestoreSource(string source) => _restoreSources.Add(source);
+        public void AddRestoreSource(string source)
```
jonsequitur, line 57: Given then change in the implementation, should this be `TryAddRestoreSource`?

src/Microsoft.DotNet.Interactive.Tests/LanguageKernelPackageTests.cs
```diff
@@ -541,13 +526,13 @@ public async Task Pound_i_nuget_with_multi_submissions_combines_the_Text_Updates
...
-            events.OfType<DisplayedValueUpdated>()
-                  .Last().FormattedValues
+            events.OfType<DisplayedValueProduced>()
                   .Should()
-                  .ContainSingle(e => e.MimeType == HtmlFormatter.MimeType)
+                  .ContainSingle(v => v.Value is InstallPackagesMessage)
                   .Which.Value
+                  .As<InstallPackagesMessage>()
+                  .RestoreSources
+                  .Aggregate((s, acc) => acc + " & " + s)
```
jonsequitur, line 535: Instead of `Aggregate` why not use `BeEquivalentTo`? It's stricter and simpler.

KevinRansom: done

src/Microsoft.DotNet.Interactive.PackageManagement/PackageRestoreContext.cs
KevinRansom: There is not really a failure mode, it either adds the source or it was already there.  So Try is not really valuable on the name ... but GetOrAdd is indeed a bit clunky, however, it's the form I have seen most commonly ...

KevinRansom: Got to say GetOrAdd on ConcurrentDictionary is a bit tricky to parse too.

jonsequitur: Not critical. I do think `Try*` helps message the intention a bit better.

KevinRansom: The intention is to add the restore, whether it is in or not.  It's not to Try and add the restore, but sure I can make the change since you think it makes it clearer.

jonsequitur: I think it only makes it clearer because a bunch of other APIs with this behavior use the same naming convention. If I _really_ wanted to make it clearer, I'd call it something like `AddRestoreSourceIfSomeoneDidntAlreadyAddItAndDontThrowIfTheyDid`. 😂

jonsequitur: > Got to say GetOrAdd on ConcurrentDictionary is a bit tricky to parse too.

Yeah, it seems to imply that if it's going to be added, then you can't get it. _**It's one or the other!**_

KevinRansom:
````
AddRestoreSourceIfSomeoneDidntAlreadyAddItAndDontThrowIfTheyDid
````
Now that is a proper DotNet Api name.

jonsequitur (approved):

colombod (approved):

PR #1251, opened 2021-04-15
`is null` instead of `== null`

WhiteBlackGoose:
I don't know if it's useful for the project.

- [x] `== null` -> `is null`, `!= null` -> `is not null` (it's not only nicer, but also a bit more secure)

### DIFF
--- a/src/Microsoft.DotNet.Interactive.CSharp/CSharpKernel.cs
+++ b/src/Microsoft.DotNet.Interactive.CSharp/CSharpKernel.cs
@@ -153,7 +153,7 @@ namespace Microsoft.DotNet.Interactive.CSharp
             var service = QuickInfoService.GetService(document);
             var info = await service.GetQuickInfoAsync(document, cursorPosition, context.CancellationToken);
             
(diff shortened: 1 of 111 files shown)


WhiteBlackGoose:
Like it was recommended by @jonsequitur, the only change performed in this PR is changing `== null` into `is null`, `!= null` into `is not null`. There are a couple of minor changes in tests (to be precise, Linq.Expression doesn't allow for `is`-syntax, so we need to use `string.IsNullOrEmpty` instead).

WhiteBlackGoose:
@colombod could you give a quick review?

Also, I don't know if I need to ping to get a review, so please tell me. Should I silently wait or better ping a maintainer when a PR's ready?

colombod (approved):

WhiteBlackGoose:
@colombod thank you, but what about my question? 😄 . I always feel a bit guilty for pinging you or Jon every time, but I don't know whether it's normal. Maintainers normally request reviews, but I don't have permissions for it. So should I keep pinging or should I wait for review silently?

Issue #847, opened 2020-10-21
Add Save Notebook in PDF format option


RandomFractals:
Would be nice to have this option for .dib notebooks.

See this article on Jupyter Notebook  to PDF for more info:

https://towardsdatascience.com/jupyter-notebook-to-pdf-in-a-few-lines-3c48d68a7a63

Doesn't have to be restricted to just Py notebooks.

I would like to see this option for all the .net languages & polyglot notebooks this .net interactive extension supports.

PR #1270, opened 2021-04-19
Minor fixes of languages and versions for Jupyter

WhiteBlackGoose:
~~I do **not** know whether I will succeed at adding it XD. Let's see.~~

~~#847~~

- [x] Minor fix to languages names (so that you can see proper buttons in File>Download as)
As far as I could understand, it might also make sense to make the default language versions to 9.0 and 5.0 for C# and F# respectively. At least, that's the languages which work. Did I do it right?
~~- [ ] Exporting a notebook into `html` to ~a temporary file~ to a variable.~~
~~This command~~
~~```~~
~~jupyter-nbconvert CSharpTest.ipynb --output-dir=. --stdout~~
~~```~~
~~will return an `html` code.~~
~~- [ ] Converting the exported html ~file~ code into pdf~~
~~

Alright, this PR only performs minor changes to versions and language mismatch. 

**Why this PR doesn't close 847**: there are outside tools which already do it for any ipynb notebook, so no need to add it to dotnet interactive explicitly. It might be also added as a frontend extension to Jupyter lab, but it will be still a separate tool.

### DIFF
--- a/src/Microsoft.DotNet.Interactive.Jupyter/Protocol/LanguageInfo.cs
+++ b/src/Microsoft.DotNet.Interactive.Jupyter/Protocol/LanguageInfo.cs
@@ -64,14 +64,14 @@ namespace Microsoft.DotNet.Interactive.Jupyter.Protocol
 
     public class CSharpLanguageInfo : LanguageInfo
     {
-        public CSharpLanguageInfo(string version = "8.0") : base("C#", version, "text/x-csharp", ".cs", pygmentsLexer: "csharp")
+        public CSharpLanguageInfo(string version = "9.0") : base("C#", version, "text/x-csharp", ".cs", pygmentsLexer: "csharp")
         {
         }
     }
 
     public class FSharpLanguageInfo : LanguageInfo
     {
-        public FSharpLanguageInfo(string version = "4.5") : base("F#", version, "text/x-fsharp", ".fs", pygmentsLexer: "fsharp")
+        public FSharpLanguageInfo(string version = "5.0") : base("F#", version, "text/x-fsharp", ".fs", pygmentsLexer: "fsharp")
         {
            
         }
--- a/src/Microsoft.DotNet.Interactive.Jupyter/Protocol/LanguageInfo.cs
+++ b/src/Microsoft.DotNet.Interactive.Jupyter/Protocol/LanguageInfo.cs
@@ -71,7 +71,7 @@ namespace Microsoft.DotNet.Interactive.Jupyter.Protocol
 
     public class FSharpLanguageInfo : LanguageInfo
     {
-        public FSharpLanguageInfo(string version = "4.5") : base("C#", version, "text/x-fsharp", ".fs", pygmentsLexer: "fsharp")
+        public FSharpLanguageInfo(string version = "4.5") : base("F#", version, "text/x-fsharp", ".fs", pygmentsLexer: "fsharp")
         {
            
         }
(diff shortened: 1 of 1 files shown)


src/Microsoft.DotNet.Interactive.Jupyter/Protocol/LanguageInfo.cs
```diff
@@ -71,7 +71,7 @@ public CSharpLanguageInfo(string version = "8.0") : base("C#", version, "text/x-
 
     public class FSharpLanguageInfo : LanguageInfo
     {
-        public FSharpLanguageInfo(string version = "4.5") : base("C#", version, "text/x-fsharp", ".fs", pygmentsLexer: "fsharp")
+        public FSharpLanguageInfo(string version = "4.5") : base("F#", version, "text/x-fsharp", ".fs", pygmentsLexer: "fsharp")
```
colombod, line 74: not sure if these are the right version at this point in time

WhiteBlackGoose: Got it. I already commited a fix of them, just haven't pushed yet.

src/Microsoft.DotNet.Interactive.Jupyter/Protocol/LanguageInfo.cs
```diff
@@ -64,14 +64,14 @@ public LanguageInfo(string name, string version, string mimeType, string fileExt
 
     public class CSharpLanguageInfo : LanguageInfo
     {
-        public CSharpLanguageInfo(string version = "8.0") : base("C#", version, "text/x-csharp", ".cs", pygmentsLexer: "csharp")
+        public CSharpLanguageInfo(string version = "9.0") : base("C#", version, "text/x-csharp", ".cs", pygmentsLexer: "csharp")
```
WhiteBlackGoose, line 67: If I understand it correctly, C# 9 is supported too

src/Microsoft.DotNet.Interactive.Jupyter/Protocol/LanguageInfo.cs
```diff
@@ -64,14 +64,14 @@ public LanguageInfo(string name, string version, string mimeType, string fileExt
...
         {
         }
     }
 
     public class FSharpLanguageInfo : LanguageInfo
     {
-        public FSharpLanguageInfo(string version = "4.5") : base("F#", version, "text/x-fsharp", ".fs", pygmentsLexer: "fsharp")
+        public FSharpLanguageInfo(string version = "5.0") : base("F#", version, "text/x-fsharp", ".fs", pygmentsLexer: "fsharp")
```
WhiteBlackGoose, line 74: So is F# 5

src/Microsoft.DotNet.Interactive.Jupyter/Protocol/LanguageInfo.cs
jonsequitur: Correct, with some caveats that not all C# 9 features may be fully supported in C# scripting.

WhiteBlackGoose:
I thought I would finally do something great, but, as usual, just super-minor fixes. Anyway, here's an example of fix:

WhiteBlackGoose:
@colombod why did it exceed a time limit? Anyway, judging by the fact that everything passes on Windows, could you review it?

jonsequitur:
We've been seeing intermittent deadlocks in the Linux tests. Sixty minutes is the cutoff for the build but it should only take about half that long. I'll restart it.

WhiteBlackGoose:
Now it passes

WhiteBlackGoose:
@colombod could you please review?

jonsequitur (approved):

jonsequitur:
Thanks, @WhiteBlackGoose!

PR #1279, opened 2021-04-20
Displaying bitmap?

WhiteBlackGoose:
It's an experiment. Do **NOT** look at the code now (you will want to ban me after looking at it). But here's the thought: what if we can display bitmap via base64 encoding? Here is the result of what I just did:
image

So to sum up: we create a default html formatter for bitmap, which would convert a bitmap into bytes, then into base64, and thanks to html we can store it right in the notebook. And display it.

### DIFF
--- a/src/Microsoft.DotNet.Interactive.Formatting.Tests/HtmlFormatterTests.cs
+++ b/src/Microsoft.DotNet.Interactive.Formatting.Tests/HtmlFormatterTests.cs
@@ -12,6 +12,7 @@ using FluentAssertions;
 using FluentAssertions.Extensions;
 using Xunit;
 using static Microsoft.DotNet.Interactive.Formatting.Tests.Tags;
+using System.Drawing;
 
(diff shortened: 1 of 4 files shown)


WhiteBlackGoose:
@jonsequitur @colombod thoughts?

WhiteBlackGoose:
Related: #67

colombod:
Look at this [file](https://github.com/pi-top/pi-top-4-.NET-SDK/blob/master/src/PiTop.Camera.InteractiveExtension/KernelExtension.cs) from pi-top[4] sdk where we do formatters for few image formats.

I do same job but probably here you are looking at making extensions for nuget packages like `System.Drawing.Common` via extension pacakges `System.Drawing.Common.Interactive`

And possibly `OpenCvSharp.Interactive` and `SharpImage.Interactive`.

Those packages would be useful as out of band extensions users can bring along with their favourite image manipulation package. `SharpImage` is intersting as doesn't have any native dll requirement.

PR #1286, opened 2021-04-21
update to new apis

brettfo:
This is a big one.

The VS Code notebook APIs have taken a large change that's mostly incompatible with the current Stable release.  Because of that a bunch of common code had to be moved into the `stable/src` and `insiders/src` directories and other helpers rearranged as necessary.

These changes aren't meant to be used in Stable currently, only in Insiders so we can unblock our users.  The differences between the two will be reconciled before the next Stable release which will then bring both back into close alignment.

There are some bugs yet to be filed against VS Code once this has been merged, specifically:

- [x] Notebook controller webview not getting `postMessage` calls via `onDidReceiveMessage()`.  **Because of this bug, JavaScript cells don't behave as expected.** (microsoft/vscode#121895)
- [ ] `onDidChangeNotebookAssociation` should fire even on initial notebook open (microsoft/vscode#121904).
- [ ] cannot set cell language using mouse on the cell itself, only via command palette (microsoft/vscode#121927)

### DIFF
--- a/src/dotnet-interactive-vscode/common/ipynbUtilities.ts
+++ b/src/dotnet-interactive-vscode/common/ipynbUtilities.ts
@@ -19,7 +19,7 @@ function isDotNetCellMetadata(arg: any): arg is DotNetCellMetadata {
         && typeof arg.language === 'string';
 }
 
-export function getDotNetMetadata(metadata: CellMetadata | undefined): DotNetCellMetadata {
+export function getDotNetMetadata(metadata: any): DotNetCellMetadata {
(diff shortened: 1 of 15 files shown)


colombod (approved):

colombod:
microsoft/vscode#121895 was closed and latest insiders addressed the webview ipc issue

colombod (changes requested):
Can't select cell language with mouse gesture, it flashes only, using the command to select the cell language works and also triggers the `onDidChangeNotebookAssociation`

PR #1295, opened 2021-04-22
remove custom .ipynb handling and update apis

brettfo:
Remove all custom Jupyter/.ipynb handling.  The default new notebook is still .ipynb until microsoft/vscode#121974 is fixed, but then we'll default to .dib.

Scenarios to validate in Insiders:
1. Double-click existing .NET `.ipynb`; our controller should be selected and cell languages should be correct. **Our controller is properly selected based on our specified affinity, but due to microsoft/vscode#121904 the cell languages aren't yet correct.  There is currently no workaround.**
2. Double-click existing `.dib`; everything should work as before. **This is good.**
3. New Jupyter notebook, select one of our languages.  Our kernel controller should be selected.  **Our controller is properly selected, but again, due to microsoft/vscode#121904 the cell languages aren't yet correct.**


colombod:
in my environment I don't have python extension so our controller is the only option
So far:
1. open dib works
2. new blank notebook using the command `Jupyter : create new blank notebook` selects our controller but languages are wrong
3. open ipynb our controller is selected but languages are wrong
4. new blank notebook using the command `.NET Interactive : create new blank notebook` selects our controller but languages are wrong

colombod:
Our notebook handling setting medatada on cells when working with dib files makes the document dirty and that causes the infinite loop of reopen when closing a dirty document even discarding the edits, the document seems to be modified and vscode will reopen

PR #1296, opened 2021-04-23
wip while waiting on Brett

colombod:
Work on top of #1295 "

Scenarios to validate in Insiders:
1. [ ] Double-click existing .NET `.ipynb`; our controller should be selected and cell languages should be correct. 
2. [x] Double-click existing `.dib`; everything should work as before. (now works)
3. [ ] New Jupyter notebook, select one of our languages.  Our kernel controller should be selected. 

Issues
sql cell metadata not persisted:
Save a notebook with this cell
image

Close notebook, then reopen:
image


src/dotnet-interactive-vscode/common/vscode/extension.ts
```diff
@@ -142,7 +142,7 @@ export async function activate(context: vscode.ExtensionContext) {
     context.subscriptions.push(vscode.workspace.onDidRenameFiles(e => handleFileRenames(e, clientMapper)));
 
     // language registration
-    context.subscriptions.push(vscode.workspace.onDidOpenTextDocument(async e => await updateNotebookCellLanguageInMetadata(e)));
```
colombod, line 170: Thank you @jrieken!
The ` vscode.workspace.onDidOpenTextDocument` causes the dirty and unsaved notebook to be reopen forever. The suggestion to use the `vscode.notebook.onDidOpenNotebookDocument` event also let us batch the edits as a single apply call

brettfo:
Issues addressed in #1295.

Issue #1266, opened 2021-04-17
After loading existing notebook, all cells default to C# language


pkese:
If I restart the kernel or load an existing notebook,  
all cells with F# code are marked as C# and won't run until I manually switch them back to F#.

I've tested this in vscode-insiders.

This is so annoying, that it practically renders the whole interactive concept useless.

colombod:
Hi, we are tracking this issue (microsoft/vscode-jupyter#5477) as it is surfaced by the Jupyter file handling and not the .NET Interactive itself. We are actively working with the team to address this issues as soon as possible

jhoneill:
> Hi, we are tracking this issue ([microsoft/vscode-jupyter#5477](https://github.com/microsoft/vscode-jupyter/issues/5477)) as it is surfaced by the Jupyter file handling and not the .NET Interactive itself. We are actively working with the team to address this issues as soon as possible

Unsurprisingly this also breaks PowerShell.. 
Lots of these messages 
![image](https://user-images.githubusercontent.com/16209593/115250407-71319680-a121-11eb-97af-8f574547b009.png)


Things might not be quite so bad if .NET Interactive in VS Code did not treat every  notebook as C# and then say every cell is an 
exception.

gbaydin:
I have faced the same problem as well. I agree that this is a major problem affecting usability.

jonsequitur:
We're trying to address this issue tactically in the short term via #1281. This is partly because while this issue is frustrating, we want to make the workaround apparent. 

The actual solution will take a bit longer for us to work through with the VS Code team and other teams building notebook experiences, which is why we spent time on the tactical bandage.

jonsequitur:
(And even without #1281, the workaround is available. From the command pallet, select `View: Reopen editor with` and then `.NET Interactive for Jupyter`.

jhoneill:
I think this is being mis-diagnosed. 

.Net Interactive can mix cells of multiple languages in a single notebook.  
Jupyter server cannot

Jupyter will create a notebook like this 

```
"cells": [
  {   "cell_type": "code",   "metadata": {}  }
],
"metadata": {
  "kernelspec": {   "display_name": ".NET (PowerShell)",   "language": "PowerShell",   "name": ".net-powershell"  },
  "language_info": {   "file_extension": ".ps1",   "mimetype": "text/x-powershell",   "name": "PowerShell"   }
```

When loaded in VS code this changes we get   
[...]

PR #1299, opened 2021-04-26
Preserve original kernel spec if present

colombod:
Addressing issue #1266

### DIFF
--- a/src/dotnet-interactive-vscode/common/ipynbUtilities.ts
+++ b/src/dotnet-interactive-vscode/common/ipynbUtilities.ts
@@ -127,20 +127,29 @@ export const requiredLanguageInfoData = {
 };
 
 export function withDotNetKernelMetadata(metadata: { [key: string]: any } | undefined): any | undefined {
-    const result = {
+    let result = {
         ...metadata,
         custom: {
             ...metadata?.custom,
             metadata: {
                 ...metadata?.custom?.metadata,
                 kernelspec: {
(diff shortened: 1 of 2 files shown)


jonsequitur (approved):

src/dotnet-interactive-vscode/common/ipynbUtilities.ts
```diff
@@ -127,20 +127,29 @@ export const requiredLanguageInfoData = {
...
     };
+    let name = result?.custom?.metadata?.kernelspec?.name
+
+    if (typeof name !== 'string' || !name.startsWith(".net-")) {
+        result.custom.metadata.kernelspec = {
+            ...metadata?.custom?.metadata?.kernelspec,
+            ...requiredKernelspecData,
+        };
+
+        result.custom.metadata.language_info = { ...requiredLanguageInfoData };
```
brettfo, line 150: This isn't quite correct, because `requiredLanguageInfoData` always defaults to C#.  I think the best solution would be to leave the metadata untouched if the kernelspec name was determined to be good, e.g.,

``` typescript
const kernelspecName = metadata?.custom?.metadata?.kernelspec?.name;
if (typeof kernelspecName === 'string' && kernelspecName.startsWith('.net-')) {
    return metadata; // don't change anything
}

// otherwise the previous contents of this method as they were before
```

brettfo (approved):

Issue #1288, opened 2021-04-22
Comment after "r" magic command causes unexpected behaviour


WhiteBlackGoose:
### Describe the bug
Let's consider installation of Plotly.NET:
```cs
#r "nuget: Plotly.NET.Interactive, *-*"
open Plotly.NET
let xData = [0. .. 10.]
let yData = [0. .. 10.]
let myFirstChart = Chart.Point(xData,yData)
myFirstChart
```
The output of the first magic command is
```
Installed package Plotly.NET.Interactive version 2.0.0-beta8
Loading extensions from Plotly.NET.Interactive.dll

Added Kernel Extension including formatters for GenericChart
```
<details><summary><strong>Screenshot</strong></summary>
<img src="https://user-images.githubusercontent.com/31178401/115667838-f77ef000-a34e-11eb-8d2a-fd9fccbe66df.png">
</details>

Now, let me add a comment right after the magic command and keep all other commands (restart beforehand):
```cs
#r "nuget: Plotly.NET.Interactive, *-*" // ducks best
open Plotly.NET
let xData = [0. .. 10.]
let yData = [0. .. 10.]
let myFirstChart = Chart.Point(xData,yData)
myFirstChart
```
Now, the package is **still installed**. But
1) It doesn't produce the log message (like "package installing...")
2) What's more important, it does not apply the kernel extension formatter.
<details><summary><strong>Screenshot</strong></summary>
<img src="https://user-images.githubusercontent.com/31178401/115668256-8ee44300-a34f-11eb-8c7e-2053b38d81fa.png">
</details>

**Which version of .NET Interactive are you using?**: 2021-04-20T18:25:10.9573324Z
- OS
    - [x] Windows 10
 - Browser 
    - [x] Opera
 - Frontend
    - [x] Jupyter Notebook

brettfo:
Magic commands, including `#r`, etc., don't allow for inline comments.

PR #1303, opened 2021-04-26
Fixing #1288

WhiteBlackGoose:


WhiteBlackGoose:
Oh, alright, forgot that I can't comment on snippets that I haven't changed.

I found the place where it cannot parse `//`. [This](https://github.com/WhiteBlackGoose/interactive/blob/59f98f0d840566bb25a228fb891d454429f12ede/src/Microsoft.DotNet.Interactive/Parsing/SubmissionParser.cs#L267) thing creates a parser, which I cannot debug [here](https://github.com/WhiteBlackGoose/interactive/blob/59f98f0d840566bb25a228fb891d454429f12ede/src/Microsoft.DotNet.Interactive/Parsing/DirectiveNode.cs#L35).

After this thing parses the given submit code, it returns three errors and three unmatched tokens (`//`, `ducks`, `best!`). Since I don't know what's inside `commandLineBuilder` and how to debug, could you please give some tips on it? @jonsequitur, @colombod.

jonsequitur:
That parser is implemented by the System.CommandLine library. The repo is here: https://github.com/dotnet/command-line-api.

WhiteBlackGoose:
So what do we do? It seems to be an expected behaviour for that library, but doesn't seem so for the `r` magic command. I can think of a workaround, but that's probably not we want.

jonsequitur:
We could potentially trim off the comment before passing it to the parser here: https://github.com/dotnet/interactive/blob/9d899fb9fc1643210b85b64cce727b1b8319c416/src/Microsoft.DotNet.Interactive/Parsing/DirectiveNode.cs#L35

WhiteBlackGoose:
> The thing to keep in mind is that // is perfectly valid within the magic command itself, e.g. #!value --from-url https://example.com.

That's exactly why I'd call trimming as a workaround. Alright, thanks, I'll think

WhiteBlackGoose:
I noticed that you rely on an external `Parser`. I think it might be a good idea to have a wrapper over it, since this `Parser` is only used in the context of directive parsers, this wrapper is `DirectiveParser`. It has a private field of type `Parser`. Now, we can customize its behaviour without ugly workarounds. It also imho looks better (since fields like `directiveParser` are now exactly of the `DirectiveParser` type).

@jonsequitur I would like to hear your thoughts.

src/Microsoft.DotNet.Interactive/Parsing/DirectiveParser.cs
```diff
@@ -0,0 +1,59 @@
...
+        }
+
+        public ParseResult Parse(string input)
+        {
+            if (FindComment(input) is { } realPos)
+                input = input.Substring(0, realPos);
+            return _commandLineParser.Parse(input);
+        }
+
+        private static int? FindComment(string input)
```
WhiteBlackGoose, line 40: Oh, and unlike the rest if the code, this part... well, it might be implemented better than I did, so any recommendations are welcomed.

src/Microsoft.DotNet.Interactive.Jupyter.Tests/MagicCommandTests.lsmagic.cs
```diff
@@ -115,5 +118,23 @@ public async Task lsmagic_does_not_list_hidden_commands()
                   .Should()
                   .NotContain("#!hidden");
         }
+
+        [Fact]
+        public async Task r_executes_even_with_comment()
```
jonsequitur, line 123: Rather than this more integrated test, it would make sense to test `DirectiveParser` directly.

WhiteBlackGoose: This makes sense, thanks.

src/Microsoft.DotNet.Interactive/Parsing/DirectiveParser.cs
```diff
@@ -0,0 +1,59 @@
...
+            var secondQuote = input.IndexOf('"', firstQuote + 1);
+            if (secondQuote is -1)
+                return null;
+
+            var doubleSlash = input.IndexOf("//", secondQuote + 1);
+            if (doubleSlash is -1)
+                return null;
+
+            return doubleSlash;
+        }
```
jonsequitur, line 55: It would be useful to start with a few test cases, possibly using `[Theory]`. Quote-counting might not be valuable, since it's a) not required for `#r`, e.g. `#r nuget:package` is valid, and b) it's valid to surround any token with quotes, and it's commonly used to indicate a single token when there's a space in it, e.g. `#!blah "this is parsed as one token"`.

As for test cases, none of the following should be identified as containing a comment:

```console
#!value --from-file //volume/path/to/file
#!value --from-url https://example.com
```

WhiteBlackGoose: I thought they aren't directives. If they are, the thing becomes way less trivial. Here's what I see as possible solutions:
- Make a wrapper around `ParseResult` too. The reason is that after we got a `ParseResult` from the command line parser, we need to see if the unmatched tokens are in fact comments. Since `ParseResult` is immutable (which is, in fact, good), we need to extract the necessary data from it, draw our own conclusions regarding comments, and then return our `DirectiveParseResult`.
- Change the behaviour. Make `#r nuget:DuckPackage //quack` raise an error as with invalid syntax (now it swallows errors with no feedback).

What do you think?

jonsequitur: Magic commands and directives are the same thing at the implementation level. If there's any semantic distinction it's that a "directive" refers to something that the compiler is or someday will be aware of, i.e. `#i` and `#r`. (Support for these is currently not at parity across C# and F#.)

Since a comment like this is valid for a directive in the underlying language, I think it's useful to fix this. A more targeted fix, just for `#i` and `#r`, might be easier. Distinguishing a comment from valid inputs above will still be challenging though.

WhiteBlackGoose: Given that the command line parser can distinguish, I can reinterpret its parse result and see, if the first unmatched token starts with `//`. Okay, let me see if I actually can. I will prepare a number of sets, and you will say if you think that's the behaviour you'd expect.

src/Microsoft.DotNet.Interactive.Tests/Parsing/DirectiveParserTests.cs
```diff
@@ -0,0 +1,38 @@
...
+using System.Threading.Tasks;
+using Xunit;
+using System;
+using System.CommandLine;
+
+namespace Microsoft.DotNet.Interactive.Tests.Parsing
+{
+    public class DirectiveParserTests
+    {
+        private static Command GetRDirective()
```
WhiteBlackGoose, line 17: `r` is private, so I can make it internal and make the assembly visible to the test. Another option is to test `SubmissionParser`, but it's a bit more complicated and high-level than `DirectiveParser`

src/Microsoft.DotNet.Interactive/Parsing/DirectiveParser.cs
```diff
@@ -0,0 +1,79 @@
...
+            commandLineBuilder.EnableDirectives = false;
+            _commandLineParser = commandLineBuilder.Build();
+        }
+
+        public DirectiveParseResult Parse(string input)
+        {
+            var parseResult = _commandLineParser.Parse(input);
+            IReadOnlyList<string> unmatchedTokens;
+            IReadOnlyCollection<ParseError> errors;
+            if (parseResult.UnmatchedTokens.Count == parseResult.Errors.Count && parseResult.UnmatchedTokens.Count > 0 && parseResult.UnmatchedTokens[0].StartsWith("//"))
```
WhiteBlackGoose, line 59: @jonsequitur is it a bad solution, or *very* bad? By this trick when I get as many errors as unmatched tokens, all those errors look like `Unrecognized token bla-bla`, so we can check if the first token starts with `//`.

jonsequitur: Interesting solution. It's not too, too bad, given the alternative would probably be to introduce a whole other parser.

This could be tightened up further by making sure that the unmatched tokens all come after the matched tokens. It's possible, I think, for there to be unmatched tokens in between matched tokens.

src/Microsoft.DotNet.Interactive.Jupyter.Tests/MagicCommandTests.lsmagic.cs
jonsequitur: You can grab the directive's `Command` directly and parse whatever you like:

image

src/Microsoft.DotNet.Interactive/Parsing/DirectiveParser.cs
WhiteBlackGoose: Aha, I see. Alright, then I'll tighten it up and add tests.

src/Microsoft.DotNet.Interactive.Jupyter.Tests/MagicCommandTests.lsmagic.cs
WhiteBlackGoose: Got it. I currently extract it via reflection, I'll switch to this way then.

WhiteBlackGoose: Although it would require having a kernel, which we wanted to avoid for those tests

jonsequitur: If there's a refactoring that makes it easier to test this in isolation, that would be useful. Otherwise, I'd prefer to work through the existing public API rather than use reflection, `InternalsVisibleTo`, etc. even if it pulls in a `Kernel`.

WhiteBlackGoose:
@jonsequitur despite that the parser now works correctly, the reaction to inputs is still strange. Here:
image

In both cases it *applied* the kernel extension, which is good. But in the first case it somehow managed to see those tokens *even though* the parser returned 0 errors (so the parser returns the same thing for both `bla-bla //comment` and `bla-bla // comment` (when there's no space and when there is)).

It seems like there's some other parser for all of it... not sure why you would need two parsers simultaneously...

Another *possible* reason is `KeyValueOptions`. Instead of rewriting it into `DirectiveParseResult`, I just take `ParseResult` and pass there where needed, hoping that `KeyValueOptions` doesn't affect the behaviour. It's hard to rewrite it, because it also abuses some model binder from the CLI API.

To sum up: there are two possible reasons for this behaviour. It's either that there's a second parser which gives a result different from `DirectiveParser`. Or that `KeyValueOptions`, which still take the old `ParseResult` (which contains errors) affect this behaviour. What do you think?

jonsequitur:
Hard to tell. Is the `SubmissionParser` putting the comment into a separate node?

I wouldn't expect `KeyValueOptions` to be involved unless `#!value` is being invoked.

WhiteBlackGoose:
Okay, so we need to do something with this PR. It is clear that this hack... I'd be scared to have it in the actual code. What I made helped to reveal a few problems with the parser, so for example, if you build the code from my PR, you will see, that
```cs
#somemagiccommand --its --arguments // comment with spaces
```
gives an error (whereas the code in `master` would swallow it, at least, the time back when I was working on this PR),
and
```cs
#somemagiccommand --its --arguments //commentwithnospaces
```
would work correctly (unlike the code in `master`, at least that time back). In both cases, as can be seen in the [screenshot](https://github.com/dotnet/interactive/pull/1303#issuecomment-834612948) it calls the kernel extension (which is the goal, *but* there's still a fake parsing error).

I'm not sure I'm able to do something with it without rewriting a lot of things... but maybe this experience will help someone to fix it in a different way.

@jonsequitur if you don't want to have this PR opened, you can close it.

PR #1317, opened 2021-05-01
Improve contextual HTML formatting

jonsequitur:
This change plumbs `FormatContext` through all of the `PocketView` APIs and removes the need to pass the `TextWriter`, which is now a property of the `FormatContext`. Depth tracking is part of the `FormatContext` and nested tables are no longer allowed by default.

It also adds support for expressing dependencies so that, for example, a single `<style>` element can be added from within a nested formatter such that it won't create duplicate style elements. 

With some styles applied, the JSON formatter output now looks like this:

image


src/Microsoft.DotNet.Interactive.FSharp/FSharpHtml.fs
```diff
@@ -41,14 +42,7 @@ module Html =
...
                     |> List.toArray 
-                    |> Array.map (fun content -> 
-                         match content with 
-                         | :? HtmlElement as s -> 
-                            // Prefer to pass the object over for 
-                            match s with 
-                            | Obj o -> o
-                            | _ -> box s
-                         | _ -> box content)
+                    |> Array.map (fun content -> content)
```
TIHan, line 45: This line could be removed as it's mapping to itself.

src/Microsoft.DotNet.Interactive.PowerShell/Microsoft.DotNet.Interactive.PowerShell.csproj
```diff
@@ -34,40 +34,6 @@
     	CopyToOutputDirectory="PreserveNewest" />
   </ItemGroup>
 
-  <ItemGroup Condition="'$(NCrunch)' != '1'">
```
colombod, line 37: Careful :)

src/Microsoft.DotNet.Interactive.Formatting/Formatter.cs
```diff
@@ -403,13 +392,12 @@ public static void Register(ITypeFormatter formatter)
         /// <param name="formatter">The formatter.</param>
         /// <param name="mimeType">The MimeType for this formatter. If it is not specified it defaults to <see cref="PlainTextFormatter.MimeType"/></param>
         public static void Register<T>(
-            Func<FormatContext, T, TextWriter, bool> formatter,
+            FormatDelegate<T> formatter,
```
colombod, line 395: We need to inform external contributors, this will break XPlot.Plotly and Plotly.NET.Interactive as we merge. Is it possible to retain the old Register and label as deprecated?

jonsequitur: Do you know that they were using this overload? We can check easily enough.

colombod: this one for example [FormatterKernelExtension](https://github.com/plotly/Plotly.NET/blob/63cadbe361fea10c0dcd585ca78a2d08c682a178/src/Plotly.NET.Interactive/Extension.fs)

colombod: And [this](https://github.com/pi-top/pi-top-4-.NET-SDK/blob/master/src/PiTop.Camera.InteractiveExtension/KernelExtension.cs) is the code in pi-top

jonsequitur: That's using a different overload.

colombod (approved):

PR #1333, opened 2021-05-11
use DisplayAs instead of Display in value kernel

jonsequitur:


src/Microsoft.DotNet.Interactive/KernelInvocationContextExtensions.cs
```diff
@@ -35,6 +35,30 @@ public static class KernelInvocationContextExtensions
...
+        public static DisplayedValue DisplayAs(
+            this KernelInvocationContext context,
+            string value,
+            string mimeType = null)
+        {
+           
+            var displayId = Guid.NewGuid().ToString();
+
+            var formattedValue = new FormattedValue(
+                mimeType,
```
colombod, line 48: should use here the preferred mimeType if the one passed is null or whitespaces?

colombod (approved):

PR #1346, opened 2021-05-14
Formatter self registration and tabular data improvements

jonsequitur:
The goal here is to enable a cleaner pattern for formatters to self-register and enable multiple data explorers to provide HTML outputs over SQL and other data queries data rather than hard-coding a specific one as is currently the case.


src/Microsoft.DotNet.Interactive.Formatting.Tests/FormatterTests.cs
```diff
@@ -211,468 +197,538 @@ public void Check_default_HTML_formatting_of_unicode()
...
-        public class ConcreteClass1InheritingAbstractClass : AbstractClass
-        {
-        };
+                formatted.Should().Be($"Hello from {nameof(TypeWithCustomFormatter)} using MIME type {mimeType}");
+            }
 
-        public class ConcreteClass2InheritingAbstractClass : AbstractClass
-        {
-        };
+            [TypeFormatterSource(typeof(CustomFormatterSource))]
```
jonsequitur, line 626: I like the idea of allowing types to say "Here's how I want to be formatted!" but I don't love using attributes to do it. Unless we write an analyzer, there's no compiler feedback for specifying the wrong type. 

Any other ideas?

src/Microsoft.DotNet.Interactive.Formatting/TabularData/TabularDataFormatterSource.cs
```diff
@@ -0,0 +1,32 @@
...
+            {
+                IReadOnlyList<IHtmlContent> headers =
+                    value.Schema
+                         .Fields
+                         .Select(f => (IHtmlContent) td(span(f.Name)))
+                         .ToArray();
+
+                IReadOnlyList<IHtmlContent> rows =
+                    value.Data
+                         .Select(d => (IHtmlContent) tr(d.Values.Select(v => td(v))))
```
colombod, line 25: How do we ensure that the values are preserving the order of the column names?

jonsequitur: Nice catch!

Updating the test to capture this.

src/Microsoft.DotNet.Interactive.Formatting.Tests/TabularDataFormatterTests.cs
```diff
@@ -136,10 +139,34 @@ public void can_generate_tabular_json_from_dictionary()
             this.Assent(formattedData, _configuration);
         }
 
+        [Fact]
+        public void Tabular_data_resource_with_is_formatted_as_a_table()
```
colombod, line 143:
```suggestion
        public void Tabular_data_resource_is_formatted_as_a_table()
```

src/Microsoft.DotNet.Interactive.PowerShell/Microsoft.DotNet.Interactive.PowerShell.csproj
```diff
@@ -34,39 +34,6 @@
     	CopyToOutputDirectory="PreserveNewest" />
   </ItemGroup>
 
-  <ItemGroup Condition="'$(NCrunch)' != '1'">
```
colombod, line 37: we need this back

colombod (changes requested):

src/Microsoft.DotNet.Interactive.Formatting/TabularData/TabularDataFormatterSource.cs
jonsequitur: Actually I think the answer here is we don't. They're expected to be in the same order, otherwise we go down a kind of unfortunate perf path.

colombod (approved):

PR #1381, opened 2021-06-01
respond to latest api changes

brettfo:
More VS Code API changes, the most notable is that the property `NotebookCellOutput.outputs` has been renamed to `NotebookCellOutput.items`.  The old property name is still runtime compatible, but we should respond to the changes as soon as possible.

~~N.b., that there are 3 warnings when building the insiders extension:~~

- ~~`common/vscode/commands.ts`:134 - The namespace `notebook` was renamed to `notebooks`, but to enable this to build in both stable and insiders, the old name was preserved.  This will be runtime compatible for a short while.~~
- ~~`common/vscode/extension.ts`:174 - Same `notebook`/`notebooks` namespace issue as above.~~
- ~~`common/vscode/extension.ts`:208 - `viewType` was renamed to `notebookType`, but like the namespace change mentioned above, the old name will be runtime compatible for a bit.~~

Test notebook script was verified against both stable and insiders.


src/dotnet-interactive-vscode/common/tests/unit/utilities.ts
```diff
@@ -69,7 +69,7 @@ export function decodeNotebookCellOutputs(outputs: vscodeLike.NotebookCellOutput
     jsonLikeMimes.add('application/json');
     jsonLikeMimes.add(vscodeLike.ErrorOutputMimeType);
```
colombod, line 70: `jsonLikeMimes` -> `jsonLikeMimeTypes`

src/dotnet-interactive-vscode/insiders/src/notebookControllers.ts
```diff
@@ -67,8 +67,8 @@ export class DotNetNotebookKernel {
             }
         });
         this.commonControllerInit(jupyterController);
-        this.disposables.push(vscode.notebook.onDidOpenNotebookDocument(async notebook => {
-            if (notebook.viewType === jupyterViewType && isDotNetNotebook(notebook)) {
+        this.disposables.push(vscode.notebooks.onDidOpenNotebookDocument(async notebook => {
```
colombod, line 70: With today breaking changes this `vscode.notebooks.onDidOpenNotebookDocument` becomes `vscode.workspace.onDidOpenNotebookDocument`

colombod (changes requested):

src/dotnet-interactive-vscode/insiders/src/notebookControllers.ts
brettfo: This namespace change isn't live yet, but I'll do it as soon as it comes down.

colombod (approved):

colombod (approved):

PR #1385, opened 2021-06-03
Make DotNet.Interactive netstandard2.0

KevinRansom:
Yes it looks yuk.

### DIFF
--- a/src/Microsoft.DotNet.Interactive/KernelScheduler.cs
+++ b/src/Microsoft.DotNet.Interactive/KernelScheduler.cs
@@ -12,7 +12,7 @@ namespace Microsoft.DotNet.Interactive
 {
     static class DotNetStandardHelpers
     {
-#if NETSTANDARD2_1_OR_GREATER
+#if !NETSTANDARD2_0
         static public bool GetIsCompletedSuccessfully(this Task task)
     {
         return task.IsCompletedSuccessfully;
--- a/src/Microsoft.DotNet.Interactive/KernelScheduler.cs
+++ b/src/Microsoft.DotNet.Interactive/KernelScheduler.cs
@@ -10,6 +10,28 @@ using Pocket;
 
 namespace Microsoft.DotNet.Interactive
 {
+    static class DotNetStandardHelpers
+    {
+#if NETSTANDARD2_1_OR_GREATER
+        static public bool GetIsCompletedSuccessfully(this Task task)
+    {
+        return task.IsCompletedSuccessfully;
+    }
+#else
+        // NetStandard 2.1
+        // internal const int TASK_STATE_RAN_TO_COMPLETION = 0x1000000;                          // bin: 0000 0001 0000 0000 0000 0000 0000 0000
+        // public bool IsCompletedSuccessfully => (m_stateFlags & TASK_STATE_COMPLETED_MASK) == TASK_STATE_RAN_TO_COMPLETION;
+        // <see cref="IsCompleted"/> will return true when the Task is in one of the three
+        // final states: <see cref="System.Threading.Tasks.TaskStatus.RanToCompletion">RanToCompletion</see>,
+        // <see cref="System.Threading.Tasks.TaskStatus.Faulted">Faulted</see>, or
+        // <see cref="System.Threading.Tasks.TaskStatus.Canceled">Canceled</see>.
+        static public bool GetIsCompletedSuccessfully(this Task task)
+        {
+            return task.IsCompleted && !task.IsFaulted && !task.IsCanceled;
+        }
+#endif
+    }
+
     public class KernelScheduler<T, TResult> : IDisposable, IKernelScheduler<T, TResult>
     {
         private static readonly Logger Log = new("KernelScheduler");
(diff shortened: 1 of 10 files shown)


src/Microsoft.DotNet.Interactive/KernelScheduler.cs
```diff
@@ -128,7 +140,7 @@ private void Run(ScheduledOperation operation)
                                     {
                                         if (!operation.TaskCompletionSource.Task.IsCompleted)
                                         {
-                                            if (t.IsCompletedSuccessfully)
+                                            if (IsCompletedSuccessfully(t))
```
jonsequitur, line 143: This can still be an extension method.

src/Microsoft.DotNet.Interactive/KernelScheduler.cs
```diff
@@ -111,6 +111,18 @@ private void ScheduledOperationRunLoop(object _)
...
         }
 
+        // NetStandard 2.1
+        // internal const int TASK_STATE_RAN_TO_COMPLETION = 0x1000000;                          // bin: 0000 0001 0000 0000 0000 0000 0000 0000
+        // public bool IsCompletedSuccessfully => (m_stateFlags & TASK_STATE_COMPLETED_MASK) == TASK_STATE_RAN_TO_COMPLETION;
+        // <see cref="IsCompleted"/> will return true when the Task is in one of the three
+        // final states: <see cref="System.Threading.Tasks.TaskStatus.RanToCompletion">RanToCompletion</see>,
+        // <see cref="System.Threading.Tasks.TaskStatus.Faulted">Faulted</see>, or
+        // <see cref="System.Threading.Tasks.TaskStatus.Canceled">Canceled</see>.
+        static bool IsCompletedSuccessfully(Task task)
```
jonsequitur, line 121: Please specify the visibility.

src/Microsoft.DotNet.Interactive/KernelScheduler.cs
KevinRansom: I know, I will clean things up, much of this is ugly right now.

jonsequitur: You call this ugly? Pffff. I've seen uglier.

src/Microsoft.DotNet.Interactive/Microsoft.DotNet.Interactive.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>netstandard2.1</TargetFramework>
```
colombod, line 4: I thought we were multi targeting here

KevinRansom: @colombod , we will be, I wanted to run all of the tests against the netstandard2.0 assembly, to ensure nothing was broken.

jonsequitur: We were discussing whether running tests against both netstandard2.0 and netstandard2.1 is worth the trouble. The netstandard2.0 version will be for an internal-only prototype and not supported long-term, so I'm inclined to just fix bugs if/when they occur rather than add to the maintenance burden.

KevinRansom:
Yes, this isn't final, I repeat, I wanted to ensure that all of the tests passed.  You will note, that this doesn't build a package for us to give to partners.


Sent from Outlook<http://aka.ms/weboutlook>

KevinRansom:
@jonsequitur , @colombod , I think this is ready now.  We build a nuget package called ... Microsoft.DotNet.Interactive.Netstandard20

src/Microsoft.DotNet.Interactive/Microsoft.DotNet.Interactive.Netstandard20.csproj
```diff
@@ -0,0 +1,12 @@
+<Project>
+
+  <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>
+    <PackageId>Microsoft.DotNet.Interactive.Netstandard20</PackageId>
+    <PackageDescription>Core types for building applications providing interactive programming for .NET. for the desktop</PackageDescription>
+    <PackageTags>interactive desktop</PackageTags>
+  </PropertyGroup>
+
+  <Import Project="Microsoft.DotNet.Interactive.csproj" />
```
brettfo, line 10: If you do the `<Import>` before the `<PropertyGroup>` then you can remove a bunch of the parts that look like this `Condition="$(PackageId) == ''"` because they'll just get overwritten.

src/Microsoft.DotNet.Interactive/Microsoft.DotNet.Interactive.csproj
```diff
@@ -1,16 +1,16 @@
...
+    <TargetFramework Condition="'$(TargetFramework)'==''">netstandard2.1</TargetFramework>
```
brettfo, line 4: Why not multitarget here and remove the `*.Netstandard20` project altogether?  The resultant NuGet package will have both netstandard2.0 and 2.1 binaries.

brettfo (approved):

jonsequitur:
@KevinRansom Is this ready to merge?

PR #1392, opened 2021-06-04
more API updates and improve reporting of cell cancellation

brettfo:
There's one more VS Code API change.

This also fixes a scenario where cell cancellation (`CommandFailed`) was getting reported twice with the `reject()` function.

Test notebook script verified on both stable and insiders.


colombod (approved):

brettfo:
Looks like I broke some tests, oops.

PR #1393, opened 2021-06-04
load script extensions from NuGet packages

brettfo:
This is a purely source-based version of loading interactive extensions.  The existing binary-based one loads all assemblies from `<NUGET_PACKAGE_PATH>/interactive-extensions/dotnet/*.dll` so this one plays off of that scenario by simply loading and executing the file `<NUGET_PACKAGE_PATH>/interactive-extensions/dotnet/extension.dib` if it was found.

This takes advantage of the fact that a multi-cell `.dib` file can be executed with a single call to `SubmitCode`.

### DIFF
--- a/src/Microsoft.DotNet.Interactive.Tests/Utility/KernelExtensionTestHelper.cs
+++ b/src/Microsoft.DotNet.Interactive.Tests/Utility/KernelExtensionTestHelper.cs
@@ -9,6 +9,7 @@ using System.Runtime.CompilerServices;
 using System.Text;
 using System.Text.RegularExpressions;
 using System.Threading.Tasks;
+using System.Xml.Linq;
 using Microsoft.DotNet.Interactive.Utility;
 
 namespace Microsoft.DotNet.Interactive.Tests.Utility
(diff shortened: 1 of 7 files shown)


src/Microsoft.DotNet.Interactive.Tests/Utility/KernelExtensionTestHelper.cs
```diff
@@ -71,18 +74,54 @@ public static ExtensionPackage GetOrCreateFileProviderExtension()
...
+            {
+                if (_scriptBasedExtensionPackage is null)
+                {
+                    var projectDir = DirectoryUtility.CreateDirectory();
+                    var packageName = $"MyTestExtension.{Path.GetRandomFileName()}";
+                    var packageVersion = "2.0.0-" + Guid.NewGuid().ToString("N");
+
+                    var extensionScriptPath = new FileInfo(Path.Combine(projectDir.FullName, "extension.dib"));
+                    var extensionScriptContent = @"
+#! csharp
```
jonsequitur, line 89:
```suggestion
#!csharp
```

src/Microsoft.DotNet.Interactive.Tests/Utility/KernelExtensionTestHelper.cs
```diff
@@ -101,7 +140,12 @@ public static ExtensionPackage GetOrCreateFileProviderExtension()
   </PropertyGroup>
 
   <ItemGroup>
-    <None Include=""$(OutputPath)/{packageName}.dll"" Pack=""true"" PackagePath=""interactive-extensions/dotnet"" />
+    {string.Join("\n", allPackageFiles.Select(item =>
```
jonsequitur, line 143: For your future angle bracket needs, you can also do XML with PocketView. 😂

jonsequitur (approved):

PR #1395, opened 2021-06-07
Stable api update

brettfo:


colombod (approved):

NotebookTestScript.dib
```diff
@@ -116,7 +116,9 @@ let x = 1 +
    "outputs": [
     {
      "data": {
-      "text/html": "<div class=\"dni-plaintext\">2</div>"
```
colombod, line 119: Why is there an output cell in a dib?

PR #1397, opened 2021-06-08
Use NameInCode to backtick-escape necessary F# completions when inserting

cartermp:
Before this change, some F# completions that require backticks would show up correctly in the completion list, but would insert as the display name rather than the "name in code".

After this change, the display remains the same, but when inserting the completion, the `NameInCode` value is used, which has backticks. This is what we do in the VS tools for F# as well (although it is a much more complex process... 🙂)


cartermp:
@jonsequitur @colombod pls review

colombod (approved):

colombod:
@cartermp looks good to me

PR #1425, opened 2021-06-15
add simple host kernel

brettfo:


src/Microsoft.DotNet.Interactive.VSCode/KernelExtensions.cs
```diff
@@ -0,0 +1,32 @@
...
+
+namespace Microsoft.DotNet.Interactive.VSCode
+{
+    public static class KernelExtensions
+    {
+        public static CSharpKernel UseVSCodeHelpers(this CSharpKernel kernel)
+        {
+            var command = new SubmitCode($@"
+#r ""{typeof(VSCodeInteractiveHost).Assembly.Location.Replace("\\", "/")}""
+{typeof(IInteractiveHost).FullName} host = new {typeof(VSCodeInteractiveHost).FullName}();
```
jonsequitur, line 16: I'm not sure we should have this `host` variable in the long term.

That said, `Host` would be a more conventional name.

src/Microsoft.DotNet.Interactive.VSCode/VSCodeInteractiveHost.cs
```diff
@@ -0,0 +1,45 @@
...
+        /// </summary>
+        /// <param name="prompt">The prompt to show.</param>
+        /// <param name="isPassword">Whether the input should be treated as a password.</param>
+        /// <returns>The value.</returns>
+        public Task<string> GetInputAsync(string prompt = "", bool isPassword = false, CancellationToken cancellationToken = default)
+        {
+            var command = new GetInput(prompt, isPassword, targetKernelName: "vscode");
+            var completionSource = new TaskCompletionSource<string>();
+            var token = command.GetToken();
+            Kernel.Root.KernelEvents.Where(e => e.Command.GetToken() == token).Subscribe(e =>
```
jonsequitur, line 26: We can typically reference the kernel directly (e.g. pass it into the constructor) rather than relying on the static method, which in turn relies on `AsyncLocal`.

src/dotnet-interactive-vscode/common/stdioKernelTransport.ts
```diff
@@ -160,45 +162,87 @@ export class StdioKernelTransport implements KernelTransport {
...
+            for (let i = this.eventSubscribers.length - 1; i >= 0; i--) {
+                this.eventSubscribers[i](envelope);
+            }
+        } else if (this.isKernelCommandEnvelope(envelope)) {
+            for (let i = this.commandSubscribers.length - 1; i >= 0; i--) {
+                // TODO: pass in context with shortcut methods for publish, etc.
+                // TODO: wrap and return succeed/failed
+                try {
+                    const handled = await this.commandSubscribers[i](envelope);
+                    if (handled) {
```
jonsequitur, line 193: This becomes dependent on the subscription order when there are multiple subscribers that might be added in a different order. We should enforce a exactly one command handler.

colombod: I think here we ensure a single subscriber for commands, that would actually be the kernel handling them , same job we do in the .NET World using a composite kernel. here we need a kernel with routing responsibility too.

src/dotnet-interactive/CommandLine/CommandLineParser.cs
```diff
@@ -395,15 +396,32 @@ Command StdIO()
...
+                                        case CSharpKernel ck:
+                                            ck.UseVSCodeHelpers();
+                                            break;
+                                        case FSharpKernel fk:
+                                            fk.UseVSCodeHelpers();
+                                            break;
+                                    }
+                                });
+                                // TODO: move this
+                                kernel.RegisterCommandType<GetInput>();
```
jonsequitur, line 421: The fact that we're doing this here means that for embedded kernels, these gestures won't work.

I wonder if a more declarative approach to command and event serialization registration would be easier at this point.

src/Microsoft.DotNet.Interactive.VSCode/VSCodeInteractiveHost.cs
brettfo: Added a commit where we directly pass in and use the kenel.

src/Microsoft.DotNet.Interactive.VSCode/KernelExtensions.cs
brettfo: Renamed to `InteractiveHost` and also made it work with PowerShell.

src/Microsoft.DotNet.Interactive.VSCode/VSCodeInteractiveHost.cs
brettfo: Done.

src/dotnet-interactive-vscode/common/stdioKernelTransport.ts
brettfo: Updated to only have a single command handler instead of a list.

brettfo:
@colombod I got it to work properly with `DotNetKernel.SetVariableAsync()` instead of the nasty code-gen which means we now only have one implementation that works with C#, F#, and PowerShell.

src/dotnet-interactive/CommandLine/CommandLineParser.cs
brettfo: Moved to a more appropriate location.

colombod (approved):

PR #1431, opened 2021-06-17
allow routing to specific remote kernels

brettfo:
This will shortly be re-written when the composite kernel knows more about proxy kernel subkernels and commands, but it at least adds a useful test.

### DIFF
--- a/src/Microsoft.DotNet.Interactive/Connection/ProxyKernel.cs
+++ b/src/Microsoft.DotNet.Interactive/Connection/ProxyKernel.cs
@@ -9,6 +9,7 @@ using System.Threading.Tasks;
 using Microsoft.DotNet.Interactive.Commands;
 using Microsoft.DotNet.Interactive.Events;
 using Microsoft.DotNet.Interactive.Parsing;
+using Microsoft.DotNet.Interactive.Utility;
 
 namespace Microsoft.DotNet.Interactive.Connection
 {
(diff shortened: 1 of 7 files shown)


colombod (approved):
I was there

src/Microsoft.DotNet.Interactive/Extensions/KernelUriExtensions.cs
```diff
@@ -0,0 +1,35 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Linq;
+
+namespace Microsoft.DotNet.Interactive.Extensions
```
jonsequitur, line 6: This namespace is intended to be about kernel extensions, not extension methods.

src/Microsoft.DotNet.Interactive/Connection/ProxyKernel.cs
```diff
@@ -74,8 +75,10 @@ internal override async Task HandleAsync(KernelCommand command, KernelInvocation
                     return;
             }
 
+            var kernelUri = KernelUri.Parse(command.TargetKernelName);
+            var remoteTargetKernelName = kernelUri.GetRemoteKernelName();
             var targetKernelName = command.TargetKernelName;
```
jonsequitur, line 80: Renaming this to `localTargetKernelName` might make it a bit more meaningful in context given these changes.

src/Microsoft.DotNet.Interactive/Extensions/KernelUriExtensions.cs
brettfo: Moved to `Microsoft.DotNet.Interactive.Utility` alongside other extension methods.

jonsequitur (approved):

PR #1451, opened 2021-06-28
add `JupyterInteractiveHost`

brettfo:
This adds a jupyter interactive host like we already have for vscode, which means from either environment and from C#, F#, or PowerShell you can do:

C#:

``` csharp
var value = await InteractiveHost.GetInputAsync();
```

F#:

``` fsharp
let value = InteractiveHost.GetInputAsync().Result
```

Pwsh:

``` pwsh
$value = Read-Host
```


colombod (approved):

src/Microsoft.DotNet.Interactive.Jupyter/Shell.cs
```diff
@@ -111,9 +111,18 @@ public Task StartAsync(CancellationToken cancellationToken)
                                 request,
                                 kernelIdentity);
 
-                            await _scheduler.Schedule(context);
-
-                            await context.Done();
+                            try
+                            {
+                                JupyterRequestContext.Current = context;
```
jonsequitur, line 116: This isn't concurrency safe. Given that this change exposes this as global state, this could become a problem.

src/Microsoft.DotNet.Interactive.Jupyter/JupyterInteractiveHost.cs
```diff
@@ -0,0 +1,44 @@
...
+
+namespace Microsoft.DotNet.Interactive.Jupyter
+{
+    public class JupyterInteractiveHost : IInteractiveHost
+    {
+        public JupyterInteractiveHost()
+        {
+        }
+
+        public Task AddCellAsync(string language, string contents, CancellationToken cancellationToken = default)
```
jonsequitur, line 17: A good test of the naming here is whether this works conceptually with a REPL as well. 

I think this should likely not be part of the interface but instead be just another command that kernels can choose to implement.

And then maybe there's a generalization that can work across notebooks and REPLs, `DisplayEditableSubmission` or something.

brettfo:
@colombod I had to re-add the type `PasswordString` because it was preventing clear-text passwords from showing up in the diagnostic log (because `new PasswordString(...).ToString()` shows up as `Microsoft.DotNet.Interactive.PasswordString` in the log instead of `hunter2`.

src/Microsoft.DotNet.Interactive.Jupyter/JupyterInteractiveHost.cs
brettfo: This is purely a feature of the front end, and not of the kernel.  Cells can be added in VS Code, but not in Jupyter.  Once @colombod and I have finished the various kernel refactorings, this probably _should_ move to the VS Code Kernel and GetInput should move to the VS Code and Jupyter kernels, but in the mean time we needed a place to dump the methods.

jonsequitur: In the Jupyter namespace, "cell" is a reasonable word. 

But `IInteractiveHost` (which is in `Microsoft.DotNet.Interactive`) has "cell"-related methods, which I think is the wrong level of abstraction. We should look for something that makes sense in interfaces other than notebooks. The command naming should align to that. It would be better to get the naming right sooner than later as other code will be depending on this soon.

PR #1461, opened 2021-07-01
update tool versions

colombod:
Use tool version  1.0.233002 to address nuget source bug


brettfo (changes requested):
This is correct for at least the F# packages, but we should wait to hear from @jonsequitur regarding the System.CommandLine and ExtensionLab packages.

brettfo (approved):

PR #1470, opened 2021-07-07
Print only new packages added

colombod:


KevinRansom:
They are indeed very invasive, I wonder if we can add some form diagnostic mode.

KevinRansom (approved):

src/Microsoft.DotNet.Interactive.NetStandard20/Microsoft.DotNet.Interactive.Netstandard20.csproj
```diff
@@ -11,6 +11,8 @@
 
   <ItemGroup>
     <Compile Include="$(MSBuildThisFileDirectory)../Microsoft.DotNet.Interactive/**/*.cs" />
+    <Compile Remove="$(MSBuildThisFileDirectory)../Microsoft.DotNet.Interactive/bin/**/*.cs" />
```
KevinRansom, line 14: Why is this necessary, shouldn't generated files be included?

colombod: @brettfo found that they get included twice, running a dotnet build is causing that issue

brettfo: In Diego's environment, the auto-generated files are under `bin/` and `obj/` and the netstandard2.0 project should certainly include its own generated files, but not those from the non-netstandard2.0 project.

brettfo (approved):

Issue #799, opened 2020-09-27
Console out renders a distinct block per call in VS Code


jhoneill:
### Describe the bug
I think this may be a generic re-statement of #730 
When PowerShell generates output it sends one or more items to the pipeline, and each one is rendered as a separate block and in the ipynb file each one becomes an object under `outputs` with      
```
"output_type": "execute_result",
"data": {
      "text/plain": 
```
This is different from .net interactive in Jupyter which outputs a single object with an `output type` of `stream`

The objects are then double spaced in the print out and, if there enough objects, VS Code hangs. 
Piping objects through `Out-String` produces a more desirable  behaviour spacing like this:
![image](https://user-images.githubusercontent.com/16209593/94362624-a0fd4c00-00b4-11eb-8e0b-7e3741d76732.png)

When there are **many objects**, using `| Out-string` successfully inserts a large, multi-line text block, where **vs-code hangs** (c.f #730), to repro try `dir c:\windows | out-string` and then re-run the command  removing `| out-String`  On my machine this has 119 files which is sufficient to cause the hang. 

- Version of .NET Interactive 
    - v1.0.146902
- OS
    - [x] Windows 10 -  10.0.19041.0
- Frontend
    - [x] Visual Studio Code 1.150.0

jonsequitur:
Related: https://github.com/microsoft/vscode/issues/107613

jonsequitur:
Related: https://github.com/microsoft/vscode/issues/103661

bmitc:
As detailed in #951 (and also #879 which contains a workaround), when using `printf` and `printfn` in an F# cell, newlines get entered before and after parameterized values. For example, if one runs:

```
printfn "Number = %.2f, String = %s" 2.123 "testing"
```

Then the result looks like this:
![image](https://user-images.githubusercontent.com/65685447/103321548-c7a68800-49ff-11eb-8e77-ae53121f32c3.png)

jessiewestlake:
Does anyone know where on the roadmap fixing this issue is?

Being able to run code and see it's output is the only useful aspect of using Interactive Notebooks. The output is virtually unreadable with this problem. I spent more time setting up the environment to use the .NET Interactive Notebooks than I spent running my first `Get-ChildItem` command and realizing the output is basically unusable.

.NET Interactive Notebooks is an amazing feature and something I could really use to get my team into using VS Code and teaching them PowerShell coding. Please continue working on this. Thank you!

jonsequitur:
[...]

Issue #1471, opened 2021-07-07
Ansi output not supported


DonJayamanne:
The following cell doesn't work in Notebooks:
```csharp
Console.Write("\x1b[38;2;0;255;0m"); // set text color to green
Console.WriteLine("green");
Console.Write("\x1b[0m"); // reset colors
```

Basically any ANSI codes are not supported (because all output is treated as as plain text).

jonsequitur:
![image](https://user-images.githubusercontent.com/547415/128542002-9724c844-f927-475a-ab6a-48684a99bdc1.png)

PR #1472, opened 2021-07-07
Add support for Ansi codes in outputs

DonJayamanne:
https://github.com/dotnet/interactive/issues/1471

* Fixes support for ansi output
* Also ensures the generated ipynb is identicatal (if not similar) to the one generated by jupyter
	* This is because we're using VS Codes stderr & stdout output items.

Screen Shot 2021-07-07 at 12 14 26

### DIFF
--- a/src/dotnet-interactive-vscode/common/interactiveClient.ts
+++ b/src/dotnet-interactive-vscode/common/interactiveClient.ts
@@ -127,8 +127,7 @@ export class InteractiveClient {
             };
 
             let reportOutputs = () => {
-                outputObserver([...outputs]);
-                outputs = [];
+                outputObserver(outputs);
             };
 
             let failureReported = false;
--- a/src/dotnet-interactive-vscode/common/interactiveClient.ts
+++ b/src/dotnet-interactive-vscode/common/interactiveClient.ts
@@ -49,7 +49,8 @@ import {
     SubmitCode,
     SubmitCodeType,
     CancelType,
-    Cancel
+    Cancel,
+    StandardErrorValueProduced
 } from './interfaces/contracts';
 import { Eol } from './interfaces';
 import { clearDebounce, createOutput } from './utilities';
@@ -126,7 +127,8 @@ export class InteractiveClient {
             };
 
             let reportOutputs = () => {
-                outputObserver(outputs);
+                outputObserver([...outputs]);
+                outputs = [];
             };
 
             let failureReported = false;
@@ -163,7 +165,8 @@ export class InteractiveClient {
                     case StandardOutputValueProducedType:
                         {
                             let disp = <DisplayEvent>eventEnvelope.event;
-                            let output = this.displayEventToCellOutput(disp);
+                            const stream = eventEnvelope.eventType === StandardErrorValueProducedType ? 'stderr' : 'stdout';
+                            let output = this.displayEventToCellOutput(disp, stream);
                             outputs.push(output);
                             reportOutputs();
                         }
@@ -390,16 +393,20 @@ export class InteractiveClient {
         }
     }
 
-    private displayEventToCellOutput(disp: DisplayEvent): vscodeLike.NotebookCellOutput {
+    private displayEventToCellOutput(disp: DisplayEvent, stream?: 'stdout' | 'stderr'): vscodeLike.NotebookCellOutput {
         const encoder = new TextEncoder();
         let outputItems: Array<vscodeLike.NotebookCellOutputItem> = [];
         if (disp.formattedValues && disp.formattedValues.length > 0) {
             for (let formatted of disp.formattedValues) {
                 let data = encoder.encode(formatted.value);
-                outputItems.push({
+                const outputItem:vscodeLike.NotebookCellOutputItem = {
                     mime: formatted.mimeType,
-                    data,
-                });
+                    data
+                }
+                if (stream){
+                    outputItem.stream = stream;
+                }
+                outputItems.push(outputItem);
             }
         }
 
--- a/src/dotnet-interactive-vscode/insiders/src/notebookControllers.ts
+++ b/src/dotnet-interactive-vscode/insiders/src/notebookControllers.ts
@@ -126,7 +126,9 @@ export class DotNetNotebookKernel {
                 const controllerErrors: vscodeLike.NotebookCellOutput[] = [];
 
                 function outputObserver(outputs: Array<vscodeLike.NotebookCellOutput>) {
-                    outputUpdatePromise = outputUpdatePromise.finally(() => updateCellOutputs(executionTask!, [...outputs])).catch(() => {});
+                    outputUpdatePromise = outputUpdatePromise.catch(ex => {
+                        console.error('Failed to update output', ex);
+                    }).finally(() => updateCellOutputs(executionTask!, [...outputs]));
                 }
                 const client = await this.config.clientMapper.getOrAddClient(cell.notebook.uri);
                 executionTask.token.onCancellationRequested(() => {
--- a/src/dotnet-interactive-vscode/insiders/src/notebookControllers.ts
+++ b/src/dotnet-interactive-vscode/insiders/src/notebookControllers.ts
@@ -126,7 +126,7 @@ export class DotNetNotebookKernel {
                 const controllerErrors: vscodeLike.NotebookCellOutput[] = [];
 
                 function outputObserver(outputs: Array<vscodeLike.NotebookCellOutput>) {
-                    outputUpdatePromise = outputUpdatePromise.finally(() => updateCellOutputs(executionTask!, [...outputs]));
+                    outputUpdatePromise = outputUpdatePromise.finally(() => updateCellOutputs(executionTask!, [...outputs])).catch(() => {});
                 }
                 const client = await this.config.clientMapper.getOrAddClient(cell.notebook.uri);
                 executionTask.token.onCancellationRequested(() => {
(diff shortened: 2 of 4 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/interactive?pullRequest=1472) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/interactive?pullRequest=1472) before we can accept your contribution.<br/><br/>:x: DonJayamanne [sign now](https://cla.dotnetfoundation.org/dotnet/interactive?pullRequest=1472)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/interactive?pullRequest=1472) it.</sub>

src/dotnet-interactive-vscode/common/interactiveClient.ts
```diff
@@ -126,7 +127,8 @@ export class InteractiveClient {
             };
 
             let reportOutputs = () => {
-                outputObserver(outputs);
+                outputObserver([...outputs]);
+                outputs = [];
```
DonJayamanne, line 131: Not sure we should re-broadcast everything.
Generally with observer pattern, I'd expect to see only the new stuff, not re-emit of everything all over agian (this makes the changes I introduced easier)

src/dotnet-interactive-vscode/common/interactiveClient.ts
```diff
@@ -163,7 +165,8 @@ export class InteractiveClient {
                     case StandardOutputValueProducedType:
                         {
                             let disp = <DisplayEvent>eventEnvelope.event;
-                            let output = this.displayEventToCellOutput(disp);
+                            const stream = eventEnvelope.eventType === StandardErrorValueProducedType ? 'stderr' : 'stdout';
+                            let output = this.displayEventToCellOutput(disp, stream);
```
DonJayamanne, line 169: We need this so we can tell VS Code whether its stdout or stderr
Also required when saving ipynb

src/dotnet-interactive-vscode/insiders/src/notebookControllers.ts
```diff
@@ -118,28 +118,23 @@ export class DotNetNotebookKernel {
...
+            let outputUpdatePromise = Promise.resolve();
             try {
                 const startTime = Date.now();
                 executionTask.start(startTime);
                 executionTask.clearOutput(cell);
                 const controllerErrors: vscodeLike.NotebookCellOutput[] = [];
 
                 function outputObserver(outputs: Array<vscodeLike.NotebookCellOutput>) {
-                    updateCellOutputs(executionTask!, [...outputs, ...controllerErrors]).then(() => { });
+                    outputUpdatePromise = outputUpdatePromise.finally(() => updateCellOutputs(executionTask!, [...outputs]));
```
DonJayamanne, line 129: We must wait on previous updates, so that we can concatenate the output items

src/dotnet-interactive-vscode/insiders/src/notebookControllers.ts
```diff
@@ -204,8 +202,21 @@ export async function updateCellLanguages(document: vscode.NotebookDocument): Pr
 }
 
 async function updateCellOutputs(executionTask: vscode.NotebookCellExecution, outputs: Array<vscodeLike.NotebookCellOutput>): Promise<void> {
-    const reshapedOutputs = outputs.map(o => new vscode.NotebookCellOutput(o.items.map(oi => generateVsCodeNotebookCellOutputItem(oi.data, oi.mime))));
-    await executionTask.replaceOutput(reshapedOutputs);
```
DonJayamanne, line 208: Removed replacing, as appending is more efficient

src/dotnet-interactive-vscode/insiders/src/notebookControllers.ts
```diff
@@ -204,8 +202,21 @@ export async function updateCellLanguages(document: vscode.NotebookDocument): Pr
...
+    const previousOutput = executionTask.cell.outputs.length ? executionTask.cell.outputs[executionTask.cell.outputs.length - 1] : undefined;
+    const previousOutputItem = previousOutput?.items.length ? previousOutput.items[previousOutput.items.length - 1] : undefined;
+    await Promise.all(outputs.map(async (o, index) => {
+        const items = o.items.map(oi => generateVsCodeNotebookCellOutputItem(oi.data, oi.mime, oi.stream));
+        // If all of these items are of the same stream type & previous item is the same stream, then append it.
+        if (index === 0 && previousOutput && previousOutputItem?.mime && ['application/vnd.code.notebook.stderr', 'application/vnd.code.notebook.stdout'].includes(previousOutputItem?.mime)){
```
DonJayamanne, line 210: If we get output from stdout or std err, then concatenate them (so that we have a single entry in generated output file)

DonJayamanne: These mime types are documented in VS Code, hence safe to hardcode

src/dotnet-interactive-vscode/common/interactiveClient.ts
brettfo: We support the concept of updating/replacing outputs, so we need to broadcast everything, otherwise 1 display output then 4 updates would be incorrectly rendered as 5 output items.

src/dotnet-interactive-vscode/insiders/src/notebookControllers.ts
```diff
@@ -149,14 +144,17 @@ export class DotNetNotebookKernel {
                     diagnosticCollection.set(cell.document.uri, diags.filter(d => d.severity !== contracts.DiagnosticSeverity.Hidden).map(vscodeUtilities.toVsCodeDiagnostic));
                 }
 
-                return client.execute(source, getSimpleLanguage(cell.document.languageId), outputObserver, diagnosticObserver, { id: cell.document.uri.toString() }).then(() => {
+                return client.execute(source, getSimpleLanguage(cell.document.languageId), outputObserver, diagnosticObserver, { id: cell.document.uri.toString() }).then(async () => {
+                    await outputUpdatePromise;
```
brettfo, line 148: This `await` and the one a few lines below should be wrapped in a `try`/`catch` to ensure we always call `endExecution()` so the notebook doesn't look like it hung.

brettfo:
Could you add a scenario (probably the green console output?) to this notebook and include it in this PR?

https://github.com/dotnet/interactive/blob/main/NotebookTestScript.dib

This is our pre-release smoke-test and I'd like to see the console out covered.

src/dotnet-interactive-vscode/insiders/src/notebookControllers.ts
```diff
@@ -118,28 +118,25 @@ export class DotNetNotebookKernel {
...
+            let outputUpdatePromise = Promise.resolve();
             try {
                 const startTime = Date.now();
                 executionTask.start(startTime);
                 executionTask.clearOutput(cell);
                 const controllerErrors: vscodeLike.NotebookCellOutput[] = [];
 
                 function outputObserver(outputs: Array<vscodeLike.NotebookCellOutput>) {
-                    updateCellOutputs(executionTask!, [...outputs, ...controllerErrors]).then(() => { });
+                    outputUpdatePromise = outputUpdatePromise.catch(ex => {
```
DonJayamanne, line 129: Capture & log all errors,
this way we won't need a try..catch..

src/dotnet-interactive-vscode/common/interactiveClient.ts
DonJayamanne: For my benefit can you give me an example of where that would be necessary.
because The following didn't make sense
>  otherwise 1 display output then 4 updates would be incorrectly rendered as 5 output items.

DonJayamanne: @brettfo @jonsequitur 
I personally don't think we should create outputs, instead we should be returning output items.
At least for streams.
I.e if you have `Console.Write(1)`, and `Console.Write(2)` one after the other, 
if we have a single output & in that we have two output items, hen VS Code will render them as `12`.
However if its two sepate outputs each with their own otuput items, then it will be rendered as `1` <New Line> `2`
VSCode has API to append outputs as well.

The current PR tries to work around this by appending them if the previous output has a stream outputitem of the same type as the current one (i.e. concatenate them).

But that's a bigger change, & I didn't want to introduce that here.
I.e. VS Code will render `1` & `2` in the same line if they belong to the same output object.
Currently its all in seprate outputs

DonJayamanne: FYI - this PR does ensure `Console.Write(1)` & `Console.Write(2)` will be rendered as `12`

jonsequitur: This will fix #799.

src/dotnet-interactive-vscode/insiders/src/notebookControllers.ts
brettfo, line 129: I don't see `outputUpdatePromise` ever being assigned any work other than the pre-resolved promise which means the `catch()` block will never get hit and it really reduces to just the `finally()` call; let's collapse this to be similar to the original behavor, but with this new `catch()` block attached, e.g.,

``` typescript
outputUpdatePromise = updateCellOutputs(...).catch(ex => {
    console.error(...);
});
```

(or whatever makes the most sense)

jonsequitur:
@DonJayamanne The test that's been failing here is fixed in `main`. If you rebase, we should be able to merge this.

src/dotnet-interactive-vscode/insiders/src/notebookControllers.ts
DonJayamanne: There's a little bit more to the way its used. We update the promise everytime, so as to chain the promises. basically the subsequent update can only be done after the previous update has been completed (this is achieved by chaining the promises).

DonJayamanne:
Oh sorry, totally forgot about this PR.
Will work on this today.

colombod:
This is merged via #1514

PR #1485, opened 2021-07-09
WIP on using shared variable to parameterize SQL query

jonsequitur:


colombod:
We can close this one,no?

Issue #1456, opened 2021-06-29
Cells that get stuck running can sometimes not be stopped.


Treit:
### Describe the bug
Cells that get stuck running can sometimes not be stopped by clicking the 'Stop cell' button.

I have seen multiple times a cell that gets stuck in a running state and clicking the little square 'stop cell' button does nothing.

I recently saw this with the following trivial notebook:

    #!csharp

    int i = 1234;

    #!javascript

    let x = 0
    interactive.csharp.getVariable('i').then(v => console.log(v))
    console.log(x)

    #!markdown

    <h1>Test!</h1>
    Test test test

For some reason this showed several cells that were 'stuck' running and clicking the stop button did nothing.

## .NET Interactive Version
Version: 1.0.230903+1a272ef3782d6e88b6d36ebd648fb1882da0a530
Build date: 2021-06-28T20:51:38.7685995Z

- OS
    - [x] Windows 11 (build 22000.51)
    - [ ] Windows 10
    - [ ] macOS
    - [ ] Linux (Please specify distro)
    - [ ] iOS
    - [ ] Android
 - Browser 
    - [ ] Chrome
    - [x] Edge
    - [ ] Firefox
    - [ ] Safari
 - Frontend
    - [ ] Jupyter Notebook
    - [ ] Jupyter Lab
    - [ ] nteract
    - [x] Visual Studio Code 
    - [ ] Visual Studio Code Insiders
    - [ ] Other (please specify)

### Screenshots
This animation shows the issue:

![vscode2](https://user-images.githubusercontent.com/8271812/123876941-1b335900-d8f1-11eb-991c-3eb2ce9454a6.gif)

RandomFractals:
I hit this issue too after creating simple data pull C# cell in a .Net interactive .ipynb notebook.

It worked the first time I created it. Restarting and running that notebook in both latest stable and insiders either does nothing or gets still in infinite cell run loop. 

You can try my test from this ticket: https://github.com/RandomFractals/vscode-data-table/issues/26#issuecomment-875798847

jonsequitur:
Is the code snippet in the initial comment relevant to this bug? That looks the same as the code presented in #1452 (which I can repro) but the screen shot seems to be showing something different (which I cannot repro). 

I'm using version v1.0.2351010 of the extension.

RandomFractals:
[...]

PR #1489, opened 2021-07-12
defer setting host variable

brettfo:
The setting of the `InteractiveHost` variable should be deferred to when the kernel is actually used.  Eventually `SetVariableAsync` will turn into a proper command that can be deferred through the normal mechanism, but as a stop-gap, we can wrap it in an `AnonymousKernelCommand` and defer that.

We had two internal-only implementations of  `AnonymousKernelCommand` so I collapsed them into one.


brettfo:
@RandomFractals I think this will help with the PowerShell kernel issues you've been seeing in #1456.  Can you give this a shot on your machine to see if it fixes the issue?

To test locally you'll need to:

1. Pull down these changes to your machine:
   ```
   git clone https://github.com/brettfo/interactive
   ```
2. Build:
   ```
   git checkout defer-set-variable
   build.cmd
   ```
3. Launch VS Code and type `Ctrl+,` (control comma) to open the settings dialog.
4. Type `dotnet-interactive` to filter the settings.
5. Locate the setting called "Dotnet-interactive: Kernel Transport Args".
6. Click "Edit in settings.json" and set its value to:
``` json
    "dotnet-interactive.kernelTransportArgs": [
        "{dotnet_path}",
        "D:/brettfo/interactive/artifacts/bin/dotnet-interactive/Debug/net5.0/Microsoft.DotNet.Interactive.App.dll",
        "[vscode]",
        "stdio",
        "--working-dir",
        "{working_dir}"
    ],
```
7. Save the settings file and restart VS Code for good measure.

At this point re-run your scenario and see if it no longer gets stuck when launching the PowerShell kernel.

To use the bundled version of the interactive tool again, simply delete the `"dotnet-interactive.kernelTransportArgs"` entry in your settings.json and restart VS Code.

src/Microsoft.DotNet.Interactive.Jupyter/KernelExtensions.cs
```diff
@@ -22,10 +22,14 @@ namespace Microsoft.DotNet.Interactive.Jupyter
...
     public static class KernelExtensions
     {
-        public static Task UseJupyterHelpersAsync<TKernel>(this TKernel kernel) where TKernel : DotNetKernel
+        public static void UseJupyterHelpers<TKernel>(this TKernel kernel) where TKernel : DotNetKernel
         {
             var interactiveHost = new JupyterInteractiveHost();
-            return kernel.SetVariableAsync("InteractiveHost", interactiveHost, typeof(IInteractiveHost));
+            var command = new AnonymousKernelCommand((_command, context) =>
+            {
+                return kernel.SetVariableAsync("InteractiveHost", interactiveHost, typeof(IInteractiveHost));
```
colombod, line 30: does this address the completion too? we can do the using statements too here if we defer those commands

src/Microsoft.DotNet.Interactive/Commands/AnonymousKernelCommand.cs
```diff
@@ -3,7 +3,7 @@
 
 namespace Microsoft.DotNet.Interactive.Commands
 {
-    internal class AnonymousKernelCommand : KernelCommand
+    public class AnonymousKernelCommand : KernelCommand
```
jonsequitur, line 6: We shouldn't make this class available as part of our public interface as it can't be serialized.

jonsequitur:
@brettfo Please rebase this so we can get it merged for our next release.

src/Microsoft.DotNet.Interactive.Jupyter/KernelExtensions.cs
jonsequitur: Instead of an anonymous command, you can now do `KernelInvocationContext.ScheduleAsync`.

PR #1503, opened 2021-07-22
add kusto grammar to vs code extension

brettfo:
Kusto language grammar adapted from [here](https://github.com/madskristensen/TextmateBundleInstaller/blob/master/src/Bundles/kusto/Syntaxes/kusto.plist).

Draft because there's not yet a backing kernel, but once there is, this'll be ready.

image


colombod:
should we call it and match over kql instead?

colombod (approved):

PR #1525, opened 2021-08-02
update stable vscode target to 1.59

brettfo:
This is our standard migration of `insiders` code to `stable` for the VS Code extension.  This time we've also added documentation for what we're doing.


colombod (approved):

brettfo:
GitHub doesn't seem to be reporting the status of the checks, but it's all green [here](https://dev.azure.com/dnceng/public/_build/results?buildId=1272150&view=results).

Issue #1518, opened 2021-07-30
nuget command displays no output for requests to load a package that is already loaded


KevinRansom:
This PR reduces the nuget displayed message: https://github.com/dotnet/interactive/pull/1510

However, it does not display output for requests that have already been loaded, producing a display like this:
![image](https://user-images.githubusercontent.com/5175830/127709632-34dc1c1a-1e7b-49ba-a450-49aa664ae4b4.png)

Note that the second cell requests:
````
#r "nuget: FSharp.Data"
#r "nuget: FSharp.Compiler.Service"
````

The output shows only:
![image](https://user-images.githubusercontent.com/5175830/127709879-9a2bccbe-5978-4ac7-89bf-c270c9c2a040.png)

This can be confusing to the user, did it or did it not load FSharp.Compiler.Service, why can't I see it? What did I do wrong?

In the image above it is clear what has happened but what about this one:
![image](https://user-images.githubusercontent.com/5175830/127710309-dbe01b62-215e-479f-b6db-9d624837ed98.png)

I doubt anyone would be explain why google.protobuf isn't showed as loaded, without inferring that it was a dependency of OnnxTransformer and so had already been loaded.

I propose that we change the display to always show the set of requested packages for the submission that caused the Restore.

KevinRansom:
Okay so I have changes that produce this:

![image](https://user-images.githubusercontent.com/5175830/128072350-bc70fe29-75a0-42be-b33c-1c467fdf9b83.png)


![image](https://user-images.githubusercontent.com/5175830/128072650-ddf46d31-4788-4e6a-8f58-ab42f05539a5.png)

colombod:
I think this is a good user experience improvement, is there a PR open?

brettfo:
Fixed in #1529.

PR #1529, opened 2021-08-04
#r nuget output

KevinRansom:
Fixes issue: https://github.com/dotnet/interactive/issues/1518

Now produces Installing: and Installed for #r "nuget ...." when it requests an already requested package.

image

And ...
image

### DIFF
--- a/src/Microsoft.DotNet.Interactive.PackageManagement/PackageRestoreContext.cs
+++ b/src/Microsoft.DotNet.Interactive.PackageManagement/PackageRestoreContext.cs
@@ -91,6 +91,11 @@ namespace Microsoft.DotNet.Interactive
                     {
                         resolvedPackage = requested;
                     }
+                    else
+                    {
+                        // Package already loaded with a different version
+                        return null;
+                    }
                 }
             }
 
--- a/src/Microsoft.DotNet.Interactive.PackageManagement/PackageRestoreContext.cs
+++ b/src/Microsoft.DotNet.Interactive.PackageManagement/PackageRestoreContext.cs
@@ -103,7 +103,7 @@ namespace Microsoft.DotNet.Interactive
 
             PackageReference AppendPathSeparator(PackageReference resolved)
             {
-                if (resolved == null)
+                if (resolved is null)
                 {
                     var newPackageRef = new PackageReference(packageName, packageVersion);
                     _requestedPackageReferences.TryAdd(key, newPackageRef);
--- a/src/Microsoft.DotNet.Interactive.PackageManagement/PackageRestoreContext.cs
+++ b/src/Microsoft.DotNet.Interactive.PackageManagement/PackageRestoreContext.cs
@@ -67,37 +67,50 @@ namespace Microsoft.DotNet.Interactive
             // Package names are case insensitive.
             var key = packageName.ToLower(CultureInfo.InvariantCulture);
 
-            if (_resolvedPackageReferences.TryGetValue(key, out var resolvedPackage))
+            PackageReference resolvedPackage = null;
+            if (_resolvedPackageReferences.TryGetValue(key, out var resolved))
             {
                 if (string.IsNullOrWhiteSpace(packageVersion) ||
                     packageVersion == "*" ||
-                    string.Equals(resolvedPackage.PackageVersion.Trim(), packageVersion.Trim(), StringComparison.OrdinalIgnoreCase))
+                    packageVersion == "*-*" ||
+                    string.Equals(resolved.PackageVersion.Trim(), packageVersion.Trim(), StringComparison.OrdinalIgnoreCase))
                 {
-                    return resolvedPackage;
+                    resolvedPackage = resolved;
                 }
                 else
                 {
-                    // It was previously resolved at a different version than the one requested
+                    // Package already loaded with a different version
                     return null;
                 }
             }
-
-            if (_requestedPackageReferences.TryGetValue(key, out PackageReference existingPackage))
+            else
             {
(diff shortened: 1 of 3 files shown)


src/Microsoft.DotNet.Interactive.Tests/LanguageKernelPackageTests.cs
```diff
@@ -1148,5 +1198,121 @@ public async Task Pound_r_nuget_should_display_only_requested_packages(Language
...
+            var result2 = await kernel.SendAsync(command2);
+
+            var codeThirdSubmission = @"
+#r ""nuget:Google.Protobuf, 3.5.1""
+#r ""nuget:Microsoft.ML.OnnxTransformer, 1.4.0""
+";
+            var command3 = new SubmitCode(codeThirdSubmission);
+            var result3 = await kernel.SendAsync(command3);
+
+            using var events = result3.KernelEvents.ToSubscribedList();
```
colombod, line 1228: I would subscribe to the event before submitting the command that is to be tested

src/Microsoft.DotNet.Interactive.Tests/LanguageKernelPackageTests.cs
```diff
@@ -1148,5 +1198,121 @@ public async Task Pound_r_nuget_should_display_only_requested_packages(Language
...
+            var command1 = new SubmitCode(codeFirstSubmission);
+            var _result1 = await kernel.SendAsync(command1);
+
+            var codeSecondSubmission = @"
+#r ""nuget:Google.Protobuf, 3.5.0""
+";
+            var command2 = new SubmitCode(codeSecondSubmission);
+            var result2 = await kernel.SendAsync(command2);
+
+            using var events = result2.KernelEvents.ToSubscribedList();
```
colombod, line 1281: I would subscribe to the event before submitting the command that is to be tested

src/Microsoft.DotNet.Interactive.Tests/LanguageKernelPackageTests.cs
```diff
@@ -1101,40 +1101,90 @@ public async Task Pound_r_nuget_works_immediately_after_a_language_selector(Lang
...
+#r ""nuget:Microsoft.ML.OnnxTransformer,1.4.0""
+";
+            var command1 = new SubmitCode(codeFirstSubmission);
+            var _result1 = await kernel.SendAsync(command1);
+
+            var codeSecondSubmission = @"
+#r ""nuget:Google.Protobuf, 3.5.1""
+";
+            var command2 = new SubmitCode(codeSecondSubmission);
+            var result2 = await kernel.SendAsync(command2);
```
colombod, line 1166: I would subscribe to the event before submitting the command that is to be tested

colombod (changes requested):

src/Microsoft.DotNet.Interactive.PackageManagement/PackageRestoreContext.cs
```diff
@@ -67,37 +67,50 @@ public void TryAddRestoreSource(string source)
...
+                }
+            }
+
+            return AppendPathSeparator(resolvedPackage);
 
-            if (_requestedPackageReferences.TryGetValue(key, out PackageReference existingPackage))
+            PackageReference AppendPathSeparator(PackageReference resolved)
             {
-                if (string.Equals(existingPackage.PackageVersion.Trim(), packageVersion.Trim(), StringComparison.OrdinalIgnoreCase))
+                if (resolved == null)
```
jonsequitur, line 101:
```suggestion
                if (resolved is null)
```

src/Microsoft.DotNet.Interactive.Tests/LanguageKernelPackageTests.cs
```diff
@@ -1101,40 +1101,90 @@ public async Task Pound_r_nuget_works_immediately_after_a_language_selector(Lang
...
+            };
+
+            events.OfType<DisplayedValueUpdated>()
+                  .Where(v => v.Value is InstallPackagesMessage)
+                  .Last().Value
+                  .As<InstallPackagesMessage>()
+                  .InstalledPackages
+                  .Aggregate((s, acc) => acc + " & " + s)
+                  .Should()
+                  .ContainAll(expectedDisplayed)
```
jonsequitur, line 1144: You can use `BeEquivalentTo` without the aggregation, which will be more concise and provide better error output. This will also avoid having to specify the packages you don't expect to see.

src/Microsoft.DotNet.Interactive.Tests/LanguageKernelPackageTests.cs
KevinRansom: updated

src/Microsoft.DotNet.Interactive.Tests/LanguageKernelPackageTests.cs
KevinRansom: updated

src/Microsoft.DotNet.Interactive.Tests/LanguageKernelPackageTests.cs
KevinRansom: It's the third submission I want events for, and that is updated, per your comment.

src/Microsoft.DotNet.Interactive.PackageManagement/PackageRestoreContext.cs
KevinRansom: Doh .... yes.

src/Microsoft.DotNet.Interactive.Tests/LanguageKernelPackageTests.cs
KevinRansom: Updated

colombod (approved):

PR #1531, opened 2021-08-04
add better detection for bad symlinks

brettfo:
Our build script has been updated to ensure all necessary symlinks are present and text has been added to the top of our dev guides indicating what needs to be done and the workaround.

### DIFF
--- a/src/dotnet-interactive-vscode/DEVGUIDE.md
+++ b/src/dotnet-interactive-vscode/DEVGUIDE.md
@@ -1,6 +1,8 @@
 The organization of this extension is a bit complicated.
 ===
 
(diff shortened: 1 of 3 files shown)


src/dotnet-interactive-vscode/DEVGUIDE.md
```diff
@@ -1,6 +1,8 @@
 The organization of this extension is a bit complicated.
 ===
 
+This repo depends on symbolic links between directories.  By default Windows doesn't support this feature.  To work around this scenario, please run the PowerShell script `<root>/src/ensure-symlinks.ps1` as an administrator.  This usually only needs to be run once.
+
```
jonsequitur, line 5: Is it useful to have two different developer guides?

colombod (approved):

src/dotnet-interactive-vscode/DEVGUIDE.md
brettfo: The root dev guide contains most of this info.  I'll collapse them both into the same one.

PR #1555, opened 2021-08-18
Do not run cells in untrusted workspace

colombod:
image

### DIFF
--- a/src/dotnet-interactive-vscode/insiders/package.json
+++ b/src/dotnet-interactive-vscode/insiders/package.json
@@ -55,7 +55,8 @@
   ],
   "capabilities": {
     "untrustedWorkspaces": {
-      "supported": true
+      "supported": "limited",
+      "description": "Execution of cells in Interactive Window and Notebooks is not supported in untrusted workspaces."
     },
     "virtualWorkspaces": true
   },
--- a/src/dotnet-interactive-vscode/stable/package.json
+++ b/src/dotnet-interactive-vscode/stable/package.json
@@ -55,7 +55,8 @@
   ],
   "capabilities": {
     "untrustedWorkspaces": {
-      "supported": true
+      "supported": "limited",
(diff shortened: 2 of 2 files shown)


src/dotnet-interactive-vscode/insiders/package.json
```diff
@@ -55,7 +55,8 @@
   ],
   "capabilities": {
     "untrustedWorkspaces": {
-      "supported": true
+      "supported": "limited",
+      "description": "Execution of cells in Interactive Window and Notebooks is not supported in untrusted workspaces."
```
brettfo, line 59: Same change is needed in stable `package.json`, unless untrusted workspaces isn't a thing in stable.

src/dotnet-interactive-vscode/insiders/src/notebookControllers.ts
```diff
@@ -146,8 +146,10 @@ export class DotNetNotebookKernel {
     }
 
     private async executeHandler(cells: vscode.NotebookCell[], document: vscode.NotebookDocument, controller: vscode.NotebookController): Promise<void> {
-        for (const cell of cells) {
-            await this.executeCell(cell, controller);
+        if (vscode.workspace.isTrusted) {
+            for (const cell of cells) {
+                await this.executeCell(cell, controller);
+            }
```
brettfo, line 152: Might be good to show an information dialog that execution won't happen because it's untrusted, otherwise a user won't have any idea why nothing happens.  Would be good to coordinate with Python on the experience.

colombod: if you hit the run button in vscode you get the dialog you see in the screenshot above

colombod: removed

src/dotnet-interactive-vscode/insiders/package.json
colombod: Done

brettfo (approved):

PR #1556, opened 2021-08-20
Proxy Kernel should not split code submissions

colombod:
This is to address the problem that causes the proxykernel to split a single SubmitCommand into multiple ones. This fixes all JQuery


brettfo (approved):

src/Microsoft.DotNet.Interactive.Tests/Parsing/SubmissionParserTests.cs
```diff
@@ -366,6 +367,69 @@ public void Submissions_targeting_proxy_kernels_are_not_split_prior_to_sending()
...
+            var _ = proxyKernel.RunAsync();
+            var code = @"#i ""nuget:source1""
+#i ""nuget:source2""
+#r ""nuget:package1""
+#r ""nuget:package2""
+
+Console.WriteLine(1);";
+             await proxyKernel.SendAsync(new SubmitCode(code, proxyKernel.Name));
+
+            commandAndEventSender.Commands.Count.Should().Be(1);
```
jonsequitur, line 398: Do we want to assert that the command is in fact the original command?

src/Microsoft.DotNet.Interactive.Tests/Parsing/SubmissionParserTests.cs
```diff
@@ -366,6 +367,69 @@ public void Submissions_targeting_proxy_kernels_are_not_split_prior_to_sending()
                 .Be(code);
         }
 
+        [Fact]
+        public async Task proxyKernel_does_not_perform_split_if_all_parts_go_to_same_targetKernel_as_the_original_command()
+        {
+            
+            var commandAndEventReceiver = new BlockingCommandAndEventReceiver();
+            var commandAndEventSender = new RecordingKernelCommandAndEventSender();
```
jonsequitur, line 375: Why this new testing approach?

src/Microsoft.DotNet.Interactive/Connection/ProxyKernel.cs
```diff
@@ -91,6 +91,8 @@ internal override async Task HandleAsync(KernelCommand command, KernelInvocation
             }
 
             _executionContext = ExecutionContext.Capture();
+            command.TargetKernelName ??= Name;
```
jonsequitur, line 94: Why is this needed?

src/Microsoft.DotNet.Interactive.Tests/Parsing/SubmissionParserTests.cs
colombod: indeed!

src/Microsoft.DotNet.Interactive.Tests/Parsing/SubmissionParserTests.cs
colombod: moved to routing

PR #1558, opened 2021-08-23
publish CommandSucceeded and CommandFailed 

colombod:
commands submitted using SendAsync explicitly will produce completion events


src/Microsoft.DotNet.Interactive/Commands/KernelCommand.cs
```diff
@@ -25,6 +25,9 @@ protected KernelCommand(string targetKernelName = null, KernelCommand parent = n
         [JsonIgnore]
         public KernelCommand Parent { get; internal set; }
 
+        [JsonIgnore]
+        internal bool ShouldPublishCompletionEvents { get; set; }
```
jonsequitur, line 29:
```suggestion
        internal bool ShouldPublishCompletionEvent { get; set; }
```

src/Microsoft.DotNet.Interactive.Tests/LanguageKernelTests.cs
```diff
@@ -651,20 +651,25 @@ public async Task when_a_sequence_of_diagnostics_requests_is_fired_diagnostics_a
...
+            await  Task.WhenAll(
+                kernel.SendAsync(requestCompletionsCommand),
                 kernel.SendAsync(requestDiagnosticsCommand)
             );
 
-            var events = results.SelectMany(r => r.KernelEvents.ToSubscribedList()).ToList();
+            using var _ = new AssertionScope();
+            
+            events.Should()
+                .ContainSingle<CommandSucceeded>(e => e.Command == requestDiagnosticsCommand);
```
jonsequitur, line 669: Is this a change in behavior?

colombod: ops sorry .. forgot that change in

PR #1589, opened 2021-09-10
fix issue with multiple addition of markdown 

colombod:


brettfo (approved):

src/Microsoft.DotNet.Interactive.Documents/CodeSubmission.cs
```diff
@@ -58,7 +58,13 @@ void AddElement()
             }
 
             var splittingKernelNames = kernelNames.ToList();
-            splittingKernelNames.Add(new KernelName("markdown", new[] { "md" })); // not a kernel language, but still a valid cell splitter
+
```
jonsequitur, line 61: We're going to want to enforce uniqueness including aliases here  (and probably use a `HashSet<string>`) and change the naming so we're clear about "cell dividers" and not "kernel names".

PR #1603, opened 2021-09-16
fix issue with casting to ISupportNugetPackage

colombod:
When performing #!connect is kql and sql connection assume that the command is handled in a kernel that supports nuget packages.

### DIFF
--- a/src/Microsoft.DotNet.Interactive.SqlServer/SqlToolsServiceDiscovery.cs
+++ b/src/Microsoft.DotNet.Interactive.SqlServer/SqlToolsServiceDiscovery.cs
@@ -13,23 +13,11 @@ namespace Microsoft.DotNet.Interactive.SqlServer
         public static ResolvedPackageReference FindResolvedPackageReference(this Kernel rootKernel)
         {
             var runtimePackageIdSuffix = "native.Microsoft.SqlToolsService";
-            foreach (var kernel in rootKernel.SubkernelsAndSelf().OfType<ISupportNuget>())
-            {
-                var resolvedPackageReferences = kernel.ResolvedPackageReferences;
-
-                // Walk through the packages looking for the package that ends with the name "Microsoft.SqlToolsService"
-                // and grab the packageroot
-
-                var resolved = resolvedPackageReferences.FirstOrDefault(p => p.PackageName.EndsWith(runtimePackageIdSuffix, StringComparison.OrdinalIgnoreCase));
-
-                if (resolved is { })
-                {
-                    return resolved;
-                }
-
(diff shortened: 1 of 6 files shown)


colombod:
cc @JustinMDotNet

src/Microsoft.DotNet.Interactive.SqlServer/SqlToolsServiceDiscovery.cs
```diff
@@ -0,0 +1,65 @@
...
+using System.Runtime.InteropServices;
+
+namespace Microsoft.DotNet.Interactive.SqlServer
+{
+    internal static class SqlToolsServiceDiscovery
+    {
+        public static ResolvedPackageReference FindResolvedPackageReference(this Kernel rootKernel)
+        {
+            var runtimePackageIdSuffix = "native.Microsoft.SqlToolsService";
+            foreach (var kernel in rootKernel.SubkernelsAndSelf().OfType<ISupportNuget>())
```
jonsequitur, line 16: This whole block can be a `return ... SingleOrDefault`

src/Microsoft.DotNet.Interactive/KernelExtensions.cs
```diff
@@ -424,5 +424,22 @@ public static T LogEventsToPocketLogger<T>(this T kernel)
 
             await VisitSubkernelsAsync(kernel, onVisit, recursive);
         }
+
+        public static IEnumerable<Kernel> SubkernelsAndSelf(
```
jonsequitur, line 428: I'd prefer to avoid the duplicate logic here. The `Visit*` methods can probably just call this now, no?

src/Microsoft.DotNet.Interactive.SqlServer/SqlToolsServiceDiscovery.cs
colombod: ok

colombod: Fsharp could load the package too, We need to make sure we find the one that has the package

jonsequitur: `FirstOrDefault` then?

src/Microsoft.DotNet.Interactive/KernelExtensions.cs
colombod: makes sense

jonsequitur (approved):

PR #1604, opened 2021-09-17
add notebook parser server command

brettfo:
This adds a new command `dotnet-interactive notebook-parser` which only exists to (de)serialize notebooks.

The old method had special `ParseNotebook`/`SerializeNotebook` commands on `CompositeKernel`, but that required nearly 5s of startup time before we could start processing.  With the latest changes I'm seeing slightly under 300ms startup time.

This also includes a change to how we generate the TypeScript interfaces; enums are now done by string name instead of numerical value.  E.g., this

``` typescript
export enum SubmissionType {
    Run = 0,
    Diagnose = 1,
}
```

became this

``` typescript
export enum SubmissionType {
    Run = "run",
    Diagnose = "diagnose",
}
```


src/Microsoft.DotNet.Interactive.Documents/InteractiveDocument.cs
```diff
@@ -7,11 +7,11 @@ namespace Microsoft.DotNet.Interactive.Documents
 {
     public partial class InteractiveDocument
     {
-        public InteractiveDocumentElement[] Elements { get; }
+        public InteractiveDocumentElement[] Elements { get; set; }
```
jonsequitur, line 10: Let's make this a `List<InteractiveDocumentElement>` and remove the setter. The property shouldn't be allowed to be null.

src/Microsoft.DotNet.Interactive.Tests/Server/SerializationTests.cs
```diff
@@ -147,6 +148,103 @@ public void All_event_types_are_tested_for_round_trip_serialization()
...
+            serializeRequest.NewLine.Should().Be("\r\n");
+            serializeRequest.Document.Should().BeEquivalentTo(new InteractiveDocument(new[]
+            {
+                new InteractiveDocumentElement("csharp", "var x = 1;"),
+                new InteractiveDocumentElement("fsharp", "let y = 2"),
+            }));
+        }
+
+        [Fact]
+        public void Notebook_parse_response_can_be_serialized()
```
jonsequitur, line 195: Assent tests are best used for avoiding accidental breaking changes to the serialization format. I'd name the test accordingly because "can be serialized" doesn't indicate concern for the specific output.

src/Microsoft.DotNet.Interactive/Server/NotebookParserServerExtensions.cs
```diff
@@ -0,0 +1,22 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Text.Json;
+
+namespace Microsoft.DotNet.Interactive.Server
+{
+    public static class NotebookParserServerExtensions
```
jonsequitur, line 8: I don't think these belong in Microsoft.DotNet.Interactive.

src/dotnet-interactive/CommandLine/ParseNotebookCommand.cs
```diff
@@ -0,0 +1,18 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Threading.Tasks;
+using Microsoft.DotNet.Interactive.Server;
+
+namespace Microsoft.DotNet.Interactive.App.CommandLine
+{
+    internal static class ParseNotebookCommand
+    {
+        public static async Task<int> Do(NotebookParserServer parserServer)
```
jonsequitur, line 11: FYI, you can return `Task` and skip the `return 0`. It'll product a non-zero exit code if an exception is thrown.

src/Microsoft.DotNet.Interactive/Server/NotebookParserServer.cs
```diff
@@ -0,0 +1,409 @@
...
+using System.Collections.Generic;
+using System.IO;
+using System.Text.Json;
+using System.Threading.Tasks;
+using Microsoft.DotNet.Interactive.Documents;
+using Microsoft.DotNet.Interactive.Documents.Jupyter;
+
+namespace Microsoft.DotNet.Interactive.Server
+{
+    public class NotebookParserServer
```
jonsequitur, line 14: I don't think this belongs in Microsoft.DotNet.Interactive.

src/Microsoft.DotNet.Interactive/Server/NotebookParserServer.cs
```diff
@@ -0,0 +1,409 @@
...
+                            {
+                                ".dib" => CodeSubmission.Read(contentStream, request.DefaultLanguage, WellKnownKernelNames),
+                                ".ipynb" => Notebook.Read(contentStream, WellKnownKernelNames),
+                                _ => throw new NotSupportedException($"Unable to parse an interactive document with extension '{request.Extension}'"),
+                            };
+                            return new NotebookParseResponse(request.Id, document);
+                        }
+                    case NotebookSerializeRequest serialize:
+                        {
+                            var resultStream = new MemoryStream();
```
jonsequitur, line 87: This needs to be disposed.

src/Microsoft.DotNet.Interactive/Server/NotebookParserServer.cs
```diff
@@ -0,0 +1,409 @@
...
+
+        public static NotebookParserServerResponse HandleRequest(NotebookParseOrSerializeRequest request)
+        {
+            try
+            {
+                switch (request)
+                {
+                    case NotebookParseRequest parse:
+                        {
+                            var contentStream = new MemoryStream(parse.RawData);
```
jonsequitur, line 76: This needs to be disposed.

src/Microsoft.DotNet.Interactive/Server/NotebookParserServer.cs
```diff
@@ -0,0 +1,409 @@
...
+                            var document = request.Extension.ToLowerInvariant() switch
```
jonsequitur, line 77: Is there a way to specify the format and override the expectation based on file extension?

src/Microsoft.DotNet.Interactive/Server/NotebookParserServer.cs
```diff
@@ -0,0 +1,409 @@
...
+    }
+
+    public abstract class NotebookParseOrSerializeRequest
+    {
+        public string Type { get; set; }
+        public string Id { get; set; }
+        public string Extension { get; set; }
+        public string DefaultLanguage { get; set; }
+
+        protected NotebookParseOrSerializeRequest(string type, string id, string extension, string defaultLanguage)
```
jonsequitur, line 122: What's the point of the constructor if the properties have setters?

src/Microsoft.DotNet.Interactive/Server/NotebookParserServer.cs
```diff
@@ -0,0 +1,409 @@
...
+                }
+            }
+            catch (Exception ex)
+            {
+                return new NotebookErrorResponse(request.Id, ex.ToString());
+            }
+        }
+    }
+
+    public abstract class NotebookParseOrSerializeRequest
```
jonsequitur, line 115: Separate file please.

src/Microsoft.DotNet.Interactive/Server/NotebookParserServer.cs
brettfo: This is the result of me fighting with the JSON deserializer and trying to make it happy.  I should be able to get rid of the setters.

src/Microsoft.DotNet.Interactive/Server/NotebookParserServer.cs
brettfo: The format is dictated by the `Extension` property, but it sounds like calling it `Format` with values of `dib` and `ipynb` would be more clear.

src/Microsoft.DotNet.Interactive.Documents.Tests/NotebookParserServerTests_ObjectInterface.cs
```diff
@@ -0,0 +1,92 @@
...
+using Microsoft.DotNet.Interactive.Documents.ParserServer;
+using Xunit;
+
+namespace Microsoft.DotNet.Interactive.Documents.Tests
+{
+    public class NotebookParserServerTests_ObjectInterface
+    {
+        [Theory]
+        [InlineData(DocumentSerializationType.Dib, "#!csharp\nvar x = 1;")]
+        [InlineData(DocumentSerializationType.Ipynb, @"{""cells"":[{""cell_type"":""code"",""source"":[""var x = 1;""]}]}")]
```
jonsequitur, line 17: Think these names are inconsistent with what we're calling "CodeSubmission" and "Jupyter" elsewhere.

brettfo: You mean for the enum `DocumentSerializationType`?

Would you prefer this?

``` csharp
public enum DocumentSerializationType
{
    CodeSubmission,
    Jupyter,
}
```

jonsequitur: That looks good.

colombod (approved):

PR #1610, opened 2021-09-25
proxy kernel handles its own anonymous commands for fast-tracked requests

brettfo:
Includes:
- all instances of `KernelCommand == KernelCommand` and replace with the appropriate comparison.
- normalization of `"command-id"` to `"id"`.
- ~~preserve original language service command if no split is needed.~~ Change caused problems and pulled out.
- publish events for remotely split commands by token.


colombod (approved):

src/Microsoft.DotNet.Interactive/Kernel.cs
```diff
@@ -189,6 +189,14 @@ private bool TryPreprocessLanguageServiceCommand(LanguageServiceCommand command,
             }
 
             commands = postProcessCommands;
+            if (commands.Count == 1 &&
+                commands[0] is LanguageServiceCommand languageCommand &&
```
jonsequitur, line 193: The intent here isn't clear. We should extract a common way here and elsewhere to identify commands to be fast-tracked.

brettfo: This isn't about fast-tracked commands, it's just a place we found where language service commands were unnecessarily being split.  The normal submission parser does something similar, we just forgot to do it here.

jonsequitur: It's still worth consolidating this logic. It's not really clear what the "why" is here. Extracting a named method could help.

brettfo: I had to pull this change out, since it caused problems with how we compute language service commands and offsets.

jonsequitur (approved):

PR #1615, opened 2021-09-27
update typescript api for vscode

colombod:
using the id property from the new contract

### DIFF
--- a/src/dotnet-interactive-vscode/common/interactiveClient.ts
+++ b/src/dotnet-interactive-vscode/common/interactiveClient.ts
@@ -47,6 +47,7 @@ import { clearDebounce, createOutput } from './utilities';
 import * as vscodeLike from './interfaces/vscode-like';
 import { CompositeKernel } from './interactive/compositeKernel';
 import { ProxyKernel } from './interactive/proxyKernel';
+import { Guid } from './interactive/tokenGenerator';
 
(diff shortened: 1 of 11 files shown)


src/dotnet-interactive-vscode/common/interactiveClient.ts
```diff
@@ -316,8 +320,9 @@ export class InteractiveClient {
         return new Promise<TEvent>(async (resolve, reject) => {
             let handled = false;
             token = token || this.getNextToken();
+            const id = Guid.create().toString();
             let disposable = this.subscribeToKernelTokenEvents(token, eventEnvelope => {
-                if (eventEnvelope.command?.token === token) {
+                if (eventEnvelope.command?.id === id) {
```
jonsequitur, line 325: Why this change and are there tests that demonstrate that this is correct?

colombod: We have tests covering that part of the code. The change is because we want to terminate only when we obtain termination events from the command vscode extension generated via the client. We also tested using the notebook we have to exercise the feature set.

colombod: needs better testing

jonsequitur (approved):

PR #1617, opened 2021-09-28
address cancel command tests timeouts

colombod:


src/Microsoft.DotNet.Interactive.Tests/CancelCommandTests.cs
```diff
@@ -86,18 +86,30 @@ public async Task can_cancel_user_loop_using_CancellationToken()
 
             var commandToCancel = new SubmitCode(@"
 using Microsoft.DotNet.Interactive;
-
-while(!KernelInvocationContext.Current.CancellationToken.IsCancellationRequested){ await Task.Delay(10); }", targetKernelName: "csharp");
+var cancellationToken = KernelInvocationContext.Current.CancellationToken;
+while(!cancellationToken.IsCancellationRequested){ 
+    await Task.Delay(10); 
+    if(cancellationToken != KernelInvocationContext.Current.CancellationToken){
+        throw new Exception(""The token is different"");
```
jonsequitur, line 93: We should remove this exception.

PR #1619, opened 2021-09-30
update api surface for frontend kernel

colombod:
perform input gestures as static import, no variable injection


src/Microsoft.DotNet.Interactive.VSCode/KernelExtensions.cs
```diff
@@ -1,20 +1,53 @@
...
+#r ""{typeof(VSCodeInteractiveHost).Assembly.Location.Replace("\\", "/")}""
+open {typeof(VSCodeInteractiveHost).FullName};
+");
+            kernel.DeferCommand(command);
+            return kernel;
         }
+
+        public static PowerShellKernel UseVSCodeHelpers(this PowerShellKernel kernel)
+        {
+            kernel.ReadInput = prompt => VSCodeInteractiveHost.GetInputAsync(prompt: prompt).Result;
```
jonsequitur, line 42: Can this handler be made `async`? Otherwise please wrap this in `Task`.

src/Microsoft.DotNet.Interactive.Jupyter/JupyterInteractiveHost.cs
```diff
@@ -8,13 +8,9 @@
...
 {
-    public class JupyterInteractiveHost : IInteractiveHost
+    public static class JupyterInteractiveHost 
     {
-        public JupyterInteractiveHost()
-        {
-        }
-
-        public Task<string> GetInputAsync(string prompt = "", bool isPassword = false, CancellationToken cancellationToken = default)
+        public static  Task<string> GetInputAsync(string prompt = "", bool isPassword = false, CancellationToken cancellationToken = default)
```
jonsequitur, line 13: How will we enforce that this method signature is the same across hosts? Keep in mind notebooks need to be round tripped between VS Code, Jupyter, etc.

src/Microsoft.DotNet.Interactive.VSCode/KernelExtensions.cs
```diff
@@ -1,20 +1,53 @@
...
+            return kernel;
+        }
 
-            var interactiveHost = new VSCodeInteractiveHost(rootKernel);
-            return kernel.SetValueAsync("InteractiveHost", interactiveHost, typeof(IInteractiveHost));
+        public static FSharpKernel UseVSCodeHelpers(this FSharpKernel kernel)
+        {
+            var command = new SubmitCode($@"
+#r ""{typeof(VSCodeInteractiveHost).Assembly.Location.Replace("\\", "/")}""
+open {typeof(VSCodeInteractiveHost).FullName};
```
jonsequitur, line 34: Don't need the closing semicolon.

src/Microsoft.DotNet.Interactive.Jupyter/JupyterInteractiveHost.cs
colombod: that api is exposed in both vscode and jupyter mode

src/Microsoft.DotNet.Interactive.VSCode/KernelExtensions.cs
colombod: you mean like 
```csharp
kernel.ReadInput = prompt => Task.Run(() =>VSCodeInteractiveHost.GetInputAsync(prompt: prompt)).Result;
            kernel.ReadPassword = prompt =>
            {
                var value = Task.Run( () => VSCodeInteractiveHost.GetInputAsync(prompt: prompt, isPassword: true)).Result;
                return new PasswordString(value);
            };
```

src/Microsoft.DotNet.Interactive.VSCode/KernelExtensions.cs
colombod: we do that all over the place in case location contains a white space in the path I think

src/Microsoft.DotNet.Interactive.Jupyter/JupyterInteractiveHost.cs
jonsequitur: Understood but what enforces that the API is the same for both?

src/Microsoft.DotNet.Interactive.VSCode/KernelExtensions.cs
jonsequitur: Yeah, that won't risk a deadlock.

src/Microsoft.DotNet.Interactive.Jupyter/JupyterInteractiveHost.cs
colombod: both experiences use the exact same GetInputAsync method

colombod: sorry i see now what you mean

src/Microsoft.DotNet.Interactive.Jupyter/KernelExtensions.cs
```diff
@@ -44,6 +39,26 @@ public static T UseDefaultMagicCommands<T>(this T kernel)
...
+
+        public static FSharpKernel UseJupyterHelpers(
+            this FSharpKernel kernel)
+        {
+            var command = new SubmitCode($@"
+#r ""{typeof(TopLevelMethods).Assembly.Location.Replace("\\", "/")}""
+
+[<AutoOpen>]
+module JupyterTopLevelModule =
+    let GetInputAsync (prompt : string, isPassword : bool) =
```
brettfo, line 57: Give this a shot:

```suggestion
    let GetInputAsync ([<Optional; DefaultParameterValue("""")>] prompt : string, [<Optional; DefaultParameterValue(false)>] isPassword : bool) = 
```

colombod: it did not work well :)

brettfo (approved):

PR #1631, opened 2021-10-05
removing kernel server and frontend kernel

colombod:
use new api and proxy kernels to remove the kernel server and frontend kernel in
stdio 
vscode

### DIFF
--- a/src/Microsoft.DotNet.Interactive.VSCode/VSCodeInteractiveHost.cs
+++ b/src/Microsoft.DotNet.Interactive.VSCode/VSCodeInteractiveHost.cs
@@ -6,8 +6,6 @@ using System.Reactive.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 
-using Microsoft.DotNet.Interactive.Events;
-
 namespace Microsoft.DotNet.Interactive.VSCode
 {
     public static class VSCodeInteractiveHost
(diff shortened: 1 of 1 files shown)


src/Microsoft.DotNet.Interactive/Connection/CommandOrEventExtensions.cs
```diff
@@ -0,0 +1,25 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.DotNet.Interactive.Connection
+{
+    public static class CommandOrEventExtensions
+    {
+        public static Task DispatchAsync(this CommandOrEvent commandOrEvent, Kernel kernel, CancellationToken cancellationToken)
```
jonsequitur, line 11: This doesn't seem like an API we should expose publicly.

src/dotnet-interactive/CommandLine/VSCodeCommand.cs
```diff
@@ -17,5 +20,36 @@ public static async Task<int> Do(StartupOptions startupOptions, KernelServer ker
...
+                    if (commandOrEvent.IsParseError)
+                    {
+                        var _ = sender.SendAsync(commandOrEvent.Event, cancellationTokenSource.Token);
+                    }
+                    else
+                    {
+                        await commandOrEvent.DispatchAsync(kernel, cancellationTokenSource.Token);
+                    }
+                }
+            }, cancellationTokenSource.Token);
```
jonsequitur, line 48: Why is any of this orchestration unique to `vscode` mode?

src/dotnet-interactive/dotnet-interactive.csproj
```diff
@@ -36,6 +36,14 @@
     <None Remove="resources\*.js" />
   </ItemGroup>
 
+  <PropertyGroup>
+    <MicrosoftCodeAnalysisAnalyzersVersion>2.9.6</MicrosoftCodeAnalysisAnalyzersVersion>
+    <SystemRuntimeLoaderVersion>4.3.0</SystemRuntimeLoaderVersion>
+    <SystemCollectionsImmutableVersion>5.0.0</SystemCollectionsImmutableVersion>
+    <SystemReflectionMetadataVersion>5.0.0</SystemReflectionMetadataVersion>
+    <SystemCommandLineVersion>2.0.0-beta1.21308.1</SystemCommandLineVersion>
+  </PropertyGroup>
```
brettfo, line 45: Where did all these values come from and where are they used?  The Arcade way of setting package values like this is to put them at the end of `eng/Versions.props` to keep them collected.

PR #1632, opened 2021-10-06
Updating all Microsoft.CodeAnalysis.* packages to version 4.0.0-4.final to support C# 10

MoaidHathot:
C# 10 is supported starting from version `4` and the current `3.11.0` version does not support it yet.
This PR updates all `Microsoft.CodeAnalysis.*` packages to the prerelease version `4.0.0-4.final` which enables C# 10 support in Interactive Notebooks running on .NET 6

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -16,7 +16,7 @@
 
   <PropertyGroup>
     <SystemReactiveVersion>5.0.0</SystemReactiveVersion>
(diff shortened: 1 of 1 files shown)


brettfo:
Merge and signed build are complete, instructions have been updated [here](Running%20a%20.NET%206%20kernel.md).  In short, package version changed from `1.0.250102` to `1.0.250602`.

PR #1660, opened 2021-10-14
Add input var support via #!share for MS SQL extension

Charles-Gagnon:
Closes https://github.com/dotnet/interactive/issues/1647

KQL coming up next, just a matter of converting the types and constructing the declare statements.

For the unsupported types - we should be able to handle some of those (uint for example) fine if we just cast them to an appropriate type first. I'll investigate as a follow up task - wanted to keep this as straightforward as possible for now.

Variables being added normally
image

(note - there's an issue where we're getting two CodeSubmitted commands when we share multiple times which is why you see the duplicate messages. I'll file an issue as a follow up to investigate why this is happening)

Don't support nulls error
image

Unsupported type error
image

### DIFF
--- a/src/Microsoft.DotNet.Interactive/KernelExtensions.cs
+++ b/src/Microsoft.DotNet.Interactive/KernelExtensions.cs
@@ -205,7 +205,7 @@ namespace Microsoft.DotNet.Interactive
                             await ((ISupportSetValue)kernel).SetValueAsync(name, shared);
                         } catch (Exception ex)
                         {
-                            throw new InvalidOperationException($"Error setting value '{name}' from kernel '{from}'. {ex.Message}", ex);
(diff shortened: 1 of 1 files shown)


Charles-Gagnon:
@colombod @jonsequitur

Charles-Gagnon:
@jonsequitur @colombod Could one of you restart the Windows pipeline? Looks like the flakey test that @colombod was telling me about.

Charles-Gagnon:
Looks like a couple transient test failures again - would one of you two mind restarting those? @jonsequitur @colombod

colombod:
> Added screenshots. Also note the issue I called out where I'm seeing multiple CodeSubmitted commands (one of them being empty) when we have multiple `#!share` directives. I'll file an issue for that once this gets in as long as we're ok with that for now, but if you think this might be an issue with my code let me know.

Looks cool, what about making the error even more informative like "Error sharing value 'x' from kernel 'csharp' into kernel 'sql-sqltools2019' ........."

PR #1674, opened 2021-10-27
support multiple mimeType

colombod:
when using `Displa()` it is possible to produce multiple formatted outputs specifying multiple mimetypes

`DataExplorer<T>` registers formatters for json and tabular json if the type T is TabularDataResource

### DIFF
--- a/src/Microsoft.DotNet.Interactive.ExtensionLab.Tests/NteractKernelExtensionTests.cs
+++ b/src/Microsoft.DotNet.Interactive.ExtensionLab.Tests/NteractKernelExtensionTests.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
+using System.Linq;
 using System.Threading.Tasks;
 
 using Assent;
@@ -11,7 +12,7 @@ using Assent;
 using FluentAssertions;
 
 using Microsoft.DotNet.Interactive.Formatting;
-
+using Microsoft.DotNet.Interactive.Formatting.TabularData;
 using Xunit;
 
 namespace Microsoft.DotNet.Interactive.ExtensionLab.Tests
@@ -28,7 +29,23 @@ namespace Microsoft.DotNet.Interactive.ExtensionLab.Tests
         }
 
         [Fact]
-        public async Task it_registers_formatters()
+        public async Task it_configures_preferred_mimeTypes()
+        {
+            using var kernel = new CompositeKernel();
+
+            var kernelExtension = new NteractKernelExtension();
+
+            await kernelExtension.OnLoadAsync(kernel);
+
+            var mimetypes = Formatter.GetPreferredMimeTypesFor(typeof(NteractDataExplorer));
+
+            mimetypes
+                .Should().BeEquivalentTo(HtmlFormatter.MimeType, TabularDataResourceFormatter.MimeType);
+        }
+
+
+        [Fact]
+        public async Task it_registers_html_formatter()
         {
             using var kernel = new CompositeKernel();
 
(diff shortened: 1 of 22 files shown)


src/Microsoft.DotNet.Interactive.ExtensionLab.Tests/NteractKernelExtensionTests.cs
```diff
@@ -44,9 +61,52 @@ public async Task it_registers_formatters()
             };
 
 
-            var formatted = data.ExploreWithNteract().ToDisplayString(HtmlFormatter.MimeType);
+            var formattedValue = data.ExploreWithNteract().ToDisplayString(HtmlFormatter.MimeType);
+            formattedValue.Should().Contain("configureRequireFromExtension('nteract','1.0.0')(['nteract/nteractapi'], (nteract) => {");
+        }
+
+        [Fact]
+        public async Task it_registers_tabularResource_formatter()
```
jonsequitur, line 69:
```suggestion
        public async Task it_registers_TabularDataResourceFormatter()
```

src/Microsoft.DotNet.Interactive.ExtensionLab.Tests/SandDanceKernelExtensionTests.cs
```diff
@@ -48,6 +64,50 @@ public async Task it_registers_formatters()
             formatted.Should().Contain("configureRequireFromExtension('SandDance','1.0.0')(['SandDance/sanddanceapi'], (sandDance) => {");
         }
 
+        [Fact]
+        public async Task it_registers_tabularResource_formatter()
```
jonsequitur, line 68:
```suggestion
        public async Task it_registers_TabularDataResourceFormatter()
```

src/Microsoft.DotNet.Interactive.Formatting.Tests/FormatterTests.cs
```diff
@@ -170,10 +171,10 @@ public void Check_default_formatting_of_strings_with_escapes()
             {
                 var value = "hola! \n \t \" \" ' ' the joy of escapes! and    white  space  ";
 
-                var mimeType = Formatter.GetPreferredMimeTypeFor(typeof(string));
-                var text = value.ToDisplayString(mimeType);
+                var mimeType = Formatter.GetPreferredMimeTypesFor(typeof(string));
+                var text = value.ToDisplayString(mimeType.FirstOrDefault());
```
jonsequitur, line 175: The fact that it's first isn't important. Maybe select it out by name to make the test more robust?

src/Microsoft.DotNet.Interactive.Formatting/Formatter.cs
```diff
@@ -530,9 +554,9 @@ internal static ITypeFormatter InferPreferredFormatter(Type actualType, string m
 
             // Last resort use preferred mimeType
 
-            var preferredMimeType = GetPreferredMimeTypeFor(actualType);
+            var preferredMimeType = GetPreferredMimeTypesFor(actualType);
 
-            return GetPreferredFormatterFor(actualType, preferredMimeType);
+            return GetPreferredFormatterFor(actualType, preferredMimeType.FirstOrDefault());
```
jonsequitur, line 559: Can we really assume the first one is the preferred one? Since these are intended to be registered from different extensions that have no knowledge of one another, we should probably make this method return multiple formatters just like we've changed the other methods to return multiple MIME types.

src/Microsoft.DotNet.Interactive/Commands/RequestValue.cs
```diff
@@ -32,7 +33,7 @@ public override Task InvokeAsync(KernelInvocationContext context)
                     if (value is { })
                     {
                         var valueType = value.GetType();
-                        var mimeType = MimeType ?? Formatter.GetPreferredMimeTypeFor(valueType);
+                        var mimeType = MimeType ?? Formatter.GetPreferredMimeTypesFor(valueType).FirstOrDefault();
```
jonsequitur, line 36: Again, `FirstOrDefault` makes me nervous. If we need to converge to a single one when multiple are registered, what should be the behavior? Fall back to the overall preferred MIME type if available?

I also think some MIME types are more suitable for data and some for display. In the case of `RequestValue` we might prefer a data MIME type as a convention.

src/Microsoft.DotNet.Interactive/Commands/RequestValue.cs
```diff
@@ -46,7 +47,7 @@ public override Task InvokeAsync(KernelInvocationContext context)
                     }
                     else
                     {
-                        var mimeType = MimeType ?? Formatter.GetPreferredMimeTypeFor(typeof(object));
+                        var mimeType = MimeType ?? Formatter.GetPreferredMimeTypesFor(typeof(object)).FirstOrDefault();
```
jonsequitur, line 50: Again, this needs more thought.

src/Microsoft.DotNet.Interactive.Formatting/Formatter.cs
colombod: i like the idea

src/Microsoft.DotNet.Interactive.Formatting.Tests/FormatterTests.cs
colombod: right

src/Microsoft.DotNet.Interactive/Commands/RequestValue.cs
colombod: makes sense

src/Microsoft.DotNet.Interactive/Commands/RequestValue.cs
colombod: changed

src/Microsoft.DotNet.Interactive.Formatting/Formatter.cs
```diff
@@ -530,9 +603,9 @@ internal static ITypeFormatter InferPreferredFormatter(Type actualType, string m
 
             // Last resort use preferred mimeType
 
-            var preferredMimeType = GetPreferredMimeTypeFor(actualType);
+            var preferredMimeType = GetPreferredDisplayMimeTypesFor(actualType);
```
jonsequitur, line 606: Why are we renaming this?

src/Microsoft.DotNet.Interactive.Formatting/Formatter.cs
```diff
@@ -145,20 +172,35 @@ public static void SetPreferredMimeTypeFor(Type type, string preferredMimeType)
...
+            set => _defaultDataMimeType = value ??
+                                          throw new ArgumentNullException(nameof(value));
+        }
+
+        public static IReadOnlyCollection<string> GetPreferredDisplayMimeTypesFor(Type type) =>
+            _preferredDisplayMimeTypesTable.GetOrAdd(
+                type ??= typeof(object), 
+                key => InferMimeTypes(key, _preferredDisplayMimeTypes, _defaultPreferredDisplayMimeTypes, _defaultDisplayMimeType));
+
+        public static string GetPreferredDataMimeTypeFor(Type type) =>
```
jonsequitur, line 200: I don't think splitting this into two methods makes things clearer. If I'm understanding the change, this keeps the `FirstOrDefault` ambiguity in `GetPreferredDisplayMimeTypesFor` and introduces a new ambiguity around which of these methods to use at any given time.

My thought about "display" versus "data" MIME types that we discussed wasn't intended to be reflected in the API. I was just trying to draw attention to the idea that when multiple are registered, they have different purposes. For example, say both `text/html` and `application/json` are registered. We can choose between them based on the context. Are we displaying something? We could use `text/html`. Are we transferring data? We could use `application/json`.

jonsequitur (approved):

PR #1685, opened 2021-11-02
default to use cdn files

colombod:


src/Microsoft.DotNet.Interactive.ExtensionLab/NteractKernelExtension.cs
```diff
@@ -11,7 +12,7 @@ public class NteractKernelExtension : IKernelExtension, IStaticContentSource
     {
         public Task OnLoadAsync(Kernel kernel)
         {
-            kernel.UseNteractDataExplorer();
+            kernel.UseNteractDataExplorer(libraryUri:new Uri(@"https://colombod.github.io/dotnet-interactive-cdn/extensionlab/1.0.252001/nteract/nteractapi.js", UriKind.Absolute), libraryVersion: "1.0.252001");
```
brettfo, line 15: Same parameterization request with the version number.

colombod: is a constant value at the moment. we will change it once we get a build to generate and publish those js files

PR #1687, opened 2021-11-02
add --name option to MsKql and MsSql kernels to store query results

colombod:
Can use --name to store query resutls for variable sharing

In MsSql
```
#!sql-adventureworks --name my_data_result
select * from sys.databases
```

then get it from csharp
```
#!csharp
#!share --from sql-adventureworks my_data_result
my_data_result
```


src/Microsoft.DotNet.Interactive.SqlServer/MsSqlKernel.cs
```diff
@@ -122,5 +99,29 @@ protected override bool CanDeclareVariable(string name, object value, out string
 
             return true;
         }
+
+        protected override void StoreQueryResults(List<TabularDataResource> results, ParseResult commandKernelChooserParseResult)
```
jonsequitur, line 103: Does this need to be `List<TabularDataResource>` as opposed to say `IReadOnlyList<TabularDataResource>`?

colombod: fixed

jonsequitur (approved):

PR #1690, opened 2021-11-03
update logic for valuekernel chooser

colombod:
value kernel is using parse invoke when options --from-file or --from-url are used
if there is a command for the kernel then at that point errors are handled
errors won't store value
errors will retain previous value if present

### DIFF
--- a/src/Microsoft.DotNet.Interactive/KeyValueStoreKernel.cs
+++ b/src/Microsoft.DotNet.Interactive/KeyValueStoreKernel.cs
@@ -4,10 +4,6 @@
 using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
-using System.CommandLine;
-using System.CommandLine.Binding;
-using System.CommandLine.Invocation;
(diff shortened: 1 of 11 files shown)


src/Microsoft.DotNet.Interactive/KeyValueStoreKernel.cs
```diff
@@ -26,6 +22,8 @@ public class KeyValueStoreKernel :
         internal const string DefaultKernelName = "value";
 
         private readonly ConcurrentDictionary<string, object> _values = new();
+        private ChooseKeyValueStoreKernelDirective _chooseKernelDirective;
+        private (bool hadValue, object previousValue, string newValue)? _lastOperation;
```
jonsequitur, line 26: Are these stored values? Why is `previousValue` an `object` while `newValue` is a `string`?

colombod: type inference from refactoring tools, fixing that now

PR #1701, opened 2021-11-04
add journey to main product

colombod:
moving journey to main product


src/Microsoft.DotNet.Interactive.Journey/Outcome.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.DotNet.Interactive.Journey
+{
+    public enum Outcome
```
jonsequitur, line 6: Is this needed?

jonsequitur (approved):

src/Microsoft.DotNet.Interactive.Journey/Outcome.cs
colombod: not sure

PR #1738, opened 2021-11-13
Command and event API support for requesting kernel info

jonsequitur:


src/Microsoft.DotNet.Interactive/Commands/RequestKernelInfo.cs
```diff
@@ -0,0 +1,32 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Threading.Tasks;
+
+namespace Microsoft.DotNet.Interactive.Commands;
+
+public class RequestKernelInfo : KernelCommand
+{
+    public override async Task InvokeAsync(KernelInvocationContext context)
```
colombod, line 10: no need for async

src/Microsoft.DotNet.Interactive/Commands/RequestKernelInfo.cs
```diff
@@ -0,0 +1,32 @@
...
+            if (host.TryGetKernelInfo(kernel, out var kernelInfo))
+            {
+
+
+
+
+
+            }
+        }
+    }
```
colombod, line 31: need to return a completed task

colombod (approved):

src/microsoft-dotnet-interactive/src/contracts.ts
```diff
@@ -105,7 +110,7 @@ export interface SendEditableCode extends KernelCommand {
 
 export interface SubmitCode extends KernelCommand {
     code: string;
-    submissionType?: SubmissionType;
```
colombod, line 108: this is now mandatory but used to be optional

src/microsoft-dotnet-interactive/src/contracts.ts
```diff
@@ -239,7 +246,7 @@ export interface CompleteCodeSubmissionReceived extends KernelEvent {
 }
 
 export interface CompletionsProduced extends KernelEvent {
-    linePositionSpan?: LinePositionSpan;
```
colombod, line 242: this is now mandatory but used to be optional

colombod (changes requested):

src/microsoft-dotnet-interactive/src/contracts.ts
```diff
@@ -272,7 +279,7 @@ export interface ErrorProduced extends DisplayEvent {
 
 export interface HoverTextProduced extends KernelEvent {
     content: Array<FormattedValue>;
-    linePositionSpan?: LinePositionSpan;
```
colombod, line 275: this is now mandatory but used to be optional

PR #1757, opened 2021-11-16
move all npm-related stuff to build scripts

brettfo:
With this change running `build.cmd`/`./build.sh` will do `npm i` and `npm run compile` before `dotnet` is invoked which means the project files could be simplified.

This also means the dev instructions can be simplified to just "Run the build script" (pending).

PR marked as draft while I run some internal builds to make sure everything still works, including the `SkipTests=true` flag.

N.b., The Linux portion of this PR was done entirely in a Codspaces VM and because of this some devcontainer stuff was updated to make the experience easier.


brettfo:
This _mostly_ works.  For reasons unknown to me, setting `SkipTests=true` in the build still causes the NPM tests to run.  I have no idea why and I'm willing to explain my bafflement to anybody that wants to be sad.  The .NET tests, on the other hand, _can_ be skipped appropriately.

At @colombod's request I also added a `build-js.cmd` script that will _only_ build the NPM packages.

colombod (approved):

brettfo:
Got it all working as expected.

PR #1760, opened 2021-11-16
if there is no challenge there should be no error

colombod:


src/Microsoft.DotNet.Interactive.Journey.Tests/LessonTests.cs
```diff
@@ -163,6 +163,32 @@ public async Task explicitly_starting_the_next_challenge_at_last_challenge_does_
             Lesson.CurrentChallenge.Should().Be(null);
         }
 
+        [Fact]
+        public async Task when_student_reaches_the_end_submissions_work_as_normal()
```
jonsequitur, line 167:
```suggestion
        public async Task when_student_reaches_the_end_then_submissions_work_as_normal()
```

colombod: done

jonsequitur (approved):

PR #1761, opened 2021-11-17
fix result set sharing

colombod:


colombod:
@Charles-Gagnon this should do

src/Microsoft.DotNet.Interactive.SqlServer/ToolsServiceKernel.cs
```diff
@@ -131,7 +131,7 @@ public async Task HandleAsync(SubmitCode command, KernelInvocationContext contex
             {
                 try
                 {
-                   
+                    var results = new List<TabularDataResource>();
```
Charles-Gagnon, line 134: You're calling StoreQueryResults for each summary still which is at best unnecessary and at worst could result in multiple names with the same list (if it's even possible for the name to change...which seems unlikely but still). 

You could move that call out of the outer foreach right?

jonsequitur (approved):

Charles-Gagnon (approved):
See comment, but functionally this looks great. Thanks for the quick fix! 😄

src/Microsoft.DotNet.Interactive.SqlServer/ToolsServiceKernel.cs
colombod: name collision is left to the implementation. Storing at each bach ensure nothing is lost dueto exceptions or cancellation

Charles-Gagnon: Why not call store right after creating the list then? It just seems unnecessary to call store each time when you're passing in the same list every time

PR #1762, opened 2021-11-17
Update SqlToolsService to 157

Charles-Gagnon:
NOTE: These packages are NOT published as of the opening of this PR. 

This brings in the full set of Linux RIDs as well as many other bug fixes/improvements to the service itself
Also : 

1. Fixed a couple of places that were still referencing net5.0 
2. Changed result handler to only store the query results once at the end (regardless of whether an error occurred or not)


Charles-Gagnon:
@colombod @KevinRansom @jonsequitur

colombod (approved):

PR #1773, opened 2021-11-17
updating packages and keep newtonsoft in tool project

colombod:


src/dotnet-interactive/JupyterInstallCommand.cs
```diff
@@ -75,14 +75,16 @@ private static void ComputeKernelSpecArgs(HttpPortRange httpPortRange, Directory
...
-                var argv = parsed["argv"].Value<JArray>();
+                argv.Add( "--http-port-range");
+                argv.Add($"{httpPortRange.Start}-{httpPortRange.End}");
 
-                argv.Insert(argv.Count - 1, "--http-port-range");
-                argv.Insert(argv.Count - 1, $"{httpPortRange.Start}-{httpPortRange.End}");
+                newKernelSpec["argv"] = JsonSerializer.SerializeToElement(argv);
 
-                File.WriteAllText(kernelSpec.FullName, parsed.ToString(Newtonsoft.Json.Formatting.Indented));
+                File.WriteAllText(kernelSpec.FullName, JsonSerializer.Serialize(newKernelSpec));
```
jonsequitur, line 87: Should this be opinionated about which serialization settings are used? Are we contract-testing this?

colombod: this is jsut editing a json document ( the kernel spec file) leaving all untouched and jsut patching the argv array, the types are left as jsonelement all the way

colombod: makes sense, added contract pinning test

jonsequitur (approved):

PR #1778, opened 2021-11-18
set CI as Frontend telemetry 

colombod:
Added api to get frontend name from environment variable, CI is using itsown so we can also remove it from telemetry count

### DIFF
--- a/src/dotnet-interactive.Tests/CommandLine/TelemetryTests.cs
+++ b/src/dotnet-interactive.Tests/CommandLine/TelemetryTests.cs
@@ -199,13 +199,17 @@ public class TelemetryTests : IDisposable
     }
 
     [Fact]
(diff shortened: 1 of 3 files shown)


src/dotnet-interactive.Tests/CommandLine/TelemetryTests.cs
```diff
@@ -199,13 +199,17 @@ public async Task Jupyter_command_sends_frontend_telemetry()
     }
 
     [Fact]
-    public async Task Jupyter_install_command_does_not_send_frontend_telemetry()
+    public async Task Jupyter_install_command_sends_default_fronted_telemetry()
```
jonsequitur, line 202:
```suggestion
    public async Task Jupyter_install_command_sends_default_frontend_telemetry()
```

jonsequitur (approved):

PR #1779, opened 2021-11-19
change behavior for repeated load

colombod:
When using a wildcard to load a package already loaded the already loaded version is used

### DIFF
--- a/src/Microsoft.DotNet.Interactive.PackageManagement/KernelSupportsNugetExtensions.cs
+++ b/src/Microsoft.DotNet.Interactive.PackageManagement/KernelSupportsNugetExtensions.cs
@@ -98,8 +98,22 @@ namespace Microsoft.DotNet.Interactive
 
                     if (alreadyGotten is { } && !string.IsNullOrWhiteSpace(pkg.PackageVersion) && pkg.PackageVersion != alreadyGotten.PackageVersion)
                     {
-                        var errorMessage = GenerateErrorMessage(pkg, alreadyGotten).ToString(OutputMode.NonAnsi);
-                        context.Fail(context.Command, message: errorMessage);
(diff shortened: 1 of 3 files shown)


src/Microsoft.DotNet.Interactive.PackageManagement/KernelSupportsNugetExtensions.cs
```diff
@@ -98,8 +98,22 @@ private static Command r()
 
                     if (alreadyGotten is { } && !string.IsNullOrWhiteSpace(pkg.PackageVersion) && pkg.PackageVersion != alreadyGotten.PackageVersion)
                     {
-                        var errorMessage = GenerateErrorMessage(pkg, alreadyGotten).ToString(OutputMode.NonAnsi);
-                        context.Fail(context.Command, message: errorMessage);
+                        if (!pkg.IsPackageVersionSpecified || pkg.PackageVersion is "*-*" or "*")
```
brettfo, line 101: This shouldn't have the condition `!pkg.IsPackageVersionSpecified` because the default behavior in `.csproj` files is to use the _lowest_ version if nothing was specified, where the wildcard is to use the highest.  Otherwise the `"*-*" || "*"` looks good.

There might be an issue around `"*-*"` vs `"*"`, because `"*-*"` will match something like "1.0.0-beta.123456", but `"*"` will only match non-preview versions.

brettfo (approved):

src/Microsoft.DotNet.Interactive.PackageManagement/KernelSupportsNugetExtensions.cs
colombod: that would require check what unset or wildcard would resolve now.

PR #1785, opened 2021-11-24
Update vscode insiders and api update

colombod:
vscode repo has folder and file structure changes for d.ts files


src/dotnet-interactive-vscode/insiders/.eslintignore
```diff
@@ -1,3 +1,3 @@
 jupyter.d.ts
-vscode.proposed.d.ts
+vscode.proposed.*.d.ts
```
brettfo, line 2: Keep both; if I understand it correctly, `vscode.proposed.d.ts` won't be caught by the pattern `vscode.proposed.*.d.ts`; there would have to be 2 dots `.` next to each other, e.g., `vscode.proposed..d.ts`.

brettfo (approved):

src/dotnet-interactive-vscode/insiders/.eslintignore
colombod: done! thanks for spotting the issue!

Issue #1788, opened 2021-11-24
Error thrown when SqlToolsService package isn't able to be found is not very helpful


Charles-Gagnon:
Currently if the MS SQL or KQL extensions aren't able to find the SQL Tools Service exe to run then they'll fail when attempting to connect with some error about "PathToToolsService can't be null or empty".

1. This isn't very useful to the end-user
2. It would be ideal to have this happen as soon as the extension is loaded instead of on connect

PR #1791, opened 2021-11-24
Resolve SQL Tools Service path upon extension load and improve errors

Charles-Gagnon:
Fixes https://github.com/dotnet/interactive/issues/1788

Instead of resolving the path every time we connect we'll do it once upon extension load (since it should never be expected to change after loading). 

Also improved the error checking, there's 3 things we specifically check now : 

1. Can't find native STS package

NoPackage

2. Can't resolve path to STS exe within package

NoPath

3. Exe doesn't exist

NoExe

@colombod and I discussed adding a check on every connect to ensure that the exe still exists (like we essentially were doing) but the error message in that case seems clear enough so I think it's fine to leave it as is (if it disappeared that'd be quite strange anyways)

NoExeConnect

I am wondering about the way I'm displaying the error currently - it isn't really noticeable currently. Is there some standard formatting I should use? Or should I even throw Exceptions instead? The extension still loaded and so is available for things to reference, it's just the connect command that wouldn't be available. But it seems kind of odd since if that can't be found the whole point of the extension is kind of lost anyways...

### DIFF
--- a/src/Microsoft.DotNet.Interactive.Kql/ConnectKqlCommand.cs
+++ b/src/Microsoft.DotNet.Interactive.Kql/ConnectKqlCommand.cs
@@ -9,7 +9,7 @@ namespace Microsoft.DotNet.Interactive.Kql
 {
     public class ConnectKqlCommand : ConnectKernelCommand<KqlKernelConnector>
     {
-        internal string ResolvedToolsServicePath { get; set; }
+        private readonly string ResolvedToolsServicePath;
 
         public ConnectKqlCommand(string resolvedToolsServicePath)
             : base("kql", "Connects to a Microsoft Kusto Server database")
--- a/src/Microsoft.DotNet.Interactive.Kql/ConnectKqlCommand.cs
+++ b/src/Microsoft.DotNet.Interactive.Kql/ConnectKqlCommand.cs
@@ -4,13 +4,13 @@
 using System.CommandLine;
 using System.Threading.Tasks;
 using Microsoft.DotNet.Interactive.Connection;
-using Microsoft.DotNet.Interactive.Documents;
-using Microsoft.DotNet.Interactive.SqlServer;
 
 namespace Microsoft.DotNet.Interactive.Kql
 {
     public class ConnectKqlCommand : ConnectKernelCommand<KqlKernelConnector>
     {
+        public static string ResolvedToolsServicePath { get; internal set; }
+
         public ConnectKqlCommand()
             : base("kql", "Connects to a Microsoft Kusto Server database")
         {
@@ -25,11 +25,7 @@ namespace Microsoft.DotNet.Interactive.Kql
         public override async Task<Kernel> ConnectKernelAsync(KernelInfo kernelInfo, KqlKernelConnector connector,
             KernelInvocationContext context)
         {
-            var root = Kernel.Root.FindResolvedPackageReference();
-
-            var pathToService = root.PathToService("MicrosoftKustoServiceLayer");
-
-            connector.PathToService = pathToService;
+            connector.PathToService = ResolvedToolsServicePath;
 
             var kernel = await connector.ConnectKernelAsync(kernelInfo);
 
--- a/src/Microsoft.DotNet.Interactive.Kql/ConnectKqlCommand.cs
+++ b/src/Microsoft.DotNet.Interactive.Kql/ConnectKqlCommand.cs
@@ -9,11 +9,12 @@ namespace Microsoft.DotNet.Interactive.Kql
 {
     public class ConnectKqlCommand : ConnectKernelCommand<KqlKernelConnector>
     {
-        public static string ResolvedToolsServicePath { get; internal set; }
+        internal string ResolvedToolsServicePath { get; set; }
 
-        public ConnectKqlCommand()
+        public ConnectKqlCommand(string resolvedToolsServicePath)
             : base("kql", "Connects to a Microsoft Kusto Server database")
         {
+            ResolvedToolsServicePath = resolvedToolsServicePath;
             Add(new Option<string>(
                 "--cluster",
                 "The cluster used to connect") {IsRequired = true});
--- a/src/Microsoft.DotNet.Interactive.Kql/KqlKernelExtension.cs
+++ b/src/Microsoft.DotNet.Interactive.Kql/KqlKernelExtension.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System;
 using System.IO;
 using System.Runtime.InteropServices;
 using System.Threading.Tasks;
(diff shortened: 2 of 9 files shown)


src/Microsoft.DotNet.Interactive.Kql/KqlKernelExtension.cs
```diff
@@ -13,14 +15,41 @@ public Task OnLoadAsync(Kernel kernel)
...
+
+                            KernelInvocationContext.Current?.Display(
+                                new HtmlString(@"<details><summary>Query Microsoft Kusto Server databases.</summary>
+        <p>This extension adds support for connecting to Microsoft Kusto Server databases using the <code>#!connect kql</code> magic command. For more information, run a cell using the <code>#!kql</code> magic command.</p>
+        </details>"),
+                                "text/html");
+                        }
+                        else
+                        {
+                            KernelInvocationContext.Current?.DisplayStandardError($"The KQL extension was loaded successfully and resolved the path to the Kusto Tools Service but the file {pathToService} does not exist. The connect command will not be available.");
```
Charles-Gagnon, line 39: As mentioned I'm not sure this is going to be noticeable enough as an error (warning?). Maybe I could just stuff it in a span and make the text yellow/red?

Charles-Gagnon:
@colombod @jonsequitur

colombod:
You could use Context.Fail() and that will display error for the execution too, right now you are still “successful” but writing error message

Dr Diego Colombo PhD

src/Microsoft.DotNet.Interactive.Kql/KqlKernelExtension.cs
colombod:
```suggestion
                            throw new InvalidOperationException($"The KQL extension was loaded successfully and resolved the path to the Kusto Tools Service but the file {pathToService} does not exist. The connect command will not be available.");
```

src/Microsoft.DotNet.Interactive.Kql/KqlKernelExtension.cs
```diff
@@ -13,14 +15,41 @@ public Task OnLoadAsync(Kernel kernel)
...
+                        }
+
+                    }
+                    else
+                    {
+                        KernelInvocationContext.Current?.DisplayStandardError("The KQL extension was loaded successfully but was unable to determine the path to the KQL Tools Service. The connect command will not be available.");
```
colombod, line 45:
```suggestion
                        throw new InvalidOperationException("The KQL extension was loaded successfully but was unable to determine the path to the KQL Tools Service. The connect command will not be available.");
```

src/Microsoft.DotNet.Interactive.Kql/KqlKernelExtension.cs
```diff
@@ -13,14 +15,41 @@ public Task OnLoadAsync(Kernel kernel)
...
+                    }
+                }
+                else
+                {
+                    KernelInvocationContext.Current?.DisplayStandardError($"The KQL extension was loaded successfully but was unable to find the KQL Tools Service package. The connect command will not be available. (RID: {RuntimeInformation.RuntimeIdentifier})");
```
colombod, line 50:
```suggestion
                    throw new InvalidOperationException($"The KQL extension was loaded successfully but was unable to find the KQL Tools Service package. The connect command will not be available. (RID: {RuntimeInformation.RuntimeIdentifier})");
```

src/Microsoft.DotNet.Interactive.SqlServer/MsSqlKernelExtension.cs
```diff
@@ -12,15 +15,40 @@ public Task OnLoadAsync(Kernel kernel)
...
+                            KernelInvocationContext.Current?.Display(
+                                new HtmlString(@"<details><summary>Query Microsoft SQL Server databases.</summary>
     <p>This extension adds support for connecting to Microsoft SQL Server databases using the <code>#!connect mssql</code> magic command. For more information, run a cell using the <code>#!sql</code> magic command.</p>
     </details>"),
-                    "text/html");
+                                "text/html");
+                        }
+                        else
+                        {
+                            KernelInvocationContext.Current?.DisplayStandardError($"The SQL Server extension was loaded successfully and resolved the path to the SQL Tools Service but the file {pathToService} does not exist. The connect command will not be available.");
```
colombod, line 39:
```suggestion
                            throw new InvalidOperationException($"The SQL Server extension was loaded successfully and resolved the path to the SQL Tools Service but the file {pathToService} does not exist. The connect command will not be available.");
```

src/Microsoft.DotNet.Interactive.SqlServer/MsSqlKernelExtension.cs
```diff
@@ -12,15 +15,40 @@ public Task OnLoadAsync(Kernel kernel)
...
+                        }
+
+                    }
+                    else
+                    {
+                        KernelInvocationContext.Current?.DisplayStandardError($"The SQL Server extension was loaded successfully but was unable to determine the path to the SQL Tools Service. The connect command will not be available.");
```
colombod, line 45:
```suggestion
                        throw new InvalidOperationException($"The SQL Server extension was loaded successfully but was unable to determine the path to the SQL Tools Service. The connect command will not be available.");
```

src/Microsoft.DotNet.Interactive.SqlServer/MsSqlKernelExtension.cs
```diff
@@ -12,15 +15,40 @@ public Task OnLoadAsync(Kernel kernel)
...
+                    }
+                }
+                else
+                {
+                    KernelInvocationContext.Current?.DisplayStandardError($"The SQL Server extension was loaded successfully but was unable to find the SQL Tools Service package. The connect command will not be available. (RID: {RuntimeInformation.RuntimeIdentifier})");
```
colombod, line 50:
```suggestion
                    throw new InvalidOperationException($"The SQL Server extension was loaded successfully but was unable to find the SQL Tools Service package. The connect command will not be available. (RID: {RuntimeInformation.RuntimeIdentifier})");
```

src/Microsoft.DotNet.Interactive.Kql/ConnectKqlCommand.cs
```diff
@@ -4,13 +4,13 @@
...
 using System.Threading.Tasks;
 using Microsoft.DotNet.Interactive.Connection;
-using Microsoft.DotNet.Interactive.Documents;
-using Microsoft.DotNet.Interactive.SqlServer;
 
 namespace Microsoft.DotNet.Interactive.Kql
 {
     public class ConnectKqlCommand : ConnectKernelCommand<KqlKernelConnector>
     {
+        public static string ResolvedToolsServicePath { get; internal set; }
```
jonsequitur, line 12: If this can be made non-`static` and/or non-`public` that would be good.

src/Microsoft.DotNet.Interactive.Kql/KqlKernelExtension.cs
jonsequitur: I'd also suggest for these errors being more specific: `The #!connect kql command will not be available`

colombod (approved):

src/Microsoft.DotNet.Interactive.SqlServer/ConnectMsSqlCommand.cs
```diff
@@ -5,15 +5,17 @@
...
 using Microsoft.DotNet.Interactive.Connection;
 using Microsoft.DotNet.Interactive.CSharp;
-using Microsoft.DotNet.Interactive.Documents;
 
 namespace Microsoft.DotNet.Interactive.SqlServer
 {
     public class ConnectMsSqlCommand : ConnectKernelCommand<MsSqlKernelConnector>
     {
-        public ConnectMsSqlCommand()
+        internal string ResolvedToolsServicePath { get; set; }
```
colombod, line 13: as this is set via constructor, should this jsut be reaonly private field?

PR #1844, opened 2022-01-11
reusable stdio connector 

colombod:
stdio connector should be reused for multiple proxies
kernelHost needs to be able to track proxies created on connectors other than the default connector

### DIFF
--- a/src/Microsoft.DotNet.Interactive/Connection/StdIoKernelConnector.cs
+++ b/src/Microsoft.DotNet.Interactive/Connection/StdIoKernelConnector.cs
@@ -14,107 +14,106 @@ using System.Threading.Tasks;
 using Microsoft.DotNet.Interactive.Events;
 using Microsoft.DotNet.Interactive.Utility;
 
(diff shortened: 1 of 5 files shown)


src/Microsoft.DotNet.Interactive/Connection/StdIoKernelConnector.cs
```diff
@@ -14,15 +14,27 @@
...
+public class StdIoKernelConnector : IKernelConnector, IDisposable
 {
-   public class StdIoKernelConnector : IKernelConnector
-    {
-        public string[] Command { get; }
+    private MultiplexingKernelCommandAndEventReceiver? _receiver;
+    private KernelCommandAndEventTextStreamSender? _sender;
+    private Process? _process;
+
+    public string[] Command { get; }
```
jonsequitur, line 25: This is a somewhat confusing property name. I had to look at how it's used to understand what it meant, since the word "command" means something very different in the context of .NET Interactive's APIs.

jonsequitur (approved):

PR #1846, opened 2022-01-13
refactor directory shape of NPM projects

brettfo:
~~Marked as draft until the internal signed build has also been verified.~~
Internal build and generated artifacts were good.


colombod (approved):
Awesome work

PR #1887, opened 2022-02-03
removing lib outputs

colombod:


brettfo (approved):

jonsequitur:
Should we also be adding certain paths to `.gitignore` so this doesn't happen again?

PR #1890, opened 2022-02-04
report error if dotnet(.exe) couldn't be found

brettfo:
When the extension starts up we first call `dotnet --version` to determine if the SDK is up-to-date.  If it's not even there we used to just blow up, but now we can at least give the user a pointer.

screenshot

### DIFF
--- a/src/dotnet-interactive-vscode-common/src/extension.ts
+++ b/src/dotnet-interactive-vscode-common/src/extension.ts
@@ -84,12 +84,17 @@ export async function activate(context: vscode.ExtensionContext) {
     registerAcquisitionCommands(context, diagnosticsChannel);
 
     // check sdk version
-    const dotnetVersion = await getDotNetVersionOrThrow(DotNetPathManager.getDotNetPath(), diagnosticsChannel);
-    if (!isVersionSufficient(dotnetVersion, minDotNetSdkVersion)) {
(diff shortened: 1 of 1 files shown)


src/dotnet-interactive-vscode-common/src/extension.ts
```diff
@@ -84,12 +84,17 @@ export async function activate(context: vscode.ExtensionContext) {
...
+    try {
+        const dotnetVersion = await getDotNetVersionOrThrow(DotNetPathManager.getDotNetPath(), diagnosticsChannel);
+        if (!isVersionSufficient(dotnetVersion, minDotNetSdkVersion)) {
+            const message = `The .NET SDK version ${dotnetVersion} is not sufficient. The minimum required version is ${minDotNetSdkVersion}.`;
+            diagnosticsChannel.appendLine(message);
+            vscode.window.showErrorMessage(message);
+            throw new Error(message);
+        }
+    } catch (e) {
+        vscode.window.showErrorMessage(`Unable to determine the .NET SDK version.  Please download the ${minDotNetSdkVersion} SDK from https://dotnet.microsoft.com/en-us/download`);
```
jonsequitur, line 96: Does this happen when the SDK is missing? The error reads as something more specific than that, like maybe there are multiple versions and we're not sure which to use.

jonsequitur (approved):

PR #1908, opened 2022-02-25
add LanguageName and LanguageVersion to KernelInfo

jonsequitur:


samples/ExtensionLab/ExtensionLab-EF.ipynb
```diff
@@ -8,8 +8,59 @@
           "language": "csharp"
         }
       },
-      "outputs": [],
```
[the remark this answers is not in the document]
jonsequitur: I've been thinking we should check in outputs but it's not something we've done systematically. It's a bit of a pain from a PR standpoint but it's also more useful as documentation. Thoughts?

src/Microsoft.DotNet.Interactive.Formatting/DictionaryExtensions.cs
```diff
@@ -4,7 +4,7 @@
 using System;
 using System.Collections.Generic;
 
-namespace Microsoft.DotNet.Interactive.Formatting
```
colombod, line 7: some tools will complain about namespace not matching folders.

src/Microsoft.DotNet.Interactive.SqlServer/MsSqlKernel.cs
```diff
@@ -30,6 +30,8 @@ internal class MsSqlKernel : ToolsServiceKernel
             _connectionString = connectionString;
         }
 
+        public override string LanguageName => "MSSQL";
```
colombod, line 33: should we sue T-SQL as language name?

jonsequitur: Good call.

colombod (approved):

samples/ExtensionLab/ExtensionLab-EF.ipynb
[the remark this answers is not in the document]
jonsequitur: It will block rendering of the whole notebook?

src/Microsoft.DotNet.Interactive.Formatting/DictionaryExtensions.cs
jonsequitur: We can suppress that or use global namespace declarations for the shared files. We're not getting a warning in our CI so I'm not too concerned.

PR #1929, opened 2022-03-16
complete protocol objects

colombod:
Adding missing types and enforcing consistency


src/Microsoft.DotNet.Interactive.Jupyter.Tests/CompleteRequestHandlerTests.cs
```diff
@@ -27,7 +29,7 @@ public CompleteRequestHandlerTests(ITestOutputHelper output) : base(output)
         public async Task send_completeReply_on_CompleteRequest()
         {
             var scheduler = CreateScheduler();
-            var request = ZeroMQMessage.Create(new CompleteRequest("System.Console.", 15));
+            var request = ZeroMQMessage.Create(new CompleteRequest("System.Console.", 15), new Session());
```
jonsequitur, line 32: This boilerplate change in all of the callers suggests this parameter should be nullable and coalesce to a default.

PR #1932, opened 2022-03-16
set handler in kernel using custom interface

colombod:
When kernels implement custom command handler via the IKernelCommadnHandler interface it is enough to resolve the handler for the custom command


src/Microsoft.DotNet.Interactive.Tests/CompositeKernelTests.cs
```diff
@@ -525,6 +526,28 @@ public void When_child_kernel_is_added_then_its_frontend_environment_is_obtained
                 .BeSameAs(compositeKernel.FrontendEnvironment);
         }
 
+        [Fact]
+        public async Task When_kernel_implements_IKernelCommandHandler_for_a_custom_command_it_is_correctly_selected_has_handler()
```
jonsequitur, line 530: An interesting test case that I'm not sure if we have: What if there's an implemented `IKernelCommandHandler<Blah>` and then someone registers a custom handler for `Blah`? I think this should override the declared one. Do you know if this works?

PR #1933, opened 2022-03-18
refactor

colombod:
move csharpproject commands and events in the right project
ensure all commands and events are tested
ensure public  api contract tests check csharp project too


src/Microsoft.DotNet.Interactive/Server/KernelCommandEnvelope.cs
```diff
@@ -45,7 +45,7 @@ protected KernelCommandEnvelope(KernelCommand command)
 
         KernelCommand IKernelCommandEnvelope.Command => _command;
 
-        internal static void RegisterCommandTypeForSerialization<T>() where T : KernelCommand
```
colombod, line 48: This was a leftover from Ian pr, but no is useful (there is an equivalent counterpart on events) I am wondering if the registration of events / commands for serialization could be automated at write time. Yet this was useful in ensuring all commands and events were tested. we missed a couple :)

jonsequitur: How did we miss them, given the test that scans for them and makes sure they're all covered? Were they in other assemblies?

src/Microsoft.DotNet.Interactive.ApiCompatibility.Tests/ApiCompatibilityTests.cs
```diff
@@ -69,6 +70,13 @@ public void csharp_api_is_not_changed()
         this.Assent(contract, _configuration);
     }
 
+    [FactSkipLinux]
+    public void csharpProject_api_is_not_changed()
+    {
+        var contract = ApiContract.GenerateContract<CSharpProjectKernel>();
+        this.Assent(contract, _configuration);
```
jonsequitur, line 77: We need serialization contract tests in place for the commands and events from day 1 but otherwise I think this project is too new to both full API surface contract tests. A lot of these types might better be made internal. Let's review those later.

src/Microsoft.DotNet.Interactive.Tests/Server/SerializationTests.cs
```diff
@@ -134,11 +161,26 @@ public void All_command_types_are_tested_for_round_trip_serialization()
         [Fact]
         public void All_event_types_are_tested_for_round_trip_serialization()
         {
-            var eventTypes = typeof(KernelEvent)
-                             .Assembly
-                             .ExportedTypes
-                             .Concrete()
-                             .DerivedFrom(typeof(KernelEvent));
+
+            var interactiveEvents = typeof(Kernel)
```
jonsequitur, line 165: Don't we discover the others using reflection? We should do the same for these.

src/Microsoft.DotNet.Interactive.CSharpProject/ProjectFile.cs
```diff
@@ -0,0 +1,16 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.DotNet.Interactive.CSharpProject;
+
+public class ProjectFile
+{
+    public string Path { get; }
```
jonsequitur, line 8: Didn't we say we wanted to enforce this to be the `RelativeFilePath` type? An unvalidated `string` here will be a problem.

src/Microsoft.DotNet.Interactive.CSharpProject/Commands/CompileProject.cs
```diff
@@ -0,0 +1,16 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Microsoft.DotNet.Interactive.Commands;
+
+namespace Microsoft.DotNet.Interactive.CSharpProject.Commands;
+
+public class CompileProject : KernelCommand
+{
+    public string Code { get; }
```
jonsequitur, line 10: What is an example of this code? This command can be clearer in intent.

src/Microsoft.DotNet.Interactive.CSharpProject/CSharpProjectKernel.cs
```diff
@@ -1,109 +1,120 @@
...
+        // register commands and event with serialization
+        KernelCommandEnvelope.RegisterCommand<OpenProject>();
+        KernelCommandEnvelope.RegisterCommand<OpenDocument>();
+        KernelCommandEnvelope.RegisterCommand<CompileProject>();
+
+        KernelEventEnvelope.RegisterEvent<DocumentOpened>();
+        KernelEventEnvelope.RegisterEvent<AssemblyProduced>();
+    }
+}
+//public class CSharpProjectKernel :
```
jonsequitur, line 28: Can we delete this commented out code?

src/Microsoft.DotNet.Interactive.ApiCompatibility.Tests/ApiCompatibilityTests.cs
colombod: ok will take this part off

colombod: done

src/Microsoft.DotNet.Interactive.Tests/Server/SerializationTests.cs
colombod: what do you mean, they are discovered scanning the assemblies using reflection

src/Microsoft.DotNet.Interactive.CSharpProject/CSharpProjectKernel.cs
colombod: done

src/Microsoft.DotNet.Interactive.CSharpProject/ProjectFile.cs
```diff
@@ -1,16 +1,23 @@
...
 
+using System;
+using System.IO;
+
 namespace Microsoft.DotNet.Interactive.CSharpProject;
 
 public class ProjectFile
 {
-    public string Path { get; }
+    public string RelativePath { get; }
```
jonsequitur, line 11: Why not the `dotnet try` `RelativeFilePath` type?

src/Microsoft.DotNet.Interactive.CSharpProject/Commands/CompileProject.cs
colombod: not sure I understand. That code is anything user want to be compiled with the context of the project. I would argue that that command is not compileproject, is CompleCode. The fact that the project kernel is compiling that code against a project sounds like a n implementation detail of the proejct kernel.

Should this question be addressed in the pr @brettfo  is working on with the actual kernel implementation? This pr goal is to move things in their place. Your point is something I would like to address with @brettfo too

src/Microsoft.DotNet.Interactive.CSharpProject/ProjectFile.cs
colombod: is used via the constructor but exposed as string

src/Microsoft.DotNet.Interactive.CSharpProject/Commands/CompileProject.cs
jonsequitur: If it's more correctly `CompileCode`, what is the context for the code? Is it to be inlined somehow into an existing project or file? Is it a complete program?

src/Microsoft.DotNet.Interactive.ApiCompatibility.Tests/ApiCompatibilityTests.csharpProject_api_is_not_changed.approved.txt
```diff
@@ -0,0 +1,940 @@
+
+  public static class InstrumentationEmitter
```
jonsequitur, line 2: This file can be omitted for now.

jonsequitur (approved):

PR #1938, opened 2022-03-22
register new event with serialization

colombod:


src/Microsoft.DotNet.Interactive.CSharpProject/CSharpProjectKernel.cs
```diff
@@ -1,303 +1,318 @@
...
+namespace Microsoft.DotNet.Interactive.CSharpProject
+{
+    public class CSharpProjectKernel
+        : Kernel
+    {
+        private RoslynWorkspaceServer _workspaceServer;
+        private Protocol.Workspace _workspace;
+        private Protocol.Buffer _buffer;
+
+        static CSharpProjectKernel()
```
jonsequitur, line 27: I'd avoid doing this in a static constructor since these things can potentially be reset. The more explicit code is also more performant.

If there's a maintainability issue here let's fix the design bug that prevents `IKernelCommandHandler<T>` from working as expected.

src/Microsoft.DotNet.Interactive.CSharpProject/CSharpProjectKernel.cs
```diff
@@ -1,303 +1,318 @@
...
+                default:
+                    return CodeAnalysis.DiagnosticSeverity.Warning;
+            }
+        }
+
+        private void EnsureProjectOpened()
+        {
+            if (_workspaceServer is null || _workspace is null)
+            {
+                throw new Exception($"Project must be opened, send the command '{nameof(OpenProject)}' first.");
```
jonsequitur, line 295: We should never throw `Exception`.

The naming for these is a little unconventional. `Ensure*` usually means we're going to implicitly perform some prerequisite action. If you're going to throw, I would go with e.g. `ThrowIfProjectWasNotOpened`.

src/Microsoft.DotNet.Interactive.CSharpProject/CSharpProjectKernel.cs
```diff
@@ -1,303 +1,318 @@
...
+        }
+
+        private static async Task<Package> CreateConsoleWorkspacePackage()
+        {
+            var packageBuilder = new PackageBuilder("console");
+            packageBuilder.CreateUsingDotnet("console");
+            packageBuilder.TrySetLanguageVersion("8.0");
+            packageBuilder.AddPackageReference("Newtonsoft.Json", "13.0.1");
+            var package = packageBuilder.GetPackage() as Package;
+            await package.CreateRoslynWorkspaceForRunAsync(new Budget());
```
jonsequitur, line 314: We can get rid of the `Budget` stuff I believe.

brettfo (approved):

PR #1944, opened 2022-03-24
update-versions.ps1 should also update tool for ADS

brettfo:


brettfo:
Converted to a draft while we figure out the publish script in #1945.

colombod (approved):

PR #1967, opened 2022-04-14
Add quotes around GUIDs when importing them into SQL cells.

corivera:
Fixes https://github.com/dotnet/interactive/issues/1941


src/Microsoft.DotNet.Interactive.SqlServer/MsSqlKernel.cs
```diff
@@ -72,6 +72,7 @@ value switch
                 {
                     string s => $"N{s.AsSingleQuotedString()}",
                     char c => $"N{c.ToString().AsSingleQuotedString()}",
+                    Guid g => $"N{g.ToString().AsSingleQuotedString()}",
```
Charles-Gagnon, line 75: Add test to https://github.com/dotnet/interactive/blob/main/src/Microsoft.DotNet.Interactive.SqlServer.Tests/MsSqlConnectionTests.cs#L300

corivera: Done.

Charles-Gagnon (approved):

jonsequitur (approved):

PR #1980, opened 2022-04-21
fix extension samples

colombod:
updates sample notebooks


jonsequitur:
I think maybe we should leave the outputs in the notebook. It's more useful that way if people are viewing them online.

colombod:
> I think maybe we should leave the outputs in the notebook. It's more useful that way if people are viewing them online.

Done

jonsequitur (approved):

PR #1989, opened 2022-04-26
add variable explorer

brettfo:
Adds a new pane at the bottom of VS Code titled `.NET Interactive: Values`.

When top-level variables are defined in a kernel, they'll appear in the variable explorer:
image

The `Share to...` link asks the user the kernel they'd like to receive the value:
image

When the target kernel is selected, a new cell will be added to the notebook with the variable sharing magic command pre-filled:
image

And values can be shared both into and out of the JavaScript kernel:
image


colombod (commented):
can we upgrade the table look ???

brettfo:
Short-term I'd like to get this merged so it'll be easier to test the table style updates; that way the PR won't contain any functionality changes, so it'll be easier to review and test.

src/Microsoft.DotNet.Interactive.CSharp.Tests/CSharpKernelTests.cs
```diff
@@ -38,5 +38,19 @@ public async Task Script_state_is_available_within_middleware_pipeline()
             variableCountBeforeEvaluation.Should().Be(0);
             variableCountAfterEvaluation.Should().Be(1);
         }
+
+        [Fact]
+        public async Task GetValueInfos_only_returns_newly_bindable_objects()
```
jonsequitur, line 43: The name is a little confusing. Does this mean it does not return variables that have been shadowed?

brettfo: Yes, that is correct.  I'll update the name.

src/Microsoft.DotNet.Interactive.PowerShell.Tests/PowerShellKernelTests.cs
```diff
@@ -240,5 +240,26 @@ public async Task Powershell_customobject_is_formatted_for_outdisplay()
                 e => e.Should().BeOfType<CommandSucceeded>()
             );
         }
+
+        [Fact]
+        public async Task RequestValueInfos_only_returns_user_defined_values()
```
jonsequitur, line 245: Can we explicitly request a non-user defined value?

brettfo: Yes, it can be requested, but it's not reported by default, otherwise the variable explorer is flooded with crap.

brettfo: E.g., PowerShell has a bunch of pre-defined values like `CurrentEncoding`, `True`, `False`, etc.

colombod (approved):

PR #2002, opened 2022-05-10
Await clearOutput promise when starting cell execution.

corivera:
I was doing some testing with Run All Cells in Azure Data Studio, and occasionally when running an Interactive notebook I would see that cell outputs wouldn't get updated when the cell execution completed. When I dug into it I saw that the cell's output was getting cleared after the cell execution had completed. Looks like the promise for the clearOutput call in DotNetNotebookKenel.executeCell wasn't being awaited, so a very fast code cell (e.g. immediately returning a number) could finish before that promise resolved.

### DIFF
--- a/src/dotnet-interactive-vscode-ads/src/notebookControllers.ts
+++ b/src/dotnet-interactive-vscode-ads/src/notebookControllers.ts
@@ -139,7 +139,7 @@ export class DotNetNotebookKernel {
             try {
                 const startTime = Date.now();
                 executionTask.start(startTime);
-                executionTask.clearOutput(cell);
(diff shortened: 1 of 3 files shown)


jonsequitur:
Is there a good way to test for this?

corivera:
Maybe, but I'm not sure how consistent the testing would be. I only repro'd this issue after running the same notebook several times, and the cells with missed outputs varied each time.

brettfo:
There's not a way to test this other than manually, but the fix looks like it's correct.

brettfo (approved):

PR #2003, opened 2022-05-10
complete on dispose

colombod:


jonsequitur (approved):

src/microsoft-dotnet-interactive/src/kernel.ts
```diff
@@ -27,10 +28,26 @@ export class Kernel {
     private readonly _tokenGenerator: TokenGenerator = new TokenGenerator();
     public rootKernel: Kernel = this;
     public parentKernel: CompositeKernel | null = null;
+    private _scheduler?: KernelCommandScheduler | null = null;
 
     constructor(readonly name: string) {
     }
 
+    private getScheduler(): KernelCommandScheduler {
```
colombod, line 36: @jonsequitur i suspect we need to replicate the fast track behaviour too here

PR #2015, opened 2022-05-13
Integrate Azure Data Studio connection dialog with Interactive input tokens

corivera:
I added a new input token, "connectionString", which causes the connection dialog to be opened when run in an Azure Data Studio context. A connection string for the resulting connection is then returned as the input result. If running in VS Code, then this token resolves to normal input. I also added an error message ("No input was provided.") when receiving empty or undefined input from the user, since it was causing obscure errors when passing that value to the underlying SQL kernel. It makes sense that we fail whatever operation is running if the user doesn't provide any input for it. So far this works for the SQL kernel, but I need to do some testing with Kusto.

Edit: The Kusto kernel doesn't support a connection string option currently, so no impact there.


src/Microsoft.DotNet.Interactive/Commands/RequestInput.cs
```diff
@@ -5,14 +5,17 @@ namespace Microsoft.DotNet.Interactive.Commands;
 
 public class RequestInput : KernelCommand
 {
-    public RequestInput(string prompt, bool isPassword = false, string targetKernelName = null)
+    public RequestInput(string prompt, bool isPassword = false, bool isConnectionString = false, string targetKernelName = null)
```
Charles-Gagnon, line 8: Do you think you'll have other types of input? (bool, int, enum, etc.) If so may want to look into having this all consolidated into an `inputType` param instead

src/dotnet-interactive-vscode-ads/src/versionSpecificFunctions.ts
```diff
@@ -3,6 +3,7 @@
 
 import * as path from 'path';
 import * as vscode from 'vscode';
+import * as azdata from 'azdata';
```
Charles-Gagnon, line 6: Is this file only included in the ADS version of the extension or something?

src/dotnet-interactive-vscode-common/src/extension.ts
```diff
@@ -139,14 +139,30 @@ export async function activate(context: vscode.ExtensionContext) {
...
+                if (isAzureDataStudio(context) && isConnectionString) {
+                    value = await versionSpecificFunctions.getSqlConnectionString();
+                } else {
+                    value = await vscode.window.showInputBox({ prompt, password });
+                }
+                if (!value) {
+                    commandInvocation.context.publish({
+                        eventType: contracts.ErrorProducedType,
+                        event: {
+                            message: 'No input was provided.'
```
Charles-Gagnon, line 153: localize?

src/dotnet-interactive-vscode-ads/src/versionSpecificFunctions.ts
corivera: Yes, every extension has a file like this. The common code references this file, so they all need to have the same function signatures.

src/dotnet-interactive-vscode-common/src/extension.ts
corivera: I'm not seeing localization for any other strings in the extension.

jonsequitur: We aren't yet localizable.

src/Microsoft.DotNet.Interactive/Commands/RequestInput.cs
```diff
@@ -3,16 +3,25 @@
 
 namespace Microsoft.DotNet.Interactive.Commands;
 
+public enum InputType
```
jonsequitur, line 6: We shouldn't introduce an enum for this.

src/dotnet-interactive-vscode-common/src/extension.ts
```diff
@@ -139,14 +139,29 @@ export async function activate(context: vscode.ExtensionContext) {
...
-                const value = await vscode.window.showInputBox({ prompt, password });
-                commandInvocation.context.publish({
-                    eventType: contracts.InputProducedType,
-                    event: {
-                        value
-                    },
-                    command: commandInvocation.commandEnvelope,
-                });
+                let value: string | undefined;
+                if (isAzureDataStudio(context) && requestInput.inputType === contracts.InputType.ConnectionString) {
```
jonsequitur, line 143: ADS-specific logic should not be in common.

src/Microsoft.DotNet.Interactive/Commands/RequestInput.cs
corivera: Would you prefer a new input token for connection strings e.g. "@connectionString:myConnectionString", or to use the prompt part after the existing input token e.g. "@input:connectionString"?

corivera: Would you prefer a new token for connection strings e.g. "@ connectionString:myConnectionString", or to use the prompt part after the existing input token e.g. "@ input:connectionString"?

jonsequitur: The goal is for `@input:x` to be able to work in different frontends, whether a REPL, automation, VS Code, Jupyter, or ADS. ADS currently has a specific experience for connection strings but this should be able to fall back gracefully rather than block people in other environments. 

With that in mind, we also need a way to differentiate between a text input prompt and a connection chooser in ADS. This seems more like we need a way to hint at a preference rather than say `@input` and `@connectionString` are fundamentally different.

corivera: Switched to using "@input:connectionString"

src/Microsoft.DotNet.Interactive/Kernel.Static.cs
```diff
@@ -35,19 +35,19 @@ public partial class Kernel
         /// </summary>
         /// <param name="prompt">The prompt to show.</param>
         /// <returns>The user input value.</returns>
-        public static async Task<string> GetInputAsync(string prompt = "")
+        public static async Task<string> GetInputAsync(string inputName = "", string prompt = "")
```
jonsequitur, line 38: This is a breaking change. Currently, we don't require an input name, so it seems correct for this to be optional, but I also think the intent of this parameter is unclear. 

Yes, it's metadata that the UI can interpret in order to provide more richer interactions if supported. Any suggestions for naming that helps make that more understandable?

jonsequitur: I wonder if `inputType` would be more appropriate to the use case. I think each frontend will use this to do a sort of negotiation falling back to the most appropriate UX that it supports.

corivera: Fixed this yesterday to to have the new property be the second argument.

corivera:
Looks like that last test failure is from a host process crash, rather than an actual test failure.

src/Microsoft.DotNet.Interactive/Kernel.Static.cs
corivera: @jonsequitur changed the new field to `inputType`. Any other comments?

src/Microsoft.DotNet.Interactive/Parsing/SubmissionParser.cs
```diff
@@ -358,19 +358,26 @@ internal Parser GetDirectiveParser()
 
             if (targetKernelName is "input" or "password")
             {
-                var inputRequest = new RequestInput($"Please enter a value for field \"{valueName}\".", isPassword: targetKernelName == "password");
+                var inputRequest = new RequestInput($"Please enter a value for field \"{valueName}\".", isPassword: targetKernelName == "password", null, valueName);
```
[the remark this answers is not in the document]
corivera: We still need to pass it to the front end to check what the input type is. `valueName` is the "connectionString" part of "@ input:connectionString"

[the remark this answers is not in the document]
corivera: `inputType `is what eventually gets checked to determine if we should open the connection dialog for an input request: https://github.com/dotnet/interactive/pull/2015/files#diff-606ee82b02ae83922defbbedbb1a85a4da2cd4a926472bb4b5322b1ab2cab711R35

PR #2030, opened 2022-05-25
powershell input gestures

colombod:


src/Microsoft.DotNet.Interactive.VSCode/VSCodeClientKernelExtension.cs
```diff
@@ -2,35 +2,55 @@
...
-                                     new[] { "frontend" });
-                hostKernel.KernelInfo.SupportedKernelCommands.Add(new(nameof(RequestInput)));
-                root.SetDefaultTargetKernelNameForCommand(typeof(RequestInput), "vscode");
-                hostKernel.KernelInfo.SupportedKernelCommands.Add(new(nameof(SendEditableCode)));
+                if (subkernel is PowerShell.PowerShellKernel powerShellKernel)
+                {
+                    powerShellKernel.ReadInput = (prompt) =>
+                    {
+                        var result = Kernel.Root.SendAsync(new RequestInput(prompt)).GetAwaiter().GetResult();
+                        return result.KernelEvents.ToEnumerable().OfType<InputProduced>().FirstOrDefault()?.Value;
```
jonsequitur, line 44: I wonder if there's a simpler / safer pattern for this.

We had also discussed `Try*` type methods for getting user input. The user canceling should be a different result than, say, the user entering `""`, and using nullness to convey that distinction is something we should avoid.

colombod:
closing because @brettfo got better ideas

brettfo (approved):

Issue #1927, opened 2022-03-15
Extension loading requires to have matching versions between dependencies and runtime


colombod:
When loading extensions there are failures when version used at build time and runtime do not match

hmendozaDBA:
Does this could explain, why my extensions are not being currently loaded?

jonsequitur:
We can break this down into a few different tasks:

- [ ] #2053
- [ ] Allow option to load proxy kernels in a separate procress or assembly load context
- [ ] Enable a magic command extension point that does not depend on Microsoft.DotNet.Interactive

PR #2053, opened 2022-06-07
add support for convention based formatters

jonsequitur:
This addresses #1927 for the case of formatters. Using this approach, formatters can be registered from libraries that have no reference to Microsoft.DotNet.Interactive.Formatting, and therefore won't have assembly version conflicts with the version of  Microsoft.DotNet.Interactive.Formatting that's packaged into `dotnet-interactive`.


colombod (approved):
Nice design as usual.

PR #2077, opened 2022-06-15
implement kernel info handling in typescript

colombod:


src/microsoft-dotnet-interactive/src/javascriptKernel.ts
```diff
@@ -10,7 +10,7 @@ export class JavascriptKernel extends kernel.Kernel {
     private suppressedLocals: Set<string>;
 
     constructor(name?: string) {
-        super(name ?? "javascript");
+        super(name ?? "javascript", "javascript");
```
jonsequitur, line 13:
```suggestion
        super(name ?? "javascript", "JavaScript");
```

PR #2089, opened 2022-06-22
Adds js implementation of the html kernel

colombod:


src/microsoft-dotnet-interactive/src/htmlKernel.ts
```diff
@@ -0,0 +1,82 @@
...
+            if (mutation.type === "childList") {
+
+                const nodes = Array.from(mutation.addedNodes);
+                for (const addedNode of nodes) {
+                    const element = addedNode as HTMLDivElement;
+                    element.id;//?
+                    container.id;//?
+                    if (element?.id === container.id) {//?
+                        completionPromise.resolve();
+                        completionPromise.resolve();
```
jonsequitur, line 67: Why twice?

colombod: by mistake

src/microsoft-dotnet-interactive/src/javascriptKernel.ts
```diff
@@ -74,10 +74,18 @@ export class JavascriptKernel extends kernel.Kernel {
 
     private allLocalVariableNames(): string[] {
         const result: string[] = [];
-        for (const key in globalThis) {
-            if (typeof (<any>globalThis)[key] !== 'function') {
-                result.push(key);
+        try {
```
colombod, line 77: @brettfo I added this because we are getting dom security exceptions during this scanning and they break the constructor.

This is the kind of error we get in headless js kernel 
https://github.com/microsoft/playwright/issues/9934

PR #2102, opened 2022-06-23
Browser package

jonsequitur:
This introduces the basics for executing JavaScript when running .NET Interactive without a UI.


src/Microsoft.DotNet.Interactive.Browser/Microsoft.DotNet.Interactive.Browser.csproj
```diff
@@ -0,0 +1,36 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net6.0</TargetFramework>
```
brettfo, line 4: Why not `netstandard2.1`?

src/Microsoft.DotNet.Interactive.Browser/PlaywrightKernelConnector.cs
```diff
@@ -0,0 +1,132 @@
...
+
+namespace Microsoft.DotNet.Interactive.Browser;
+
+public class PlaywrightKernelConnector : IKernelConnector
+{
+    public Task<Kernel> CreateKernelAsync(string kernelName)
+    {
+        var disposables = new CompositeDisposable();
+
+        var browserChannel = "msedge";
```
brettfo, line 23: What does this do on Mac/Linux?

src/microsoft-dotnet-interactive/lib/compositeKernel.d.ts
```diff
@@ -0,0 +1,19 @@
+import * as contracts from "./contracts";
```
brettfo, line 1: Rebase so this get's excluded.  Same for all other files in this directory.

brettfo (changes requested):

src/Microsoft.DotNet.Interactive.Browser/PlaywrightKernelConnector.cs
jonsequitur: We'll make this configurable later.

src/Microsoft.DotNet.Interactive.Browser/Microsoft.DotNet.Interactive.Browser.csproj
jonsequitur: I'm not sure it's useful. The functionality here is never going to work in WASM, which is the primary reason we  support `netstandard2.1` in some packages.

colombod (changes requested):

PR #2105, opened 2022-06-24
remove http start mode from cli

colombod:


AbhitejJohn:
Could this be a good candidate to add to the changelog?

jonsequitur:
It will be generated automatically when we create a release.

jonsequitur (approved):

PR #2113, opened 2022-06-28
Upgrade packages

colombod:


NuGet.config
```diff
@@ -11,6 +11,7 @@
     <add key="dotnet-libraries" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-libraries/nuget/v3/index.json" />
     <add key="dotnet5" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet5/nuget/v3/index.json" />
     <add key="MachineLearning" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/MachineLearning/nuget/v3/index.json" />
+    <add key="nuget org" value="Https://api.nuget.org/v3/index.json" />
```
jonsequitur, line 14:
```suggestion
    <add key="nuget.org" value="Https://api.nuget.org/v3/index.json" />
```

Is a `.` not allowed in the key?

colombod: I actually need to delete that source ! Just getting the package versions that work, not all of them are on internal feeds

PR #2123, opened 2022-07-01
document model API improvements

jonsequitur:
This PR updates the Microsoft.DotNet.Interactive.Documents API to more closely follow the `.ipynb` spec and observed behavior in Jupyter Lab. It also makes more concepts explicit in the API and makes it mutable so that creating documents incrementally is easier.

This addresses #2148.


src/Microsoft.DotNet.Interactive/Server/Serializer.cs
```diff
@@ -23,10 +21,8 @@ static Serializer()
             NumberHandling = JsonNumberHandling.AllowReadingFromString | JsonNumberHandling.AllowNamedFloatingPointLiterals,
             Encoder = JavaScriptEncoder.UnsafeRelaxedJsonEscaping
         };
-        JsonSerializerOptions.Converters.Add(new ByteArrayConverter());
```
brettfo, line 26: We need to keep this.  When a notebook is saved it's sent back as a byte array, but serialized as a base64 string.

PR #2169, opened 2022-07-26
force package upgrades to take care of component governance alerts

brettfo:


Directory.Build.targets
```diff
@@ -8,12 +8,20 @@
...
     <PackageReference Include="xunit" Version="2.4.1" />
     <PackageReference Include="xunit.runner.visualstudio" Version="2.4.3" />
   </ItemGroup>
 
+  <!--
+  Microsoft.NET.Test.Sdk pulls in an older version of Newtonsoft.Json that _must_ be updated to 13.0.1 to address
+  component goverance alerts.
+  -->
+  <ItemGroup>
+    <PackageReference Update="Microsoft.NET.Test.Sdk" Version="$(MicrosoftNETTestSdkVersion)" />
```
AbhitejJohn, line 22: So would adding a PackageReference Update to Newtonsoft.Json in this central targets save us from including it in all the test projects?

AbhitejJohn (approved):

Directory.Build.targets
brettfo: We don't want this package in our shipping binaries and while we could do some fancy filtering, it seemed easier just to explicitly set it in only the few places we need it.

AbhitejJohn: Oh, wouldn't that mean the Microsoft.Net.Test.Sdk addition here would be included in shipped binaries then?

brettfo: Not in this case; I'm using `Update=` instead of `Include=`.

PR #2173, opened 2022-07-27
Use dotnet tool for SQL Tools and Kusto services

corivera:
Currently our SQL Tools Service nuget package takes up almost 2 GB of space due to having to download all the platform runtimes. With these dotnet tool versions the space usage is reduced to less than 100 MB in total. I opted to install them as global tools so we don't need to do any path checking.

I'm currently trying to publish the new builds to nuget.org, so I'll mark this PR as a draft for now so it doesn't get merged by mistake.


src/Microsoft.DotNet.Interactive/Utility/Dotnet.cs
```diff
@@ -86,8 +86,17 @@ public Task<CommandLineResult> AddReference(FileInfo projectToReference, TimeSpa
 
             var versionArg = version is not null ? $"--version {version}" : "";
             
-            var args = $@"{packageName} --tool-path ""{toolPath.FullName.TrimTrailingSeparators()}"" {versionArg}";
-            
+            var args = $@"{packageName}";
+            if (toolPath != null)
```
colombod, line 90:
```suggestion
            if (toolPath is not null)
```

src/Microsoft.DotNet.Interactive.SqlServer/MsSqlKernelExtension.cs
```diff
@@ -3,54 +3,40 @@
...
                 {
-                    var pathToService = root.PathToToolsService("MicrosoftSqlToolsServiceLayer");
-
-                    if (!string.IsNullOrEmpty(pathToService))
-                    {
-                        if (File.Exists(pathToService))
-                        {
-                            compositeKernel
-                            .AddKernelConnector(new ConnectMsSqlCommand(pathToService));
+                    var commandLineResult = await dotnet.ToolInstall("Microsoft.SqlServer.SqlToolsServiceLayer.Tool");
```
colombod, line 27: is it safe to always install latest version available? What if there are protocol breaking changes?

src/Microsoft.DotNet.Interactive.Kql/KqlKernelExtension.cs
```diff
@@ -3,56 +3,41 @@
...
-                if (root != null)
+                // Check if the required Sql Tools Service tool is installed, and then install it if necessary
+                var dotnet = new Dotnet();
+                var installedGlobalTools = await dotnet.ToolList();
+                const string kqlToolName = "MicrosoftKustoServiceLayer";
+                bool kqlToolInstalled = installedGlobalTools.Any(tool => string.Equals(tool, kqlToolName, StringComparison.InvariantCultureIgnoreCase));
+                if (!kqlToolInstalled)
                 {
-                    var pathToService = root.PathToToolsService("MicrosoftKustoServiceLayer");
+                    var commandLineResult = await dotnet.ToolInstall("Microsoft.SqlServer.KustoServiceLayer.Tool");
```
colombod, line 28: same question here about always installing latest

src/Microsoft.DotNet.Interactive.SqlServer/MsSqlKernelExtension.cs
corivera: We try to maintain backwards compatibility with existing APIs.

colombod: but this is always downloading latest version of the tool, does not protect the user from getting in a broken state when changes like that gets in without also releasing the .NET Interactive extension immediately

corivera: Added a specific version number when installing.

corivera:
@colombod Looks like those CI test failures are unrelated to these changes. Are we good to merge this PR?

colombod:
> @colombod Looks like those CI test failures are unrelated to these changes. Are we good to merge this PR?

running the failied part now

src/Microsoft.DotNet.Interactive.Kql/KqlKernelExtension.cs
```diff
@@ -3,56 +3,41 @@
...
-                if (root != null)
+                // Check if the required Sql Tools Service tool is installed, and then install it if necessary
+                var dotnet = new Dotnet();
+                var installedGlobalTools = await dotnet.ToolList();
+                const string kqlToolName = "MicrosoftKustoServiceLayer";
+                bool kqlToolInstalled = installedGlobalTools.Any(tool => string.Equals(tool, kqlToolName, StringComparison.InvariantCultureIgnoreCase));
+                if (!kqlToolInstalled)
                 {
-                    var pathToService = root.PathToToolsService("MicrosoftKustoServiceLayer");
+                    var commandLineResult = await dotnet.ToolInstall("Microsoft.SqlServer.KustoServiceLayer.Tool", null, null, "1.0.0");
```
jonsequitur, line 28: It seems like we shouldn't hard code the version. How often do we expect this to be updated?

corivera: It doesn't have a dedicated release cadence, so any time we have any useful improvements, basically. Diego wanted a fixed version here in case of breaking changes in the future.

jonsequitur (approved):

PR #2185, opened 2022-08-03
Routing command from javascript to other kernels

colombod:


src/Microsoft.DotNet.Interactive/Commands/KernelCommand.cs
```diff
@@ -5,60 +5,81 @@
...
-        [JsonIgnore] public RoutingSlip RoutingSlip { get; }
+    [JsonIgnore]
+    public IReadOnlyCollection<Uri> RoutingSlip
+    {
+        get
+        {
+            var unique = new HashSet<Uri>();
+            var list = new List<Uri>(Parent?.RoutingSlip ?? Array.Empty<Uri>());
+            list.AddRange(_routingSlip.Where(unique.Add));
+            return list;
```
jonsequitur, line 63: Properties shouldn't return a new instance on each call. Is there a reason why this state isn't being built up only once?

src/Microsoft.DotNet.Interactive/Commands/KernelCommand.cs
```diff
@@ -5,60 +5,81 @@
...
+    internal bool? ShouldPublishCompletionEvent { get; set; }
 
-        [JsonIgnore] 
-        public ParseResult KernelChooserParseResult { get; internal set; }
+    [JsonIgnore] 
+    public ParseResult KernelChooserParseResult { get; internal set; }
 
-        [JsonIgnore] public RoutingSlip RoutingSlip { get; }
+    [JsonIgnore]
+    public IReadOnlyCollection<Uri> RoutingSlip
```
jonsequitur, line 56: This should be a `RoutingSlip`.

src/Microsoft.DotNet.Interactive/Kernel.cs
```diff
@@ -405,18 +405,21 @@ public void RegisterCommandType<TCommand>()
...
                         default:
-                            await Scheduler.RunAsync(
-                                               c,
-                                               InvokePipelineAndCommandHandler,
-                                               c.SchedulingScope.ToString(),
-                                               cancellationToken: cancellationToken)
-                                           .ContinueWith(t =>
-                                           {
-                                               if (t.IsCanceled)
+                            if (!context.IsFailed)
```
jonsequitur, line 408: Why is this check specific to the `default` case?

src/Microsoft.DotNet.Interactive/Kernel.cs
```diff
@@ -529,7 +532,24 @@ internal async Task<KernelCommandResult> InvokePipelineAndCommandHandler(KernelC
...
+                            {
+                                return false;
+                            }
+
+                            if (outer == inner.Parent)
+                            {
+                                return true;
+                            }
+
+                            var isInRoutingSlip = outer.RoutingSlip.Zip(inner.RoutingSlip, (o, i) => o.Equals(i)).All(x => x);
```
jonsequitur, line 548: This logic looks like a good candidate to move into `RoutingSlip`.

src/Microsoft.DotNet.Interactive/KernelInvocationContext.cs
```diff
@@ -62,14 +62,18 @@ private KernelInvocationContext(KernelCommand command)
...
-        
+
             _disposables.Add(operation);
         }
 
         public KernelCommand Command { get; }
 
         public bool IsComplete { get; private set; }
 
+        private bool _isFailed;
```
jonsequitur, line 73: Nitpick: Move field declarations to the top of the class.

src/Microsoft.DotNet.Interactive/KernelScheduler.cs
```diff
@@ -115,32 +117,32 @@ private void ScheduledOperationRunLoop(object _)
...
+                    _currentTopLevelOperation = default;
                     _currentlyRunningOperation = null;
                 }
             }
         }
 
         private void Run(ScheduledOperation operation)
         {
-            _currentTopLevelOperation.Value ??= operation;
+            _currentTopLevelOperation = _currentTopLevelOperation is null? operation.Value : _currentTopLevelOperation;
```
jonsequitur, line 128: There's a potential race condition here.

```suggestion
            _currentTopLevelOperation = _currentTopLevelOperation ?? operation.Value;
```

src/Microsoft.DotNet.Interactive/RoutingSlip.cs
```diff
@@ -38,9 +38,9 @@ IEnumerator<Uri> IEnumerable<Uri>.GetEnumerator()
 
     IEnumerator IEnumerable.GetEnumerator()
     {
-        return _uris.GetEnumerator();
+        return ((IEnumerable)_uris).GetEnumerator();
```
jonsequitur, line 47: This should not be necessary.

src/Microsoft.DotNet.Interactive/RoutingSlip.cs
```diff
@@ -38,9 +38,9 @@ IEnumerator<Uri> IEnumerable<Uri>.GetEnumerator()
...
     }
 
-    public Uri this[int index] => _uris[index];
+    public Uri this[int i] => _uris[i];
```
jonsequitur, line 50: I think `index` is preferred.

src/Microsoft.DotNet.Interactive/Commands/KernelCommand.cs
colombod: changed that to be built when parenting is changed

src/Microsoft.DotNet.Interactive/Commands/KernelCommand.cs
colombod: Done

src/Microsoft.DotNet.Interactive/Kernel.cs
colombod: Correct! Done

src/Microsoft.DotNet.Interactive/KernelInvocationContext.cs
colombod: Done

src/Microsoft.DotNet.Interactive/RoutingSlip.cs
colombod: done

src/Microsoft.DotNet.Interactive/RoutingSlip.cs
colombod: done

brettfo (approved):

PR #2214, opened 2022-08-21
Update the formatting of the _style attribute's value

bmitc:
The `_style` attribute function already took in a list of strings, however it wasn't clear how the strings in the list should be formatted. I found that something like

```fsharp
_style [
    "width: 3em";
    $"background: {colorString}";
    "display: inline-block";
    "border: 3px solid black";
]
```

did not work as expected, while

```fsharp
_style [
    "width: 3em;";
    $"background: {colorString};";
    "display: inline-block;";
    "border: 3px solid black;";
]
```

worked.

I think that since `_style` takes in a list of strings, it should automatically handle the formatting. If it only took a string, then the formatting should be explicitly given of course.

The change I made is to have the `_style` function automatically normalize the formatting of the strings with regards to handling the `;`. The change should be backwards compatible with existing uses of `_style`, and I have added documentation and a test.

### DIFF
--- a/src/Microsoft.DotNet.Interactive.FSharp.Tests/HtmlFormatterTests.fs
+++ b/src/Microsoft.DotNet.Interactive.FSharp.Tests/HtmlFormatterTests.fs
@@ -4,7 +4,6 @@
 namespace Microsoft.DotNet.Interactive.FSharp.Tests
 
 open Microsoft.DotNet.Interactive
-open Microsoft.DotNet.Interactive.Formatting
 open Microsoft.DotNet.Interactive.FSharp.FSharpKernelHelpers.Html
 open Xunit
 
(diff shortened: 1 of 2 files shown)


src/Microsoft.DotNet.Interactive.FSharp.Tests/HtmlFormatterTests.fs
```diff
@@ -4,7 +4,6 @@
 namespace Microsoft.DotNet.Interactive.FSharp.Tests
 
 open Microsoft.DotNet.Interactive
-open Microsoft.DotNet.Interactive.Formatting
```
bmitc, line 7: This was not being used.

jonsequitur (approved):

bmitc:
Thanks for the approval and merge @jonsequitur!

jonsequitur:
Thanks for your help, @bmitc!

bmitc:
No problem!

PR #2234, opened 2022-08-27
Don't share by ref when using `#!share --mime-type`

jonsequitur:


colombod (approved):

colombod:
I wonder if value should be set according to mime type. If we share as image/png should be return the raw image as byte array in the value field?

jonsequitur:
I think that's going to be very unintuitive. You really won't know what you're getting from one case to another.

PR #2241, opened 2022-08-31
add support for RequestInput type hints and a file picker

jonsequitur:
This adds support for `RequestInput` commands to carry type hints based on the magic command parser's type expectations. One example is implemented here, which is triggering a file dialog when the expected type is `FileSystemInfo`:

image

### DIFF
--- a/src/Microsoft.DotNet.Interactive/Parsing/SubmissionParser.cs
+++ b/src/Microsoft.DotNet.Interactive/Parsing/SubmissionParser.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Generic;
 using System.CommandLine;
+using System.CommandLine.Binding;
 using System.CommandLine.Builder;
 using System.CommandLine.Parsing;
 using System.IO;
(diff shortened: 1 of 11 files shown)


colombod (approved):

src/Microsoft.DotNet.Interactive/Parsing/SubmissionParser.cs
```diff
@@ -476,6 +511,17 @@ static bool ContainsInvalidCharactersForValueReference(ReadOnlySpan<char> tokenT
             }
         }
 
+        private static string GetTypeHint(Symbol symbol) =>
+            symbol switch
+            {
+                IValueDescriptor<DateTime> => "datetime-local",
+                IValueDescriptor<int> => "number",
```
AbhitejJohn, line 518: Curious, how did we populate this list and how do we keep it updated with new types? I'm sure it's not the end of the world if this doesn't cover everything as well.

AbhitejJohn (approved):

src/Microsoft.DotNet.Interactive/Parsing/SubmissionParser.cs
jonsequitur: #2245 makes this mechanism extensible.

Issue #2190, opened 2022-08-05
HTML display for JSON is not left-aligned


jonsequitur:
<img width="906" alt="image" src="https://user-images.githubusercontent.com/547415/183126878-8ed4d335-3c8d-4790-b364-6a03325d657d.png">

PR #2250, opened 2022-09-05
Setting text in JSON formatter to align to the right instead of left

jasiozet:
Fixes: https://github.com/dotnet/interactive/issues/2190

How it looks after the fix:
image
How it have looked before:
image

Json file used: https://jsonplaceholder.typicode.com/posts/1/comments

Let me know if I need to do anything more


jonsequitur:
Thanks, @jasiozet!

jasiozet:
My pleasure, I really like these notebooks :)

PR #2259, opened 2022-09-14
.dib metadata support

### LINKED ISSUE #2184: Add `#!metadata` kernel
# Current state of the world

The addition of a metadata kernel will allow `.dib` notebooks to have cell specifiers separate from the well-known ones of `#!csharp`, `#!fsharp`, etc.  E.g., if one cell of a notebook connects to a SQL database with:

```
#!connect mssql --kernel-name adventureworks "<my connection string>"
```

then the user adds their queries, the current resultant .dib will look like this:

```
#!csharp

#!connect mssql --kernel-name adventureworks "<my connection string>"

#!csharp

#!sql-adventureworks
SELECT * FROM CUSTOMERS
```

This gives an odd view to the user because the query cell isn't _really_ C# (or SQL, or F#, ...), and the user specified that they wanted the new kernel to be named `adventureworks`, not `sql-adventureworks`.  The ideal would be something like this:

```
#!csharp

#!connect mssql --kernel-name adventureworks "<my connection string>"

#!adventureworks
SELECT * FROM CUSTOMERS
```

however, it's not known at parse time that `#!adventureworks` is a kernel selector, so when loaded, the notebook would have a single cell that looked like this:

```
#!connect mssql --kernel-name adventureworks "<my connection string>"

#!adventureworks
SELECT * FROM CUSTOMERS
```

# Proposal

Add a kernel named `#!metadata` which at runtime is a no-op, but at parse time can specify various bits of document metadata, possibly expected to be valid JSON.  The example above would result in something like this:

```
#!metadata
{
  "kernels": [
    {
      "name": "csharp",
      "language": "csharp"
    },
    {
      "name": "fsharp",
      "language": "fsharp"
    },
    {
      "name": "adventureworks",
      "language": "sql"
    },
  ]
}

#!csharp

#!connect mssql --kernel-name adventureworks "<my connection string>"

#!adventureworks
SELECT * FROM CUSTOMERS
```

The parser knows about the special `#!metadata` kernel, parses its contents, and it suddently knows that further down the file that `#!adventureworks` is a cell splitter and its language is SQL.

jonsequitur:
This fixes #2184.


src/Microsoft.DotNet.Interactive.Documents.Tests/CodeSubmissionFormatTests.dib_file_can_be_round_tripped_through_read_and_write_without_the_content_changing.approved.dib
```diff
@@ -0,0 +1,27 @@
+#!meta
+
+{"kernelInfo":{"defaultKernelName":"csharp","items":[{"aliases":["cs","C#","c#"],"name":"csharp"},{"aliases":["fs","F#","f#"],"name":"fsharp"},{"aliases":["powershell"],"name":"pwsh"},{"name":"mermaid"},{"name":"javascript"}]}}
```
brettfo, line 3: Can we format/indent this?  It's kind-of meaningful to human eyes and if/when this changes, the diff will be more manageable.

src/Microsoft.DotNet.Interactive.Documents/InteractiveDocumentElement.cs
```diff
@@ -29,7 +29,10 @@ public InteractiveDocumentElement()
         [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
         public string? Id { get; set; }
 
-        public string? Language { get; set; }
+        public string? KernelName { get; set; }
+
+        // FIX: (InteractiveDocumentElement) add this back
```
brettfo, line 34: Language won't need to be added back; it can be looked up via the metadata.

brettfo:
The bottom of `NotebookTestScript.dib` has a dib-writing scenario that'll need to be updated.

colombod (approved):

src/Microsoft.DotNet.Interactive.Documents/InteractiveDocumentElement.cs
jonsequitur: Do you mean kernel info? The goal here is to not have to spin up the kernel in order to know some of these things.

src/Microsoft.DotNet.Interactive.Documents.Tests/CodeSubmissionFormatTests.dib_file_can_be_round_tripped_through_read_and_write_without_the_content_changing.approved.dib
jonsequitur: Yes it should work if formatted over multiple lines, but I need to add a test for that.

PR #2261, opened 2022-09-14
update lock files

colombod:


AbhitejJohn:
@colombod : Would we need to check these files in? My impression is that these are build generated artifacts.

colombod:
These files are generated at dev time and used at ci time to ensure that only the locked version are installed ignoring cachesDr Diego Colombo PhDOn 15 Sep 2022, at 21:23, Abhitej John ***@***.***> wrote:﻿
@colombod : Would we need to check these files in? My impression is that these are build generated artifacts.

PR #2268, opened 2022-09-15
Rename to Polyglot Notebook

colombod:


src/dotnet-interactive-vscode-insiders/package.json
```diff
@@ -208,31 +208,31 @@
       },
       {
         "command": "dotnet-interactive.createNewInteractive",
-        "title": ".NET Interactive: Create Interactive Window"
+        "title": "Polyglot Notebook: Create Interactive Window"
```
brettfo, line 211: Keep the old name here and instead add a new command with a new name.  The old command will simply forward to the new one.

src/dotnet-interactive-vscode-insiders/package.json
```diff
@@ -256,67 +256,67 @@
...
-          "C# (.NET Interactive)"
+          "C# Script"
         ],
         "configuration": "./syntaxes/csharp.language-configuration.json"
       },
       {
         "id": "dotnet-interactive.fsharp",
         "aliases": [
-          "F# (.NET Interactive)"
+          "F#"
```
brettfo, line 266: Can't have this; it clashes with the "real" F# language.  Similar for the others.

colombod: @jonsequitur  what do we do? `F# (Polyglot Notebook)`? @claudiaregio ??

src/dotnet-interactive-vscode-insiders/package.json
colombod: yup, will create new commands contributions and register old commands as delegate but with the dialog.

src/dotnet-interactive-vscode-insiders/package.json
claudiaregio: My comment below comes with the following assumptions, please let me know if they're wrong. 
1. This is referring to the language that gets saved in the metadata of the notebook
2. This is what is displayed in the cell status bar in the bottom right of a cell in a VS Code notebook

My personal opinion is that it should be F# (.NET Interactive), C# (.NET Interactive), etc. The reason being that the special version of C#, F# that is being run is being run by .NET Interactive. In addition, if Jupyter one day decides to support a language picker, a Jupyter/JupyterLab user would have no notion of the "Polyglot Notebooks" extension in VS Code since they can just install the .NET Interactive kernel. This way for the Jupyter use case it would also be more clear what is running that C#, F# code etc.

src/dotnet-interactive-vscode-insiders/package.json
claudiaregio: Do the command prefixes have to match the extension name? The extension name was "Polyglot Notebooks" with an s

src/dotnet-interactive-vscode-insiders/README.md
```diff
@@ -2,31 +2,31 @@ This extension is currently **_in preview_**.
 _The Visual Studio Code notebook support that this extension uses is also in preview and design is ongoing, so the extension might not work._
 ---
 
-# .NET Interactive Notebooks
+# Polyglot Notebooks
 
-This extension adds support for using .NET Interactive in a Visual Studio Code notebook.
+This extension adds support for using Polyglot in a Visual Studio Code notebook.
```
jonsequitur, line 7:
```suggestion
This extension adds support for Polyglot Notebooks to Visual Studio Code.
```

src/dotnet-interactive-vscode-insiders/README.md
```diff
@@ -2,31 +2,31 @@ This extension is currently **_in preview_**.
...
 
 2.  Install the latest [.NET 6 SDK](https://dotnet.microsoft.com/download/dotnet/6.0)
 
-3.  Install the .NET Interactive Notebooks extension from the [marketplace](https://marketplace.visualstudio.com/items?itemName=ms-dotnettools.dotnet-interactive-vscode).
+3.  Install the Polyglot Notebooks extension from the [marketplace](https://marketplace.visualstudio.com/items?itemName=ms-dotnettools.dotnet-interactive-vscode).
 
 ## Opening and Creating notebooks
 
-To open an existing .NET notebook, bring up the Command Palette(`Ctrl+Shift+P`) and select **.NET Interactive: Open notebook**.  Now, navigate to a local `.ipynb` file.
+To open an existing .NET notebook, bring up the Command Palette(`Ctrl+Shift+P`) and select **Polyglot: Open notebook**.  Now, navigate to a local `.ipynb` file.
```
jonsequitur, line 19:
```suggestion
To open an existing .NET notebook, bring up the Command Palette(`Ctrl+Shift+P`) and select **Polyglot Notebook: Open notebook**.  Now, navigate to a local `.ipynb` file.
```

src/dotnet-interactive-vscode-insiders/README.md
```diff
@@ -2,31 +2,31 @@ This extension is currently **_in preview_**.
...
 
 ![opennotebook](https://user-images.githubusercontent.com/2546640/94441970-67d6e180-0171-11eb-8319-c12ba82c3d30.gif)
 
-To create a new notebook, open the Command Palette(`Ctrl+Shift+P`), and select **.NET Interactive: Create new blank notebook**. You can also create a new notebook with `Ctrl+Shift+Alt+N` key combination.
+To create a new notebook, open the Command Palette(`Ctrl+Shift+P`), and select **Polyglot: Create new blank notebook**. You can also create a new notebook with `Ctrl+Shift+Alt+N` key combination.
```
jonsequitur, line 23:
```suggestion
To create a new notebook, open the Command Palette(`Ctrl+Shift+P`), and select **Polyglot  Notebook: Create new blank notebook**. You can also create a new notebook with `Ctrl+Shift+Alt+N` key combination.
```

jonsequitur: I think "Polyglot Notebook" reads better here but I believe the extension name would be "Polyglot Notebooks". I'm assuming we should align these. 

Thoughts, @claudiaregio?

src/dotnet-interactive-vscode-insiders/package.json
```diff
@@ -206,6 +206,38 @@
...
+        "command": "polyglot-notebook.stopCurrentNotebookKernel",
+        "title": "Polyglot Notebook: Stop the current notebook's kernel"
+      },
+      {
+        "command": "polyglot-notebook.stopAllNotebookKernels",
+        "title": "Polyglot Notebook: Stop all notebook kernels"
+      },
+      {
+        "command": "polyglot-notebook.shareValueTo",
+        "title": "Share value to..."
```
jonsequitur, line 239: Does this read more clearly?

```suggestion
        "title": "Share value with..."

```

src/dotnet-interactive-vscode-insiders/package.json
jonsequitur: What should Python, JavaScript, and so on say? What about different execution environments, e.g. JavaScript in the browser versus nodeJS? 

Regardless of the label, these are all unions of the language in question and the Polyglot Notebook language (which adds magic commands.)

What we're really discussing here is the kernel picker experience generally and maybe we should move the discussion over to that issue.

src/dotnet-interactive-vscode-insiders/package.json
```diff
@@ -256,67 +284,67 @@
...
-          "JavaScript (.NET Interactive)"
+          "JavaScript"
         ],
         "configuration": "./syntaxes/javascript.language-configuration.json"
       },
       {
         "id": "dotnet-interactive.magic-commands",
         "aliases": [
-          "Magic Commands (.NET Interactive)"
+          "Plyglot Notenook Magic Commands"
```
jonsequitur, line 314:
```suggestion
          "Polyglot Notebook Magic Commands"

```

src/dotnet-interactive-vscode-insiders/package.json
```diff
@@ -256,67 +284,67 @@
...
-          "PowerShell (.NET Interactive)"
+          "PowerShell"
         ],
         "configuration": "./syntaxes/powershell.language-configuration.json"
       },
       {
         "id": "dotnet-interactive.sql",
         "aliases": [
-          "SQL (.NET Interactive)"
+          "SQL"
```
jonsequitur, line 340: Does it make sense to differentiate SQL dialects?

src/dotnet-interactive-vscode-insiders/README.md
claudiaregio: I agree with you Jon that "Polyglot Notebook" reads better and there is prior art for the command prefix and extensions not matching perfectly. There is the GitHub Issue Notebooks extension but the command is "New GitHub Issue Notebook". This was the same question I was asking in the first chat too basically. 

If we can have the extension be "Polyglot Notebooks" but the commands prefixed with "Polyglot Notebook" I think that would make most sense.

Issue #322, opened 2020-03-28
Import .NET notebook as module


roberchi:
I create a simple extension here [https://github.com/roberchi/Chino.Jupyter.Extensions](https://github.com/roberchi/Chino.Jupyter.Extensions) to support the loading .NET notebook as modules in order to create a kind of library of notebooks.
The extension simply load the notebook specified **#!import [path\notebook]** by loading the json definition and execute each code cell in the same handling kernel of the extension. 
Do you think colud be a relevant feature and can be included in the core?
I'm not expert in the kernel implementation and extension so my code need to be reviewed and improved :-)

jonsequitur:
I think this would be a very useful feature to include and we'd be happy to help you with a PR.

There are a few things to consider:

* Is `#!import` the right name? The F# and C# interactive tools (CSI and FSI) both currently support a `#load` directive:

  * CSI's `#load` directive is used to load `.csx.` scripts 

  * FSI's `#load` directive is used to load `.fs` source files or `.fsx` scripts.

   So "load" might be a more familiar term here, but a distinction between `#load` and `#!load` seems confusing. We could potentially shadow the underlying `#load` directives and augment them with additional functionality to add support for `.ipynb` (and possibly other formats.)

* We might consider iterating on this in our [extension lab](https://github.com/dotnet/interactive/tree/master/src/Microsoft.DotNet.Interactive.ExtensionLab). We plan to publish these as NuGet packages (#398).

jonsequitur:
@LadyNaggaga @tmat @KevinRansom Thoughts on this?

roberchi:
@jonsequitur by my side, extends #load directive to support .ipynb is a great idea.

LadyNaggaga:
I think this would a very useful feature too. I am onboard

jonsequitur:
Maybe this should be `#r "notebook:/path/to/notebook"`

PR #2286, opened 2022-09-23
Introduce the #!import magic command

ambrose-leung:
Partially addresses #322  (import and run cells from another notebook)

nested importing still needs to be done in another PR

Thanks @jonsequitur @colombod for the coding session and getting me set up with working on this repo

### DIFF
--- a/src/Microsoft.DotNet.Interactive.Tests/ImportNotebookTests.cs
+++ b/src/Microsoft.DotNet.Interactive.Tests/ImportNotebookTests.cs
@@ -1,11 +1,14 @@
 ﻿using System;
 using System.IO;
+using System.Linq;
 using System.Threading.Tasks;
 using FluentAssertions;
+using Microsoft.AspNetCore.Http;
 using Microsoft.DotNet.Interactive.CSharp;
 using Microsoft.DotNet.Interactive.Documents;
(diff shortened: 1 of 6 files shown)


src/Microsoft.DotNet.Interactive.Tests/ImportNotebookTests.cs
```diff
@@ -0,0 +1,85 @@
+﻿using System;
```
colombod, line 1: can you add the header with silence
like
``` 
```

src/Microsoft.DotNet.Interactive.Tests/ImportNotebookTests.cs
```diff
@@ -0,0 +1,85 @@
...
+                    KernelName = "csharp"
+                },
+                new InteractiveDocumentElement
+                {
+                    Contents = "5+3",
+                    KernelName = "markdown" //should not evaluate to 8
+                },
+                new InteractiveDocumentElement
+                {
+                    Contents = "5+3",
```
colombod, line 45: the case for html and markdown is different as they won't produce that event. Html will handle the code submission. very likely you will see the events for that command.

jonsequitur (approved):

Issue #1958, opened 2022-04-08
Race Condition on port


rfrerebe:
### Describe the bug

When opening several notebooks in jupyterhub, and restarting them to execute all cells, very often this exception is raised

Port number (here 1032) is usually in the low range of the allowed range.

```
Unhandled exception: System.IO.IOException: Failed to bind to address http://100.64.5.12:1032: address already in use.
 ---> Microsoft.AspNetCore.Connections.AddressInUseException: Address already in use
 ---> System.Net.Sockets.SocketException (98): Address already in use
   at System.Net.Sockets.Socket.UpdateStatusAfterSocketErrorAndThrowException(SocketError error, String callerName)
   at System.Net.Sockets.Socket.DoBind(EndPoint endPointSnapshot, SocketAddress socketAddress)
   at System.Net.Sockets.Socket.Bind(EndPoint localEP)
   at Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets.SocketTransportOptions.CreateDefaultBoundListenSocket(EndPoint endpoint)
   at Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets.SocketConnectionListener.Bind()
   --- End of inner exception stack trace ---
   at Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets.SocketConnectionListener.Bind()
   at Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets.SocketTransportFactory.BindAsync(EndPoint endpoint, CancellationToken cancellationToken)
   at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Infrastructure.TransportManager.BindAsync(EndPoint endPoint, ConnectionDelegate connectionDelegate, EndpointConfig endpointConfig, CancellationToken cancellationToken)
   at Microsoft.AspNetCore.Server.Kestrel.Core.KestrelServerImpl.<>c__DisplayClass30_0`1.<<StartAsync>g__OnBind|0>d.MoveNext()
--- End of stack trace from previous location ---
   at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.AddressBinder.BindEndpointAsync(ListenOptions endpoint, AddressBindContext context, CancellationToken cancellationToken)
   --- End of inner exception stack trace ---
   at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.AddressBinder.BindEndpointAsync(ListenOptions endpoint, AddressBindContext context, CancellationToken cancellationToken)
[...]

PR #2288, opened 2022-09-26
Quick workaround for self-evident cross-process race condition (Issue #1958)

daviddenis-stx:
This workaround (for https://github.com/dotnet/interactive/issues/1958) allows to run multiple kernels (in separate processes) without stumbling upon Microsoft.AspNetCore.Connections.AddressInUseException.

A better fix is possible, but this uncomplicated workaround definitely helps for almost free.

### DIFF
--- a/src/dotnet-interactive/Program.cs
+++ b/src/dotnet-interactive/Program.cs
@@ -99,7 +99,7 @@ namespace Microsoft.DotNet.Interactive.App
 
             return webHost;
 
-            static HttpPort GetFreePort(StartupOptions startupOptions)
+            static HttpPort GetFreePort(StartupOptions startupOptions, int maxRetries = 50)
             {
                 using var __ = Log.OnEnterAndExit(nameof(GetFreePort));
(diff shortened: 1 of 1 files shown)


src/dotnet-interactive/Program.cs
```diff
@@ -99,30 +99,31 @@ internal static IDisposable StartToolLogging(StartupOptions options)
...
             {
                 using var __ = Log.OnEnterAndExit(nameof(GetFreePort));
                 if (startupOptions.HttpPort is not null && !startupOptions.HttpPort.IsAuto)
                 {
                     return startupOptions.HttpPort;
                 }
 
-                var currentPort = 0;
-                var endPort = 0;
+                var startPort = 49152; // RFC 6335 compatible
```
colombod, line 110: 0 values as default will let the  TCPListener pick a random one when not specified, this way is the HttpPortRance is not specified we keep scanning instead of just letting pick a random one

colombod (changes requested):

PR #2296, opened 2022-09-27
merge main

shibbas:


brettfo:
There seems to be some issues with AzDO reporting the runs to GH, but a few hours ago there was [all green run](https://dev.azure.com/dnceng-public/public/_build/results?buildId=33442&view=results) against the proper commit, so in the interest of time I'm just merging it.

PR #2301, opened 2022-09-28
Default handler on proxy kernels doesn't execute on remote

colombod:
Adding commands ay runtime is not working on proxy kernel. The command is not routed to the remote kernel

### DIFF
--- a/src/Microsoft.DotNet.Interactive/Connection/ProxyKernel.cs
+++ b/src/Microsoft.DotNet.Interactive/Connection/ProxyKernel.cs
@@ -53,6 +53,10 @@ public sealed class ProxyKernel : Kernel
         {
             if (coe.Event is { } e)
             {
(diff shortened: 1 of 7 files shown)


src/Microsoft.DotNet.Interactive/Connection/ProxyKernel.cs
```diff
@@ -53,6 +53,10 @@ public sealed class ProxyKernel : Kernel
         {
             if (coe.Event is { } e)
             {
+                if (e is KernelInfoProduced)
```
jonsequitur, line 56: You can remove the empty block.

colombod: ops done

jonsequitur (approved):

PR #2302, opened 2022-09-29
Open connection dialog when using connection string input types

corivera:
This change opens the Azure Data Studio connection dialog when using an input token for the connection string in the **connect** magic command. Since we need to use some ADS-specific extension methods here, I moved the input handling into the versionSpecificFunctions files so that we can import azdata without breaking the vscode extension.

### DIFF
--- a/src/dotnet-interactive-vscode-ads/src/versionSpecificFunctions.ts
+++ b/src/dotnet-interactive-vscode-ads/src/versionSpecificFunctions.ts
@@ -3,6 +3,8 @@
 
 import * as vscode from 'vscode';
 import * as azdata from 'azdata';
+import { CompositeKernel } from './vscode-common/dotnet-interactive';
+import * as contracts from './vscode-common/dotnet-interactive/contracts';
 
 export function getNotebookDocumentFromEditor(notebookEditor: vscode.NotebookEditor): vscode.NotebookDocument {
     return notebookEditor.document;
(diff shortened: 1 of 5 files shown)


src/dotnet-interactive-vscode-ads/src/versionSpecificFunctions.ts
```diff
@@ -27,3 +28,19 @@ export async function replaceNotebookMetadata(notebookUri: vscode.Uri, documentM
...
+
+export async function handleRequestInput(prompt: string, password: boolean, inputTypeHint: string): Promise<string | null | undefined> {
+    let result: string | null | undefined;
+    if (inputTypeHint === 'connectionstring-mssql') {
+        let connection = await azdata.connection.openConnectionDialog();
+        if (connection) {
+            result = await azdata.connection.getConnectionString(connection.connectionId, true);
+        }
+    } else {
+        result = (inputTypeHint === 'file')
```
jonsequitur, line 40: It would be ideal if this block wasn't repeated code, e.g. a map of type hint to implementation that the ADS code could just add a handler to. The code is simple now but additional use of type hints is planned and it will get more complex.

corivera: Where would that input handler need to be added? Would it be another RequestInputType command handler that gets registered from the version specific code?

jonsequitur: It doesn't need to be a kernel command handler, but just something the kernel calls into when handling an input request. If you add custom handlers there for the hinted type, then use the custom experience. Otherwise, return a value that allows the central handler to fall back to the default behaviors (i.e. currently "file" and "text").

corivera: Added a new handler

colombod (approved):

brettfo (approved):

PR #2303, opened 2022-09-29
`#!import` support in documents API

jonsequitur:
This adds support to `InteractiveDocument` for reading documents linked via `#!import` magic commands.

### DIFF
--- a/src/Microsoft.DotNet.Interactive.Documents/InteractiveDocument.cs
+++ b/src/Microsoft.DotNet.Interactive.Documents/InteractiveDocument.cs
@@ -40,19 +40,24 @@ public class InteractiveDocument : IEnumerable
     public IDictionary<string, object> Metadata =>
         _metadata ??= new Dictionary<string, object>();
 
-    public IEnumerable<InteractiveDocument> GetImportedDocuments()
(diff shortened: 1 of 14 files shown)


colombod (approved):

src/Microsoft.DotNet.Interactive.Documents/InteractiveDocument.cs
```diff
@@ -134,16 +160,19 @@ internal void NormalizeElementKernelNames(KernelInfoCollection kernelInfos)
         return null;
     }
 
-    public static async Task<InteractiveDocument> LoadInteractiveDocumentAsync(
-    FileInfo file,
-    KernelInfoCollection kernelInfos)
+    public static async Task<InteractiveDocument> LoadAsync(
```
colombod, line 163: If LoadAsync expands / resolves the imports always could this cause a notebook parsing to inline the imported notebooks as we open a ipynb / dib in vscode?

jonsequitur: It doesn't resolve them or even try to identify them until someone calls `GetImportsAsync`.

PR #2307, opened 2022-09-29
RequestKernelInfo hangs

colombod:


jonsequitur (approved):

src/Microsoft.DotNet.Interactive/RoutingSlip.cs
```diff
@@ -11,49 +12,103 @@ namespace Microsoft.DotNet.Interactive;
 public class RoutingSlip : IReadOnlyList<Uri>
 {
     private readonly HashSet<Uri> _uniqueUris ;
-    private readonly List<Uri> _uris;
-    private readonly object _lock = new();
+    private readonly List<Entry> _uris;
+    private readonly ConcurrentDictionary<Uri, Entry> _entriesByUri;
```
jonsequitur, line 16: This is pretty heavyweight for every since routing slip.

src/Microsoft.DotNet.Interactive/RoutingSlip.cs
```diff
@@ -11,49 +12,103 @@ namespace Microsoft.DotNet.Interactive;
...
-    IEnumerator<Uri> IEnumerable<Uri>.GetEnumerator() => _uris.GetEnumerator();
+    IEnumerator<Uri> IEnumerable<Uri>.GetEnumerator() => _uris.Select(e => e.KernelUri).GetEnumerator();
 
     IEnumerator IEnumerable.GetEnumerator() => _uris.GetEnumerator();
 
-    public Uri this[int index] => _uris[index];
+    public Uri this[int index] => _uris[index].KernelUri;
     public int Count => _uris.Count;
+
+    public class Entry
```
jonsequitur, line 96: This is a good use case for a record. I don't think it should be public though.

src/Microsoft.DotNet.Interactive/RoutingSlip.cs
```diff
@@ -11,49 +12,103 @@ namespace Microsoft.DotNet.Interactive;
...
+        }
+
+        public bool Handled { get; internal set; }
+
+        public Uri KernelUri { get; }
+    }
+
+    public bool HasProcessed(Uri kernelInfoUri)
+    {
+        return _entriesByUri.TryGetValue(kernelInfoUri, out var entry) && entry.Handled;
```
jonsequitur, line 111: It's better to use consistent vocabulary, and there's always a better word than "processed".

src/Microsoft.DotNet.Interactive/RoutingSlip.cs
```diff
@@ -11,49 +12,103 @@ namespace Microsoft.DotNet.Interactive;
...
         return contains;
     }
 
-    IEnumerator<Uri> IEnumerable<Uri>.GetEnumerator() => _uris.GetEnumerator();
+    IEnumerator<Uri> IEnumerable<Uri>.GetEnumerator() => _uris.Select(e => e.KernelUri).GetEnumerator();
 
     IEnumerator IEnumerable.GetEnumerator() => _uris.GetEnumerator();
 
-    public Uri this[int index] => _uris[index];
+    public Uri this[int index] => _uris[index].KernelUri;
```
jonsequitur, line 93: I don't think the indexer makes much sense anymore, and it probably no longer makes sense for the routing slip to be `IEnumerable<Uri>`.

src/Microsoft.DotNet.Interactive/Commands/KernelCommand.cs
```diff
@@ -35,11 +35,11 @@ public KernelCommand Parent
         internal set
         {
             _parent = value;
-            var currentSlip = RoutingSlip;
+            var currentSlip = RoutingSlip ?? new RoutingSlip();
```
jonsequitur, line 38: If `RoutingSlip` is null that's a bug. Let's not hide it by silently coalescing here.

src/Microsoft.DotNet.Interactive/Commands/KernelCommand.cs
```diff
@@ -35,11 +35,11 @@ public KernelCommand Parent
...
             RoutingSlip = new RoutingSlip(_parent?.RoutingSlip);
             foreach (var uri in currentSlip)
             {
-                RoutingSlip.TryAdd(uri);
+                RoutingSlip.TryMarkArrival(uri);
```
jonsequitur, line 42: This doesn't make sense. What if the `Uri` in the source slip was also already completed?

As an aside, there are other potential bugs in this logic. Is setting `Parent` to null a scenario we want to allow? Is reparenting a scenario we want to allow? If not, this should probably no longer be a property setter.

PR #2311, opened 2022-10-03
add cell execution count

### LINKED ISSUE #1852: Support Execution Count
Execution count does not show up in VS Code Notebook 
 - Frontend
    - [x] Visual Studio Code 
    - [x] Visual Studio Code Insiders

### Screenshots

image

brettfo:
Fixes #1852.

The VS Code APIs only allow us to change a cell's execution order number while it's running which means we have no way of resetting all values, etc. when the backing kernel is restarted.

### DIFF
--- a/src/dotnet-interactive-vscode-insiders/src/notebookControllers.ts
+++ b/src/dotnet-interactive-vscode-insiders/src/notebookControllers.ts
@@ -11,7 +11,7 @@ import { getSimpleLanguage, isDotnetInteractiveLanguage, jupyterViewType, notebo
 import { getCellLanguage, getDotNetMetadata, getLanguageInfoMetadata, isDotNetNotebookMetadata, withDotNetKernelMetadata } from './vscode-common/ipynbUtilities';
 import { reshapeOutputValueForVsCode } from './vscode-common/interfaces/utilities';
 import { selectDotNetInteractiveKernelForJupyter } from './vscode-common/commands';
-import { ErrorOutputCreator } from './vscode-common/interactiveClient';
+import { ErrorOutputCreator, InteractiveClient } from './vscode-common/interactiveClient';
 import { LogEntry, Logger } from './vscode-common/dotnet-interactive/logger';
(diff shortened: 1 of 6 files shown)


jonsequitur:
Got tests?

brettfo:
Most of the changes are in VS Code-specific files; the only file we can unit test is `interactiveClient.ts` which is just the functions `resetExecutionCount()` and `getNextExecutionCount()`.  I'm happy to add tests, but they didn't seem terribly important.

jonsequitur:
I guess the question is whether tests can be added that will prevent someone from accidentally regressing this later.

brettfo:
Not in any meaningful way.  All changes were from within the VS Code space which we can't unit test.

I'll add a scenario to the test notebook, though, so we remember to check it on the manual passes.

src/dotnet-interactive-vscode-insiders/src/notebookControllers.ts
```diff
@@ -170,16 +173,16 @@ export class DotNetNotebookKernel {
 
                 return client.execute(source, getSimpleLanguage(cell.document.languageId), outputObserver, diagnosticObserver, { id: cell.document.uri.toString() }).then(async () => {
                     await outputUpdatePromise;
-                    endExecution(cell, true);
+                    endExecution(client, cell, true);
```
colombod, line 176: do we need to add the execution count management to the client? Is looks something the controller should manage since is a mapping with the notebook id and also the controller manages the restart of the kernel associated with a client

brettfo: We need to have it on the kernel because:

1. We need to be able to reset to `1` on kernel restart and at that point we only have access to the client.
2. We need to keep that data in the client's metadata so if the user does a save-as .ipynb (which goes through our own code) we can populate the `execution_count` field.

colombod: ouch .. I see, approving then ..

colombod (approved):

PR #2329, opened 2022-10-07
the acquire command is still related to dotnet-interactive

colombod:


src/dotnet-interactive-vscode-ads/package.json
```diff
@@ -196,7 +196,7 @@
     },
     "commands": [
       {
-        "command": "polyglot-notebook.acquire",
+        "command": "dotnet-interactive.acquire",
```
brettfo, line 199: The education bundle uses the `dotnet-interactive.acquire` command.  In case a user gets into a state where they have an old education bundle and a new Polyglot Notebook, we should also have a copy of the old command name that silently forwards to the new one.

brettfo (changes requested):

src/dotnet-interactive-vscode-ads/package.json
colombod: i don't think we need this to be polyglot-notebook

brettfo (approved):

Issue #1440, opened 2021-06-22
Restarting kernel doesn't reset JavaScript state


jonsequitur:
Repro steps:

1) Set a variable in a JavaScript cell:

```js
myVariable = 123;
```

2) Restart the kernel

3) In another JavaScript cell, run this:

```js
console.log(myVariable);
```

Expected:

```
Error: ReferenceError: myVariable is not defined
```

Actual:

```
123
```

PR #2335, opened 2022-10-10
reset vewbiew on kernel restart

colombod:
Addresses #1440. Now Js kernel and webview state are reset on kernel restart


src/dotnet-interactive-vscode-common/src/commands.ts
```diff
@@ -139,6 +139,7 @@ export function registerKernelCommands(context: vscode.ExtensionContext, clientM
             await vscode.commands.executeCommand('dotnet-interactive.stopCurrentNotebookKernel', notebook);
             const _client = await clientMapper.getOrAddClient(notebook.uri);
             restartCompletionSource.resolve();
+            await vscode.commands.executeCommand('workbench.notebook.layout.webview.reset', notebook.uri);
```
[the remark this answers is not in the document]
colombod: the context is not there in this part of the project. I have moved the needed part in version specific functions and ads / stable are using the common one. is this reasonable?

src/dotnet-interactive-vscode-common/src/commands.ts
```diff
@@ -139,6 +139,9 @@ export function registerKernelCommands(context: vscode.ExtensionContext, clientM
             await vscode.commands.executeCommand('dotnet-interactive.stopCurrentNotebookKernel', notebook);
             const _client = await clientMapper.getOrAddClient(notebook.uri);
             restartCompletionSource.resolve();
+            if (isAzureDataStudio(context)) {
```
brettfo, line 142: Negate this; it should only execute if _not_ ADS.

colombod: ops!

brettfo (approved):

PR #2338, opened 2022-10-11
handle KernelShutdownRequest

colombod:


src/Microsoft.DotNet.Interactive.Jupyter/Shell.cs
```diff
@@ -83,7 +88,7 @@ public Task StartAsync(CancellationToken cancellationToken)
             var kernelIdentity = Guid.NewGuid().ToString();
             Task.Run(async () =>
             {
-                using var activity = Log.OnEnterAndExit();
+                using var activity = Log.OnEnterAndExit("jupyter shell channel");
```
jonsequitur, line 91: Why change this? The usual default will be the containing method name, which is less verbose and more predictable.

jonsequitur: Ah, I see. Since there are two different telemetry contexts in this method, factoring them into actual separate methods might be cleaner.

jonsequitur (approved):

PR #2346, opened 2022-10-13
Handle null values from data query

colombod:
Added support for nullable types mapping in `TabularDataResource`.
SQL and KQL kernels now do not parse `NULL` string if field is null and produce correct `TabularDataResource` instances
Improved performance by creating `TabularDataResource` directly from the `ColumnInfo` and `CellValue`.
ColumnInfo that are keys preserve that quality in `TableSchema.PrimaryKey`


colombod:
@shibbas this should do.

PR #2356, opened 2022-10-17
JS implementation for SendValue command

colombod:
Value sharing is now handled on the js kernel
image

the kernel with `#!share` is not listed as a completion for the `--from` option
image

### DIFF
--- a/src/Microsoft.DotNet.Interactive.Browser.Tests/JavaScriptKernelTests.cs
+++ b/src/Microsoft.DotNet.Interactive.Browser.Tests/JavaScriptKernelTests.cs
@@ -28,7 +28,7 @@ public class JavaScriptKernelTests : IDisposable
 
     public void Dispose() => _disposables.Dispose();
 
-    [Fact]
+    [FactSkipLinux("Requires Playwright installed")]
     public async Task It_can_execute_code()
     {
         using var kernel = await CreateJavaScriptProxyKernelAsync();
@@ -40,7 +40,7 @@ public class JavaScriptKernelTests : IDisposable
         events.Should().NotContainErrors();
     }
 
-    [Fact]
+    [FactSkipLinux("Requires Playwright installed")]
     public async Task It_can_get_a_return_value()
     {
         using var kernel = await CreateJavaScriptProxyKernelAsync();
@@ -57,7 +57,7 @@ public class JavaScriptKernelTests : IDisposable
                                   v.Value == "123");
     }
 
-    [Fact]
+    [FactSkipLinux("Requires Playwright installed")]
     public async Task It_can_get_console_log_output()
     {
         using var kernel = await CreateJavaScriptProxyKernelAsync();
@@ -74,7 +74,7 @@ public class JavaScriptKernelTests : IDisposable
                                   v.Value == "123");
     }
 
-    [Fact]
+    [FactSkipLinux("Requires Playwright installed")]
     public async Task It_can_import_value_from_another_kernel()
     {
         using var kernel = await CreateJavaScriptProxyKernelAsync();
@@ -103,7 +103,7 @@ console.log(x);", targetKernelName: kernel.Name));
                                 v.Value == "123");
     }
 
-    [Fact]
+    [FactSkipLinux("Requires Playwright installed")]
     public async Task It_can_share_value_with_another_kernel()
     {
         using var kernel = await CreateJavaScriptProxyKernelAsync();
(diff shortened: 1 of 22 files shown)


src/Microsoft.DotNet.Interactive/Commands/SendValue.cs
```diff
@@ -24,7 +24,7 @@ public class SendValue : KernelCommand
         FormattedValue = formattedValue;
     }
 
-    public FormattedValue FormattedValue { get; }
+    public FormattedValue FormattedValue { get; internal set; }
```
jonsequitur, line 27: Why is this needed?

src/Microsoft.DotNet.Interactive/ValueSharing/JavaScriptValueDeclarer.cs
```diff
@@ -34,21 +33,18 @@ static JavaScriptValueDeclarer()
...
-        out SubmitCode command)
+    public static bool TryGetValueDeclaration(object referenceValue, string declareAsName, out string code)
     {
-        if (valueProduced.Value is { } value)
+        if (referenceValue is { } value)
         {
-            var code = $"{valueProduced.Name} = {JsonSerializer.Serialize(value, _serializerOptions)};";
-            command = new SubmitCode(code);
+            code = $"{declareAsName} = {JsonSerializer.Serialize(value, _serializerOptions)};";
+            new SubmitCode(code);
```
jonsequitur, line 41: This line can go.

src/microsoft-dotnet-interactive/src/javascriptKernel.ts
```diff
@@ -11,15 +11,34 @@ export class JavascriptKernel extends Kernel {
...
     }
 
+    private handleSendValue(invocation: IKernelCommandInvocation): Promise<void> {
+        const sendValue = <contracts.SendValue>invocation.commandEnvelope.command;
+        if (sendValue.formattedValue) {
+            switch (sendValue.formattedValue.mimeType) {
+                case 'application/json':
+                    (<any>globalThis)[sendValue.name] = JSON.parse(sendValue.formattedValue.value);
+                    break;
+                case 'text/plain':
```
jonsequitur, line 31: This MIME type feels wrong to me. Maybe `text/javascript`?

src/Microsoft.DotNet.Interactive/Commands/SendValue.cs
colombod: No reason

src/Microsoft.DotNet.Interactive/ValueSharing/JavaScriptValueDeclarer.cs
colombod: done

src/Microsoft.DotNet.Interactive/Commands/SendValue.cs
jonsequitur: 🤨

src/microsoft-dotnet-interactive/src/javascriptKernel.ts
colombod: if the variable is shared as plain/text we trat that as string value. what should be the behaviour ?

jonsequitur: I'm not sure this MIME type should be valid, or else we should just declare a string.

colombod: right now if is text/plain it declares a string.

colombod: Deleted that case

src/Microsoft.DotNet.Interactive.Tests/VariableSharingTests.cs
```diff
@@ -542,7 +543,7 @@ await localCompositeKernel
                           remoteKernelUri);
 
             javascriptKernel.UseValueSharing();
-
+            javascriptKernel.RegisterCommandType<SendValue>();
```
jonsequitur, line 546: It shouldn't be necessary to register this on both the remote and the proxy. Is there a bug here?

colombod: good point, not needed

src/Microsoft.DotNet.Interactive.Browser.Tests/JavaScriptKernelTests.cs
```diff
@@ -72,6 +73,64 @@ public async Task It_can_get_console_log_output()
                                   v.Value == "123");
     }
 
+    [FactSkipLinux]
```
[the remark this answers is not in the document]
colombod: I have added the ability to give a reason why we skip.

brettfo (approved):

PR #2362, opened 2022-10-19
fix Jupyter metadata

jonsequitur:


src/Microsoft.DotNet.Interactive.Documents.Tests/JupyterFormatTests.cs
```diff
@@ -81,19 +81,22 @@ public void notebook_metadata_default_language_is_honored_in_cells_without_langu
                 });
     }
 
-    [Fact]
-    public void Default_language_can_be_specified_in_metadata()
+    [Theory]
+    [InlineData("C#", "csharp")]
+    [InlineData("F#", "fsharp")]
+    [InlineData("f#", "fsharp")]
+    [InlineData("PowerShell", "powershell")]
```
brettfo, line 88: isn't this kernel name `pwsh` and `powershell` an alias?

src/Microsoft.DotNet.Interactive.Documents.Tests/JupyterFormatTests.cs
```diff
@@ -983,6 +986,11 @@ public void serialized_notebook_has_appropriate_metadata()
                     name = "C#",
                     pygments_lexer = "csharp",
                     version = "10.0"
+                },
+                dotnet_interactive = new
+                {
+                    defaultKernelName = "csharp",
+                    items = Array.Empty<object>()
```
brettfo, line 993: if we default to `csharp`, then `items` should contain `{ name: "csharp", aliases: [] }`

src/Microsoft.DotNet.Interactive.Documents.Tests/JupyterFormatTests.cs
jonsequitur: Here's what we have in the kernelspec: 
image

src/Microsoft.DotNet.Interactive.Documents.Tests/JupyterFormatTests.cs
jonsequitur: I notice we don't have `language` in this object. Did you hit that in your PR?

colombod (approved):

PR #2381, opened 2022-10-26
proxy kernels updates kernel info on first command

colombod:
image

Now javascript kernel language is correctly used in the proxy KernelInfo

### DIFF
--- a/src/Microsoft.DotNet.Interactive.SqlServer/MsSqlKernel.cs
+++ b/src/Microsoft.DotNet.Interactive.SqlServer/MsSqlKernel.cs
@@ -20,7 +20,7 @@ namespace Microsoft.DotNet.Interactive.SqlServer
         internal MsSqlKernel(
             string name,
             string connectionString,
-            ToolsServiceClient client) : base(name, client, "SQLite")
+            ToolsServiceClient client) : base(name, client, "SQL")
         {
             if (string.IsNullOrWhiteSpace(connectionString))
             {
--- a/src/Microsoft.DotNet.Interactive.SqlServer/MsSqlKernel.cs
+++ b/src/Microsoft.DotNet.Interactive.SqlServer/MsSqlKernel.cs
@@ -20,7 +20,7 @@ namespace Microsoft.DotNet.Interactive.SqlServer
         internal MsSqlKernel(
             string name,
             string connectionString,
-            ToolsServiceClient client) : base(name, client, "SQL")
+            ToolsServiceClient client) : base(name, client, "T-SQL")
         {
             if (string.IsNullOrWhiteSpace(connectionString))
             {
--- a/src/Microsoft.DotNet.Interactive/Connection/ProxyKernel.cs
+++ b/src/Microsoft.DotNet.Interactive/Connection/ProxyKernel.cs
@@ -17,11 +17,11 @@ public sealed class ProxyKernel : Kernel
     private readonly IKernelCommandAndEventSender _sender;
     private readonly IKernelCommandAndEventReceiver _receiver;
     private ExecutionContext _executionContext;
+    private bool _requiresRequestKernelInfoOnFirstCommand = true;
+    private string _suppressCompletionsForCommandId;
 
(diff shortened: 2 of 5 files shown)


src/Microsoft.DotNet.Interactive/Connection/ProxyKernel.cs
```diff
@@ -34,24 +34,22 @@ public sealed class ProxyKernel : Kernel
 
         if (remoteUri is not null)
         {
-            _remoteUri = remoteUri;
+            KernelInfo.RemoteUri = remoteUri;
         }
         else if (sender.RemoteHostUri is { } remoteHostUri)
         {
-            _remoteUri = new(remoteHostUri, name);
+            KernelInfo.RemoteUri = new(remoteHostUri, name);
```
jonsequitur, line 41: I wonder if the bug we were looking at in the browser kernel yesterday means this the wrong behavior. It sort of works by accident, sometimes, but it's not really a safe assumption.

src/Microsoft.DotNet.Interactive/Connection/ProxyKernel.cs
```diff
@@ -121,6 +123,25 @@ private Task HandleByForwardingToRemoteAsync(KernelCommand command, KernelInvoca
         });
     }
 
+    private void EnsureKernelInfoIsLoaded(KernelCommand command, KernelInvocationContext context)
+    {
+        if (_requiresRequestKernelInfoOnFirstCommand)
```
jonsequitur, line 128: Can this be done using a deferred command?

jonsequitur: Though I guess that would produce a doubled `RequestKernelInfo` in some cases...

src/Microsoft.DotNet.Interactive/Connection/ProxyKernel.cs
```diff
@@ -184,6 +205,10 @@ private void DelegatePublication(KernelEvent kernelEvent)
                     _inflight.Remove(token);
                     pending.completionSource.TrySetResult(cs);
                     break;
+                case CommandFailed _ when kernelEvent.Command.GetOrCreateId() == _suppressCompletionsForCommandId:
+                case CommandSucceeded _ when kernelEvent.Command.GetOrCreateId() == _suppressCompletionsForCommandId:
+                    _suppressCompletionsForCommandId = null;
```
jonsequitur, line 210: Why is this needed?

src/Microsoft.DotNet.Interactive.SqlServer/MsSqlKernel.cs
```diff
@@ -20,7 +20,7 @@ internal class MsSqlKernel : ToolsServiceKernel
         internal MsSqlKernel(
             string name,
             string connectionString,
-            ToolsServiceClient client) : base(name, client, "SQLite")
+            ToolsServiceClient client) : base(name, client, "SQL")
```
jonsequitur, line 23: `T-SQL`?

colombod: @brettfo  what are you thoughts on this language

brettfo: "SQL" is good; it'll play well with my kernel picker work

brettfo (approved):

colombod:
Mssqlserver language is t-sql, SQLite is SQLite not sure how we want to handle that part. I have set the discovery kernel to be sql Dr Diego Colombo PhDOn 27 Oct 2022, at 21:50, Brett V. Forsgren ***@***.***> wrote:﻿
@brettfo commented on this pull request.



In src/Microsoft.DotNet.Interactive.SqlServer/MsSqlKernel.cs:
> @@ -20,7 +20,7 @@ internal class MsSqlKernel : ToolsServiceKernel
         internal MsSqlKernel(
             string name,
             string connectionString,
-            ToolsServiceClient client) : base(name, client, "SQLite")
+            ToolsServiceClient client) : base(name, client, "SQL")

"SQL" is good; it'll play well with my kernel picker work

PR #2382, opened 2022-10-26
Jupyter zmq connections

shibbas:


src/Microsoft.DotNet.Interactive.Jupyter/ConnectJupyterKernelCommand.cs
```diff
@@ -51,31 +56,69 @@ public class ConnectJupyterKernelCommand : ConnectKernelCommand
...
+        if (latestHash == _optionHash)
+        {
+            return _kernelSpecCompletions;
+        }
 
-        var kernel = await connector?.CreateKernelAsync(localName);
-        kernel?.RegisterForDisposal(disposables);
-        return kernel;
+        _optionHash = latestHash;
+        _kernelSpecCompletions = new List<CompletionItem>();
```
jonsequitur, line 101: Why is this a field if we're overwriting it on each call?

src/Microsoft.DotNet.Interactive.Jupyter/Connection/IJupyterConnection.cs
```diff
@@ -8,5 +8,7 @@ namespace Microsoft.DotNet.Interactive.Jupyter.Connection;
 
 public interface IJupyterConnection : IDisposable
 {
+    Task<string[]> GetKernelSpecsAsync();
```
jonsequitur, line 11:
```suggestion
    Task<IReadOnlyCollection<string>> GetKernelSpecsAsync();
```

jonsequitur: And would `GetKernelSpecNamesAsync` be more appropriate?

src/Microsoft.DotNet.Interactive.Jupyter/Http/JupyterHttpConnection.cs
```diff
@@ -60,6 +74,29 @@ public void Dispose()
...
+            HttpResponseMessage response = await SendWebRequestAsync(
+                apiPath: "api/kernelspecs",
+                body: null,
+                contentType: "application/json",
+                method: HttpMethod.Get
+            );
+
+            if (!response.IsSuccessStatusCode)
+            {
+                return null;
```
jonsequitur, line 90:
```suggestion
                return Array.Empty<string>();
```

It's good to avoid expecting that `null` will have some special meaning distinct from an empty sequence.

src/Microsoft.DotNet.Interactive.Jupyter/IJupyterKernelSpecModule.cs
```diff
@@ -11,5 +12,6 @@ public interface IJupyterKernelSpecModule
     {
         Task<CommandLineResult> InstallKernel(DirectoryInfo sourceDirectory);
         DirectoryInfo GetDefaultKernelSpecDirectory();
+        IReadOnlyDictionary<string, DirectoryInfo> GetInstalledKernelDirectories();
```
jonsequitur, line 15: How does this relate to `GetDefaultKernelSpecDirectory`?

shibbas: The default is being used for installing the .NET kernel spec in this case whereas the list of installed kernel directories is returning all the directories where all the kernels are installed based on the search order specified here: https://docs.jupyter.org/en/latest/use/jupyter-directories.html where JUPYTER_PATH is searched first and then default.

jonsequitur: Would it make sense to consolidate them to ensure consistency?
.

shibbas: Jupyter seems to actually separate the concept here. There is a list of jupyter paths that it searches for when looking for kernels vs. a default data_dir that is either set or can be overridden via an environment variable JUPYTER_DATA_DIR. So it's aligned with jupyter and is designed to be used for two different scenarios.

src/Microsoft.DotNet.Interactive.Jupyter/ConnectJupyterKernelCommand.cs
```diff
@@ -51,31 +56,69 @@ public class ConnectJupyterKernelCommand : ConnectKernelCommand
...
+        var connection = GetJupyterConnection(commandLineContext.ParseResult);
+        JupyterKernelConnector connector = new JupyterKernelConnector(connection, kernelType);
+        CompositeDisposable disposables = new CompositeDisposable
+        {
+            connection
+        };
+
+        var localName = commandLineContext.ParseResult.GetValueForOption(KernelNameOption);
+
+        var kernel = await connector?.CreateKernelAsync(localName);
```
jonsequitur, line 69:
```suggestion
        var kernel = await connector.CreateKernelAsync(localName);
```

jonsequitur (approved):

PR #2406, opened 2022-11-01
js kernel can evaluate javascript code

colombod:
html kernel can evaluate javascript in when <script> tags are added and contains code

### DIFF
--- a/src/microsoft-dotnet-interactive/src/htmlKernel.ts
+++ b/src/microsoft-dotnet-interactive/src/htmlKernel.ts
@@ -3,6 +3,7 @@
 
 import * as contracts from "./contracts";
 import { Kernel, IKernelCommandInvocation } from "./kernel";
+import { Logger } from "./logger";
 import { PromiseCompletionSource } from "./promiseCompletionSource";
 
 export class HtmlKernel extends Kernel {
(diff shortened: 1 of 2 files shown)


src/microsoft-dotnet-interactive/src/htmlKernel.ts
```diff
@@ -45,7 +46,8 @@ export interface HtmlDomFragmentInserterConfiguration {
     getOrCreateContainer?: () => HTMLElement,
     updateContainerContent?: (container: HTMLElement, htmlFragment: string) => void,
     createMutationObserver?: (callback: MutationCallback) => MutationObserver,
-    normalizeHtmlFragment?: (htmlFragment: string) => string
+    normalizeHtmlFragment?: (htmlFragment: string) => string,
+    jsEvaluator?: (js: string) => void
```
brettfo, line 50: is this ever explicitly set anywhere?

colombod: 17 lines later ...

brettfo: I meant by specifying it in the config value.  I don't ever see it being specified, so why not leave it as `eval(...)`?

colombod: if done that way is the eval of node during testing and not the one from the browser engine

brettfo (approved):

PR #2416, opened 2022-11-02
Handle ErrorProduced Events in Extension

### LINKED ISSUE #2415: explicitly handle `ErrorProduced` event type in `InteractiveClient.Execute`
[This switch statement](https://github.com/dotnet/interactive/blob/d8d2990c6830b948525d558ab46849b8c88e0877/src/dotnet-interactive-vscode-common/src/interactiveClient.ts#L157) needs to be expanded to handle `ErrorProducedType` and to pass the error through `this.config.createErrorOutput(...)`.

colombod:
fixes #2415

### DIFF
--- a/NotebookTestScript.dib
+++ b/NotebookTestScript.dib
@@ -108,6 +108,17 @@ value.Display("application/json", "text/html");
 
 #!markdown
 
+# Execute the next cell, the expression should evaluate but the cell fails with the message:
+`Error: Value 'undefinedVariableName' not found in kernel pwsh`
+
+#!csharp
+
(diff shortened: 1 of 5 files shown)


brettfo (approved):

NotebookTestScript.dib
```diff
@@ -108,6 +108,17 @@ value.Display("application/json", "text/html");
 
 #!markdown
 
+# Execute the next cell, the expression should evaluate but the cell fails with the message:
```
jonsequitur, line 111:
```suggestion
# Execute the next cell. It should fail with the message:
```

PR #2445, opened 2022-11-07
Adding more tags

colombod:


jonsequitur (approved):

AbhitejJohn (approved):

src/dotnet-interactive-vscode-insiders/package.json
```diff
@@ -32,6 +32,8 @@
     ".NET Core",
```
AbhitejJohn, line 32: Would adding the list of kernel names we support as tags improve visibility as well?

colombod: not sure because kernels will also be dynamic. some tags are added based on some languages we support. Not sure what happens if we over tag it ...

PR #2448, opened 2022-11-07
Routing slip improvement

colombod:
Events and Commands have different routing slip behaviours


src/Microsoft.DotNet.Interactive.Tests/CommandRoutingSlipTests.cs
```diff
@@ -0,0 +1,159 @@
...
+    public void the_uri_array_contains_only_stamped_kernel_uri()
+    {
+        var routingSlip = new CommandRoutingSlip();
+        routingSlip.StampAsArrived(new Uri("kernel://pid-1234"));
+        routingSlip.Stamp(new Uri("kernel://pid-1234"));
+        routingSlip.StampAsArrived(new Uri("kernel://pid-5678"));
+        routingSlip.Stamp(new Uri("kernel://pid-5678"));
+        routingSlip.StampAsArrived(new Uri("kernel://pid-91011"));
+
+        routingSlip.ToUriArray().Should().ContainInOrder(
```
jonsequitur, line 39: We'll maybe want to name this in a more informative way.

src/Microsoft.DotNet.Interactive.Tests/CommandRoutingSlipTests.cs
```diff
@@ -0,0 +1,159 @@
...
+        toBeAppended.FullStamp(new Uri("kernel://pid-4321"));
+        toBeAppended.FullStamp(new Uri("kernel://pid-8765"));
+
+        original.Append(toBeAppended);
+
+        original.ToUriArray().Should().ContainInOrder(
+            new Uri("kernel://pid-1234"),
+            new Uri("kernel://pid-5678"),
+            new Uri("kernel://pid-4321"),
+            new Uri("kernel://pid-8765"));
```
jonsequitur, line 94: These numbers make this unnecessarily hard to read. How about something like `kernel://1`, etc.?

src/Microsoft.DotNet.Interactive.Tests/CommandRoutingSlipTests.cs
```diff
@@ -0,0 +1,159 @@
...
+    }
+
+    [Fact]
+    public void can_append_a_routingSlip_to_another_skipping_entries_if_the_other_contains_it()
```
jonsequitur, line 98: Why this behavior? This looks logically inconsistent with tests that throw if you try to stamp more than once.

src/Microsoft.DotNet.Interactive.Tests/CommandRoutingSlipTests.cs
```diff
@@ -0,0 +1,159 @@
...
+        toBeAppended.StampAsArrived(new Uri("kernel://pid-8765"));
+
+        original.Append(toBeAppended);
+
+        original.ToUriArray().Should().NotContain(
+            new Uri("kernel://pid-8765"));
+    }
+
+    [Fact]
+    public void fails_to_append_a_routingSlip_to_another_if_they_do_not_start_with_same_uris()
```
jonsequitur, line 141: I would be more specific than "fails" in naming the test, and say that it throws.

src/Microsoft.DotNet.Interactive/Connection/KernelCommandEnvelope.cs
```diff
@@ -187,7 +187,8 @@ public static IKernelCommandEnvelope Deserialize(JsonElement json)
             {
                 var uri = new Uri(routingSlipItem.GetString(), UriKind.Absolute);
 
-                command.TryAddToRoutingSlip(uri);
+                command.RoutingSlip.StampAsArrived(uri);
+                command.RoutingSlip.Stamp(uri);
```
jonsequitur, line 191: This looks wrong. I would think that on deserialization, a command will frequently not be completed yet. 

Can these two states be distinguished after deserialization?

src/Microsoft.DotNet.Interactive/Kernel.cs
```diff
@@ -638,7 +636,16 @@ protected internal void PublishEvent(KernelEvent kernelEvent)
                 throw new ArgumentNullException(nameof(kernelEvent));
             }
 
-            kernelEvent.TryAddToRoutingSlip(this.GetKernelUri());
+            var kernelUri = this.GetKernelUri();
+            if (!kernelEvent.RoutingSlip.StartsWith(kernelUri))
```
jonsequitur, line 640: This implies that cycles are allowed. Is that a bug?

src/Microsoft.DotNet.Interactive/RoutingSlipBase.cs
```diff
@@ -0,0 +1,88 @@
...
+    
+    protected class Entry
+    {
+        public Uri Uri { get; set; }
+        public bool Completed { get; set; }
+    }
+
+    public abstract void Stamp(Uri uri);
+
+    public Uri[] ToUriArray()
```
jonsequitur, line 33: I'm not sure this method should exist because it seems to lend itself to externalizing the core responsibilities of this class.

src/Microsoft.DotNet.Interactive/RoutingSlipBase.cs
```diff
@@ -0,0 +1,88 @@
...
+    {
+        var entries = _entries.Where(e => e.Completed).Select(e => e.Uri).ToArray();
+        return entries;
+    }
+
+    public bool Contains(Uri uri)
+    {
+        return _entries.Any(e => e.Uri == uri);
```
jonsequitur, line 41: We should look at more performant ways to do this given how hot a path this will be.

src/Microsoft.DotNet.Interactive.Tests/CommandRoutingSlipTests.cs
colombod: doing it

src/Microsoft.DotNet.Interactive.Tests/Connection/SerializationTests.cs
```diff
@@ -233,23 +235,22 @@ public static IEnumerable<object[]> Events()
 
             IEnumerable<KernelEvent> events()
             {
-                var submitCode = new SubmitCode("123");
 
                 yield return new CodeSubmissionReceived(
-                    submitCode);
```
jonsequitur, line 239: Any reason you inlined this?

colombod: because reusing same command over and over is making problems with the routing slip

jonsequitur: Interesting. That's probably a good thing. What kind of problem?

colombod:
It is reused Dr Diego Colombo PhDOn 9 Nov 2022, at 19:13, Jon Sequeira ***@***.***> wrote:﻿
@jonsequitur commented on this pull request.



In src/Microsoft.DotNet.Interactive.Tests/Connection/SerializationTests.cs:
>  
                 yield return new CodeSubmissionReceived(
-                    submitCode);

Interesting. That's probably a good thing. What kind of problem?

jonsequitur (approved):

PR #2466, opened 2022-11-14
register everything for disposal

brettfo:
This _should_ clean up the processes that VS Code launches.


src/Microsoft.DotNet.Interactive/KernelExtensions.cs
```diff
@@ -580,5 +580,16 @@ public static T LogEventsToPocketLogger<T>(this T kernel)
                 }
             }
         }
+
+        public static bool TryRegisterForDisposal<T>(this Kernel kernel, T candidateDisposable)
```
jonsequitur, line 584: This is really just a convenience method and shouldn't be in the public API.

colombod (approved):

PR #2471, opened 2022-11-15
better api surface for routing slips

colombod:


src/Microsoft.DotNet.Interactive.Tests/CommandRoutingSlipTests.cs
```diff
@@ -25,7 +25,9 @@ public void can_be_stamped_with_kernel_uri_on_arrival()
     {
         var routingSlip = new CommandRoutingSlip();
         routingSlip.StampAsArrived(new Uri("kernel://1"));
-        routingSlip.StartsWith(new Uri("kernel://1/?tag=arrived")).Should().BeTrue();
+        var other = new CommandRoutingSlip();
+        other.StampAsArrived(new Uri("kernel://1"));
+        routingSlip.StartsWith(other).Should().BeTrue();
```
brettfo, line 30: Re-add `?tag=arrived`, because if it _didn't_ get stamped, `"kernel://1".StartsWith("kernel://1")` still returns true.

colombod: not sure I follow. the `StampAsArrived` adds the `arrived` tag to the query. We removed all api with uris and strings, `StartsWith` now requires another `RoutingSlip` and checks using the absolute uri including the query

jonsequitur (approved):

PR #2482, opened 2022-11-18
show help page if dotnet sdk version is insufficient

brettfo:
If the user's SDK version wasn't sufficient an error toast is shown indicating that, but it can go unnoticed and isn't really actionable.  This PR will show a rendered markdown file that is more visible, has more information, and has actionable links.  Eventually this feature can be expanded to include a host of various fixes.

This will be particularly useful given that we're about to ship a version of the extension that will require a newer SDK.

Example:
image

Future work will be to detect multiple different issues and dynamically build a help page that can be used as a check list, but since the only issue we currently detect is a bad SDK, I opted to not add that complexity just yet.


claudiaregio:
NOICE SEND IT

colombod (approved):

PR #2500, opened 2022-11-22
Fix input request

colombod:


brettfo (commented):
let's add a test for this

brettfo (approved):

PR #2532, opened 2022-11-30
http request kernel

colombod:
image
MicrosoftTeams-image (4)

### DIFF
--- a/src/Microsoft.DotNet.Interactive.HttpRequest.Tests/HttpRequestKernelTests.cs
+++ b/src/Microsoft.DotNet.Interactive.HttpRequest.Tests/HttpRequestKernelTests.cs
@@ -239,6 +239,33 @@ Content-Type: application/json
     }
 
     [Fact]
+    public async Task comments_can_be_placed_before_a_variable_expanded_request()
+    {
+        HttpRequestMessage request = null;
+        var handler = new InterceptingHttpMessageHandler((message, _) =>
+        {
+            request = message;
+            var response = new HttpResponseMessage(HttpStatusCode.OK);
+            return Task.FromResult(response);
+        });
+        var client = new HttpClient(handler);
+        using var kernel = new HttpRequestKernel(client: client);
+        kernel.SetValue("theHost", "example.com");
+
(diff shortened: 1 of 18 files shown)


src/Microsoft.DotNet.Interactive.HttpRequest.Tests/HttpRequestKernelTests.cs
```diff
@@ -0,0 +1,256 @@
...
+        using var kernel = new HttpRequestKernel(client: client);
+        kernel.BaseAddress = new Uri("http://baseAddress.com");
+
+        var result = await kernel.SendAsync(new SubmitCode("get  https://{{host}}:1200/endpoint"));
+
+        var events = result.KernelEvents.ToSubscribedList();
+
+        events.Should().NotContainErrors();
+
+        request.RequestUri.Should().Be("https://baseaddress.com:1200/endpoint");
```
brettfo, line 129: I'm not understanding this one.  Shouldn't the result be `https://http://baseAddress.com:1200/endpoint`?

colombod:
Host uses only the host part Dr Diego Colombo PhDOn 30 Nov 2022, at 17:55, Brett V. Forsgren ***@***.***> wrote:﻿
@brettfo commented on this pull request.



In src/Microsoft.DotNet.Interactive.HttpRequest.Tests/HttpRequestKernelTests.cs:
> +        {
+            request = message;
+            return Task.FromResult(response);
+        });
+        var client = new HttpClient(handler);
+        using var kernel = new HttpRequestKernel(client: client);
+        kernel.BaseAddress = new Uri("http://baseAddress.com");
+
+        var result = await kernel.SendAsync(new SubmitCode("get  https://{{host}}:1200/endpoint"));
+
+        var events = result.KernelEvents.ToSubscribedList();
+
+        events.Should().NotContainErrors();
+
+        request.RequestUri.Should().Be("https://baseaddress.com:1200/endpoint");

I'm not understanding this one.  Shouldn't the result be https://http://baseAddress.com:1200/endpoint?

PR #2550, opened 2022-12-06
modify list

colombod:


src/Microsoft.DotNet.Interactive.Documents/ParserServer/NotebookParserServer.cs
```diff
@@ -33,6 +33,7 @@ public NotebookParserServer(TextReader input, TextWriter output)
         new("sql", languageName: "SQL"),
         new("kql", languageName: "KQL"),
         new("mermaid", languageName: "Mermaid"),
+        new("httpRequest", languageName: "HTTP Request"),
```
brettfo, line 36: Double check in the kennel itself, but the language name should be "http".

brettfo (changes requested):

src/Microsoft.DotNet.Interactive.Documents/ParserServer/NotebookParserServer.cs
colombod: correct''!

PR #2552, opened 2022-12-07
Fix Object formatters link

WeihanLi:


AbhitejJohn (approved):
Thank you for fixing this up.

PR #2569, opened 2022-12-19
fix importing notebooks

colombod:
importing notebooks with markdown cells adds those to the ui.

image

This will add all markdown cells after the cell executing the import and all outputs that the imported notebook will generate

### DIFF
--- a/src/Microsoft.DotNet.Interactive.Tests/ImportNotebookTests.cs
+++ b/src/Microsoft.DotNet.Interactive.Tests/ImportNotebookTests.cs
@@ -6,6 +6,7 @@ using System.IO;
 using System.Linq;
 using System.Threading.Tasks;
 using FluentAssertions;
+using Microsoft.DotNet.Interactive.Commands;
 using Microsoft.DotNet.Interactive.CSharp;
 using Microsoft.DotNet.Interactive.Documents;
 using Microsoft.DotNet.Interactive.Documents.Jupyter;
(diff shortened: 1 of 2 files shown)


colombod:
@marckruzik I still see a problem as the markdown cell will be added below the cell executing the `#!import` command. This will cause to see all other output events just under the `#!import` and then all the markdown after that, this  looks confusing.

Can you describe the scenario you are after?

marckruzik:
> Can you describe the scenario you are after?

I saw that markdown cells are causing an error display, so I supposed (maybe wrongly) that markdown should be displayed as well. If markdown is not displayed, that's fine by me.

src/Microsoft.DotNet.Interactive.Tests/ImportNotebookTests.cs
```diff
@@ -84,4 +85,71 @@ public async Task It_imports_and_runs_well_known_polyglot_file_formats(string no
             ((ReturnValueProduced)returnedValues[i]).Value.Should().Be(results[i]);
         }
     }
+
+    [Theory]
+    [InlineData(".ipynb")]
+    [InlineData(".dib")]
+    public async Task It_imports_and_forwards_markdown_content_to_frontend(string notebookExt)
```
jonsequitur, line 92: I'm not sure I understand the test name here. What if there is no frontend?

colombod:
I think this is the right shape:
image

brettfo (approved):

src/Microsoft.DotNet.Interactive/KernelExtensions.cs
```diff
@@ -93,15 +93,29 @@ public static TKernel UseImportMagicCommand<TKernel>(this TKernel kernel)
...
                 foreach (var element in document.Elements)
                 {
-                    await kernel.RootKernel.SendAsync(new SubmitCode(element.Contents, element.KernelName));
+                    switch (element.KernelName?.ToLowerInvariant())
+                    {
+                        case "markdown":
+                            var @event = new DisplayedValueProduced(element.Contents, context.Command, new[]
+                            {
+                                new FormattedValue("text/markdown", element.Contents),
+                                new FormattedValue(PlainTextFormatter.MimeType, element.Contents)
```
jonsequitur, line 110: I don't see the plain text output being tested, and I'm not convinced this is the correct value for it. I could make the argument that the plain text version should remove the formatting characters from the Markdown, e.g. convert it to HTML and then grab the inner text from the HTML.

src/Microsoft.DotNet.Interactive/KernelExtensions.cs
```diff
@@ -93,15 +93,29 @@ public static TKernel UseImportMagicCommand<TKernel>(this TKernel kernel)
...
+            async (FileInfo notebookFile, KernelInvocationContext context) =>
             {
                 var document = await InteractiveDocument.LoadAsync(
                     notebookFile,
                     CreateKernelInfos(kernel.RootKernel as CompositeKernel));
 
                 foreach (var element in document.Elements)
                 {
-                    await kernel.RootKernel.SendAsync(new SubmitCode(element.Contents, element.KernelName));
+                    switch (element.KernelName?.ToLowerInvariant())
```
jonsequitur, line 104: At this point this should be based on the language name, not the kernel name.

src/Microsoft.DotNet.Interactive/KernelExtensions.cs
colombod: Makes sense

src/Microsoft.DotNet.Interactive/KernelExtensions.cs
colombod: Correct, will do this in a new pr

PR #2572, opened 2022-12-19
remove usage of dictionary

colombod:
reduce dictionary usage from tabular data resource


AbhitejJohn:
For posterity, what would this help with? Couldn't find much in the description, so my best guess is that this would help us with conversions to and from the Tabular data resource helping with data frame sharing (https://github.com/dotnet/interactive/issues/2060). Is that accurate?

PR #2573, opened 2022-12-20
Improve integration of data frame

colombod:
With this pr I am moving some code from extensionlab and from other extension into the main produce.
the goal is to provide easy transformations between DataFrame and TabularDataResource.

Also Tabular data resource serializers can perform both read and write operations


src/Microsoft.DotNet.Interactive.ApiCompatibility.Tests/ApiCompatibilityTests.Formatting_api_is_not_changed.approved.txt
```diff
@@ -292,6 +293,10 @@ Microsoft.DotNet.Interactive.Formatting.TabularData
     public System.Void Add(TableSchemaFieldDescriptor descriptor)
     public System.Boolean Contains(System.String name)
     public System.Collections.Generic.IEnumerator<TableSchemaFieldDescriptor> GetEnumerator()
+  public class TableDataFieldDescriptorsJsonConverter : Microsoft.DotNet.Interactive.Formatting.JsonConverter<TableDataFieldDescriptors>
```
jonsequitur, line 296: Does this need to be public?

src/Microsoft.DotNet.Interactive.ApiCompatibility.Tests/ApiCompatibilityTests.Formatting_api_is_not_changed.approved.txt
```diff
@@ -314,20 +323,20 @@ Microsoft.DotNet.Interactive.Formatting.TabularData
     DateTime=8
     GeoPoint=9
     GeoJson=10
-  public class TableSchemaFieldTypeConverter : Microsoft.DotNet.Interactive.Formatting.JsonConverter<TableSchemaFieldType>
+  public class TableSchemaFieldTypeJsonConverter : Microsoft.DotNet.Interactive.Formatting.JsonConverter<TableSchemaFieldType>
```
jonsequitur, line 326: Does this need to be public?

src/Microsoft.DotNet.Interactive.ApiCompatibility.Tests/ApiCompatibilityTests.Formatting_api_is_not_changed.approved.txt
```diff
@@ -314,20 +323,20 @@ Microsoft.DotNet.Interactive.Formatting.TabularData
...
     .ctor()
     public TableSchemaFieldType Read(ref System.Text.Json.Utf8JsonReader& reader, System.Type typeToConvert, System.Text.Json.JsonSerializerOptions options)
     public System.Void Write(System.Text.Json.Utf8JsonWriter writer, TableSchemaFieldType value, System.Text.Json.JsonSerializerOptions options)
+  public class TableSchemaJsonConverter : Microsoft.DotNet.Interactive.Formatting.JsonConverter<TableSchema>
```
jonsequitur, line 330: Does this need to be public?

src/Microsoft.DotNet.Interactive.ApiCompatibility.Tests/ApiCompatibilityTests.Formatting_api_is_not_changed.approved.txt
```diff
@@ -337,6 +346,10 @@ Microsoft.DotNet.Interactive.Formatting.TabularData
     .ctor(Action<T,Microsoft.DotNet.Interactive.Formatting.FormatContext> format)
     public System.String MimeType { get;}
     public System.Boolean Format(T value, Microsoft.DotNet.Interactive.Formatting.FormatContext context)
+  public class TabularDataResourceJsonConverter : Microsoft.DotNet.Interactive.Formatting.JsonConverter<TabularDataResource>
```
jonsequitur, line 349: Does this need to be public?

src/Microsoft.DotNet.Interactive.Data/TabularDataExtensions.cs
```diff
@@ -0,0 +1,59 @@
...
+using System;
+using System.Linq;
+using Microsoft.Data.Analysis;
+
+// ReSharper disable CheckNamespace
+namespace Microsoft.DotNet.Interactive.Formatting.TabularData;
+
+public static class TabularDataExtensions
+{
+    public static DataFrame ToDataFrame(this TabularDataResource tabularDataResource)
```
jonsequitur, line 13: Let's discuss before we add a new package just for a few extension methods.

colombod: The package is added as that will contain also some additional formatters for it and will be consumed by kql and sql packages. This could be part of the main Dotnet.Interactive package

src/Microsoft.DotNet.Interactive.ApiCompatibility.Tests/ApiCompatibilityTests.Formatting_api_is_not_changed.approved.txt
colombod: Suppose that as they are used as attributes they don't need to be

src/Microsoft.DotNet.Interactive.Data/TabularDataExtensions.cs
jonsequitur: If we're considering removing `TabularDataResource`, then adding a package to create a method that lets us work with both `DataFrame` and `TabularDataResource` together is just extra work that will have to be undone later.

Also, I only see this being used in tests, which tells me it can be a utility method in the test code for now.

colombod:
This method is needed in the notebooks to create a dataframe from a tabulardata resource. 

Unti we can share a dataframe directly user will need this method to create an instance of dataframe with results sets they shared out of kql/sql or python .

Is a public api Luis and I had in a notebook and then in an extension to make it easy to create data frames after doing query in sql kernel

PR #2576, opened 2022-12-22
handle the case kernelinfo si deserialised as dicittionary

colombod:
From vscode invocation the document is deserialized and the medatadata now is a dictionary. This fixes saving notebooks

### DIFF
--- a/src/Microsoft.DotNet.Interactive.Documents/InteractiveDocument.cs
+++ b/src/Microsoft.DotNet.Interactive.Documents/InteractiveDocument.cs
@@ -241,13 +241,68 @@ public class InteractiveDocument : IEnumerable
     {
         if (metadata is not null)
         {
-            if (metadata.TryGetValue("kernelInfo", out var kernelInfoObj) &&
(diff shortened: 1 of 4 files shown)


src/Microsoft.DotNet.Interactive.Documents/InteractiveDocument.cs
```diff
@@ -241,13 +241,68 @@ internal static void MergeKernelInfos(KernelInfoCollection destination, KernelIn
...
+                if (kernelInfoObj is JsonElement kernelInfoJson &&
+                    kernelInfoJson.Deserialize<KernelInfoCollection>(ParserServerSerializer.JsonSerializerOptions) is
+                        { } kernelInfoDeserialized)
+                {
+                    kernelInfo = kernelInfoDeserialized;
+                    return true;
+                }
+
+                // todo: the kernelInfo should not deserialize as a dictionary
+                if (kernelInfoObj is Dictionary<string,object> kernelInfoAsDictionary)
```
jonsequitur, line 255: Is there a test for the issue this is intended to resolve?

colombod: Just added a test that uses serialize and parse request. Serialize touches the new code. The document used in test comes from a vscode serialize request.

jonsequitur (approved):

PR #2586, opened 2023-01-03
FAQ, docs, samples updates

jonsequitur:


docs/FAQ.md
```diff
@@ -0,0 +1,224 @@
...
+
+### What is .NET Interactive?
+
+.NET Interactive is a [.NET tool](https://learn.microsoft.com/en-us/dotnet/core/tools/global-tools) ([`dotnet-interactive`](https://www.nuget.org/packages/Microsoft.dotnet-interactive)) containing a general-purpose engine for interactive programming including, but not limited to, use with notebooks. It includes suport for a number of languages, including C#, F#, PowerShell, and JavaScript, as well as the ability to load additional languages such as SQL and Python.
+
+### What is the difference between .NET Interactive and Jupyter?
+
+The .NET Interactive kernel (i.e. the `dotnet-interactive` [.NET tool](https://learn.microsoft.com/en-us/dotnet/core/tools/global-tools)) is a Jupyter-compliant kernel that can be used with Project Jupyter just like any other kernel. Like other Jupyter kernels, .NET Interactive isn't opinionated about which frontend you use. 
+
+The Polyglot Notebooks extension for VS Code isn't required to use the .NET Interactive kernel, but it does provide access to some additional functionality that isn't typical of Jupyter frontends, such as the ability to switch languages (i.e. subkernels) on a per-cell basis.
```
AbhitejJohn, line 31: It would be great if we could add a link for the polyglot notebooks reference to the section below.

docs/FAQ.md
```diff
@@ -0,0 +1,224 @@
...
+
+### What is a kernel?
```
AbhitejJohn, line 33: Perhaps adding a link to the sample kernel we have to the end of this section could provide readers with more details if they'd like to dive in a little more.

docs/FAQ.md
```diff
@@ -0,0 +1,224 @@
...
+
+### What is a frontend?
+
+The term "frontend" is often used to describe the UI for a notebook editor. Jupyter Notebook, JupyterLab, nteract Desktop, and the [Jupyter Extension](https://marketplace.visualstudio.com/items?itemName=ms-toolsai.jupyter) for VS Code and GitHub Codespaces are examples of notebook frontends.
+
+### What is a subkernel?
+
+A subkernel is a concept in .NET Interactive that describes one among many kernels within a single notebook session. Subkernels correspond to the different languages available in a Polyglot Notebook. Subkernels can be added dynamically at runtime, for example when connecting to a data source or a kernel running in a remote process.
+
+### What is a proxy kernel?
```
AbhitejJohn, line 51: Like the section above, it would be really nice if there was an example we could call out here too. Maybe that could be deferred to when we have the python kernel as well.

AbhitejJohn (commented):
This is looking great 👍

docs/FAQ.md
jonsequitur: Python isn't a proxy kernel. But any kernel can be proxied. The most common example currently is the JavaScript kernel. I can add an example explaining that.

docs/FAQ.md
jonsequitur: I'll come back and link to the more detailed API docs as I get them ready. The kernel APIs are a little more complex than can be squeezed into the FAQ.

colombod (approved):

PR #2592, opened 2023-01-06
apply the appropriate language configuration as the selected cell changes

### LINKED ISSUE #2559: Commenting-out code in cells does not work after last update
### Commenting-out code does not work after last update

Cannot comment-out code using `CTRL+/` nor `CTRL + K, CTRL + C.` There is no possibility to comment-out code other than typing `//` manually. 

Also auto closing brackets stopped working.

Tested for both C# and JavaScript cells. It worked perfectly before update to the last version.

**Version**
> Version: 1.0.360602+9bf026dabac44a6256a65168fa93dcb7e2edcca4
> Library version: 1.0.0-beta.22606.2+9bf026dabac44a6256a65168fa93dcb7e2edcca4
> Build date: 2022-12-08T07:51:24.4686906Z

- OS
    -  Windows 10
    - Windows 11

 - Frontend
    - Jupyter Notebook
    - Polyglot Notebook
    - Visual Studio Code

brettfo:
This allows the comment/uncomment keyboard shortcuts `Ctrl`+`K`,`C`/`Ctrl`+`K`,`U` to work based on the language of the current cell.  This also allows things like auto-complete to fill in `)` after a `(` is typed, etc.

This doesn't work, however, if the cursor is after a kernel selector magic command like this:

```
// this is a C# cell

#!python
print("hello from python")
```

If the user tries to comment the `print(..)` line it'll be replaced by `// print(...)` instead of `# print(...)` because we have no way of detecting the cursor's location, and thus the current language at the cursor.  If we did it would be trivial to see that we were in a `#!python` selector and to the right thing, but since polyglot cells aren't as common, this is a shortcoming we'll have to deal with.

Fixes #2559.


colombod (approved):

AbhitejJohn:
> we have no way of detecting the cursor's location, and thus the current language at the cursor

Would this be in the VS Code layer? Perhaps we could file a feature request, so this can be made simpler in the long term.

src/dotnet-interactive-vscode-insiders/grammars/http.language-configuration.json
```diff
@@ -0,0 +1,5 @@
+{
+    "comments": {
+        "lineComment": "//"
+    }
+}
```
jonsequitur, line 5: Arguably there should be brackets support here as well because of our support for interpolated variables.

brettfo: fixed

brettfo:
@AbhitejJohn Slight correction to my statement above: we have no way of detecting when the cursor location changes, so we can't respond to the user arrowing around in the file.  I'll think about exactly what we would need and submit a suggestion.

PR #2607, opened 2023-01-10
ValueInfo improvement

colombod:
The valueInfo event contains type information and formatted value to be used in variable explorer.
There is no need to query each individual variable
image

### DIFF
--- a/src/Microsoft.DotNet.Interactive.Browser/PlaywrightKernelConnector.cs
+++ b/src/Microsoft.DotNet.Interactive.Browser/PlaywrightKernelConnector.cs
@@ -153,7 +153,7 @@ Added browser kernels:
                 {
                     context.Publish(new ValueInfosProduced(new KernelValueInfo[]
                     {
-                        new(":root", typeof(ILocator))
+                        new(":root", new FormattedValue(PlainTextFormatter.MimeType,"document root"),typeof(ILocator))
                     }, request));
                     return Task.CompletedTask;
                 });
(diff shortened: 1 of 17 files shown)


src/Microsoft.DotNet.Interactive.Browser/PlaywrightKernelConnector.cs
```diff
@@ -153,7 +153,7 @@ public Task<Kernel> CreateKernelAsync(string kernelName, BrowserKernelLanguage l
                 {
                     context.Publish(new ValueInfosProduced(new KernelValueInfo[]
                     {
-                        new(":root", typeof(ILocator))
+                        new(":root", new FormattedValue(PlainTextFormatter.MimeType,"document root"),typeof(ILocator))
```
jonsequitur, line 156: I don't think types for the HTML kernel shared values makes sense yet.

src/Microsoft.DotNet.Interactive.CSharp/CSharpKernel.cs
```diff
@@ -138,7 +138,7 @@ Task IKernelCommandHandler<RequestValue>.HandleAsync(RequestValue command, Kerne
     public IReadOnlyCollection<KernelValueInfo> GetValueInfos() =>
         ScriptState?.Variables
             .GroupBy(v => v.Name)
-            .Select(g => new KernelValueInfo(g.Key, g.Last().Type))
+            .Select(g => new KernelValueInfo(g.Key, new FormattedValue(PlainTextFormatter.MimeType, g.Last().Value?.ToDisplayString(PlainTextFormatter.MimeType)), g.Last().Type))
```
jonsequitur, line 141:
```suggestion
            .Select(g => new KernelValueInfo(g.Key, new FormattedValue(PlainTextFormatter.MimeType, g.LastOrDefault().Value?.ToDisplayString(PlainTextFormatter.MimeType)), g.Last().Type))
```

src/Microsoft.DotNet.Interactive.Browser/PlaywrightKernelConnector.cs
colombod: what should be the formatted type here?

brettfo (approved):

PR #2608, opened 2023-01-10
Add environment variables to enable service logging in the MSSQL and Kusto kernels

corivera:
Currently we don't have any way to collect logs from the SQL and Kusto service processes if there are any issues. This change adds an environment variable to provide a log file to dump traces to for those services. Tracing is only enabled if the environment variables are set.

### DIFF
--- a/src/Microsoft.DotNet.Interactive.Kql/KqlKernelConnector.cs
+++ b/src/Microsoft.DotNet.Interactive.Kql/KqlKernelConnector.cs
@@ -35,8 +35,8 @@ public class KqlKernelConnector : IKernelConnector
 
         var connectionDetails = await BuildConnectionDetailsAsync();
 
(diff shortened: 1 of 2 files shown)


src/Microsoft.DotNet.Interactive.Kql/KqlKernelConnector.cs
```diff
@@ -35,7 +35,15 @@ public async Task<Kernel> CreateKernelAsync(string kernelName)
 
         var connectionDetails = await BuildConnectionDetailsAsync();
 
-        var client = new ToolsServiceClient(PathToService);
+        var serviceArgs = $"--parent-pid {Environment.ProcessId}";
```
corivera, line 38: This parent-pid argument here is the ID of the current process, so that we can close the service process after the parent process has exited. This was added to the SQL kernel a while ago, but was missed in the Kusto kernel.

PR #2614, opened 2023-01-13
`text/plain+summary` formatter

jonsequitur:
This adds support for a `text/plain+summary` formatter, to be used in the variable explorer and similar places where more concise plain text output is desirable.

This also reduces the indentation for multi-line plaintext output formatting.


src/Microsoft.DotNet.Interactive.Tests/Connection/SerializationTests.Command_contract_has_not_been_broken.approved.RequestValueInfos.json
```diff
@@ -3,6 +3,7 @@
   "id": "command-id",
   "commandType": "RequestValueInfos",
   "command": {
+    "mimeType": "text/plain\u002Bsummary",
```
brettfo, line 6: We should fix our serializer to allow the `+` character.

src/Microsoft.DotNet.Interactive.Tests/Connection/SerializationTests.Event_contract_has_not_been_broken.approved.ValueInfosProduced.json
```diff
@@ -2,36 +2,36 @@
...
+          "mimeType": "text/plain\u002Bsummary",
+          "value": "a"
         },
         "preferredMimeTypes": [
           "text/plain"
         ]
       },
       {
-        "typeName": "String",
+        "typeName": "System.Collections.Generic.List\u003CSystem.String\u003E",
```
brettfo, line 16: Similar comment as above, the serializer should allow `<` without having to be escaped.

src/Microsoft.DotNet.Interactive/Commands/SendEditableCode.cs
```diff
@@ -5,13 +5,14 @@ namespace Microsoft.DotNet.Interactive.Commands;
 
 public class SendEditableCode : KernelCommand
 {
-    public string KernelName { get; }
-    public string Code { get;  }
-
-    public SendEditableCode(string kernelName, string code, string targetKernelName = "vscode")
+    public SendEditableCode(string kernelName, string code, string targetKernelName = null)
```
brettfo, line 8: Because of this, will you run the last cell in `NotebookTestScript.dib` to make sure we didn't break this feature?

src/Microsoft.DotNet.Interactive/KernelInfo.cs
```diff
@@ -52,8 +52,12 @@ public KernelInfo(string localName, string[]? aliases = null)
...
     private string CreateDisplayName()
     {
-        var displayName = $"{LanguageName} {LanguageVersion}".Trim();
-        return displayName;
+        if (string.IsNullOrWhiteSpace(LanguageName))
+        {
+            return LocalName;
+        }
+
+        return $"{LocalName} - {LanguageName}";
```
brettfo, line 60: The local name is always displayed in VS Code, so this will lead to the C# display name being `csharp - csharp - C# Script`

src/dotnet-interactive-vscode-insiders/package.json
```diff
@@ -13,7 +13,7 @@
   "//version": "The version '42.42.42' is auto-set during CI package creation.",
   "version": "42.42.42",
   "engines": {
-    "vscode": "1.75.0-insider"
+    "vscode": "1.76.0-insider"
```
brettfo, line 16: Insiders is still 1.75, so this will break local testing.

src/dotnet-interactive-vscode/package.json
```diff
@@ -13,7 +13,7 @@
   "//version": "The version '42.42.42' is auto-set during CI package creation.",
   "version": "42.42.42",
   "engines": {
-    "vscode": "^1.74.0"
+    "vscode": "^1.75.0"
```
brettfo, line 16: Stable is still 1.74, so this will break local testing.

src/Microsoft.DotNet.Interactive/KernelInfo.cs
jonsequitur: Does seem to be the case: 
image

jonsequitur: Doesn't seem to be the case: 
image

src/Microsoft.DotNet.Interactive/Commands/SendEditableCode.cs
jonsequitur: I didn't make a significant change here, just moved the constructor to the top of the class declaration.

src/Microsoft.DotNet.Interactive.Tests/Connection/SerializationTests.Event_contract_has_not_been_broken.approved.ValueInfosProduced.json
jonsequitur: The test had a different configuration than the product. Fascinating.

brettfo (approved):

PR #2630, opened 2023-01-20
Switched to AsyncKeyedLock

MarkCiliaVincenti:
I made these changes because I'm seeing some concurrent dictionaries being filled up with semaphoreslims but I don't understand if and when they're being cleared. I have switched the code to make use of my library which adds a dependency that is not in the official NuGet package sources, so I had to modify NuGet.config just so that this could build, but if it gets approved then that change would need to be dropped and I'm not sure what the procedure is to get added to the official NuGet package sources.

### DIFF
--- a/NuGet.config
+++ b/NuGet.config
@@ -1,4 +1,4 @@
-<?xml version="1.0" encoding="utf-8"?>
+﻿<?xml version="1.0" encoding="utf-8"?>
 <configuration>
   <solution>
     <add key="disableSourceControlIntegration" value="true" />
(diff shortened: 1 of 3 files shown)


jonsequitur:
Thanks, @MarkCiliaVincenti! It sounds like you might have found a bug in the existing implementation, which I'd like to know more about. We'd like to avoid adding a dependency however. 

Can you share more information in an issue and/or suggest a more targeted fix?

MarkCiliaVincenti:
> Thanks, @MarkCiliaVincenti! It sounds like you might have found a bug in the existing implementation, which I'd like to know more about. We'd like to avoid adding a dependency however.
> 
> Can you share more information in an issue and/or suggest a more targeted fix?

When you say you want to avoid adding a dependency, is that a hard no? I see the solution already has a lot of dependencies.

I'm not too keen trying to chase and understand the problem in depth. I put this solution forward because it will fix a potential problem as well as probably decreasing memory allocations.

jonsequitur:
We're in the process of refactoring and reducing dependencies, so we wouldn't take this PR at this time.

PR #2634, opened 2023-01-23
improve variable explorer view

colombod:


AbhitejJohn:
Couple of callouts from the image you shared @colombod :
- The column header separators seem to half-fill the cell. I haven't seen other areas in VS Code that has something like that. Was there a reason we chose that route? 
- I like the Filter and Search functionality on the top toolbar. It makes our feature set more visible. Do we also offer hover tips/additional information on what each of these does? For instance, I don't fully understand what `Density` means.
-  Could we make the column widths dynamic (if they aren't already)? I see that the value column from your image takes in a lot of pixels which leaves less space for the Kernel column.
- [Ignore if we've discussed this already and come to an agreement] Having the share column in the middle of that table seems a little weird given the icon. I'd expect that column to be either at the beginning or the end. That's just me though and I'd like to hear what others think.

We've run an accessibility test pass with [Accessibility Insights](https://accessibilityinsights.io/docs/windows/overview/) for this control in the past and given the revamp, I'd suggest we'd do so again so we can ensure this is accessible to all our customers.

colombod:
> Couple of callouts from the image you shared @colombod :
> 
> * The column header separators seem to half-fill the cell. I haven't seen other areas in VS Code that has something like that. Was there a reason we chose that route?
> * I like the Filter and Search functionality on the top toolbar. It makes our feature set more visible. Do we also offer hover tips/additional information on what each of these does? For instance, I don't fully understand what `Density` means.
> * Could we make the column widths dynamic (if they aren't already)? I see that the value column from your image takes in a lot of pixels which leaves less space for the Kernel column.
> * [Ignore if we've discussed this already and come to an agreement] Having the share column in the middle of that table seems a little weird given the icon. I'd expect that column to be either at the beginning or the end. That's just me though and I'd like to hear what others think.
> 
> We've run an accessibility test pass with [Accessibility Insights](https://accessibilityinsights.io/docs/windows/overview/) for this control in the past and given the revamp, I'd suggest we'd do so again so we can ensure this is accessible to all our customers.

Density let the user change how dense the table is, for ow I have removed that item and put the share along the name (same column)
image

The columns are using flex and the value takes out all remaining space.

AbhitejJohn:
I'm not a good authority on UX look and feel but it was nice seeing separation between rows with lines like you had earlier, so its easy for me to distinguish adjacent rows. I'll let @claudiaregio comment on that though. It might be good to send this out to our UX design contact for a final review in case we'd need to change anything else. 

Agreed on the call on removing Density, that was a little confusing.

src/polyglot-notebooks-ui-components/src/contracts/types.ts
```diff
@@ -1,16 +1,17 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+export interface VariableGridRow {
+    id?: string;
+    name: string;
+    value: string;
+    typeName: string;
+    kernelDisplayName: string;
+    kernelName: string;
+    link: string;
```
jonsequitur, line 11: What does this field refer to?

src/polyglot-notebooks-ui-components/src/index.tsx
```diff
@@ -0,0 +1,44 @@
+// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+import { createRoot } from 'react-dom/client';
+import { VariableGrid } from "./components/VariableGrid";
+import './style.css';
+
+const container = document.getElementById("VariableGridContainer");
+const root = createRoot(container!);
+const data = [
```
jonsequitur, line 10: Why is the dummy data here?

src/polyglot-notebooks-ui-components/src/contracts/types.ts
colombod: gone

src/polyglot-notebooks-ui-components/package.json
```diff
@@ -0,0 +1,53 @@
+{
+  "name": "@microsoft/polyglot-notebooks-ui-components",
+  "displayName": ".NET Interactive Notebooks UI Components",
```
jonsequitur, line 3: This should also be Polyglot Notebooks, not .NET Interactive, yeah?

src/polyglot-notebooks-ui-components/src/components/VariableGrid.tsx
```diff
@@ -0,0 +1,340 @@
...
+                    <tbody>
+                        <tr>
+                            <th
+                                key={0}
+                                id={`0-0`}
+                                className="header"
+                            >
+                                Name
+                                <div
+                                    className='dragger'
```
jonsequitur, line 210: Isn't this more commonly called a "grip"?

jonsequitur (approved):

Issue #2637, opened 2023-01-24
Re-enable documentation for C# completions


jonsequitur:
This feature was disabled for performance reasons while investigating API updates to make completion documentation lazier.

PR #2638, opened 2023-01-24
disable completion documentation for C#

jonsequitur:
In many cases, completion documentation isn't used. It's a slightly hidden feature in VS Code and even when you know about it, you're not likely to read it for every completion item. A lazy evaluation or caching strategy can be used in the future to put this functionality back (#2637) but in the meantime, the performance improvement gained by removing it should benefit everyone.

### DIFF
--- a/src/Microsoft.DotNet.Interactive.Formatting/HtmlFormatter.cs
+++ b/src/Microsoft.DotNet.Interactive.Formatting/HtmlFormatter.cs
@@ -132,13 +132,6 @@ public static class HtmlFormatter
             return true;
         }),
 
-        //new HtmlFormatter<decimal>((d, context) =>
-        //{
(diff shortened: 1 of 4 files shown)


src/Microsoft.DotNet.Interactive.Formatting/HtmlFormatter.cs
```diff
@@ -217,12 +210,12 @@ static HtmlFormatter()
             return true;
         }),
 
-        // decimal should be displayed as plain text
-        new HtmlFormatter<decimal>((value, context) =>
-        {
-            FormatAndStyleAsPlainText(value, context);
-            return true;
-        }),
+        // // decimal should be displayed as plain text
```
colombod, line 213: why this it taken out?

brettfo (approved):

PR #2650, opened 2023-01-26
CSS handling multiline code

### LINKED ISSUE #2649: CSS doesn't work with multiline text
this code
```
CSS(@"
    .my-style {
        color: #FF0000;
    }
");
```

does not work

colombod:
fixes #2649

### DIFF
--- a/src/Microsoft.DotNet.Interactive/Kernel.Static.cs
+++ b/src/Microsoft.DotNet.Interactive/Kernel.Static.cs
@@ -73,7 +73,7 @@ public partial class Kernel
     public static void CSS(string content) =>
         // From https://stackoverflow.com/questions/524696/how-to-create-a-style-tag-with-javascript
         Javascript($@"
-            var css = '{content}',
+            var css = `{content}`,
             head = document.head || document.getElementsByTagName('head')[0],
             style = document.createElement('style');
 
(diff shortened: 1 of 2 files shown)


src/Microsoft.DotNet.Interactive/Kernel.Static.cs
```diff
@@ -73,7 +73,7 @@ private static async Task<string> GetInputAsync(string prompt, bool isPassword,
     public static void CSS(string content) =>
         // From https://stackoverflow.com/questions/524696/how-to-create-a-style-tag-with-javascript
         Javascript($@"
-            var css = '{content}',
+            var css = `{content}`,
```
jonsequitur, line 76: I was considering removing this method altogether, along with the `Javascript` method. People should likely be emitting `HTML` elements for this or calling a JavaScript kernel if one is available.

colombod: this is blocking some notebooks development as that is public api. I'd fix it and mark as deprecated?

colombod: but I think is very convenient shortcut to be able to sue the `CSS` helper to add styles

jonsequitur: Convenient isn't sufficient if it doesn't work consistently.

jonsequitur: I'd suggest we leave it for now, with the fix, but should discuss whether this is the right design, especially given the Pythonesque top-level function name.

jonsequitur (approved):

Issue #2652, opened 2023-01-27
#!set magic command


colombod:
`#!set` magic command will enable:
  
## Importing a value directly from input
```
#!csharp
#!set --name x --value @input:give me text
``` 
The execution will prompt the user for text and immediately declare a variable `x` with the value provided

## Importing a value from another kernel
```
#!csharp
#!set --name x --value @perl:y
``` 
The execution will fetch the value of `y` from the `perl` kernel and declare the variable `x` in the `csharp` kernel

## Store the return value from a cell execution
```
#!set x --from-result
123
```
The execution will evaluate and use the trailing expression `123` to declare the variable `x`

jonsequitur:
There's a small inconsistency here that's nagging me between these two approaches:

```console
#!set --name x --value @fsharp:y 
#!set --name x --from-result 
```

In the first case, the source of the value is indicated by the argument `@fsharp:y`. 

In the second case, the source of the value is indicated by the option `--from-result`.

If the second were to follow the pattern of the second, it would look more like this:

```csharp
#!set --name x --value @result
```

PR #2654, opened 2023-01-27
Introduce `#!set` magic command

colombod:
Implements #2652.

### DIFF
--- a/src/Microsoft.DotNet.Interactive/KernelExtensions.cs
+++ b/src/Microsoft.DotNet.Interactive/KernelExtensions.cs
@@ -213,11 +213,96 @@ public static class KernelExtensions
         return kernel;
     }
 
-    public static T UseSet<T>(this T kernel) where T : Kernel
+   
+
+    private static void HandleSetMagicCommand<T>(
+        T kernel, 
+        InvocationContext cmdLineContext, 
+        Option<bool> fromResultOption,
+        Option<string> nameOption, 
+        Option<string> fromValueOption) 
+        where T : Kernel
+    {
+        var fromResult = cmdLineContext.ParseResult.GetValueForOption(fromResultOption);
+        var valueName = cmdLineContext.ParseResult.GetValueForOption(nameOption);
+        var context = cmdLineContext.GetService<KernelInvocationContext>();
+
+        if (fromResult)
+        {
+            context.OnComplete((c) =>
+            {
+                if (c.IsFailed)
+                {
+                    return;
+                }
+
+                SetValueFromReturnValueProduced(c);
+            });
+        }
+        else
+        {
+            SetValueFromValueProduced();
+        }
+
+        void SetValueFromReturnValueProduced(KernelInvocationContext c)
+        {
+            var returnValueProducedEvents = new List<ReturnValueProduced>();
+            using var eventSubscription = context.KernelEvents.OfType<ReturnValueProduced>()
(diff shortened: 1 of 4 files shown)


jonsequitur:
Additional test cases we'll want:

- [ ] Is the behavior re: reference values consistent with `#!share` when `--mime-type` is specified?
- [ ] What happens when `--set-result` is used specifying a variable that already exists within the current kernel? Should this be an error? Overwrite?

src/Microsoft.DotNet.Interactive/KernelExtensions.cs
```diff
@@ -208,6 +209,194 @@ public static T UseLogMagicCommand<T>(this T kernel)
         return kernel;
     }
 
+    public static T UseSet<T>(this T kernel) where T : Kernel
```
jonsequitur, line 212:
```suggestion
    public static T UseSetMagicCommand<T>(this T kernel) where T : Kernel
```

jonsequitur: This naming would be more informative and consistent but on second thought, I think this functionality should just be part of `UseValueSharing`. It's the same core functionality and there's no use case for enabling one and not the other.

src/Microsoft.DotNet.Interactive/KernelExtensions.cs
```diff
@@ -208,6 +209,194 @@ public static T UseLogMagicCommand<T>(this T kernel)
...
+                    $"At least one of the options [{string.Join(", ", options.Select(o => o.Name))}] must be specified.";
+
+                return true;
+            }
+
+            return false;
+        }
+    }
+
+    private static void HandleSetMagicCommand<T>(
```
jonsequitur, line 302: This method should be leveraging and/or refactoring the existing `GetValueAndSendTo` method.

`#!share --from csharp x` should be the same code path (after interpolation is done) as `#!set --name x --value @csharp:x`.

src/Microsoft.DotNet.Interactive/KernelExtensions.cs
colombod: Just moved this as part of turning on ValueSharing, refactored some code to become reused (sendValue for example)

colombod:
> Additional test cases we'll want:
> 
> * [x]  Is the behavior re: reference values consistent with `#!share` when `--mime-type` is specified?
> * [ ]  What happens when `--set-result` is used specifying a variable that already exists within the current kernel? Should this be an error? Overwrite?

I think we should overwrite, same is already happening with running #!share. users will want to be able to refresh imported values.

jonsequitur (approved):

src/Microsoft.DotNet.Interactive/KernelExtensions.cs
```diff
@@ -211,7 +213,194 @@ public static T UseLogMagicCommand<T>(this T kernel)
...
+        {
+            nameOption,
+            fromResultOption,
+            fromValueOption,
+            mimeTypeOption
+        };
+
+        set.SetHandler(cmdLineContext =>
+        {
+            HandleSetMagicCommand(kernel, cmdLineContext, fromResultOption, nameOption, fromValueOption, mimeTypeOption);
```
jonsequitur, line 367: Since this is also being used by `#!share`, a more generalized name for `HandleSetMagicCommand` would be clearer.

Issue #1257, opened 2021-04-16
Printing nested objects should be similar to printing JSON


WhiteBlackGoose:
That's what we have for nested dictionaries now.
![image](https://user-images.githubusercontent.com/31178401/114974020-4843a400-9e8a-11eb-9312-0ca3f1641c9b.png)

Imho it should have a tree format, not a table. Similar to #1233.

jonsequitur:
Simple, common objects are also very hard to read when rendered using the nested plain text formatter:

<img width="916" alt="image" src="https://user-images.githubusercontent.com/547415/182661512-f9807759-bec3-40d9-a3d4-c96ac28236a9.png">

jonsequitur:
Changes to the plain text format (merged) and styling (in PR #2610) yields the following, which is an improvement. But we're still planning to do the expanding treeview-style view soon.

<img width="1381" alt="image" src="https://user-images.githubusercontent.com/547415/211691881-e2fcb290-f079-49cd-8ace-fb653233e5b2.png">

PR #2671, opened 2023-02-01
change default HTML formatting to a tree layout rather than a table

jonsequitur:
This addresses #1257.

Here's a preview. Feedback is welcome.

image


colombod (approved):

bradengc-github:
How do I change it back to the previous default? I used to get a useful table when displaying a list of C# objects but now it's just a collapsed list of NameSpace.ObjectName.

jonsequitur:
@bradengc-github Once you expand it, is the layout useful? Would expanding it by default help, in other words?

bradengc-github:
> Once you expand it, is the layout useful? Would expanding it by default help, in other words?

Unfortunately not. There's too much vertical scrolling to make that useful. (For an example, a list of objects with 10 properties can only display 3.5 items while expanded even on my large screen.) The table layout where object properties were automatically converted to columns was a much more convenient way to visualize things.

DavidZidar:
@jonsequitur Is there any way to use the old table formatter? This tree layout is only useful for nested objects, for most of my use cases the old table formatter was much better. Now I have to expand every row and can't easily compare rows.

jonsequitur:
For non-homogeneous objects in a sequence, the table formatter produced nested tables. Was that useful at all or is it mainly useful for comparing lists of objects of the same type?

DavidZidar:
I can only speak for myself but I nearly always just print lists of the same type. The old formatter was great imho. Nested tables might be useful, see LINQPad for prior art.

PR #2680, opened 2023-02-03
use formatted value in the value kernel

colombod:


brettfo (approved):

src/Microsoft.DotNet.Interactive/KeyValueStoreKernel.cs
```diff
@@ -182,7 +183,7 @@ void UndoSetValue()
         ValueDirectiveOptions options,
         KernelInvocationContext context)
     {
-        _values[options.Name] = value;
+        _values[options.Name] = new FormattedValue(options.MimeType ?? PlainTextFormatter.MimeType, value);
```
jonsequitur, line 186: Is there a test for the fallback to `text/plain`?

colombod: at the moment is a leap of faith, but a test is more inclusive, let me address that

PR #2686, opened 2023-02-04
Links for mermaid and kql added

surfmuggle:
As a convenience i added links for kql and mermaid because i assume that these are lesser well known

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,9 +9,9 @@
 - PowerShell
 - JavaScript
 - SQL 
-- KQL (Kusto Query Language)
+- KQL ([Kusto Query Language](https://learn.microsoft.com/en-us/azure/data-explorer/kusto/query/))
 - HTML*
-- Mermaid*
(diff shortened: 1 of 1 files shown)


surfmuggle:
I wonder if the interactive-ci is required for a change to a markdown readme file?
If the ci is not needed when an md file is changed then would it be possible to avoid energy consumption?

image

jonsequitur (approved):

jonsequitur:
Thanks, @surfmuggle!

> I wonder if the interactive-ci is required for a change to a markdown readme file?
> If the ci is not needed when an md file is changed then would it be possible to avoid energy consumption?

Good point. @brettfo, are you aware of any tools in the Arcade toolbox for this? 

Here's one approach that a quick search turned up:

https://pumpingco.de/blog/run-an-azure-pipelines-job-only-if-source-code-has-changed/

PR #2689, opened 2023-02-06
Enforce a minimum version for Kusto and SQL Service dotnet tools.

### LINKED ISSUE #2681: Can't connect to Kusto
All good yesterday. Today I open an notebook and can't connect to Kusto. Tried this on 2 different machines.  Wondering what changed. Any pointers? 

`Error: System.InvalidOperationException: MicrosoftKustoServiceLayer failed to start properly. Exit code: 0.
 ---> StreamJsonRpc.ConnectionLostException: The JSON-RPC connection with the remote party was lost before the request could complete.
   at StreamJsonRpc.JsonRpc.InvokeCoreAsync(JsonRpcRequest request, Type expectedResultType, CancellationToken cancellationToken)
   at StreamJsonRpc.JsonRpc.InvokeCoreAsync[TResult](RequestId id, String targetName, IReadOnlyList`1 arguments, IReadOnlyList1 positionalArgumentDeclaredTypes, IReadOnlyDictionary`2 namedArgumentDeclaredTypes, CancellationToken cancellationToken, Boolean isParameterObject)
   at Microsoft.DotNet.Interactive.SqlServer.ToolsServiceClient.ConnectAsync(ConnectParams connectParams)
   --- End of inner exception stack trace ---
   at Microsoft.DotNet.Interactive.SqlServer.ToolsServiceClient.ConnectAsync(ConnectParams connectParams)
   at Microsoft.DotNet.Interactive.Kql.ToolsServiceClientExtensions.ConnectAsync(ToolsServiceClient serviceClient, Uri ownerUri, KqlConnectionDetails kqlDetails)
   at Microsoft.DotNet.Interactive.Kql.MsKqlKernel.ConnectAsync()
   at Microsoft.DotNet.Interactive.Kql.KqlKernelConnector.CreateKernelAsync(String kernelName)
   at Microsoft.DotNet.Interactive.Kql.ConnectKqlCommand.ConnectKernelAsync(KernelInvocationContext context, InvocationContext commandLineContext)
   at Microsoft.DotNet.Interactive.CompositeKernel.<>c__DisplayClass26_0.<<AddKernelConnector>b__0>d.MoveNext() in D:\a\_work\1\s\src\Microsoft.DotNet.Interactive\CompositeKernel.cs:line 329`

corivera:
A while back, I checked in a version bump for the dotnet tools we install as part of starting up the Kusto and SQL kernels. However, I forgot to add a mechanism for enforcing a minimum version of those tools, which is a problem because we changed how we launch the tool in the Kusto kernel by adding an extra argument, which is not supported in the previous version of the tool. This change checks if the package is installed, and then whether it needs to be updated if it's already present. 

I verified the following scenarios:
1. Old 1.0.0 tool installed for Kusto and SQL
2. Current 1.1.0 tool installed for both
3. No version of either tool installed

I also made sure to keep the changes contained to the KQL and SqlServer projects, so we only need to update those nuget packages.

Fixes #2681

### DIFF
--- a/src/Microsoft.DotNet.Interactive.SqlServer/Utils.cs
+++ b/src/Microsoft.DotNet.Interactive.SqlServer/Utils.cs
@@ -69,7 +69,7 @@ public static class Utils
             });
     }
 
-    public static async Task CheckAndInstallGlobalToolAsync(string toolName, string minimumVersion)
+    public static async Task CheckAndInstallGlobalToolAsync(string toolName, string minimumVersion, string nugetPackage)
     {
         var dotnet = new Dotnet();
         var installedGlobalTools = await Utils.GetGlobalToolListAsync();
@@ -88,7 +88,7 @@ public static class Utils
         });
         if (!toolInstalled)
         {
-            var commandLineResult = await dotnet.ToolInstall("Microsoft.SqlServer.KustoServiceLayer.Tool", null, null, minimumVersion);
+            var commandLineResult = await dotnet.ToolInstall(nugetPackage, null, null, minimumVersion);
             commandLineResult.ThrowOnFailure();
         }
     }
--- a/src/Microsoft.DotNet.Interactive.SqlServer/Utils.cs
+++ b/src/Microsoft.DotNet.Interactive.SqlServer/Utils.cs
@@ -41,8 +41,7 @@ public static class Utils
     public static async Task<IEnumerable<DotnetToolInfo>> GetGlobalToolListAsync()
     {
         var dotnet = new Dotnet();
-        var args = "tool list --global";
-        var result = await dotnet.Execute(args);
+        var result = await dotnet.Execute("tool list --global");
         if (result.ExitCode != 0)
         {
             return new DotnetToolInfo[0];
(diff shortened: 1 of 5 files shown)


src/Microsoft.DotNet.Interactive.SqlServer/Utils.cs
```diff
@@ -1,8 +1,21 @@
...
 
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using Microsoft.DotNet.Interactive.Utility;
+
 namespace Microsoft.DotNet.Interactive.SqlServer;
 
+public class DotnetToolInfo {
```
jonsequitur, line 12: It's best to keep this internal.

src/Microsoft.DotNet.Interactive.SqlServer/Utils.cs
```diff
@@ -1,8 +1,21 @@
...
+    public string PackageId { get; set; }
+    public string PackageVersion { get; set; }
+    public string CommandName { get; set; }
+
+}
+
 public static class Utils
```
jonsequitur, line 19: Can this be made internal?

src/Microsoft.DotNet.Interactive.SqlServer/Utils.cs
corivera: Done

src/Microsoft.DotNet.Interactive.SqlServer/Utils.cs
corivera: Done

jonsequitur (approved):

PR #2699, opened 2023-02-08
`KernelCommandResult.KernelEvents` type change

jonsequitur:
On the way to improving diagnostics, we decided to change `KernelCommandResult.KernelEvents` from `IObservable<KernelEvent>` to `IReadOnlyList<KernelEvent>`, since this fits the common usage pattern much more closely.

In practice, the observable returned by `KernelCommandResult.KernelEvents` is always fully populated by the time the `CommandResult` is returned from `SendAsync`. Certain internal methods respond to the emitted events as they're produced but the external user of the API will never see them until the sequence is completed, which is why it was implemeted as a `ReplaySubject`.

This change should make the API clearer and easier to use.


colombod (approved):

shyamnamboodiripad:
@jonsequitur This change ended up breaking the workbench prototype. Not a significant issue and the fix was obviously simple (and a definite improvement considering that `KernelEvents` was always populated before `SendAsync` returns a `CommandResult`). But as we have been discussing, I think it would be great if we can front load as many such API changes as possible before we start consuming the API in more places :)

PR #2708, opened 2023-02-09
add tests for submit code and it's results 

shibbas:


src/Microsoft.DotNet.Interactive.Jupyter.Tests/JupyterKernelCommandTests.cs
```diff
@@ -476,4 +476,64 @@ public async Task can_submit_code_and_get_update_display_produced(JupyterConnect
 
         options.SaveState();
     }
+
+    [Theory]
+    [JupyterHttpTestData("print (\"test\")", 3, "Docstring:\nprint(value, ..., sep=' ', end='\\n', file=sys.stdout, flush=False)\n\nPrints the values to a stream, or to sys.stdout by default.\n", KernelSpecName = "python3", AllowPlayback = RECORD_FOR_PLAYBACK)]
```
jonsequitur, line 481: Not critical by any means but I've started using C# raw strings for things like this. I think they're easier to read given they don't need escaping, especially in multi-line cases.

jonsequitur (approved):

PR #2712, opened 2023-02-09
allow persisting default values for new notebooks

### LINKED ISSUE #1883: Contribute to New File... menu item
When creating a new notebook in VS Code you can go through the New File...-> (item) menu. 
We should contribute a Polyglot notebook option 

image

### LINKED ISSUE #2024: Allow default settings for .ipynb vs. .dib and default setting for default new notebook language
When creating a new notebook through the command palette, a user always has to select whether they want to use the .ipynb format or the .dib format. 

Would be nice if there were a setting that allowed users to select a default file type such that they don't have to select every time they create a new notebook. For users who stick to one file type (my assumption is most do this), it would reduce the creation of notebooks by one click. 

### Currently: 
image
image
image

### With setting: 
image
image

brettfo:
The command `Polyglot Notebook: Create new blank notebook` was renamed to `Polyglot Notebook: Create default notebook` and it's been modified to use user-saved defaults for the notebook type ('.dib', '.ipynb') and the default language/kernel ('csharp', 'fsharp', etc.).

If those defaults have already been set, the user experience is that they invoke the command and they're immediately given a new untitled notebook with the specified type and language.

If those defautls have _not_ been set, the user is asked (like the old experience), but then a toast is shown in the bottom corner that allows them to persist it.  If they choose to do so, the selected values are set as the default, but the user is then immediately shown the settings UI to change the values if desired.

image

Values saved in UI view:

image

N.b., the two values (type and language) are stored separately and it's possible for the user to, e.g., _not_ specify the notebook type, but to set C#.  In this example, they'll be asked each time if they want .dib or .ipynb, but once that selection is made C# will be auto-selected.

I also added a new command, `Polyglot Notebook: Create new blank notebook` which will always prompt the user for both values, even if they've already set one.  In this case, their saved values are not overwritten.

I've also added the "new" command (the one that honors defaults) to the File -> New menu:

image

Fixes #1883.
Fixes #2024.

### DIFF
--- a/src/polyglot-notebooks-vscode-common/src/commands.ts
+++ b/src/polyglot-notebooks-vscode-common/src/commands.ts
@@ -179,39 +179,89 @@ export function registerFileCommands(context: vscode.ExtensionContext, parserSer
         'Jupyter Notebooks': ['ipynb'],
     };
 
-    context.subscriptions.push(vscode.commands.registerCommand('polyglot-notebook.newNotebook', async () => {
-        // offer to create either `.dib` or `.ipynb`
-        const newDibNotebookText = `Create as '.dib'`;
-        const newIpynbNotebookText = `Create as '.ipynb'`;
-        const selected = await vscode.window.showQuickPick([newDibNotebookText, newIpynbNotebookText]);
-        switch (selected) {
-            case newDibNotebookText:
-                await vscode.commands.executeCommand('polyglot-notebook.newNotebookDib');
-                break;
-            case newIpynbNotebookText:
-                await vscode.commands.executeCommand('polyglot-notebook.newNotebookIpynb');
-                break;
-            default:
-                break;
+    async function newNotebookCommandHandler(preferDefaults: boolean): Promise<void> {
+        const extension = await getNewNotebookExtension(preferDefaults);
+        if (!extension) {
+            return;
         }
-    }));
 
-    context.subscriptions.push(vscode.commands.registerCommand('polyglot-notebook.newNotebookDib', async () => {
-        await newNotebook('.dib');
-    }));
+        const language = await getNewNotebookLanguage(preferDefaults);
+        if (!language) {
+            return;
+        }
 
-    context.subscriptions.push(vscode.commands.registerCommand('polyglot-notebook.newNotebookIpynb', async () => {
-        // note, new .ipynb notebooks are currently affected by this bug: https://github.com/microsoft/vscode/issues/121974
-        await newNotebook('.ipynb');
-        await selectDotNetInteractiveKernelForJupyter();
-    }));
+        await newNotebookWithLanguage(extension, language);
 
-    async function newNotebook(extension: string): Promise<void> {
-        const viewType = extension === '.dib'
-            ? constants.NotebookViewType
-            : constants.JupyterViewType;
+        if (preferDefaults) {
+            // if the defaults were even in play, ask the user if they want to save them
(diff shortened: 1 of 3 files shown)


src/polyglot-notebooks-vscode-common/src/commands.ts
```diff
@@ -229,12 +232,34 @@ export function registerFileCommands(context: vscode.ExtensionContext, parserSer
...
+        await newNotebookWithLanguage(viewType, kernelName, useDefaults);
+    }
+
+    async function newNotebookWithLanguage(viewType: string, kernelName: string, saveDefault: boolean | undefined): Promise<void> {
+        // save the default
+        if (saveDefault) {
+            // If we got here, the user entered via a path that expected to use a default, but it wasn't present.  Show
+            // an info box letting them know that a default can be specified.  The result isn't awaited because they'll
+            // likely want to continue with their current action.
+            promptToSaveConfigValue(`Would you like all future new notebooks to be created with the '${kernelName}' language?`, kernelName, constants.PolyglotConfigurationSectionName, 'defaultNotebookLanguage');
```
jonsequitur, line 260:
```suggestion
            promptToSaveConfigValue(`Would you like future notebooks to default to the '${kernelName}' kernel?`, kernelName, constants.PolyglotConfigurationSectionName, 'defaultNotebookLanguage');
```

jonsequitur: Not sure about my wording either so I'll explain what I'm getting at. The kernel name isn't the language name. We could go the other way from my suggestion as well and say e.g. `C# Script language.`

The word "all" seemed a little strong, since the setting can always be changed, and it doesn't add necessary information.

src/polyglot-notebooks-vscode-common/src/commands.ts
```diff
@@ -179,34 +179,37 @@ export function registerFileCommands(context: vscode.ExtensionContext, parserSer
...
-    }));
 
-    context.subscriptions.push(vscode.commands.registerCommand('polyglot-notebook.newNotebookDib', async () => {
-        await newNotebook('.dib');
-    }));
+        if (useDefaults) {
+            // If we got here, the user entered via a path that expected to use a default, but it wasn't present.  Show
+            // an info box letting them know that a default can be specified.  The result isn't awaited because they'll
+            // likely want to continue with their current action.
+            promptToSaveConfigValue(`Would you like all future new notebooks to be created with the '${selectedExtension}' extension?`, selectedExtension, constants.PolyglotConfigurationSectionName, 'defaultNotebookExtension');
```
jonsequitur, line 206:
```suggestion
            promptToSaveConfigValue(`Would you like future new notebooks to be created with the ${selectedExtension} extension?`, selectedExtension, constants.PolyglotConfigurationSectionName, 'defaultNotebookExtension');
```

src/polyglot-notebooks-vscode-common/src/commands.ts
claudiaregio: How do we feel about "Would you like future notebooks to start with __kernel language__ by default?"

colombod (approved):

src/polyglot-notebooks-vscode-common/src/commands.ts
claudiaregio: If we like the one above, following the format of the suggestion above, "Would you like future notebooks to use the *file extension* by default?"

PR #2739, opened 2023-02-14
Rename the netstandard2.0 package

shyamnamboodiripad:
Also include some additional code that we want to share.


src/dotnet-interactive/Connection/StdIoKernelConnector.cs
```diff
@@ -176,8 +176,13 @@ private void KillRemoteKernelProcess()
     {
         if (_process is { HasExited: false })
         {
-            // todo: ensure killing process tree
-            _process?.Kill(true);
+#if NETSTANDARD2_0
+            // TODO: Kill entrie process tree.
+            _process?.Kill();
```
shyamnamboodiripad, line 181: @jonsequitur I think this should be ok for now considering the set of kernels that we mainly care about in the VS scenario don't involve child processes. However, I wonder if we can use the same reflection trick you mentioned earlier (for cancellation) over here.

jonsequitur: This is probably fine, though we do have a helper class somewhere that will kill the whole process tree on older TFMs if you're worried about it.

shyamnamboodiripad: Ah do you remember where that helper lives? I just tried FAR on Process.Kill() but that didn't come up with any matches.

I did find helper code online that calls into System.Management (WMI), but I didn't want to copy that code over without more testing. (I couldn't find references to System.Management in the solution either.)

brettfo (approved):

PR #2743, opened 2023-02-16
Cancel commands when split

colombod:
image

### DIFF
--- a/src/Microsoft.DotNet.Interactive/KernelInvocationContext.cs
+++ b/src/Microsoft.DotNet.Interactive/KernelInvocationContext.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
+using System.CommandLine;
 using System.Reactive.Linq;
 using System.Reactive.Subjects;
 using System.Threading;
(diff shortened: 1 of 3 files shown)


src/Microsoft.DotNet.Interactive/KernelInvocationContext.cs
```diff
@@ -47,16 +51,22 @@ private KernelInvocationContext(KernelCommand command)
             category: nameof(KernelInvocationContext),
             logOnStart: true);
 
-        _cancellationTokenSource = new CancellationTokenSource();
+        _cancellationTokenSource =
+            _cancellationTokenSources.GetOrAdd(
+                command.GetOrCreateToken(),
+                s =>
+                {
+                    _ownsCancellationTokenSource = true;
```
jonsequitur, line 59: Does this assume the first command we see here is the original/parent?

jonsequitur (approved):

src/Microsoft.DotNet.Interactive/KernelInvocationContext.cs
colombod: the assumption is : **the first one to be used to establish will be the parent**

PR #2745, opened 2023-02-16
improve kql and sql connection experience 

colombod:


src/Microsoft.DotNet.Interactive.Kql/ConnectKqlCommand.cs
```diff
@@ -40,6 +41,14 @@ public ConnectKqlCommand(string resolvedToolsServicePath)
 
         var localName = commandLineContext.ParseResult.GetValueForOption(KernelNameOption);
 
+        var found = context?.HandlingKernel?.RootKernel.FindKernelByName($"kql-{localName}") is not null;
+
+        if (found)
+        {
+            throw new InvalidOperationException(
+                $"A kernel with name {localName} is already present. Use a different value for the {KernelNameOption.Name} option.");
```
jonsequitur, line 49:
```suggestion
                $"A kernel with name {localName} is already present. Use a different value for the --{KernelNameOption.Name} option.");
```

src/Microsoft.DotNet.Interactive.SqlServer.Tests/MsSqlConnectionTests.cs
```diff
@@ -70,6 +70,64 @@ public async Task It_can_connect_and_query_data()
...
+        result = await kernel.SubmitCodeAsync(
+            $"#!connect mssql --kernel-name adventureworks \"{connectionString}\"");
+
+        result.Events
+            .Should()
+            .ContainSingle<CommandFailed>()
+            .Which
+            .Message
+            .Should()
+            .Contain("\"A kernel with name adventureworks is already present. Use a different value for the kernel-name option.\"");
```
jonsequitur, line 128:
```suggestion
            .Contain("\"A kernel with name adventureworks is already present. Use a different value for the --kernel-name option.\"");
```

brettfo (approved):

PR #2747, opened 2023-02-16
report all errors when generating dynamic semantic tokens

brettfo:
The diff is mostly whitespace, the only real change was to wrap the entire method body in a try/catch and report anything that got thrown.


brettfo:
All other cases where this could fail would result in the entire extension not loading and VS Code reporting the exception at runtime.

brettfo:
I've also added a commit to report when no tokens were produced, but the cell text was non-empty, e.g., tokens probably _should_ have been produced.

jonsequitur (approved):

colombod (approved):

PR #2769, opened 2023-02-27
hunting ProxyKernel-over-stdio deadlocks

jonsequitur:


src/dotnet-interactive/Connection/StdIoKernelConnector.cs
```diff
@@ -137,7 +138,7 @@ public async Task<Kernel> CreateKernelAsync(string kernelName)
             proxyKernel.RegisterForDisposal(_refCountDisposable);
 
             _process.BeginOutputReadLine();
-            _process.BeginErrorReadLine();
+           // _process.BeginErrorReadLine();
```
jonsequitur, line 141: I think we need this.

colombod: opssss

PR #2785, opened 2023-03-01
simplify code types

colombod:
image

### DIFF
--- a/src/Microsoft.DotNet.Interactive.CSharp/CSharpKernel.cs
+++ b/src/Microsoft.DotNet.Interactive.CSharp/CSharpKernel.cs
@@ -63,7 +63,6 @@ public class CSharpKernel :
     {
         KernelInfo.LanguageName = "C#";
         KernelInfo.LanguageVersion = "11.0";
-        KernelInfo.LanguageName = "C#";
         KernelInfo.DisplayName = $"{KernelInfo.LocalName} - C# Script";
         _workspace = new InteractiveWorkspace();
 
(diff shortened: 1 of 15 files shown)


ocallesp (approved):

src/Microsoft.DotNet.Interactive.CSharp/CSharpKernel.cs
```diff
@@ -63,7 +63,8 @@ public CSharpKernel(string name) : base(name)
     {
         KernelInfo.LanguageName = "C#";
         KernelInfo.LanguageVersion = "11.0";
-        KernelInfo.DisplayName = "C# Script";
+        KernelInfo.LanguageName = "C#";
```
shyamnamboodiripad, line 66: Looks like we are setting the language name twice...

shyamnamboodiripad:
```suggestion
```

src/polyglot-notebooks-vscode-common/src/kernelSelectorUtilities.ts
```diff
@@ -12,49 +12,53 @@ export interface KernelSelectorOption {
...
 
-export function getKernelInfoDisplayValue(kernelInfo: contracts.KernelInfo): string {
+export function getKernelInfoDisplayValue(kernelInfo: { localName: string, displayName: string }): string {
     const localName = kernelInfo.localName;
     const displayName = kernelInfo.displayName;
     if (localName === displayName) {
         return localName;
     } else {
-        return `${kernelInfo.localName} - ${kernelInfo.displayName}`;
+        return kernelInfo.displayName;
```
shyamnamboodiripad, line 21:
```suggestion
        return displayName;
```

PR #2786, opened 2023-03-01
Fix kernel info display name

ocallesp:


colombod (approved):

src/Microsoft.DotNet.Interactive.Tests/KernelInfoTests.cs
```diff
@@ -38,12 +38,12 @@ public void When_LanguageName_is_not_set_then_DisplayName_is_LocalName()
         }
 
         [Fact]
-        public void By_default_DisplayName_is_derived_from_local_and_language_names()
+        public void By_default_DisplayName_is_derived_from_language_name()
```
jonsequitur, line 41: We shouldn't change the intention of the test.

PR #2791, opened 2023-03-03
Connect to jupyter kernels

shibbas:
This feature enables connecting to Jupyter kernels via the `#!connect jupyter` command and interact with them to perform the following: 

- Handle commands: `SubmitCode`, `RequestCompletions`, `RequestSignatureHelp`, `RequestHoverText`
- Add support for Variable sharing, including dataframes, using Jupyter Comms for python and R. This setup can be used to enable variable sharing for other Jupyter kernels that support comms as well by providing just the kernel side implementation to handle the `SendValue`, `RequestValue` and `RequestValueInfos` command. 
- Enable connecting to locally installed kernels directly over ZMQ using command: 
```
#!connect jupyter --kernel-name python --kernel-spec python3
```
This currently does not support selecting environment yet. vscode will need to start in the environment with Jupyter installed. 
- Enable connecting to remote kernels over HTTP and websocket using command:
```
#!connect jupyter --url <url_for_jupyter> --token <jupyter_server_token> --kernel-name python --kernel-spec python3
```

Known issues:
- stdin via input() is disabled.  Use the input prompt in polyglot and share the value instead: https://github.com/dotnet/interactive/blob/main/docs/input-prompts.md
- For ZMQ kernels, execution in kernel is not stopped when the cell run is canceled.


DEVELOPER-GUIDE.md
```diff
@@ -198,3 +198,48 @@ else
...
+### Setup full suite of tests to run
+
+Some tests require additional setup or will be skipped. `JupyterKernel` tests for e.g. are setup to have the same test run against a Jupyter server, directly against the Jupyter kernel over ZMQ and with a simulation of the messages. Jupyter server and Jupyter kernel tests require the following setup or will be skipped while the tests with simulation can be run without additional steps. 
+
+### Run tests with a local Jupyter Server
+
+1. Install [Jupyter server](https://docs.jupyter.org/en/latest/install.html) or [Anaconda](https://www.anaconda.com/products/distribution)
+2. [Install R kernel](https://docs.anaconda.com/anaconda/user-guide/tasks/using-r-language/) for R tests by calling 
+```
+conda install -c r r-irkernel
```
colombod, line 212: can we clarify if this is to be done by any terminal or the conda one?

DEVELOPER-GUIDE.md
```diff
@@ -198,3 +198,48 @@ else
...
+```
+3. Start the server locally as mentioned [here](https://docs.jupyter.org/en/latest/running.html)
+```
+jupyter notebook --no-browser --NotebookApp.token=<your_token> --port=8888
```
colombod, line 216: How is the user selecting / generating a token?

DEVELOPER-GUIDE.md
```diff
@@ -198,3 +198,48 @@ else
...
+--url http://localhost:8888 --token <your_token>
+```
+5. The tests will now use the environment variable to connect to your server. 
+
+### Run tests with a Jupyter Kernel over ZMQ
+
+1. Install [Anaconda](https://www.anaconda.com/products/distribution)
+2. [Install R kernel](https://docs.anaconda.com/anaconda/user-guide/tasks/using-r-language/) for R tests by calling 
+```
+conda install -c r r-irkernel
```
colombod, line 229: In the conda docs they say 
```
To run the commands below on Windows, use Start - Anaconda Prompt. On macOS or Linux, open a terminal.
```

do you think we should make this more clear in our docs?

DEVELOPER-GUIDE.md
shibbas: Updated to clarify

DEVELOPER-GUIDE.md
shibbas: You can pass in any token really. it doesn't need anything specific. I'll clarify in the docs.

src/Microsoft.DotNet.Interactive.Jupyter/Messaging/Comms/ICommTarget.cs
```diff
@@ -0,0 +1,11 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.DotNet.Interactive.Jupyter.Messaging.Comms;
+
+internal interface ICommTarget
+{
+    public string Name { get; }
+
+    public void OnCommOpen(CommAgent commAgent, object data);
```
jonsequitur, line 10: What kinds of things can be passed to this `object data` parameter?

src/Microsoft.DotNet.Interactive.Jupyter/Http/JupyterHttpConnection.cs
```diff
@@ -0,0 +1,164 @@
...
+
+        _disposables = new CompositeDisposable
+        {
+            _apiClient
+        };
+    }
+
+    public void Dispose()
+    {
+        Task.Run(ShutdownJupyterSessionsAsync).Wait();
```
jonsequitur, line 75: We shouldn't call `Wait`. Consider using `IAsyncDisposable`.

src/Microsoft.DotNet.Interactive.Jupyter/ZMQ/JupyterConnection.cs
```diff
@@ -0,0 +1,149 @@
...
+        Logger kernelLog = new(spec.Name);
+        var kernelProcess = CommandLine.StartProcess(command,
+                                                     arguments,
+                                                     null,
+                                                     o => kernelLog.Info(o),
+                                                     err => kernelLog.Error(err));
+        return kernelProcess;
+    }
+
+    public void Dispose()
```
jonsequitur, line 135: If the `Dispose` isn't always needed, then having `IJupyterConnection` implement `IDisposable` is probably not needed. Implementations can implement it (or `IAsyncDisposable`) on an as-needed basis.

src/Microsoft.DotNet.Interactive.Jupyter/Http/JupyterHttpConnection.cs
```diff
@@ -0,0 +1,164 @@
...
+            byte[] bytes = await response.Content.ReadAsByteArrayAsync();
+            var session = JsonSerializer.Deserialize<KernelSessionInfo>(bytes);
+            _activeSessions.Add(session.id);
+            var kernelApiClient = _apiClient.CreateClient($"api/kernels/{session.kernel.id}");
+
+            return new JupyterKernelHttpConnection(kernelApiClient, _authProvider);
+        }
+        catch (Exception e)
+        {
+            throw new KernelStartException(kernelSpecName, e.Message);
```
jonsequitur, line 143: I don't think we need to wrap this in a custom exception. Can we just drop the whole `try`/`catch` or are we handling `KernelStartException` somewhere?

src/Microsoft.DotNet.Interactive.Jupyter/Messaging/Comms/ICommTarget.cs
shibbas: this actually maps to the data returned by `comm_open` message and should match the Data property here. 
https://github.com/dotnet/interactive/blob/51c33c6ec96b3d6a468b5b4d603a472f69473611/src/Microsoft.DotNet.Interactive.Jupyter/Protocol/CommOpen.cs#L20

But this could be more specifically created as a dictionary of objects instead as defined by https://jupyter-client.readthedocs.io/en/stable/messaging.html#opening-a-comm. 

The dictionary could contain anything related to the target it's trying to open. I left it as object because there is a possibility that a kernel might not return a proper dictionary on comm_open. I could revisit this in a separate PR

src/Microsoft.DotNet.Interactive.Jupyter/ZMQ/JupyterConnection.cs
shibbas: will do in a follow up PR.

src/Microsoft.DotNet.Interactive.Jupyter/Http/JupyterHttpConnection.cs
shibbas: Makes sense. The context for kernelSpecName is already in the command being executed. Will remove this in a separate PR.

src/Microsoft.DotNet.Interactive.Jupyter/Http/JupyterHttpConnection.cs
shibbas: Will remove the Wait since it's already on a Task.Run. Will revisit the design.

jonsequitur (approved):

src/Microsoft.DotNet.Interactive.Jupyter/ZMQ/JupyterConnection.cs
```diff
@@ -0,0 +1,149 @@
...
+    }
+
+    public async Task<IJupyterKernelConnection> CreateKernelConnectionAsync(string kernelSpecName)
+    {
+        // find the related kernel spec for the kernel type 
+        var spec = await GetKernelSpecAsync(kernelSpecName);
+
+        if (spec is null)
+        {
+            throw new KernelStartException(kernelSpecName, "kernel not found");
```
shibbas, line 46: should be ArgumentException.

shibbas: Fixed with PR: https://github.com/dotnet/interactive/pull/2886

src/Microsoft.DotNet.Interactive.Jupyter/ZMQ/JupyterConnection.cs
```diff
@@ -0,0 +1,149 @@
...
+    private readonly Task<IReadOnlyDictionary<string, KernelSpec>> _getKernelSpecs;
+
+    public JupyterConnection(IJupyterKernelSpecModule kernelSpecModule)
+    {
+        if (kernelSpecModule is null)
+        {
+            throw new ArgumentNullException(nameof(kernelSpecModule));
+        }
+
+        _getKernelSpecs = Task.Run(() => kernelSpecModule.ListKernels());
```
shibbas, line 30: We might want to revisit this design to support the cold start scenario.

PR #2805, opened 2023-03-07
Update instructions

brettfo:

PR #2806, opened 2023-03-08
Set always authenticate npm

ocallesp:
The step 4 'Project setup' of the updated documentation https://github.com/dotnet/interactive/pull/2805, it suggests enabling authentication.


ocallesp:
This change is not needed.
If a contributor from the public tries to do anything with npm it'll force them to use that registry, which they don't want; they want to fall back to npmjs.org

PR #2807, opened 2023-03-08
Improve StdIoConnector

shyamnamboodiripad:
Also fix some bugs in the RequestKernelInfo flow and add tests.

### DIFF
--- a/src/Microsoft.DotNet.Interactive/Connection/ProxyKernel.cs
+++ b/src/Microsoft.DotNet.Interactive/Connection/ProxyKernel.cs
@@ -49,7 +49,7 @@ public sealed class ProxyKernel : Kernel
         {
             if (coe.Event is { } e)
             {
-                if (e is KernelInfoProduced {Command: NoCommand} kip && kip.KernelInfo.Uri == KernelInfo.RemoteUri)
+                if (e is KernelInfoProduced { Command: NoCommand } kip && kip.KernelInfo.Uri == KernelInfo.RemoteUri)
                 {
                     UpdateKernelInfoFromEvent(kip);
                     PublishEvent(new KernelInfoProduced(KernelInfo, e.Command));
@@ -64,17 +64,56 @@ public sealed class ProxyKernel : Kernel
         RegisterForDisposal(subscription);
     }
 
-    private void UpdateKernelInfoFromEvent(KernelInfoProduced kernelInfoProduced)
+    public ProxyKernel(
+        string name,
+        IKernelCommandAndEventSender sender,
+        IKernelCommandAndEventReceiver receiver,
+        Uri remoteUri,
+        string displayName,
+        bool isComposite,
+        string languageName,
+        string languageVersion,
+        ICollection<KernelCommandInfo> supportedKernelCommands,
+        ICollection<KernelDirectiveInfo> supportedDirectives) : this(name, sender, receiver, remoteUri) =>
+            UpdateKernelInfo(
+                displayName,
+                isComposite,
+                languageName,
+                languageVersion,
+                supportedKernelCommands,
+                supportedDirectives);
+
+    private void UpdateKernelInfo(
+        string displayName,
+        bool isComposite,
+        string languageName,
+        string languageVersion,
+        ICollection<KernelCommandInfo> supportedKernelCommands,
+        ICollection<KernelDirectiveInfo> supportedDirectives)
     {
         _requiresRequestKernelInfoOnFirstCommand = false;
-        KernelInfo.IsComposite = kernelInfoProduced.KernelInfo.IsComposite;
-        KernelInfo.LanguageName = kernelInfoProduced.KernelInfo.LanguageName;
-        KernelInfo.LanguageVersion = kernelInfoProduced.KernelInfo.LanguageVersion;
-        KernelInfo.DisplayName = kernelInfoProduced.KernelInfo.DisplayName;
-        ((HashSet<KernelDirectiveInfo>)KernelInfo.SupportedDirectives).UnionWith(kernelInfoProduced.KernelInfo.SupportedDirectives);
-        ((HashSet<KernelCommandInfo>)KernelInfo.SupportedKernelCommands).UnionWith(kernelInfoProduced.KernelInfo.SupportedKernelCommands);
+
+        KernelInfo.DisplayName = displayName;
+        KernelInfo.IsComposite = isComposite;
+        KernelInfo.LanguageName = languageName;
+        KernelInfo.LanguageVersion = languageVersion;
+        ((HashSet<KernelCommandInfo>)KernelInfo.SupportedKernelCommands).UnionWith(supportedKernelCommands);
+        ((HashSet<KernelDirectiveInfo>)KernelInfo.SupportedDirectives).UnionWith(supportedDirectives);
     }
-    
+
+    private void UpdateKernelInfoFromEvent(KernelInfoProduced kernelInfoProduced)
+    {
+        var kernelInfo = kernelInfoProduced.KernelInfo;
+
+        UpdateKernelInfo(
+            kernelInfo.DisplayName,
+            kernelInfo.IsComposite,
+            kernelInfo.LanguageName,
+            kernelInfo.LanguageVersion,
+            kernelInfo.SupportedKernelCommands,
+            kernelInfo.SupportedDirectives);
+    }
+
     private Task HandleByForwardingToRemoteAsync(KernelCommand command, KernelInvocationContext context)
     {
         if (command.OriginUri is null)
@@ -84,13 +123,13 @@ public sealed class ProxyKernel : Kernel
                 command.OriginUri = KernelInfo.Uri;
             }
         }
-       
+
         _executionContext = ExecutionContext.Capture();
         var token = command.GetOrCreateToken();
         command.GetOrCreateId();
 
         command.OriginUri ??= KernelInfo.Uri;
-        
+
         if (command.DestinationUri is null)
         {
             command.DestinationUri = KernelInfo.RemoteUri;
@@ -105,11 +144,15 @@ public sealed class ProxyKernel : Kernel
         }
 
         var targetKernelName = command.TargetKernelName;
-        command.TargetKernelName = null;
+        if (command.TargetKernelName == Name)
+        {
+            command.TargetKernelName = null;
+        }
+
         var completionSource = new TaskCompletionSource<KernelEvent>();
-       
+
         _inflight[token] = (command, _executionContext, completionSource, context);
-        
+
         ExecutionContext.SuppressFlow();
 
         EnsureKernelInfoIsLoaded(command, context);
@@ -230,9 +273,9 @@ public sealed class ProxyKernel : Kernel
                     {
                         UpdateKernelInfoFromEvent(kip);
                         var newEvent = new KernelInfoProduced(KernelInfo, kernelEvent.Command);
-                        
+
                         newEvent.RoutingSlip.ContinueWith(kip.RoutingSlip);
-                        
+
                         if (pending.executionContext is { } ec)
                         {
                             ExecutionContext.Run(ec, _ =>
--- a/src/Microsoft.DotNet.Interactive/Connection/ProxyKernel.cs
+++ b/src/Microsoft.DotNet.Interactive/Connection/ProxyKernel.cs
@@ -16,7 +16,6 @@ public sealed class ProxyKernel : Kernel
     private readonly IKernelCommandAndEventSender _sender;
     private readonly IKernelCommandAndEventReceiver _receiver;
     private ExecutionContext _executionContext;
-    private bool _requiresRequestKernelInfoOnFirstCommand;
     private string _suppressCompletionsForCommandId;
 
     private readonly Dictionary<string, (KernelCommand command, ExecutionContext executionContext, TaskCompletionSource<KernelEvent> completionSource, KernelInvocationContext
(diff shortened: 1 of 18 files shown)


src/dotnet-interactive.Tests/StdIoKernelConnectorTests.cs
```diff
@@ -0,0 +1,153 @@
...
+            fsharpProxyKernel.KernelInfo.SupportedKernelCommands.Should().BeEquivalentTo(fsharpKernelInfo.SupportedKernelCommands);
+            fsharpProxyKernel.KernelInfo.SupportedDirectives.Should().BeEquivalentTo(fsharpKernelInfo.SupportedDirectives);
+        }
+
+        [Fact]
+        public async Task it_throws_if_proxy_to_the_remote_composite_is_not_created_before_creating_a_proxy_to_a_specific_remote_subkernel()
+        {
+            var connector = CreateConnector();
+            var action = async () => await connector.CreateProxyKernelAsync(remoteName: "csharp");
+            await action.Should().ThrowAsync<InvalidOperationException>();
```
shyamnamboodiripad, line 117: ⁠I tried to implement the API you had proposed yesterday where `StdIoKernelConnector.CreateProxyKernelAsync()` would create both the root proxy (to the remote composite if it has not been created yet) in addition to the specified 'child' proxy (which is returned from the call e.g., proxy to remote csharp kernel).

This is straightforward to implement. However, if the caller only interacts with the above API (without ever explicitly calling `StdIoKernelConnector.CreateRootProxyKernel`) then the caller will never see the root proxy object. And since the process stays alive until the root proxy (and all other child proxies) are alive, this breaks the ref counted dispose logic. We end up with a zombie process because caller only knows to dispose all child proxies, but not the root proxy...

I considered the following alternatives -

1. Make the connector disposable and add a requirement on the caller to dispose it explicitly. This should in turn dispose the root proxy even if the object was never returned to the caller. The connector was also disposable before the current changes - but as we discussed, it can be confusing to have this requirement.
2. Make `StdIoKernelConnector.CreateProxyKernelAsync()` return both the root and child proxies. Perhaps this could return multiple proxies only for the very first call (i.e. in the case where a root proxy was not present before the call). That would mean changing the signature to retrun a collection of proxies (and potentially changing the name to `CreateProxyKernel*s*Async`). However, I think this also makes the API confusing (especially if we decide to preserve the `CreateRootProxyKernelAsync` API).
3. Require that caller should explicitly call `StdIoKernelConnector.CreateRootProxyKernelAsync()` to fetch the root proxy first. Calls to `StdIoKernelConnector.CreateProxyKernelAsync()` would fail (throw) unless `CreateRootProxyKernelAsync()` has been called explicitly first. In cases where there is just a single kernel on the remote side, `CreateRootProxyKernelAsync()` would be the only call that is needed.

I decided to go with 3 for this PR as I think it results in easier to understand API / semantics than the other options. But I could explore #2 above (or any other suggestions not covered above) if you feel strongly.

Thoughts?

src/Microsoft.DotNet.Interactive/Connection/ProxyKernel.cs
```diff
@@ -64,16 +64,56 @@ public sealed class ProxyKernel : Kernel
...
-    private void UpdateKernelInfoFromEvent(KernelInfoProduced kernelInfoProduced)
+    public ProxyKernel(
+        string name,
+        IKernelCommandAndEventSender sender,
+        IKernelCommandAndEventReceiver receiver,
+        Uri remoteUri,
+        string displayName,
+        bool isComposite,
+        string languageName,
+        string languageVersion,
```
jonsequitur, line 75: Do we need all of these properties in the constructor? I would think we only need the ones where the remote and the proxy might be different from one another. Or we could explore the earlier proposal which is just to make `Kernel.Name` mutable.

src/dotnet-interactive.Tests/StdIoKernelConnectorTests.cs
jonsequitur: The root kernel seems like an implementation detail in this case that we shouldn't expose. Another option could be to dispose the composite proxy internally before returning the subkernel the caller actually asked for. 

This might be more attractive because it gives the externally-facing behavior we want and we can potentially improve the implementation later if we determine a way to do this without needing the composite.

src/dotnet-interactive.Tests/StdioConnectionTests.cs
```diff
@@ -40,11 +42,10 @@ protected override Task<IKernelConnector> CreateConnectorAsync()
 
         var connector = new StdIoKernelConnector(
             command.ToArray(),
+            rootProxyKernelLocalName: "rootProxy",
```
jonsequitur, line 45: Does `CreateConnectorAsync` actually need to be `Task`-returning anymore?

src/dotnet-interactive/Connection/StdIoKernelConnector.cs
```diff
@@ -13,40 +14,73 @@
...
     private RefCountDisposable? _refCountDisposable = null;
 
-    public StdIoKernelConnector(string[] command, Uri kernelHostUri, DirectoryInfo? workingDirectory = null)
+    public string[] Command { get; }
+
+    public string RootProxyKernelLocalName { get; }
+
+    public Uri KernelHostUri { get; }
+
+    public DirectoryInfo WorkingDirectory { get; }
```
jonsequitur, line 46: Does this need to be public?

src/dotnet-interactive/Connection/StdIoKernelConnector.cs
```diff
@@ -13,40 +14,73 @@
...
+    public async Task<ProxyKernel> CreateRootProxyKernelAsync()
     {
-        ProxyKernel? proxyKernel;
-
-        if (_receiver is null)
+        if (_rootProxyKernel is null)
         {
+            using var activity = Log.OnEnterAndExit();
+
+            Debug.Assert(_receiver is null);
```
jonsequitur, line 79: We don't currently use `Debug.Assert` in this codebase. Let's have a discussion before deciding to introduce it.

src/dotnet-interactive.Tests/StdIoKernelConnectorTests.cs
shyamnamboodiripad: Good call - thanks! I'll look into that,

src/Microsoft.DotNet.Interactive/Connection/ProxyKernel.cs
colombod: I would support the mutable approach and just make sure the composite kernel collection is able to see that and react.

colombod: Pat of those parameters are already mutable on the kernel info

jonsequitur: Yeah, generally constructor parameters for mutable values don't add value.

shyamnamboodiripad: Yes, that totally makes sense. The main reason to create this constructor was so that we could also set `_requiresRequestKernelInfoOnFirstCommand` to false in the case where a `ProxyKernel` is being created based on results of `RequestKernelInfo` command. I previously had a constructor that accepted the corresponding `KernelInfo` directly - however based on Jon's feedback when we discussed this offline, we decided to switch to passing the underlying arguments instead.

I'll go ahead and remove the new constructor. However, I think it is still valuable to set `_requestKernelInfoOnFirstCommand` to `false` in the case where we already have the `KernelInfo`.

I can introduce an optional flag on the existing constructor for this OR we could remove this flag entirely and make it implicit (e.g. if `SupportedCommands` / `SupportedDirectives` is empty then trigger `RequestKernelInfo` implicitly before the first command inside `EnsureKernelInfoIsLoaded`).

Thoughts?

src/dotnet-interactive.Tests/StdioConnectionTests.cs
shyamnamboodiripad: Looks like all implementations use `return Task.FromResult<>` (even before the changes in the current PR). I'll go ahead and change the return type to `IKernelConnector` instead.

src/dotnet-interactive/Connection/StdIoKernelConnector.cs
shyamnamboodiripad: Looks like none of the existing public properties on `StdIoKernelConnector` (including `Command` and `KernelHostUri`) are used outside the class. I'll go ahead and convert these to `private` fields.

The newly added `ProcessId` property is the only one that needs to remain public so that tests can use it to validate process clean-up on disposal. Looks like we don't use `InternalsVisibleTo` in the repo - otherwise `ProcessId` could have been made internal for test use only. Let me know if you have any thoughts around leaving `ProcessId` as a public property...

src/dotnet-interactive/Connection/StdIoKernelConnector.cs
shyamnamboodiripad: I've gone ahead and removed them for now.

src/Microsoft.DotNet.Interactive/Connection/ProxyKernel.cs
jonsequitur: I think this should be explicit. We shouldn't do anything implicitly that introduces a network call, and there are good reasons (e.g. performance) to avoid waiting on the network at startup.

> I previously had a constructor that accepted the corresponding KernelInfo directly - however based on Jon's feedback when we discussed this offline, we decided to switch to passing the underlying arguments instead.

I think the only argument that really matters here is the kernel name, but that's only because we made the property immutable. Making it mutable is easier to understand I think.

src/dotnet-interactive/Connection/StdIoKernelConnector.cs
jonsequitur: > Looks like we don't use InternalsVisibleTo in the repo - otherwise ProcessId could have been made internal for test use only. 

We don't use `InternalsVisibleTo` because it can't be scoped to specific methods, so over time more and more API surface will become coupled to the consuming assemblies. 

The fact that `ProcessId` is useful in tests is design feedback that it might be generally useful as part of the public API. I can imagine wanting to know, for example, which process I need to kill if the proxied kernels need to be aggressively shut down.

shyamnamboodiripad: There is some analyzer tech in the Roslyn repo that can be used to restrict IVT to specific namespaces within an assembly. If we really wanted to, I think it should be possible to scope that down to type / method level.

That said, I can also see how avoiding IVT entirely can be beneficial to avoid unintended coupling in tests. So long as we can keep the public API surface from getting bloated in unnatural ways (and I agree that `ProcessId` does not fall in this category)...

jonsequitur: > So long as we can keep the public API surface from getting bloated in unnatural ways (and I agree that ProcessId does not fall in this category)...

We make sure not to make things public just to test them. If all functionality is tested through intentional public APIs, then tests don't become too coupled to implementation details, and keeps tests in the zone where they better support refactoring.

src/Microsoft.DotNet.Interactive/Connection/ProxyKernel.cs
shyamnamboodiripad: Thanks - I have reverted the constructor.

I haven't changed mutability of `Kernel.Name` (and `KernelInfo.LocalName`) yet since this does not appear necessary yet (since `StdIoKernelConnector` now allows to override name for any kernel when it is created).

One minor wrinkle with making them mutable is that we need to ensure that `KernelInfo.LocalName` and `Kernel.Name` should match. And this would also impact `Kernel.DisplayName` (which is currently copied over from the remote side) - so we may also need a way to update the `DisplayName` to match the new `Name` (if the `Name` appears inside `DisplayName`).

I don't think this is blocking for the current PR though - we can always improve how we handle collisions more in the future. But let me know if you feel strongly that we should make the names mutable in the current PR.

src/dotnet-interactive.Tests/StdIoKernelConnectorTests.cs
```diff
@@ -0,0 +1,153 @@
...
+            csharpProxyKernel.KernelInfo.SupportedDirectives.Should().BeEquivalentTo(csharpKernelInfo.SupportedDirectives);
+
+            fsharpProxyKernel.Name.Should().Be("fsharp2");
+            fsharpProxyKernel.KernelInfo.DisplayName.Should().Be(fsharpKernelInfo.DisplayName);
+            fsharpProxyKernel.KernelInfo.IsProxy.Should().BeTrue();
+            fsharpProxyKernel.KernelInfo.IsComposite.Should().BeFalse();
+            fsharpProxyKernel.KernelInfo.LanguageName.Should().Be(fsharpKernelInfo.LanguageName);
+            fsharpProxyKernel.KernelInfo.LanguageVersion.Should().Be(fsharpKernelInfo.LanguageVersion);
+            fsharpProxyKernel.KernelInfo.SupportedKernelCommands.Should().BeEquivalentTo(fsharpKernelInfo.SupportedKernelCommands);
+            fsharpProxyKernel.KernelInfo.SupportedDirectives.Should().BeEquivalentTo(fsharpKernelInfo.SupportedDirectives);
```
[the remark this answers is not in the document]
shyamnamboodiripad: Thanks - Fixed

src/Microsoft.DotNet.Interactive/Connection/ProxyKernel.cs
```diff
@@ -16,7 +16,7 @@ public sealed class ProxyKernel : Kernel
     private readonly IKernelCommandAndEventSender _sender;
     private readonly IKernelCommandAndEventReceiver _receiver;
     private ExecutionContext _executionContext;
-    private bool _requiresRequestKernelInfoOnFirstCommand = true;
+    private bool _requiresRequestKernelInfoOnFirstCommand;
```
jonsequitur, line 19: Is this being used for anything?

shyamnamboodiripad: I have removed this in 90b7dcf based on our offline conversation.

jonsequitur (approved):

PR #2812, opened 2023-03-09
can configure value grid string via commands

colombod:


src/polyglot-notebooks-vscode-common/src/variableExplorer.ts
```diff
@@ -150,7 +150,21 @@ class WatchWindowTableViewProvider implements vscode.WebviewViewProvider {
...
+                valueColumnHeader: 'Value',
+                typeColumnHeader: 'Type',
+                kernelNameColumnHeader: 'Kernel',
+                shareTemplate: 'Share value {value-name} from {kernel-name} kernel'
+            };
+
+            this.webview.postMessage({
+                command: 'set-rows',
+                rows,
+                localizationStrings: localizationStrings
```
ocallesp, line 166: I noticed that the messages to set rows are sending the column names as well. I was wondering if it would be possible to set the column names only once before sending the rows. It seems like the column names won't change often, so is there really a need to keep them in the state?

ocallesp (approved):

ocallesp:
Moved to PR https://github.com/dotnet/interactive/pull/2824

PR #2814, opened 2023-03-09
restrict webview communication to the specific notebook

brettfo:


src/polyglot-notebooks/src/webview/activation.ts
```diff
@@ -31,8 +32,9 @@ function configure(global: any, context: KernelMessagingApi) {
         }
     });
 
+    const guid = uuid();
     context.onDidReceiveKernelMessage((arg: any) => {
-        if (arg.envelope) {
+        if (arg.envelope && arg.guid === guid) {
```
colombod, line 37: can we call the variable not giud but something like webviewId ?

colombod (approved):

PR #2829, opened 2023-03-13
disable csharp project tests

colombod:


jonsequitur (approved):

AbhitejJohn:
just curious: what's causing us to do this? Trying to figure out if there is a test platform issue I should be looking into to unblock.

jonsequitur:
A couple of bugs in the .NET SDK preview are causing builds driven by the CSharpProjectKernel to fail.

colombod:
> just curious: what's causing us to do this? Trying to figure out if there is a test platform issue I should be looking into to unblock.

There are some regression in the CLI of 7.0.200 sdk

Issue #2831, opened 2023-03-14
Include KernelInfos for all kernels as part of KernelReady event


shyamnamboodiripad:
Currently `KernelInfo`s are sent after the `KernelReady` event has been sent. But the client has no way of knowing when / whether all `KernelInfo`s have been sent.

Making the initial `KernelInfos` available as part of `KernelReady` event would solve this problem. It would also eliminate the need to send a `RequestKernelInfo` command explicitly for this purpose.

shyamnamboodiripad:
FYI @jonsequitur @colombod Logging this for the enhancement you proposed.

PR #2834, opened 2023-03-14
Produce kernelinfo on kernel ready

colombod:


shyamnamboodiripad:
Fixes #2831

colombod:
> Fixes #2831

maybe ..

samples/stand-alone-kernels/perl/kernel.pl
```diff
@@ -83,17 +83,8 @@ sub main {
...
-        "routingSlip" => [
-            $kernelUri
-        ]
-    });
-
-    publish({
-        "eventType" => "KernelInfoProduced",
         "event" => {
-            "kernelInfo" => $kernelInfo,
+            "kernelInfos" => $kernelInfo
```
colombod, line 87: @brettfo am I doing this right?

jonsequitur (approved):

src/dotnet-interactive/Connection/StdIoKernelConnector.cs
```diff
@@ -133,14 +134,15 @@ public async Task<ProxyKernel> CreateRootProxyKernelAsync()
...
+            
+            _kernelReady = null;
             _receiver.Select(coe => coe.Event)
-                                   .OfType<KernelReady>()
+                                       .OfType<KernelReady>()
                                    .Take(1)
                                    .Subscribe(e =>
                                    {
-                                       kernelReadyReceived = true;
+                                       _kernelReady = e;
```
shyamnamboodiripad, line 144: Since this is all within the same method, could we use a local instead of the field for `kernelReady`?

We already cache all kernel infos that we have seen so far in `_remoteKernelInfos`. So, it feels a bit strange that at least some of those kernel infos would also be avbailable via the `_kernelReady` field.

src/dotnet-interactive/Connection/StdIoKernelConnector.cs
```diff
@@ -133,14 +134,15 @@ public async Task<ProxyKernel> CreateRootProxyKernelAsync()
             activity.Info("Process id: {0}", _process.Id);
 
             _receiver = KernelCommandAndEventReceiver.FromObservable(stdOutObservable);
-
-            bool kernelReadyReceived = false;
+            
+            _kernelReady = null;
             _receiver.Select(coe => coe.Event)
-                                   .OfType<KernelReady>()
+                                       .OfType<KernelReady>()
```
shyamnamboodiripad, line 140: nit: Indentation
```suggestion
                                   .OfType<KernelReady>()
```

src/dotnet-interactive/Connection/StdIoKernelConnector.cs
```diff
@@ -133,14 +134,15 @@ public async Task<ProxyKernel> CreateRootProxyKernelAsync()
             activity.Info("Process id: {0}", _process.Id);
 
             _receiver = KernelCommandAndEventReceiver.FromObservable(stdOutObservable);
-
-            bool kernelReadyReceived = false;
+
```
shyamnamboodiripad, line 137:
```suggestion

```

(removing redundant whitespace)

shyamnamboodiripad (approved):
Thanks for fixing! 👍🏾

src/dotnet-interactive/Connection/StdIoKernelConnector.cs
shyamnamboodiripad: (Looks like making the above change would also solve the nullability error that caused the build break in CI)

colombod: but the connector can be reused that is why is stored, you will see that event only when the process is started

shyamnamboodiripad: All kernel infos are already avaialble in _remoteInfos, so you could just fetch it from there (like we do inside CreateProxyKernelAsync below, could you not?

PR #2837, opened 2023-03-14
avoid sending proxies from dotnet interactive

colombod:
vscode extension gets confused when it sees the vscode and jsproxy to a kernelhost hat is not existing yes and creates double proxies to js

### DIFF
--- a/src/Microsoft.DotNet.Interactive/KernelHost.cs
+++ b/src/Microsoft.DotNet.Interactive/KernelHost.cs
@@ -129,7 +129,10 @@ public class KernelHost : IDisposable
 
         _kernel.VisitSubkernelsAndSelf(k =>
         {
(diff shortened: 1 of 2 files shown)


src/Microsoft.DotNet.Interactive/KernelHost.cs
```diff
@@ -129,7 +129,10 @@ public async Task ConnectAsync()
 
         _kernel.VisitSubkernelsAndSelf(k =>
         {
-            kernelInfos.Add(k.KernelInfo);
+            if (k.KernelInfo.IsProxy == false)
```
shyamnamboodiripad, line 132:
```suggestion
            if (!k.KernelInfo.IsProxy)
```

shyamnamboodiripad (approved):

PR #2852, opened 2023-03-21
Make constructor for KernelCommandResult public

shyamnamboodiripad:
And make it possible to initialize the object with an initial set of events.

### DIFF
--- a/src/Microsoft.DotNet.Interactive/KernelCommandResult.cs
+++ b/src/Microsoft.DotNet.Interactive/KernelCommandResult.cs
@@ -11,11 +11,14 @@ namespace Microsoft.DotNet.Interactive;
 [TypeFormatterSource(typeof(MessageDiagnosticsFormatterSource))]
 public class KernelCommandResult
 {
-    private readonly List<KernelEvent> _events = new();
+    private readonly List<KernelEvent> _events;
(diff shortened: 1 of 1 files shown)


src/Microsoft.DotNet.Interactive/KernelCommandResult.cs
```diff
@@ -11,11 +11,14 @@ namespace Microsoft.DotNet.Interactive;
 [TypeFormatterSource(typeof(MessageDiagnosticsFormatterSource))]
 public class KernelCommandResult
 {
-    private readonly List<KernelEvent> _events = new();
+    private readonly List<KernelEvent> _events;
 
-    internal KernelCommandResult(KernelCommand command)
+    public KernelCommandResult(KernelCommand command, IEnumerable<KernelEvent> events = null)
```
shyamnamboodiripad, line 16: @jonsequitur Another alternative would be to mark the constructor `protected internal` but I prefer `public` since I find the semantics of `protected internal` a bit weird :)

I have also left `AddEvent` below as `internal` (which means external consumers that are handed an instance of this type still won't have the ability modify the events collection after the fact).

jonsequitur (approved):

PR #2866, opened 2023-03-23
Improve memory usage

colombod:
This change improves memory behaviour for csharp kernel

### DIFF
--- a/src/Microsoft.DotNet.Interactive.CSharp/InteractiveWorkspace.cs
+++ b/src/Microsoft.DotNet.Interactive.CSharp/InteractiveWorkspace.cs
@@ -125,13 +125,6 @@ internal class InteractiveWorkspace : Workspace
             _previousSubmissionProjectId = AddProjectWithPreviousSubmissionToSolution(_currentCompilation,
                 scriptState.Script.Code, _workingProjectId, _previousSubmissionProjectId);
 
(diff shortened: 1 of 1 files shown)


jonsequitur (approved):

src/Microsoft.DotNet.Interactive.CSharp/InteractiveWorkspace.cs
```diff
@@ -125,13 +125,6 @@ public async Task UpdateWorkspaceAsync(ScriptState scriptState)
             _previousSubmissionProjectId = AddProjectWithPreviousSubmissionToSolution(_currentCompilation,
                 scriptState.Script.Code, _workingProjectId, _previousSubmissionProjectId);
 
-            //This fixes an issue where the performance of language services is degrading each time a new submission is run.
-            //Forces the creation of the Compilation object in the project of the previous submission
-            //If the compilation were not eagerly created, the entire solution would be compiled each
-            //time it is used to answer a language service question (ForkDocumentForLanguageServices).
```
shyamnamboodiripad, line 131: Is this a tradeoff between memory and throughput?

In other words, does removing this code mean that intellisense will be slower because "entire solution would be compiled each time it is used to answer a language service question"? Or is it safe to remove this code because the underlying problem has been fixed in Roslyn, and it is no longer necessary to materialize the compilation each time?

PR #2879, opened 2023-04-03
Fix null reference if content-type specified but no body

BillHiebert:
If the request has a content-type but no body is specified, it throws a null reference exception as it tries to set the ContentType header on the message.Content object. Simple change to add a Content in this case.
The issue is the RegEx which is trying to match lines containing headers, would sometimes return an empty collection for Body lines and would try to treat these headers. The simple fix is to also check that the collection is not empty.
I added a unit test to validate the first case.


shyamnamboodiripad (approved):
Thanks for fixing!

jonsequitur (approved):

PR #2788, opened 2023-03-02
Release 1.76

colombod:

PR #2882, opened 2023-04-05
Fix value sharing exception and deadlock

shyamnamboodiripad:


src/Microsoft.DotNet.Interactive/Connection/KernelCommandEnvelope.cs
```diff
@@ -74,7 +74,8 @@ public static void RegisterDefaults()
             [nameof(RequestInput)] = typeof(KernelCommandEnvelope<RequestInput>),
             [nameof(RequestValue)] = typeof(KernelCommandEnvelope<RequestValue>),
             [nameof(RequestValueInfos)] = typeof(KernelCommandEnvelope<RequestValueInfos>),
-            [nameof(RequestKernelInfo)] = typeof(KernelCommandEnvelope<RequestKernelInfo>)
+            [nameof(RequestKernelInfo)] = typeof(KernelCommandEnvelope<RequestKernelInfo>),
+            [nameof(SendValue)] = typeof(KernelCommandEnvelope<SendValue>)
```
shyamnamboodiripad, line 78: @jonsequitur The deadlock for value sharing turned out to be same underlying root cause as the `RequestKernelInfo` deadlock that @colombod had fixed in #2788.

There are two related problems that are not fixed in the current PR but would be nice to improve in the future -

1. The error for the missing registration (`KeyNotFoundException`) is being swallowed silently and is not logged anywhere. This was an easy catch once I attached to the `dotnet.exe` process and turned on first-chance exceptions. But considering that we have already stumbled into this twice now, it may be a good idea to log this so that we won't miss this error for new commands introduced in the future. I am not sure if we already have some place in the code where unhandled exceptions can be logged...

2. The reason this issue was not caught by `SerializationTests.All_command_types_are_round_trip_serializable` is because it calls `KernelCommandEnvelope.Create(command)` in the beginning for each command (which registers the command even when it is missing in the above list). We could change the test to check whether each command is registered first (perhaps by creating a new `KernelCommandEnvelope.IsRegistered` API). However, this would also not be 100% fool proof since the dictionaries are static (making it very easy for some unrelated test in the assembly to register the command before `All_command_types_are_round_trip_serializable` is executed).

Another idea may be to avoid the hardcoding in `KernelCommandEnvelope.RegisterDefaults` above and include all well-known commands (and events) by using reflection in the product code. That would potentially address both 1. and 2. above and make this class of bugs impossible to hit...

jonsequitur: > The error for the missing registration (KeyNotFoundException) is being swallowed silently and is not logged anywhere. 

This was the motivation for adding a diagnostics work item in the March sprint plan. It's worth opening a more specific issue for.

shyamnamboodiripad: I have logged #2884 for the specific issue above. I also jotted down some general logging improvements that could help in #2885.

src/dotnet-interactive.Tests/StdIoKernelConnectorTests.cs
```diff
@@ -120,11 +120,38 @@ public async Task it_can_create_a_proxy_kernel_to_more_than_one_remote_subkernel
                 SupportedKernelCommands = fsharpKernelInfo.SupportedKernelCommands
             };
 
+            const string installHttpKernel =
+                """
+                #r "Microsoft.DotNet.Interactive.HttpRequest.dll"
```
jonsequitur, line 125: This is unlikely to behave the way we expect it to in the context of the test project. We should discuss.

src/dotnet-interactive.Tests/StdIoKernelConnectorTests.cs
```diff
@@ -120,11 +120,38 @@ public async Task it_can_create_a_proxy_kernel_to_more_than_one_remote_subkernel
...
+                SupportedDirectives = httpKernelInfo.SupportedDirectives,
+                SupportedKernelCommands = httpKernelInfo.SupportedKernelCommands
+            };
+
             csharpProxyKernel.Name.Should().Be(csharpKernelInfo.LocalName);
             csharpProxyKernel.KernelInfo.Should().BeEquivalentTo(expectedCSharpKernelInfo);
 
             fsharpProxyKernel.Name.Should().Be("fsharp2");
             fsharpProxyKernel.KernelInfo.Should().BeEquivalentTo(expectedFSharpKernelInfo);
+
```
jonsequitur, line 152: When doing multiple assertions in a single test, it's best to add `using var _ = new AssertionScope();` beforehand. This will allow us to see all failures, not just the first.

src/dotnet-interactive.Tests/StdIoKernelConnectorTests.cs
```diff
@@ -135,6 +162,81 @@ public async Task it_throws_if_there_is_no_remote_subkernel_with_the_specified_n
...
+            result = await composite.SendAsync(new SubmitCode("#!set --name y --value @fsharp:x", "csharp"));
+            result.Events.Last().Should().BeOfType<CommandSucceeded>();
+
+            result = await composite.SendAsync(new SubmitCode("#!share --from fsharp x", "csharp"));
+            result.Events.Last().Should().BeOfType<CommandSucceeded>();
+
+            result = await composite.SendAsync(new SubmitCode("x + y", "csharp"));
+            result.Events.Last().Should().BeOfType<CommandSucceeded>();
+            var returnValueProduced = result.Events.OfType<ReturnValueProduced>().Should().ContainSingle().Which;
+            returnValueProduced.FormattedValues.Should().ContainSingle().Which.Value.Should().Contain("246");
```
jonsequitur, line 199: This can be chained. Ending the line with `.Which` is kind of confusing to read.

src/dotnet-interactive.Tests/StdIoKernelConnectorTests.cs
```diff
@@ -135,6 +162,81 @@ public async Task it_throws_if_there_is_no_remote_subkernel_with_the_specified_n
...
+        {
+            var connector = CreateConnector();
+            using var rootProxyKernel = await connector.CreateRootProxyKernelAsync();
+
+            var result = await rootProxyKernel.SendAsync(new RequestKernelInfo());
+            var kernelInfos = result.Events.OfType<KernelInfoProduced>().Select(e => e.KernelInfo);
+
+            using var _ = new AssertionScope();
+
+            var csharpKernelInfo = kernelInfos.Should().ContainSingle(i => i.LanguageName == "C#").Which;
```
jonsequitur, line 176: The intermediate sanity check assertions are probably not useful. They make for a lot more test code and make the intent of the test harder to grasp.

Assume happy path until you get to the assertions that prove the scenario named by the test.

src/dotnet-interactive.Tests/StdIoKernelConnectorTests.cs
shyamnamboodiripad: Ah it is already present above on line 95.

src/dotnet-interactive.Tests/StdIoKernelConnectorTests.cs
shyamnamboodiripad: This does seem to work (including in CI) :) But will discuss offline.

jonsequitur (approved):

PR #2887, opened 2023-04-06
don't require IJupyterConnection to be IDiposable

shibbas:
Addresses improvement feedback on not requiring IJupyterConnection to be IDisposable. Instead disposes them if they are IDisposable. 

https://github.com/dotnet/interactive/pull/2791

### DIFF
--- a/src/Microsoft.DotNet.Interactive.Jupyter/ConnectJupyterKernelCommand.cs
+++ b/src/Microsoft.DotNet.Interactive.Jupyter/ConnectJupyterKernelCommand.cs
@@ -69,7 +69,10 @@ public class ConnectJupyterKernelCommand : ConnectKernelCommand
         var localName = commandLineContext.ParseResult.GetValueForOption(KernelNameOption);
 
         var kernel = await connector?.CreateKernelAsync(localName);
-        kernel?.RegisterForDisposal(connection);
+        if (connection is IDisposable disposableConnection)
+        {
+            kernel?.RegisterForDisposal(disposableConnection);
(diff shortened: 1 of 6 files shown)


jonsequitur (approved):

jonsequitur (approved):

src/Microsoft.DotNet.Interactive.Jupyter/ConnectJupyterKernelCommand.cs
```diff
@@ -69,7 +69,10 @@ public ConnectJupyterKernelCommand AddConnectionOptions(IJupyterKernelConnection
         var localName = commandLineContext.ParseResult.GetValueForOption(KernelNameOption);
 
         var kernel = await connector?.CreateKernelAsync(localName);
-        kernel?.RegisterForDisposal(connection);
+        if (connection is IDisposable disposableConnection)
+        {
+            kernel?.RegisterForDisposal(disposableConnection);
```
jonsequitur, line 74: What can lead to kernel being null?

shibbas: There shouldn't be a case of kernel being null. I think this was just guarding but I think probably we would want to throw in that case. Will fix separately.

shibbas: fixed in follow up PR: https://github.com/dotnet/interactive/pull/2892

Issue #2884, opened 2023-04-05
Prevent silent deadlocks caused by missing command or event registration


shyamnamboodiripad:
We have now seen two deadlocks related to command / event registration that were fixed in the following PRs:
* #2788
* #2882

There are two related problems that would be nice to improve in the future -

1. The error for the missing registration (`KeyNotFoundException`) is being swallowed silently and is not logged anywhere. This was an easy catch once I attached to the dotnet.exe process and turned on first-chance exceptions. But considering that we have already stumbled into this twice now, it may be a good idea to log this so that we won't miss this error for new commands introduced in the future. I am not sure what would be a good place(s) in the code where (all?) unhandled exceptions can be logged...

2. The reason this issue was not caught by `SerializationTests.All_command_types_are_round_trip_serializable` is because it calls `KernelCommandEnvelope.Create(command)` in the beginning for each command (which registers the command even when it is missing in the above list). We could change the test to check whether each command is registered first (perhaps by creating a new `KernelCommandEnvelope.IsRegistered` API). However, this would also not be 100% fool proof since the dictionaries are static (making it very easy for some unrelated test in the assembly to register the command before `All_command_types_are_round_trip_serializable` is executed).

Another idea may be to avoid the hardcoding in `KernelCommandEnvelope.RegisterDefaults` and include all well-known commands (and events) by using reflection in the product code. That could potentially address both 1. and 2. above and make this class of bugs impossible to hit...

jonsequitur:
The silent deadlock here is unrelated to the serializer registration. The exception leading to the deadlock was not being handled in a way that would have made this more obvious (by throwing instead of silently deadlocking). The missing serialization registration was just one possible type of exception that could lead to this.

I'll have a PR for that issue shortly.

PR #2888, opened 2023-04-06
Prevent AnonymousKernelCommands from ending up inside serialized KernelEvents

shyamnamboodiripad:
AnonymousKernelCommands are not supposed to be serialized across process boundaries and should therefore not be included as part of KernelEvents.

### DIFF
--- a/src/Microsoft.DotNet.Interactive.Tests/Parsing/SubmissionParserTests.cs
+++ b/src/Microsoft.DotNet.Interactive.Tests/Parsing/SubmissionParserTests.cs
@@ -8,6 +8,7 @@ using FluentAssertions;
 using FluentAssertions.Execution;
 using Microsoft.DotNet.Interactive.Commands;
 using Microsoft.DotNet.Interactive.CSharp;
+using Microsoft.DotNet.Interactive.Events;
 using Microsoft.DotNet.Interactive.FSharp;
 using Microsoft.DotNet.Interactive.Jupyter;
 using Microsoft.DotNet.Interactive.Parsing;
@@ -337,7 +338,7 @@ let x = 123
             .Should()
             .AllSatisfy(child => rootSpan.Contains(child.Span).Should().BeTrue());
     }
-       
+
     private static SubmissionParser CreateSubmissionParser(string defaultLanguage = "csharp")
     {
         using var compositeKernel = new CompositeKernel
@@ -363,6 +364,15 @@ let x = 123
         compositeKernel.UseDefaultMagicCommands();
 
         return compositeKernel.SubmissionParser;
+    }
+
+    [Fact]
+    public async Task DiagnosticsProduced_events_always_point_back_to_the_original_command()
+    {
+        using var kernel = new CSharpKernel();
+        var command = new SubmitCode("#!unrecognized");
+        var result = await kernel.SendAsync(command);
+        result.Events.Should().ContainSingle<DiagnosticsProduced>().Which.Command.Should().BeSameAs(command);
     }
 
     [Fact]
@@ -447,7 +457,7 @@ let x = 123 // with some intervening code
 // language-specific code";
 
         MarkupTestFile.GetLineAndColumn(markupCode, out var code, out var _, out var _);
-            
+
         var command = new RequestDiagnostics(code);
         var commands = new CSharpKernel().UseDefaultMagicCommands().SubmissionParser.SplitSubmission(command);
 
--- a/src/Microsoft.DotNet.Interactive/Connection/ProxyKernel.cs
+++ b/src/Microsoft.DotNet.Interactive/Connection/ProxyKernel.cs
@@ -160,7 +160,6 @@ public sealed class ProxyKernel : Kernel
         switch (command)
         {
             case AnonymousKernelCommand:
-                return base.HandleAsync(command, context);
             case DirectiveCommand:
                 return base.HandleAsync(command, context);
         }
--- a/src/Microsoft.DotNet.Interactive/Events/KernelEvent.cs
+++ b/src/Microsoft.DotNet.Interactive/Events/KernelEvent.cs
@@ -11,7 +11,17 @@ public abstract class KernelEvent
 {
     protected KernelEvent(KernelCommand command)
     {
-        Command = command ?? throw new ArgumentNullException(nameof(command));
+        if (command is null)
+        {
+            throw new ArgumentNullException(nameof(command));
+        }
+
+        if (command is AnonymousKernelCommand)
+        {
+            throw new ArgumentException($"{nameof(command)} should not be an {nameof(AnonymousKernelCommand)}");
+        }
+
+        Command = command;
         RoutingSlip = new EventRoutingSlip();
     }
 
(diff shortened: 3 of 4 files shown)


src/Microsoft.DotNet.Interactive/Parsing/SubmissionParser.cs
```diff
@@ -124,7 +124,7 @@ public IReadOnlyList<KernelCommand> SplitSubmission(RequestDiagnostics requestDi
                                         CodeAnalysis.DiagnosticSeverity.Error,
                                         "NI0001", // QUESTION: (SplitSubmission) what code should this be?
                                         "Unrecognized magic command");
-                                    var diagnosticsProduced = new DiagnosticsProduced(new[] { diagnostic }, c);
+                                    var diagnosticsProduced = new DiagnosticsProduced(new[] { diagnostic }, originalCommand);
```
shyamnamboodiripad, line 127: There were some inadvertent whitespace fix ups - but this is the main part of the fix.

src/Microsoft.DotNet.Interactive/Events/KernelEvent.cs
```diff
@@ -11,7 +11,17 @@ public abstract class KernelEvent
 {
     protected KernelEvent(KernelCommand command)
     {
-        Command = command ?? throw new ArgumentNullException(nameof(command));
+        if (command is null)
+        {
+            throw new ArgumentNullException(nameof(command));
+        }
+
+        if (command is AnonymousKernelCommand)
```
jonsequitur, line 19: I don't think this is valid. This API isn't opinionated about serialization concerns and there are valid use cases for non-serializable command types.

shyamnamboodiripad: Ah it is legal to have non-serializable commands in events? My assumption was that the system isn't opinionated about whether events flow in-process or out-of-process and that therefore non-serializable commands should be avoided in all events (to avoid cases where an event works fine when the target kernel is in-process - but fails when the same kernel is moved out-of-process).

jonsequitur: The design opinion is that out-of-process communication isn't required or assumed.

src/Microsoft.DotNet.Interactive.Tests/Parsing/SubmissionParserTests.cs
```diff
@@ -365,6 +366,15 @@ private static SubmissionParser CreateSubmissionParser(string defaultLanguage =
         return compositeKernel.SubmissionParser;
     }
 
+    [Fact]
+    public async Task DiagnosticsProduced_events_always_point_back_to_the_original_command()
```
jonsequitur, line 370: This is intended to be generally true, so not specific to `DiagnosticsProduced`. The internal command (`AnonymousKernelCommand` in this case, but it could as easily be one of the public command types) is an implementation detail that the external caller shouldn't have to be aware of.

It might be useful to add another similar test for the split between subkernels because the same should be true for, for example:

```csharp
#!csharp
123
#!fsharp
456
```

shyamnamboodiripad: So looks like this invariant is not satisfied for all events in the above case. For example, `CodeSubmissionReceived`, `CompleteCodeSubmissionReceived`, `ReturnValueProduced` events for the repro code you shared above point back to the per-language `SubmitCode` commands created internally for C# and F# respectively. Only the final `CommandSucceded` event points back to the original `SubmitCode` command.

If I change the submission to the following then interestingly, the `DiagnosticsProduced` event for the erroneous line points to the F#-specific `SubmitCode` command with code `some error`. However, the line spans in the `DiagnosticsProduced.Diagnostics` reflect the spans in the original submission.
```
#!csharp
123
 #!fsharp
some error
```

This may be *reasonable* considering that the inner commands reflect the expected command splitting behavior - however, it also seems like internal implementation details of the callee are bleeding through to the caller.

If this is unexpected, I can log a separate bug to a) fix the behavior and b) add tests to ensure that inner commands are never exposed to the caller. Thoughts?

src/Microsoft.DotNet.Interactive/Events/KernelEvent.cs
shyamnamboodiripad: Got it. I've reverted the change and also updated #2884 to capture what we discussed offline (i.e. when an internal / non-serializable command is encountered, perhaps we should only serialize the properties present on the base `KernelCommand` type + add better tests to catch cases where we inadvertently end up including internal commands in `KernelEvent` payloads).

src/Microsoft.DotNet.Interactive.Tests/Parsing/SubmissionParserTests.cs
shyamnamboodiripad: @jonsequitur Btw the current PR should be ready to merge. The above issue needs more investigation / follow-up and I'd prefer not to hold up the current PR for that. Please approve if you agree :) Thanks!

src/Microsoft.DotNet.Interactive/Connection/ProxyKernel.cs
```diff
@@ -160,7 +160,6 @@ private bool CanHandleLocally(KernelCommand command)
         switch (command)
         {
             case AnonymousKernelCommand:
-                return base.HandleAsync(command, context);
```
colombod, line 163: this could lead to a deadlock if the anonymous command targets the proxy. For sure this will throw exception as it will try to forward the command on the sender and will fail at serialization. This looks like a test gap to pinpoint

colombod (changes requested):
anonymous command should not leave the current host and should be handled locally even by proxy kernel

src/Microsoft.DotNet.Interactive/Connection/ProxyKernel.cs
shyamnamboodiripad: Yes this PR is fixing one specific case where we were serializing AnonymousKernelCommand (inside SubmissionParser.cs). But the larger test gap has not been addressed yet - I have included this in #2884 and plan to address that in a follow up PR.

shyamnamboodiripad:
> anonymous command should not leave the current host and should be handled locally even by proxy kernel

@colombod Agreed. This is part of what Jon and I discussed, and I have included the larger problem in this follow up issue - #2884 which I plan to look into soon.

This PR addresses one known case where AnonymousKernelCommand ends up leaving the current host today (see the fix in SubmissionParser.cs line 127 in the current PR). The plan that @jonsequitur and I discussed was to merge this fix now and tackle the larger issue in subsequent PR. I was essentially waiting for him to take a second look and approve the PR today.

Note that I was planning to merge this change today since this is one of the issues that came up when testing the VS side integration. So would appreciate if you could remove the block on the PR :)

colombod:
If @jonsequitur approves this should auto merge.

jonsequitur:
Let's find some time to discuss how we want to do this while removing rather than adding code. There's too much complexity here.

jonsequitur (approved):

colombod (approved):

PR #2894, opened 2023-04-07
Add Preview info on `#!connect jupyter` command

shibbas:
Clarify that Jupyter subkernels are in preview support by adding a message on connect and improving the connect help doc.

image

### DIFF
--- a/src/Microsoft.DotNet.Interactive.Jupyter/JupyterKernelSpecModule.cs
+++ b/src/Microsoft.DotNet.Interactive.Jupyter/JupyterKernelSpecModule.cs
@@ -60,7 +60,7 @@ public class JupyterKernelSpecModule : IJupyterKernelSpecModule
         }
         catch (Exception exception)
         {
-            Log.Error("Failed to retrieve kernel specs", exception);
+            Log.Warning("Failed to retrieve kernel specs", exception);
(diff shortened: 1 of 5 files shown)


src/Microsoft.DotNet.Interactive.Jupyter/JupyterKernelSpecModule.cs
```diff
@@ -36,18 +37,26 @@ public Task<CommandLineResult> InstallKernel(DirectoryInfo sourceDirectory)
...
+                });
+            }
+            else
+            {
+                // fall back to custom lookup logic 
+                return LookupInstalledKernels();
+            }
         }
-        else
+        catch (Exception)
```
shyamnamboodiripad, line 59: I am not sure how swallowed exceptions are handled in other places in the code base - but should we log any exceptions that are encountered and swallowed here to a log file here?

shyamnamboodiripad (approved):

src/Microsoft.DotNet.Interactive.Jupyter/JupyterKernelSpecModule.cs
shibbas: you are correct. Added logging for this.

PR #2897, opened 2023-04-07
Enable localization

ocallesp:
TODO: Pass localization culture to dotnet interactive


src/Microsoft.DotNet.Interactive/ChooseKeyValueStoreKernelDirective.cs
```diff
@@ -18,18 +18,18 @@ public class ChooseKeyValueStoreKernelDirective : ChooseKernelDirective
     {
         NameOption = new Option<string>(
             "--name",
-            "The name of the value to create. You can use #!share to retrieve this value from another subkernel.")
+            Resources.NameOptionDescription)
```
jonsequitur, line 21: This name should probably be more specific. There are other options on other magic commands that also have a name option, so a naming convention that helps disambiguate these would be good to establish.

src/dotnet-interactive/CommandLine/CommandLineParser.cs
```diff
@@ -86,10 +86,10 @@ public static class CommandLineParser
 
         startServer ??= (startupOptions, invocationContext) =>
         {
-            operation.Info("constructing webhost");
```
jonsequitur, line 89: We don't need to localize log output.

src/dotnet-interactive/CommandLine/CommandLineParser.cs
```diff
@@ -249,17 +249,17 @@ Command StdIO()
...
                 "--http-port",
-                description: "Specifies the port on which to enable HTTP services",
+                description: Resources.StdIoHttpPortOptionDescription,
                 parseArgument: result =>
                 {
                     if (result.FindResultFor(httpPortRangeOption) is { } conflictingOption)
                     {
                         var parsed = result.Parent as OptionResult;
-                        result.ErrorMessage = $"Cannot specify both {conflictingOption.Token.Value} and {parsed.Token.Value} together";
+                        result.ErrorMessage = string.Format(Resources.StdIoErrorMessageCannotSpecifyBoth, conflictingOption.Token.Value, parsed.Token.Value);
```
jonsequitur, line 262: A design that avoids needing to use `string.Format` at the call sites is preferable here and will reduce bugs in the future. We can discuss what that looks like.

src/polyglot-notebooks-ui-components/src/components/VariableGrid.tsx
```diff
@@ -456,9 +481,9 @@ export class VariableGrid extends React.Component<VariableGridProps, VariableGri
                                     <td key={4} id={`${row.id}-${4}`} className="actions-column">
                                         <div className="actions">
                                             <button
-                                                title={`Share ${row.name} from ${row.kernelDisplayName}`}
+                                                title={this.state.gridLocalization!.shareTemplate.replace("{value-name}", row.name).replace("{kernel-name}", row.kernelDisplayName)}
```
jonsequitur, line 484: We should use string interpolation rather than `replace` here.

src/Microsoft.DotNet.Interactive/ChooseKeyValueStoreKernelDirective.cs
ocallesp: Changed to a more specific name

src/dotnet-interactive/CommandLine/CommandLineParser.cs
ocallesp: Removed this localize

src/dotnet-interactive/CommandLine/CommandLineParser.cs
ocallesp: Replaced string.Format with Replace, but based on the comment below it seems that it should use interpolation

src/Microsoft.DotNet.Interactive.Journey/KernelExtensions.cs
```diff
@@ -37,18 +37,18 @@ public static CompositeKernel UseModelAnswerValidation(this CompositeKernel kern
...
-            description: "Specify lesson source file",
+            description: Resources.CompositeKernelFromFile,
             parseArgument: result =>
             {
                 var filePath = result.Tokens.Single().Value;
                 var fromUrlResult = result.FindResultFor(fromUrlOption);
                 if (fromUrlResult is not null)
                 {
-                    result.ErrorMessage = $"The {fromUrlResult.Token.Value} and {(result.Parent as OptionResult).Token.Value} options cannot be used together";
+                    result.ErrorMessage = Resources.CompositeKernelFromFileErrorMessage.Replace("{option1}", fromUrlResult.Token.Value).Replace("{option2}", (result.Parent as OptionResult).Token.Value);
```
colombod, line 51: same here

src/Microsoft.DotNet.Interactive.Journey/KernelExtensions.cs
```diff
@@ -37,18 +37,18 @@ public static CompositeKernel UseModelAnswerValidation(this CompositeKernel kern
 
         var fromUrlOption = new Option<Uri>(
             "--from-url",
-            "Specify lesson source URL");
+            Resources.CompositeKernelFromUrlOptionDescription);
 
         var fromFileOption = new Option<FileInfo>(
             "--from-file",
-            description: "Specify lesson source file",
+            description: Resources.CompositeKernelFromFile,
```
colombod, line 44: same here

src/Microsoft.DotNet.Interactive.Journey/KernelExtensions.cs
```diff
@@ -37,18 +37,18 @@ public static CompositeKernel UseModelAnswerValidation(this CompositeKernel kern
 
         var fromUrlOption = new Option<Uri>(
             "--from-url",
-            "Specify lesson source URL");
+            Resources.CompositeKernelFromUrlOptionDescription);
```
colombod, line 40: Not sure about the resource name `CompositeKernelFromUrlOptionDescription` to eval to `Specify lesson source URL` strings. Should this be something along the line of `LessonFromUrlOptionDescription` (this should probably be `journey` instead of `lessons`)

src/dotnet-interactive/CommandLine/CommandLineParser.cs
ocallesp: Any suggestion change to fix this ?

ocallesp: Fixed

src/Microsoft.DotNet.Interactive.Journey/KernelExtensions.cs
ocallesp: Fixed

src/Microsoft.DotNet.Interactive.ExtensionLab/DataFrameKernelExtension.cs
```diff
@@ -31,14 +31,14 @@ public Task OnLoadAsync(Kernel rootKernel)
             {
                 if (kernel is CSharpKernel cSharpKernel)
                 {
-                    var showCodeOption = new Option<bool>("--show-code", "Display the C# code for the generated DataFrame types");
+                    var showCodeOption = new Option<bool>("--show-code", LocalizationResources.MagicsShowCodeDescription());
```
jonsequitur, line 34: We shouldn't do any localization in the `ExtensionLab` project.

src/Microsoft.DotNet.Interactive/ChooseKeyValueStoreKernelDirective.cs
```diff
@@ -46,7 +46,7 @@ public class ChooseKeyValueStoreKernelDirective : ChooseKernelDirective
 
                 if (!File.Exists(filePath))
                 {
-                    result.ErrorMessage = LocalizationResources.Instance.FileDoesNotExist(filePath);
+                    result.ErrorMessage = System.CommandLine.LocalizationResources.Instance.FileDoesNotExist(filePath);
```
jonsequitur, line 49: This API will be internal in the next version so we should create our own resource for it in this project.

src/Microsoft.DotNet.Interactive/KernelExtensions.cs
```diff
@@ -454,7 +454,7 @@ object ParseValueOption(ArgumentResult argResult)
 
         var fromKernelOption = new Option<string>(
             "--from",
-            "The name of the kernel to get the value from.");
+            LocalizationResources.MagicsFromDescription());
```
jonsequitur, line 457:
```suggestion
            LocalizationResources.Magics_share_from_Description());
```

Another example. Since this one is a child option of the `#!share` command, I'd suggest this naming.

src/Microsoft.DotNet.Interactive.AspNetCore/AspNetCoreCSharpKernelExtensions.cs
```diff
@@ -44,7 +44,7 @@ public static CSharpKernel UseAspNetCore(this CSharpKernel kernel)
     {
         InteractiveHost interactiveHost = null;
 
-        var directive = new Command("#!aspnet", "Activate ASP.NET Core")
+        var directive = new Command("#!aspnet", LocalizationResources.MagicsAspnetDescription())
```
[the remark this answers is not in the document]
ocallesp: removed

src/Microsoft.DotNet.Interactive/ChooseKeyValueStoreKernelDirective.cs
ocallesp: resource added to avoid using System.CommandLine.LocalizationResources

src/Microsoft.DotNet.Interactive/KernelExtensions.cs
ocallesp: I am not sure the suggested change.

Methods are already using pascal case and If there are analyzers checking for this naming convention then this suggested change can break a rule.

src/Microsoft.DotNet.Interactive/ChooseKeyValueStoreKernelDirective.cs
```diff
@@ -18,18 +18,18 @@ public class ChooseKeyValueStoreKernelDirective : ChooseKernelDirective
...
-            "The name of the value to create. You can use #!share to retrieve this value from another subkernel.")
+            LocalizationResources.MagicsStoreNameDescription())
         {
             IsRequired = true
         };
 
         FromUrlOption = new Option<Uri>(
             "--from-url",
-            description: "Specifies a URL whose content will be stored.");
+            description: LocalizationResources.MagicsFromUrlDescription());
```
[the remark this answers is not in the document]
ocallesp: changed

src/Microsoft.DotNet.Interactive/KernelExtensions.cs
ocallesp: changed

eng/Localize/LocProject.json
```diff
@@ -0,0 +1,44 @@
...
+          "CopyOption": "LangIDOnName",
+          "OutputPath": "src\\dotnet-interactive"
+        },
+        {
+          "SourceFile": "src\\Microsoft.DotNet.Interactive\\Resources.resx",
+          "CopyOption": "LangIDOnName",
+          "OutputPath": "src\\Microsoft.DotNet.Interactive"
+        },
+        {
+          "SourceFile": "src\\Microsoft.DotNet.Interactive.AspNetCore\\Resources.resx",
```
jonsequitur, line 22: Please remove this project from the loc work.

eng/Localize/LocProject.json
```diff
@@ -0,0 +1,44 @@
...
+          "CopyOption": "LangIDOnName",
+          "OutputPath": "src\\Microsoft.DotNet.Interactive.AspNetCore"
+        },
+        {
+          "SourceFile": "src\\Microsoft.DotNet.Interactive.Journey\\Resources.resx",
+          "CopyOption": "LangIDOnName",
+          "OutputPath": "src\\Microsoft.DotNet.Interactive.Journey"
+        },
+        {
+          "SourceFile": "src\\Microsoft.DotNet.Interactive.ExtensionLab\\Resources.resx",
```
jonsequitur, line 32: Please remove this project from the loc work.

src/Microsoft.DotNet.Interactive/ChooseKeyValueStoreKernelDirective.cs
```diff
@@ -18,18 +18,18 @@ public class ChooseKeyValueStoreKernelDirective : ChooseKernelDirective
...
-            "The name of the value to create. You can use #!share to retrieve this value from another subkernel.")
+            LocalizationResources.Magics_store_name_Description())
         {
             IsRequired = true
         };
 
         FromUrlOption = new Option<Uri>(
             "--from-url",
-            description: "Specifies a URL whose content will be stored.");
+            description: LocalizationResources.Magics_store_from_url_Description());
```
jonsequitur, line 28:
```suggestion
            description: LocalizationResources.Magics_value_from_url_Description());
```

The word `value` is more appropriate here because the user-facing magic command for this kernel is `#!value`.

src/Microsoft.DotNet.Interactive/Parsing/DirectiveHelpBuilder.cs
```diff
@@ -14,15 +14,15 @@ public class DirectiveHelpBuilder : HelpBuilder
...
+    public DirectiveHelpBuilder(string rootCommandName) : base(System.CommandLine.LocalizationResources.Instance)
     {
         _rootCommandName = rootCommandName;
     }
 
     public override void Write(HelpContext context)
     {
         using var writer = new StringWriter();
-        new HelpBuilder(LocalizationResources.Instance).Write(context.Command, writer);
+        new HelpBuilder(System.CommandLine.LocalizationResources.Instance).Write(context.Command, writer);
```
jonsequitur, line 25: We should not reference the System.CommandLine loc strings. It's better to have a copy of this loc string in this project, since the System.CommandLine resources won't be public in the next version.

eng/Localize/LocProject.json
ocallesp: removed

src/Microsoft.DotNet.Interactive/ChooseKeyValueStoreKernelDirective.cs
ocallesp: changed to 'value'

src/Microsoft.DotNet.Interactive/Parsing/DirectiveHelpBuilder.cs
ocallesp: The current implementation of HelpBuilder is tightly coupled to System.CommandLine.LocalizationResources. Removing this dependency would require us to create a custom implementation of HelpBuilder, which would involve a significant amount of work.

Given the current state of our project, I propose that we merge the PR as it is. Later, when we plan to remove all dependencies with classes using System.CommandLine, we can allocate time and resources to address this issue properly. This approach will allow us to move forward with the current PR and ensure that we don't introduce any breaking changes or additional complexity at this stage.

jonsequitur: The System.CommandLine.LocalizationResources API has already been made internal in System.CommandLine and that new version will be pulled into .NET Interactive this week if there's time.

I'd suggest skipping localization for any strings you're pulling from there for now.

jonsequitur (approved):

PR #2912, opened 2023-04-12
Update F# packages to latest released

vzarytovskii:
Is there anything additional needed?


brettfo (approved):

brettfo:
The F# tests are pretty good; if CI is green then it should be fine (famous last words, I'm sure.)

vzarytovskii:
> The F# tests are pretty good; if CI is green then it should be fine (famous last words, I'm sure.)

Yeah, we shouldn't have any substantial changes (also famous last words probably).

jonsequitur (approved):

PR #2915, opened 2023-04-13
Cancelation and scheduling improvements

jonsequitur:


src/Microsoft.DotNet.Interactive/Commands/KernelCommand.cs
```diff
@@ -64,4 +134,41 @@ public virtual Task InvokeAsync(KernelInvocationContext context)
...
+        {
+            return false;
+        }
+
+        return GetOrCreateId() == other.GetOrCreateId();
+    }
+
+    public override int GetHashCode()
+    {
+        return GetOrCreateId().GetHashCode();
```
shyamnamboodiripad, line 172: Shouldn't the hash code be based the token too? Or is this just here temporarily until the id concept is removed?

jonsequitur: Yeah, so far this part is a refactor with no behavioral change. This will eventually be based on the token when the id concept is gone.

src/Microsoft.DotNet.Interactive/Commands/KernelCommand.cs
```diff
@@ -64,4 +134,41 @@ public virtual Task InvokeAsync(KernelInvocationContext context)
...
+        {
+            return _id;
+        }
+
+        SetId(Guid.NewGuid().ToString("N"));
+
+        return _id;
+    }
+
+    public bool Equals(KernelCommand other)
```
shyamnamboodiripad, line 155: consider: It may be a good idea to override `object.Equals` as well (and operators `==` and `!=`)

src/Microsoft.DotNet.Interactive/Commands/KernelCommand.cs
```diff
@@ -64,4 +134,41 @@ public virtual Task InvokeAsync(KernelInvocationContext context)
...
+
+        if (other is null)
+        {
+            return false;
+        }
+
+        return GetOrCreateId() == other.GetOrCreateId();
+    }
+
+    public override int GetHashCode()
```
shyamnamboodiripad, line 170: So, I am looking at adding more logging and it looks like the logging for command objects currently relies on each command type overriding `ToString`. I was planning to override `ToString` on the base `KernelCommand` to include the token, and then update each derived `ToString` to call the base implementation (in addition to returning derived-specific strings)...

I would have preferred if we didn't need to touch every command type to add this logging - I have a concern that this may not scale well. For example, if we miss the call to `base.ToString()` when implementing `ToString` for some new command(s) in the future, this could make it difficult to debug some problem in already shipped bits where the info is missing...

Another alternative would be to log detailed info about the command (including token) in a separate message. Perhaps I could do that at the same time when we stamp the routing slip (which as we discussed would also be a good place to log the routing slip info).

However, not including the token in the `command.ToString()` would also mean that we may miss it in some places where having that info may be useful (such as when logging the command associated with an event). So, there are pros and cons...

Any thoughts on which approach may work better @jonsequitur ?

shyamnamboodiripad: I guess we could also introduce a new virtual `ToDisplayString` method to log the derived-specific info and then only override `ToString` in the base type. The base `ToString` could then call `ToDisplayString` and also log the command token to ensure that the token info is always logged whenever a command object is logged. One downside is that derived command implementations should remember to only override `ToDisplayString` (and leave `ToString` untouched) to get the desired logging behavior - but that doesn't seem too bad...

shyamnamboodiripad: Ah thanks I hadn't considered that. I guess we could just log `$"Run: {value.ToDisplayString("text/logging"}"` inside `KernelScheduler` (instead of logging `$"Run: {value}"` as we do currently). And the registration could look something like -

```
Formatter.Register(
    formatter: obj =>
    {
        if (obj is KernelCommand command)
        {
            var prefix = $"{command.GetType().Name} ({command.GetOrCreateToken()})";
    
            if (obj is SubmitCode submitCode)
            {
                return $"{prefix} {submitCode.Code}";
            }
            ...
                // Handle other command types that need custom logging
            ...
        }
    
        return obj.ToString();
    }, 
    mimeType: "text/logging");
```

Does that match what you were thinking?

jonsequitur: Yeah, that looks great!

PR #2921, opened 2023-04-18
Make sure `SendValue` always includes `FormattedValue`

### LINKED ISSUE #2839: directly setting values kernel using `#!set` fails when kernel is not in the `dotnet-interactive` process
image

jonsequitur:
This fixes #2839.

### DIFF
--- a/src/Microsoft.DotNet.Interactive.FSharp/FSharpKernel.fs
+++ b/src/Microsoft.DotNet.Interactive.FSharp/FSharpKernel.fs
@@ -220,7 +220,7 @@ type FSharpKernel () as this =
                 match result with
                 | Some(value) when value.ReflectionType <> typeof<unit> ->
                     let value = value.ReflectionValue
-                    let formattedValues = FormattedValue.FromObject(value)
+                    let formattedValues = FormattedValue.CreateManyFromObject(value)
                     context.Publish(ReturnValueProduced(value, codeSubmission, formattedValues))
                 | Some _
                 | None -> ()
@@ -377,7 +377,7 @@ type FSharpKernel () as this =
             let valueInfos =
                 script.Value.Fsi.GetBoundValues()
                 |> List.filter (fun x -> x.Name <> "it") // don't report special variable `it`
-                |> List.map (fun x -> new KernelValueInfo(x.Name, FormattedValue.FromObject(x.Value.ReflectionValue, requestValueInfos.MimeType)[0], this.getValueType(x.Name)))
+                |> List.map (fun x -> new KernelValueInfo(x.Name, FormattedValue.CreateManyFromObject(x.Value.ReflectionValue, requestValueInfos.MimeType)[0], this.getValueType(x.Name)))
                 :> IReadOnlyCollection<KernelValueInfo>
             context.Publish(new ValueInfosProduced(valueInfos, requestValueInfos))
         }
--- a/src/Microsoft.DotNet.Interactive.FSharp/FSharpKernel.fs
+++ b/src/Microsoft.DotNet.Interactive.FSharp/FSharpKernel.fs
@@ -377,7 +377,7 @@ type FSharpKernel () as this =
             let valueInfos =
                 script.Value.Fsi.GetBoundValues()
                 |> List.filter (fun x -> x.Name <> "it") // don't report special variable `it`
-                |> List.map (fun x -> new KernelValueInfo(x.Name, FormattedValue.CreateManyFromObject(x.Value.ReflectionValue, requestValueInfos.MimeType)[0], this.getValueType(x.Name)))
+                |> List.map (fun x -> new KernelValueInfo(x.Name, FormattedValue.CreateSingleFromObject(x.Value.ReflectionValue, requestValueInfos.MimeType), this.getValueType(x.Name)))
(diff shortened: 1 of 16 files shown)


colombod (approved):

src/Microsoft.DotNet.Interactive.FSharp/FSharpKernel.fs
```diff
@@ -377,7 +377,7 @@ type FSharpKernel () as this =
             let valueInfos =
                 script.Value.Fsi.GetBoundValues()
                 |> List.filter (fun x -> x.Name <> "it") // don't report special variable `it`
-                |> List.map (fun x -> new KernelValueInfo(x.Name, FormattedValue.FromObject(x.Value.ReflectionValue, requestValueInfos.MimeType)[0], this.getValueType(x.Name)))
+                |> List.map (fun x -> new KernelValueInfo(x.Name, FormattedValue.CreateManyFromObject(x.Value.ReflectionValue, requestValueInfos.MimeType)[0], this.getValueType(x.Name)))
```
shyamnamboodiripad, line 380: Should this call CreateSingleFromObject instead (considering that we only pick the first formatted value returned)?

src/Microsoft.DotNet.Interactive.FSharp/FSharpKernel.fs
```diff
@@ -220,7 +220,7 @@ type FSharpKernel () as this =
                 match result with
                 | Some(value) when value.ReflectionType <> typeof<unit> ->
                     let value = value.ReflectionValue
-                    let formattedValues = FormattedValue.FromObject(value)
+                    let formattedValues = FormattedValue.CreateManyFromObject(value)
                     context.Publish(ReturnValueProduced(value, codeSubmission, formattedValues))
```
shyamnamboodiripad, line 224: minor: Considering that a single formatted value is probably the common case, I wonder if event constructors (like `ReturnValueProduced()`) should have overloads that accept a single formatted value.

src/Microsoft.DotNet.Interactive/Commands/SendValue.cs
```diff
@@ -14,6 +15,11 @@ public class SendValue : KernelCommand
         FormattedValue formattedValue = null,
         string targetKernelName = null) : base(targetKernelName)
     {
+        if (formattedValue is null)
+        {
+            formattedValue = FormattedValue.CreateSingleFromObject(value, JsonFormatter.MimeType);
```
shyamnamboodiripad, line 20: Is it a common pattern to fall back to the json mime type by default? Should this try to use the mime type in `Microsoft.DotNet.Interactive.Formatting.Formatter.DefaultMimeType` first?

shyamnamboodiripad: Or should this pass `null` for the mime type here so that we use `Formatter.GetPreferredMimeTypesFor(value?.GetType())`?

src/Microsoft.DotNet.Interactive/FormattedValue.cs
```diff
@@ -25,7 +25,17 @@ public FormattedValue(string mimeType, string value)
 
     public string Value { get; }
 
-    public static IReadOnlyList<FormattedValue> FromObject(object value, params string[] mimeTypes)
+    public static FormattedValue CreateSingleFromObject(object value, string mimeType = null)
+    {
+        if (mimeType is null)
+        {
+            mimeType = Formatter.GetPreferredMimeTypesFor(value?.GetType()).First();
```
shyamnamboodiripad, line 32: Just thinking out loud: There is an assumption here that the order of returned mime types from `Formatter.GetPreferredMimeTypesFor()' will stay the same. Can this lead to surprising results for external callers of `CreateSingleFromObject()` if we introduce a new mime type in the list (especially if the external caller is calling `CreateSingleFromObject()` indirectly via some other public API  - not sure if this is possible)?

I guess we would not do something like that by accident - but do we have tests to catch if we do?

src/Microsoft.DotNet.Interactive.FSharp/FSharpKernel.fs
shyamnamboodiripad: Never mind - I realize that a single formatted value is probably not the common case since it depends on the set of supported mime types associated with the type of the value...

shyamnamboodiripad (approved):

src/Microsoft.DotNet.Interactive/Commands/SendValue.cs
jonsequitur: JSON is the default for value sharing. There's little use in sending text/plain or text/html except in specialized cases.

src/Microsoft.DotNet.Interactive.FSharp/FSharpKernel.fs
jonsequitur: Good catch.

src/Microsoft.DotNet.Interactive/FormattedValue.cs
jonsequitur: That's definitely expected and is intended to be reconfigured. If you want a specific MIME type, you should pass it. In practice, the reason you don't pass it is because you want whatever defaults the host has configured.

PR #2923, opened 2023-04-18
Kernel Connector can return more than one kernel

colombod:


src/Microsoft.DotNet.Interactive.Tests/ConnectDirectiveTests.cs
```diff
@@ -190,13 +191,14 @@ public class ConnectFakeKernelCommand : ConnectKernelCommand
...
-        public override Task<Kernel> ConnectKernelAsync(
+        public override async Task<IEnumerable<Kernel>> ConnectKernelsAsync(
             KernelInvocationContext context,
             InvocationContext commandLineContext)
         {
             var connector = new FakeKernelConnector();
             connector.CreateKernel = _createKernel;
-            return connector.CreateKernelAsync(commandLineContext.ParseResult.GetValueForOption(KernelNameOption));
+            var kernel = await connector.CreateKernelAsync(commandLineContext.ParseResult.GetValueForOption(KernelNameOption));
+            return new[] { kernel };
```
AbhitejJohn, line 201: ah, this is an interesting pattern change. Seems like we'll now be waiting to create each kernel in the constructor. No concerns at the moment, just calling it out. I wonder if we could do [IAsyncEnumerable](https://learn.microsoft.com/en-us/dotnet/api/system.collections.generic.iasyncenumerable-1?view=net-7.0). Dont know if we have netstandard constraints though (it only works 2.1 and above).

shyamnamboodiripad: (We do have netstd concerns since the same code is compiled to netstd2.0 as well as part of the MS.DN.Interactive.VisualStudio.csproj)

jonsequitur (approved):

PR #2930, opened 2023-04-20
Introduce custom logging formatters for commands and events

shyamnamboodiripad:
Also introduces logging for command tokens and routing slip information.

### DIFF
--- a/src/Microsoft.DotNet.Interactive/KernelScheduler.cs
+++ b/src/Microsoft.DotNet.Interactive/KernelScheduler.cs
@@ -4,8 +4,6 @@
 using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
-using System.Linq.Expressions;
-using System.Reflection;
 using System.Threading;
 using System.Threading.Tasks;
 
@@ -66,7 +64,7 @@ public class KernelScheduler<T, TResult> : IDisposable, IKernelScheduler<T, TRes
         CancellationToken cancellationToken = default)
     {
         ThrowIfDisposed();
-            
+
         ScheduledOperation operation;
         if (_isPreemptive(_currentTopLevelOperation, value))
         {
@@ -135,7 +133,7 @@ public class KernelScheduler<T, TResult> : IDisposable, IKernelScheduler<T, TRes
     {
         _currentTopLevelOperation ??= operation.Value;
 
-        using var logOp = Log.OnEnterAndConfirmOnExit($"Run: {operation.Value}");
+        using var logOp = Log.OnEnterAndConfirmOnExit(operation.Value);
 
         try
         {
@@ -196,7 +194,7 @@ public class KernelScheduler<T, TResult> : IDisposable, IKernelScheduler<T, TRes
             Log.Error(exception);
         }
     }
-        
+
     private IEnumerable<ScheduledOperation> OperationsToRunBefore(
         ScheduledOperation operation)
     {
@@ -258,7 +256,7 @@ public class KernelScheduler<T, TResult> : IDisposable, IKernelScheduler<T, TRes
                 // todo: this is still a problem with fsi
 
                 // ControlledExecution.Run isn't available in .NET Standard but since we're most likely actually running in .NET 7+, we can try to bind a delegate to it.
-                
+
                 //if (Type.GetType("System.Runtime.ControlledExecution, System.Private.CoreLib", false) is { } controlledExecutionType &&
                 //    controlledExecutionType.GetMethod("Run", BindingFlags.Static | BindingFlags.Public) is { } runMethod)
                 //{
@@ -319,7 +317,7 @@ public class KernelScheduler<T, TResult> : IDisposable, IKernelScheduler<T, TRes
 
         public Task<TResult> ExecuteAsync()
         {
-            if (_runWithControlledExecution is not null && 
+            if (_runWithControlledExecution is not null &&
                 _cancellationToken.CanBeCanceled)
             {
                 try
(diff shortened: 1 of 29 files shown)


src/Microsoft.DotNet.Interactive/KernelScheduler.cs
```diff
@@ -133,7 +134,7 @@ private void Run(ScheduledOperation operation)
     {
         _currentTopLevelOperation ??= operation.Value;
 
-        using var logOp = Log.OnEnterAndConfirmOnExit($"Run: {operation.Value}");
+        using var logOp = Log.OnEnterAndConfirmOnExit($"Run: {operation.Value.ToDisplayString("text/log")}");
```
jonsequitur, line 137: I'd suggest creating a `const` for this somewhere. We might quibble over the MIME type later and it will make it easier to change.

src/Microsoft.DotNet.Interactive/Logging/KernelEventLoggingFormatter.cs
```diff
@@ -0,0 +1,62 @@
...
+        {
+            if (@event is null)
+            {
+                context.Writer.Write($"{nameof(KernelEvent)}: <null>");
+            }
+            else
+            {
+                context.Writer.Write($"{@event.GetType().Name} (Command {@event.Command.ToDisplayString("text/log")}");
+
+                switch (@event)
```
jonsequitur, line 23: This should probably have a default case.

src/dotnet-interactive/Utility/KernelDiagnostics.cs
```diff
@@ -3,20 +3,20 @@
...
 {
-
     [DebuggerStepThrough]
     public static T LogCommandsToPocketLogger<T>(this T kernel)
         where T : Kernel
     {
         kernel.AddMiddleware(async (command, context, next) =>
         {
-            using var _ = Logger.Log.OnEnterAndExit($"Command: {command.ToString().Replace(Environment.NewLine, " ")}");
+            using var _ = Logger.Log.OnEnterAndExit($"Command: {command.ToDisplayString("text/log")}");
```
jonsequitur, line 19:
```suggestion
            using var _ = Logger.Log.OnEnterAndExit(command.ToDisplayString("text/log"));
```

It's probably easier to understand and more consistent if the formatter handles 100% of the formatting.

jonsequitur: There's a PocketLogger (and Microsoft.Extensions.Logging) perf concern with passing a string that can change over time here. We can discuss whether to address that here. You didn't introduce an issue. It was already present here.

shyamnamboodiripad: Good call. I'll removed the `Command:` here prefix in favor of the type name prefix that is already logged as part of `ToDisplayString`.

Surprisingly though, the above middleware registration does not seem to work. I couldn't find any `Command:` prefixes in the log even without the changes in the current PR. I haven't had a chance to investigate why yet - but I am also unsure whether we gain much by logging commands via the above middleware registration (on the root composite kernel) when the scheduler is already logging all the commands it sees. Would it be ok to remove the above middleware registration completely?

src/Microsoft.DotNet.Interactive/KernelScheduler.cs
shyamnamboodiripad: I considered adding a `LoggingFormatter.MimeType` const. However, in order to use it across all the projects where this is currently referenced, we would need to make it public. It seems pretty harmless, but I wasn't sure about adding to the public API for our internal logging... I will go ahead and add it.

src/Microsoft.DotNet.Interactive/Logging/KernelEventLoggingFormatter.cs
shyamnamboodiripad: For all events, we already log the event type name (and the associated command) on line 21 above. Is there any additional info we would want to log in the `default` case?

src/dotnet-interactive/Utility/KernelDiagnostics.cs
jonsequitur: Are you starting the tool with `--verbose`?

src/Microsoft.DotNet.Interactive/Logging/KernelEventLoggingFormatter.cs
jonsequitur: Ah, I see.

Some sample log output including commands and events would probably be useful to add to the PR description.

src/Microsoft.DotNet.Interactive/KernelScheduler.cs
jonsequitur: We shouldn't make it public. 

Some of our logging infrastructure code is via shared files that are directly source included and this is a good fit for that.

src/Microsoft.DotNet.Interactive/Logging/KernelEventLoggingFormatter.cs
shyamnamboodiripad: Good call - will do this soon.

src/dotnet-interactive/Utility/KernelDiagnostics.cs
shyamnamboodiripad: Ah my bad. Looks like `LogCommandsToPocketLogger` is only called from within tests (unlike - so please ignore what I said above.

PR #2938, opened 2023-04-24
enable jupyter test

colombod:


src/Microsoft.DotNet.Interactive.Jupyter.Tests/JupyterKernelTests.cs
```diff
@@ -174,8 +175,10 @@ public async Task can_cancel_submit_code_and_interrupt_kernel()
 
         var sentMessages = options.MessageTracker.SentMessages.ToSubscribedList();
         await waitForCommandReceieved;
-        cts.Cancel();
 
+        await Task.Delay(100);
```
ocallesp, line 179: Why is this delay needed ?

ocallesp:
##[error]The job running on agent NetCore-Public 89 ran longer than the maximum time of 60 minutes. For more information, see https://go.microsoft.com/fwlink/?linkid=2077134

src/Microsoft.DotNet.Interactive.Jupyter.Tests/JupyterKernelTests.cs
shyamnamboodiripad: I noticed that this made the test pass locally. I think @colombod was trying an experiment to see if this also makes the test pass in CI.... There is definitely some kind of race here.

PR #2939, opened 2023-04-25
Make value kernel extensible

colombod:


src/Microsoft.DotNet.Interactive/KeyValueStoreKernel.cs
```diff
@@ -181,19 +181,34 @@ void UndoSetValue()
...
-    private void StoreValue(
+    private async Task StoreValueAsync(
         string value,
         ValueDirectiveOptions options,
         KernelInvocationContext context,
         string mimeType = null)
+    {
+        mimeType ??= (options.MimeType ?? PlainTextFormatter.MimeType);
+
+        var shouldDisplayValue = options.MimeType is { } displayMimeType;
```
shyamnamboodiripad, line 192: nit: The `displayMimeType` variable is unused and can be dropped.

src/Microsoft.DotNet.Interactive/KeyValueStoreKernel.cs
```diff
@@ -181,19 +181,34 @@ void UndoSetValue()
...
+        _values[key] = new FormattedValue(mimeType, value);
 
-        if (options.MimeType is {} displayMimeType)
+        if (shouldDisplayValue)
         {
-            context.DisplayAs(value, displayMimeType);
+            context.DisplayAs(value, mimeType);
         }
+
+        return Task.CompletedTask;
```
shyamnamboodiripad, line 212: Can the method be made sync or is this async because we think we may need it down the line?

src/Microsoft.DotNet.Interactive/KeyValueStoreKernel.cs
```diff
@@ -181,19 +181,34 @@ void UndoSetValue()
...
+
+        await StoreValueAsync(options.Name, value, mimeType, shouldDisplayValue, context);
+    }
+
+    protected virtual Task StoreValueAsync(
+        string key,
+        string value,
+        string mimeType,
+        bool shouldDisplayValue,
+        KernelInvocationContext context)
```
shyamnamboodiripad, line 203: minor: Consider removing blank line...

shyamnamboodiripad (approved):
Change looks good - however, I am not sure why this is needed. Could you please clarify? Thanks!

PR #2950, opened 2023-04-28
Introduce object model for http requests and responses

shyamnamboodiripad:


colombod (approved):

AbhitejJohn:
Curious, how would this object model be consumed externally? Trying to understand our extensibility layer and where this fits in.

colombod:
> Curious, how would this object model be consumed externally? Trying to understand our extensibility layer and where this fits in.

The motivation is not sharing these types by reference (#!set assumes json serialization and we are out of proc anyway). this approach makes it  safe to retain and support named requests and value sharing at any time without risks with the async operation and disposable streams.

shyamnamboodiripad:
@AbhitejJohn This essentially makes it possible to register a custom formatter based on the (now 'static') object model. And as Diego mentioned, this will also form the basis for implementing scenarios like named requests (which is part of the REST client's spec) in the future.

PR #2953, opened 2023-04-28
revert scheduler changes

colombod:
fsi doesn't like the controlled execution

### DIFF
--- a/src/Microsoft.DotNet.Interactive.Tests/KernelSchedulerTests.cs
+++ b/src/Microsoft.DotNet.Interactive.Tests/KernelSchedulerTests.cs
@@ -284,7 +284,7 @@ public class KernelSchedulerTests : IDisposable
             .ThrowAsync<OperationCanceledException>();
     }
 
-    [Fact]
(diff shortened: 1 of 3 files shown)


shyamnamboodiripad (approved):

src/Microsoft.DotNet.Interactive.Tests/KernelSchedulerTests.cs
```diff
@@ -311,7 +311,7 @@ public void Infinite_loops_can_be_cancelled()
             .ThrowAsync<OperationCanceledException>();
     }
 
-    [Fact]
+    [Fact(Skip = "requires System.Runtime.ControlledExecution")]
     public async Task After_an_infinite_loop_is_cancelled_the_scheduler_can_still_be_used()
```
shyamnamboodiripad, line 315: @colombod This was one of the tests that was deadlocking in my experiment to add testing against `net472`. Based on what we were looking at earlier, I am guessing the explanation may be that the `ControlledExecution` API is not available in `net472`?

shyamnamboodiripad (approved):

PR #2961, opened 2023-05-03
vscode extension release

colombod:


src/polyglot-notebooks-vscode-insiders/package.json
```diff
@@ -10,10 +10,10 @@
     "notebookMessaging"
   ],
   "preview": false,
-  "//version": "%description.version%",
+  "//version": "The version '42.42.42' is auto-set during CI package creation.",
```
AbhitejJohn, line 13: Looks like this is resetting some of the localization efforts @ocalles added in earlier? Could we add some context on what that was breaking....

AbhitejJohn (approved):

PR #2983, opened 2023-05-12
enabling controlled execution

colombod:
Enabling controlled execution and integration with new FSI

### DIFF
--- a/src/Microsoft.DotNet.Interactive/KernelScheduler.cs
+++ b/src/Microsoft.DotNet.Interactive/KernelScheduler.cs
@@ -5,6 +5,8 @@ using System;
 using System.Collections;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
+using System.Linq.Expressions;
+using System.Reflection;
 using System.Threading;
 using System.Threading.Tasks;
 using Pocket;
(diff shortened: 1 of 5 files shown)


jonsequitur (approved):

src/Microsoft.DotNet.Interactive/KernelScheduler.cs
```diff
@@ -255,23 +255,21 @@ static ScheduledOperation()
         {
             try
             {
-                // todo: this is still a problem with fsi
-
                 // ControlledExecution.Run isn't available in .NET Standard but since we're most likely actually running in .NET 7+, we can try to bind a delegate to it.
```
shyamnamboodiripad, line 258: Please preserve this comment

src/Microsoft.DotNet.Interactive/KernelScheduler.cs
```diff
@@ -255,23 +255,21 @@ static ScheduledOperation()
...
+                    var actionParameter = Expression.Parameter(typeof(Action), "action");
+
+                    var cancellationTokenParameter = Expression.Parameter(typeof(CancellationToken), "cancellationToken");
+
+                    _runWithControlledExecution = Expression.Lambda<Action<Action, CancellationToken>>(
+                                                                Expression.Call(runMethod, actionParameter, cancellationTokenParameter),
+                                                                actionParameter,
+                                                                cancellationTokenParameter)
+                                                            .Compile();
+                }
```
shyamnamboodiripad, line 272: Wasn't there a Jupyter test that was failing / deadlocking with this change?

jonsequitur: I believe F# had an issue with it because it was also using `ControlledExecution` on the same code path. That should be fixed by the updated F# packages.

shyamnamboodiripad: Iirc the test in question was the one skipped in #2936. It originally started failing after the above `ControlledExecution` code was introduced. @colombod may remember more, but iirc it was running into a `ThreadAbortException`.

The repro was not super consistent. Diego was experimenting with enabling it again in #2938 but it continued to fail (on Linux only after adding some delay). It was eventually unskipped when the above change was reverted in #2953.

I could be wrong, but I don't think the above jupyter test had anything to do with F#, and I don't think we figured out exactly why that test was impacted by the `ControlledExecution` changes (or why it would inconsistently run into a ThreadAbortException in net core). I don't see it failing in the current CI anymore though - so perhaps something else fixed it :)

jonsequitur: If it was failing due to a deadlock it might have been the other issue that we fixed that was also deadlocking on message serialization, e.g. not completing the `Task` when channel-level exceptions occurred. That's just a guess.

PR #2996, opened 2023-05-25
Implement custom pocket logging formatters for commands and events.

shyamnamboodiripad:
Also remove the #!log magic command, improve logging of routing slips and other random improvements around logging.


src/Microsoft.DotNet.Interactive/KernelScheduler.cs
```diff
@@ -135,7 +132,8 @@ private void Run(ScheduledOperation operation)
     {
         _currentTopLevelOperation ??= operation.Value;
 
-        using var logOp = Log.OnEnterAndConfirmOnExit($"Run: {operation.Value}");
+        using var logOp = Log.OnEnterAndConfirmOnExit();
+        logOp.Info("{0}", operation.Value);
```
shyamnamboodiripad, line 136: It would be great if we could say `logOp.Infi(operation.Value)` instead. This would need a change in PocketLogger to make the message template string optional.

shyamnamboodiripad: Adding ability to pass argument(s) in `OnEnterAndConfrimOnExit` would also make the above case simpler. We could then just say `Log.OnEnterAndConfirmOnExit(operation.Value)`.

src/Microsoft.DotNet.Interactive/KernelCommandExtensions.cs
```diff
@@ -0,0 +1,35 @@
...
+using Pocket;
+
+namespace Microsoft.DotNet.Interactive;
+
+internal static class KernelCommandExtensions
+{
+    internal static void StampRoutingSlipAndLog(this KernelCommand command, Uri uri)
+    {
+        command.RoutingSlip.Stamp(uri);
+        command.LogRoutingSlipInfo();
```
jonsequitur, line 15: Nitpick: I would tend toward reversing this, e.g. `Log.RoutingSlipInfo(command);`

src/Microsoft.DotNet.Interactive/KernelEventExtensions.cs
```diff
@@ -60,4 +62,14 @@ private static IReadOnlyCollection<Diagnostic> RemapDiagnosticsFromLanguageNode(
             )
         ).ToImmutableList();
     }
+
+    internal static void StampRoutingSlipAndLog(this KernelEvent @event, Uri uri)
```
jonsequitur, line 66: Why not just add the log call into the `Stamp*` methods? Does it get too verbose?

shyamnamboodiripad: The `Stamp*` methods don't have access to the `event` /`command`. So we need to add it to the caller of those methods.

Unless you think its ok to change the public `Stamp*` methods to pass in the `event` / `command`... But even then, we only want to log when routing slip is stamped inside the tool process - but not when it is stamped within the (de)serializer.

jonsequitur (approved):

src/Microsoft.DotNet.Interactive/KernelCommandExtensions.cs
shyamnamboodiripad: Fixed.

src/Microsoft.DotNet.Interactive/KernelEventExtensions.cs
jonsequitur: The deserialization process should not call `Stamp`. If it still does that we should fix it.

shyamnamboodiripad: It does currently. See https://github.com/dotnet/interactive/blob/main/src/Microsoft.DotNet.Interactive/Connection/KernelCommandEnvelope.cs#L186.

To fix this, I guess `RoutingSlip` should have a constructor that accepts all deserialized uris (and tags), and we need a `KernelCommand.SetRoutingSlip` (similar to `KernelCommand.SetToken`) to initialize `KernelCommand.RoutingSlip` post deserialization. Is this what you were implying?

jonsequitur: Yeah, that sounds right.

PR #3003, opened 2023-05-26
Remove DiagnosticLogEntryProduced event.

shyamnamboodiripad:


src/Microsoft.DotNet.Interactive/Connection/KernelCommandAndEventReceiver.cs
```diff
@@ -69,7 +69,7 @@ public KernelCommandAndEventReceiver(ReadCommandOrEvent readCommandOrEvent)
                 {
                     Log.Error(exception);
 
-                    return new CommandOrEvent(new DiagnosticLogEntryProduced(exception.Message, KernelCommand.None), true);
+                    return new CommandOrEvent(new ErrorProduced(exception.Message, KernelCommand.None), isParseError: true);
```
shyamnamboodiripad, line 72: I'm not entirely sure if using `ErrorProduced` for this is ok. I was not able to figure out another way to handle this (other than allowing the exception to propagate). Please let me know in case you think this is not ok.

colombod:
Why deleting this? The original goal for this event was to be intercepted by the vscode extension and routed tot he diagnostics and logs outputs.

shyamnamboodiripad:
> Why deleting this? The original goal for this event was to be intercepted by the vscode extension and routed tot he diagnostics and logs outputs.

Ah this was something that @jonsequitur and I had discussed. We already removed the `#!log` magic command in an earlier change - this event seemed like a vestige that is no longer useful.

colombod (approved):

Issue #2928, opened 2023-04-20
JavaScript variables are not being registered outside of cells


IntegerMan:
### Describe the bug
Apologies if this is a misunderstanding of kernel limitations.

I can import a C# variable into JavaScript without issues. However, if I declare a JS variable in a JS cell using either `let` or `const`, that variable does not seem to be registered in the variables section, is not available in the next cell (even if it's a JS cell), does not share correctly to C# or other kernels, and perhaps most interestingly, doesn't log out at the end of the same cell it was defined in if it was the closing value as in:

``` js
let jsVar = 42;
jsVar; // jsVar is not displayed after the cell
```

JavaScript otherwise appears to function just fine, but variable persistence and sharing just isn't working for me.

See https://github.com/IntegerMan/PolyglotNotebookTalk/blob/main/cs_js.dib for my sample notebook.

### Please complete the following:

**Which version of .NET Interactive are you using?** (In a notebook, run the `#!about` magic command. ):

- OS
    - [x] Windows 11
    - [ ] Windows 10
    - [ ] macOS
    - [ ] Linux (Please specify distro)
    - [ ] iOS
    - [ ] Android
 - Browser 
    - [ ] Chrome
    - [ ] Edge
    - [ ] Firefox
    - [ ] Safari
 - Frontend
    - [ ] Jupyter Notebook
    - [ ] Jupyter Lab
    - [ ] nteract
    - [x] Visual Studio Code 
    - [ ] Visual Studio Code Insiders
    - [ ] Visual Studio
    - [ ] Other (please specify)

### Screenshots
![image](https://user-images.githubusercontent.com/5049957/233240199-2ab9b316-b1f5-45f2-86f3-c55e7c017d5f.png)

jonsequitur:
This is by design. The JavaScript kernel executes your code submission within a function, and so variables declared in that function aren't visible outside unless you use this syntax:

```javascript
x = 123;
```

By not using the `let`, `const`, or `var` keywords with your variable declaration, you're enabling JavaScript variable hoisting to make the variable visible to other cells (i.e. within other functions in the top-level scope).

As for return values from JavaScript cells, the trailing expression syntax isn't supported in JavaScript. So while in C# Script, this is equivalent to a `return` statement:

```csharp
123
```

in JavaScript you need to write this:

```javascript
return 123;
```

IntegerMan:
Thank you so much for the reply @jonsequitur. I'll play around with that tonight!

[...]

PR #3025, opened 2023-06-08
JavaScript usage documentation improvements 

jonsequitur:
This addresses #2928.

### DIFF
--- a/docs/javascript-overview.md
+++ b/docs/javascript-overview.md
@@ -43,6 +43,15 @@ try { console.log(declaredWithConst); }        catch (e) { console.log(e.toStrin
 try { console.log(declaredWithoutKeyword); }   catch (e) { console.log(e.toString()); }
 ```
 
+This code produces the following output: 
+
+```console
+ReferenceError: declaredWithLet is not defined
+ReferenceError: declaredWithVar is not defined
+ReferenceError: declaredWithConst is not defined
+4
+```
+
 So why does the fourth example work? By not using the `let`, `const`, or `var` keywords with your variable declaration, you're enabling JavaScript variable hoisting to add the variable to the top-level scope. When running in a browser (including the notebook webview), this means the variable will be added to `window`. For this reason, the final line of the above example is equivalent to the following:
 
 ```javascript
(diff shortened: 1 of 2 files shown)


docs/javascript-overview.md
```diff
@@ -96,7 +179,7 @@ When you run this code, you can see that the original C# variable gets overwritt
...
 
-One notable detail is that the type of the new C# variable has changed from the original declaration. It was originally declared as `int[]`, but after sharing the array back from JavaScript, the C# kernel has a variable called `array` which is of type `JsonDocument`. Because the JavaScript kernel runs in a different process from the .NET subkernels, sharing happens via JSON serialization. Complex types shared from JavaScript to .NET are sent as JSON and deserialized using `System.Text.Json`. There are a few simple types that will be shared as their intuitive .NET counterparts. 
+One notable detail here is that the type of the new C# variable has changed from the original declaration. It was originally declared as `int[]`, but after sharing the array back from JavaScript, the C# kernel has a variable called `array` which is of type `JsonDocument`. Because the JavaScript kernel runs in a different process from the .NET subkernels, sharing happens via JSON serialization. Complex types shared from JavaScript to .NET are sent as JSON and deserialized using `System.Text.Json`. There are a few simple types that will be shared as their intuitive .NET counterparts.
```
shyamnamboodiripad, line 182: Should we change the screenshot for the above example to use `#!set` instead of `#!share`? That may work better since the sharing is by value by default with `#!set` (and since there will be a nice error if someone tries to share by reference for JavaScript).

docs/javascript-overview.md
```diff
@@ -1,6 +1,90 @@
...
+console.log(window.declaredWithoutKeyword);
+```
+
+What does this have to do with Polyglot Notebooks?
+
+The Polyglot Notebooks JavaScript kernel executes your code submissions within an async arrow function, just like the above example:
+
+```javascript
+const doSomething = async () => {
+    // Your code here
```
shyamnamboodiripad, line 67: Just curious: Does some alternate approach exist that would allow us to execute some arbitrary piece of supplied JavaScript within a top-level scope? I am sure there is good reason to select the above approach - but was just curious if there are alternatives and what downsides these alternatives might have.

shyamnamboodiripad (approved):
Looks great 👍🏾

docs/javascript-overview.md
jonsequitur: Good call.

PR #3029, opened 2023-06-09
Consolidate formatting for http request and response types

shyamnamboodiripad:
Also switch to attribute-based formatter registration.


src/Microsoft.DotNet.Interactive.HttpRequest/Model/HttpResponse.cs
```diff
@@ -1,11 +1,19 @@
...
+#nullable enable
+
 using System.Collections.Generic;
+using Microsoft.DotNet.Interactive.Formatting;
 
+#if HTTP_REQUEST_KERNEL
 namespace Microsoft.DotNet.Interactive.HttpRequest;
-
-public class HttpResponse
+[TypeFormatterSource(typeof(HttpResponseFormatterSource))]
```
colombod, line 11: I prefer not to have ifdef. This type is intended also to be used for variable sharing and named request. Maybe it is better discuss visibility.
@jonsequitur ?

src/Microsoft.DotNet.Interactive.HttpRequest/HttpResponseFormattingExtensions.cs
```diff
@@ -1,13 +1,19 @@
...
 
+#nullable enable
+
 using System.Collections.Generic;
 using System.Linq;
 using System.Text.Json;
 using Microsoft.AspNetCore.Html;
 using Microsoft.DotNet.Interactive.Formatting;
 
+#if HTTP_REQUEST_KERNEL
```
colombod, line 12: ditto

src/Microsoft.DotNet.Interactive.HttpRequest/Microsoft.DotNet.Interactive.HttpRequest.csproj
```diff
@@ -7,6 +7,7 @@
     <!-- AssemblyInformationalVersionAttribute contains a non-standard value -->
     <Deterministic Condition="'$(NCrunch)' == '1'">false</Deterministic>
     <Nullable>enable</Nullable>
+    <DefineConstants>HTTP_REQUEST_KERNEL</DefineConstants>
```
colombod, line 10: ditto

src/Microsoft.DotNet.Interactive.HttpRequest/Model/HttpResponse.cs
jonsequitur: I think the approach is preferable to the alternatives I can think of, which are:

* Make the `HttpRequest` and `HttpResponse` types public in `Formatting`. (I don't think they belong there, and there's already more stuff there than I'd prefer.)
* Have divergent formatting for e.g. `HttpResponse` and `HttpResponseMessage`. (This is confusing for users and extra maintenance work for us.)

Other ideas?

jonsequitur:
I think a test that shows that `HttpResponseMessage` and `HttpResponse` have identical formatting output would be helpful.

src/Microsoft.DotNet.Interactive.HttpRequest/Model/HttpResponse.cs
colombod: That type is supposed to be the type to use when starting the response to other kernels

jonsequitur: Side note: Since in the context of the HTTP kernel, this is a message exchange format, we should also add some contract tests for it.

shyamnamboodiripad:
Abandoning in favor of the following as we discussed offline.
* Switch the formatter to format `HttpResponseMessage` directly.
* Introduce new async formatter API.

I will create a new set of PRs for the above.

PR #3030, opened 2023-06-09
fix #1650

### LINKED ISSUE #1650: Closing VS Code notebook orphans dotnet processes on MacOS
VS Code extension on Mac version v1.0.2462010
`#!about` Version: 1.0.246201+da749355d416da20e634e5c80073b92356b57e0e

MacOS using VS Code Version: 1.60.2
macOS Big Sur 11.6

Open/create a dib notebook. After running a C# cell, close VS Code with Cmd-Q.
There are two `dotnet` processes each consuming a whole core.

I have to kill these processes manually because they aren't killed when VS Code exits.

jonsequitur:


src/dotnet-interactive/CommandLine/CommandLineParser.cs
```diff
@@ -404,11 +405,12 @@ Command NotebookParser()
             var notebookParserCommand = new Command(
                 "notebook-parser",
                 LocalizationResources.Cli_dotnet_interactive_notebook_parserDescription());
-            notebookParserCommand.Handler = CommandHandler.Create(async () =>
+            notebookParserCommand.Handler = CommandHandler.Create(async (InvocationContext context) =>
             {
                 Console.InputEncoding = Encoding.UTF8;
                 Console.OutputEncoding = Encoding.UTF8;
                 var notebookParserServer = new NotebookParserServer(Console.In, Console.Out);
+                context.GetCancellationToken().Register(() => notebookParserServer.Dispose());
```
shyamnamboodiripad, line 413: Is it possible to test this?

src/polyglot-notebooks-vscode-common/src/extension.ts
```diff
@@ -335,6 +338,7 @@ export async function activate(context: vscode.ExtensionContext) {
 }
 
 export function deactivate() {
+    disposables.forEach(d => d());
```
shyamnamboodiripad, line 341: Same question: Can we test this?

shyamnamboodiripad (approved):

src/dotnet-interactive/CommandLine/CommandLineParser.cs
jonsequitur: Most of the implementation details are already tested in System.CommandLine. The final composed functionality involves both VS Code and .NET Interactive together, and is OS-specific.

We're planning to publish a VS Code Insiders release and seeing if this resolves the issue. Once we think this is the right fix, we'll look at how to write an effective pinning test.

PR #3032, opened 2023-06-12
js share by reference formatted values

colombod:


jonsequitur (approved):

src/Microsoft.DotNet.Interactive.Tests/VariableSharingTests.SetMagicCommand.cs
```diff
@@ -494,6 +494,43 @@ public async Task value_option_is_required()
                    .Which.Message.Should().Be("Option '--value' is required.");
         }
 
+        [Fact]
+        public async Task ProxyKernels_can_share_values()
```
jonsequitur, line 498: This test name should be more specific.

src/Microsoft.DotNet.Interactive/KernelExtensions.cs
```diff
@@ -173,51 +173,29 @@ static KernelInfoCollection CreateKernelInfos(CompositeKernel kernel)
...
-                {
-                }
-                else
-                {
-                }
+            var sourceKernel = Kernel.Root.FindKernelByName(valueOptionResult.Kernel);
+
+            ValueProduced valueProduced;
+            if (
+                sourceKernel?.KernelInfo.IsProxy == false
```
jonsequitur, line 180: I don't believe `KernelInfo` can ever be null.

src/Microsoft.DotNet.Interactive/KernelExtensions.cs
```diff
@@ -383,6 +361,8 @@ ValueOptionResult ParseValueOption(ArgumentResult argResult)
                 isByref = false;
             }
 
+            var valueSourceKernel = destinationKernel.RootKernel.FindKernelByName(sourceValueName);
```
jonsequitur, line 364: No longer needed?

src/Microsoft.DotNet.Interactive/KernelExtensions.cs
colombod: if the kernel is not found. that is the case when using @input:x

src/Microsoft.DotNet.Interactive/KernelExtensions.cs
colombod: ops

src/Microsoft.DotNet.Interactive/KernelExtensions.cs
jonsequitur: Ah right.

PR #3038, opened 2023-06-14
allow TypeFormatterSource attributes to set MIME type preferences 

jonsequitur:
This adds the ability for attribute-based formatter registrations to also set preferred formatter MIME types for the attributed type. This applies to both the public `TypeFormatterSourceAttribute` and the convention-based ones that don't require a reference to Microsoft.DotNet.Interactive.Formatting.

### DIFF
--- a/src/Microsoft.DotNet.Interactive.Formatting.Tests/FormatterTests.FormatterSources.cs
+++ b/src/Microsoft.DotNet.Interactive.Formatting.Tests/FormatterTests.FormatterSources.cs
@@ -52,6 +52,24 @@ public sealed partial class FormatterTests
         }
 
         [Fact]
+        public void Formatter_sources_can_set_preferred_MIME_type()
+        {
+            var obj = new TypeWithCustomFormatterAndMimeType();
+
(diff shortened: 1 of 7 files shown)


src/Microsoft.DotNet.Interactive.Formatting.Tests/FormatterTests.FormatterSources.cs
```diff
@@ -64,12 +82,35 @@ public void Convention_based_formatter_sources_are_still_registered_after_format
 
             formattedAfter.Should().Be(formattedBefore);
         }
-        
+
+        [Fact]
+        public void Convention_based_formatter_sources_can_set_preferred_MIME_type()
```
shyamnamboodiripad, line 87: This test seems to be a dupe of the above one. Am I missing something?

colombod (approved):

src/Microsoft.DotNet.Interactive.Formatting.Tests/FormatterTests.FormatterSources.cs
jonsequitur: The type being configured is different.

shyamnamboodiripad (approved):

PR #3039, opened 2023-06-14
Set Culture in dotnet interactive command line

ocallesp:
Three parts of this implementation:

- [ ] Add a new option to pass an specific Culture to the dotnet interactive command line
- [ ] In VS, should launch the process dotnet interactive passing a Culture.
- [ ] In VS Code, should launch the process dotnet interactive passing a Culture.

-----
It seems that we need to get a new architecture around this to correctly propagate Culture to all kernels


ocallesp:
Abandon this since we are going to take a different approach.

PR #3042, opened 2023-06-15
HTTP kernel cleanups and some unimplemented tests

jonsequitur:
This removes the `BaseAddress` property from `HttpRequestKernel` and adds some tests, some of which won't be able to pass until the HTTP parser is updated.

### DIFF
--- a/src/Microsoft.DotNet.Interactive.HttpRequest/HttpRequestKernel.cs
+++ b/src/Microsoft.DotNet.Interactive.HttpRequest/HttpRequestKernel.cs
@@ -29,10 +29,8 @@ public class HttpRequestKernel :
        IKernelCommandHandler<RequestDiagnostics>
 {
     private readonly HttpClient _client;
-    private readonly Argument<Uri> _hostArgument = new();
 
     private readonly Dictionary<string, string> _variables = new(StringComparer.InvariantCultureIgnoreCase);
-    private Uri? _baseAddress;
     private static readonly Regex IsRequest;
     private static readonly Regex IsHeader;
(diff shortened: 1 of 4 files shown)


shyamnamboodiripad (approved):

src/Microsoft.DotNet.Interactive.HttpRequest/HttpRequestKernel.cs
```diff
@@ -56,31 +54,10 @@ public HttpRequestKernel(string? name = null, HttpClient? client = null)
...
-            var host = context.ParseResult.GetValueForArgument(_hostArgument);
-            BaseAddress = host;
-        });
-        AddDirective(setHost);
-
+    
         RegisterForDisposal(_client);
     }
-
-    public Uri? BaseAddress
```
shyamnamboodiripad, line 71: Looks like the API compatibility test baseline needs to be updated for removal of this public API.

shyamnamboodiripad: There is also a public `SetValue` function below. Does it need to be part of the public API considering the same can be accomplished via a `SendValue` commands?

jonsequitur: That can be removed I suspect.

colombod: if the gesture is like value sharing then yes, we should remove that.

PR #3043, opened 2023-06-16
Consolidate formatting for http request and response types.

shyamnamboodiripad:
Also switch to attribute-based formatter registration.


src/Microsoft.DotNet.Interactive.AspNetCore/AspNetCoreCSharpKernelExtensions.cs
```diff
@@ -37,18 +38,20 @@ public static class AspNetCoreCSharpKernelExtensions
...
         typeof(InteractiveEndpointRouteBuilderExtensions).Namespace, // Microsoft.DotNet.Interactive.AspNetCore
-        typeof(HttpClient).Namespace, // System.Net.Htttp
+        typeof(HttpClient).Namespace, // System.Net.Http
     };
 
     public static CSharpKernel UseAspNetCore(this CSharpKernel kernel)
     {
         InteractiveHost interactiveHost = null;
 
+        HttpResponseMessageFormattingExtensions.RegisterFormatters();
```
shyamnamboodiripad, line 48: @jonsequitur Now that I think about it, this may not be a good idea since this could override formatters that were registered by the user after `CommandLineParser.SetUpFormatters` was called in the tool.

Perhaps this registration should happen directly in the tests (which depend on the formatter)... I'll change this.

src/Microsoft.DotNet.Interactive.HttpRequest/HttpRequestKernelExtension.cs
```diff
@@ -16,7 +16,12 @@ public static void Load(Kernel kernel, HttpClient? httpClient = null)
             compositeKernel.Add(httpRequestKernel);
             httpRequestKernel.UseValueSharing();
 
-            RegisterFormatters();
+            // Include html, plain text and json formatted outputs in display event responses.
+            Formatter.SetPreferredMimeTypesFor(
```
jonsequitur, line 20: This can be done via the `TypeFormatterSourceAttribute` now.

shyamnamboodiripad: Ah indeed - let me change that.

jonsequitur (approved):

PR #3048, opened 2023-06-19
handling number literals in value sharing

colombod:
image

### DIFF
--- a/src/polyglot-notebooks-vscode-common/src/utilities.ts
+++ b/src/polyglot-notebooks-vscode-common/src/utilities.ts
@@ -8,6 +8,7 @@ import { v4 as uuid } from 'uuid';
 import { InstallInteractiveArgs, ProcessStart } from "./interfaces";
 import { NotebookCellOutput, NotebookCellOutputItem, ReportChannel, Uri } from './interfaces/vscode-like';
 import * as commandsAndEvents from './polyglot-notebooks/commandsAndEvents';
+import * as connection from './polyglot-notebooks/connection';
 import { OutputChannelAdapter } from './OutputChannelAdapter';
 import { Logger } from './polyglot-notebooks';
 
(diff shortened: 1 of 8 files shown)


colombod:
@shibbas

src/polyglot-notebooks-vscode-common/src/utilities.ts
```diff
@@ -232,6 +232,15 @@ export function parse(text: string): any {
             return Uint8Array.from(buffer.values());
         }
 
+        //handling NaN, Infinity and -Infinity
+        if (value === "Infinity") {
+            return Infinity;
+        } else if (value === "-Infinity") {
+            return -Infinity;
+        } else if (value === "NaN") {
```
jonsequitur, line 240: Micro-optimization micro-nitpick: `NaN` will be by far the most common case here and might should be checked before the others.

src/polyglot-notebooks-vscode-common/src/utilities.ts
```diff
@@ -250,6 +259,15 @@ export function stringify(value: any): string {
...
         }
 
+        //handling NaN, Infinity and -Infinity
+        if (value === Infinity) {
+            return "Infinity";
+        } else if (value === -Infinity) {
+            return "-Infinity";
+        } else if (value !== value) {
+            return "NaN";
+        }
```
jonsequitur, line 269: Any reason to have two copies of this logic?

src/polyglot-notebooks/src/connection.ts
```diff
@@ -303,3 +303,31 @@ export function extractHostAndNomalize(kernelUri: string): string {
...
+        if (value === "Infinity") {
+            return Infinity;
+        } else if (value === "-Infinity") {
+            return -Infinity;
+        } else if (value === "NaN") {
+            return NaN;
+        }
+        return value;
+    });
+}
```
jonsequitur, line 333: Can we consolidate these multiple copies of the same logic?

src/polyglot-notebooks-vscode-common/src/utilities.ts
colombod: indeed

colombod: done

PR #3051, opened 2023-06-21
Introduce a command to replace contents of selected html elements

shyamnamboodiripad:
This makes it easier for clients to implement support for updating previously displayed values in response to `DisplayedValueUpdated` events.


ocallesp:
Abandon this PR since we are going to take a different approach.

PR #3052, opened 2023-06-21
Improve handling for http responses that are too slow and / or too large

shyamnamboodiripad:
Updates the `HttpRequestKernel` to send leading `DisplayEvent`s with partial information when the complete http response is too slow and / or too large. These leading events are then used to drive progress display to improve end-user perceived performance.

### DIFF
--- a/src/Microsoft.DotNet.Interactive.HttpRequest.Tests/HttpRequestKernelTests.cs
+++ b/src/Microsoft.DotNet.Interactive.HttpRequest.Tests/HttpRequestKernelTests.cs
@@ -6,8 +6,10 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Net;
 using System.Net.Http;
+using System.Text;
 using System.Threading.Tasks;
 using FluentAssertions;
+using FluentAssertions.Execution;
 using Microsoft.DotNet.Interactive.Commands;
 using Microsoft.DotNet.Interactive.Events;
 using Microsoft.DotNet.Interactive.Formatting;
@@ -45,6 +47,8 @@ public class HttpRequestKernelTests
 
         var result = await kernel.SendAsync(new SubmitCode($"{verb} http://testuri.ninja"));
 
+        using var _ = new AssertionScope();
+
         result.Events.Should().NotContainErrors();
 
         request.Method.Method.Should().Be(verb);
@@ -63,10 +67,12 @@ public class HttpRequestKernelTests
         var client = new HttpClient(handler);
         using var kernel = new HttpRequestKernel(client: client);
 
-        kernel.SetValue("my_host", "my.host.com");
+        using var _ = new AssertionScope();
 
-        var result = await kernel.SendAsync(new SubmitCode("get  https://{{my_host}}:1200/endpoint"));
+        var result = await kernel.SendAsync(new SendValue("my_host", "my.host.com"));
+        result.Events.Should().NotContainErrors();
 
+        result = await kernel.SendAsync(new SubmitCode("get  https://{{my_host}}:1200/endpoint"));
         result.Events.Should().NotContainErrors();
 
         request.RequestUri.Should().Be("https://my.host.com:1200/endpoint");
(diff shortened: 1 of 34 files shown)


src/Microsoft.DotNet.Interactive.HttpRequest/HttpRequestKernel.cs
```diff
@@ -97,44 +94,135 @@ async Task IKernelCommandHandler<SubmitCode>.HandleAsync(SubmitCode command, Ker
...
+        }
 
-            context.Display(response);
+        void ClearDisplayedValue()
+        {
+            if (displayedValue is not null)
+            {
+                var htmlFormattedValue = new FormattedValue(HtmlFormatter.MimeType, "<span/>");
+                var plainTextFormattedValue = new FormattedValue(PlainTextFormatter.MimeType, string.Empty);
+                displayedValue.Update(null, htmlFormattedValue, plainTextFormattedValue);
```
shyamnamboodiripad, line 187: Avoid need to pass null.

shyamnamboodiripad: (Fixed)

src/Microsoft.DotNet.Interactive/FormattedValue.cs
```diff
@@ -19,35 +19,39 @@ public FormattedValue(string mimeType, string value)
...
         }
 
-        return new FormattedValue(mimeType, value.ToDisplayString(mimeType));
+        return new FormattedValue(mimeType, value.ToDisplayString(mimeType), suppressDisplay);
     }
 
     public static IReadOnlyList<FormattedValue> CreateManyFromObject(object value, params string[] mimeTypes)
+        => CreateManyFromObject(value, suppressDisplay: false, mimeTypes);
+
+    public static IReadOnlyList<FormattedValue> CreateManyFromObject(object value, bool suppressDisplay = false, params string[] mimeTypes)
```
jonsequitur, line 44: Once we start having this many properties on the type, I'm not sure the immutability is helping any, and adding the overload is a lot more surface area than for example just making `FormattedValue.SuppressDisplay` settable.

shyamnamboodiripad: Good call. Removed the overloads and made the property settable in the latest commit.

src/Microsoft.DotNet.Interactive.HttpRequest.Tests/HttpRequestKernelTests.cs
```diff
@@ -363,12 +393,174 @@ public async Task produces_json_html_and_plain_text_formatted_values()
...
+    [Fact]
+    public async Task produces_empty_response_for_slow_responses()
+    {
+        HttpRequestMessage request = null;
+        var slowHandler = new InterceptingHttpMessageHandler(async (message, _) =>
+        {
+            request = message;
+            var response = new HttpResponseMessage(HttpStatusCode.OK);
+            response.RequestMessage = message;
+            await Task.Delay(1500);
```
jonsequitur, line 422: Is this delay a magic number of some kind that indicates a "slow" response? Putting a hard-coded delay like this into a test slows tests down while also likely being nondeterministic. Let's work on improving this test design.

src/Microsoft.DotNet.Interactive.HttpRequest/Formatting/HttpResponseFormattingExtensions.cs
```diff
@@ -48,6 +75,43 @@ internal static void FormatAsHtml(this HttpResponse? response, FormatContext con
...
+        PocketView? output = null;
+
+        if (response is HttpResponse fullResponse)
+        {
+            output = fullResponse.FormatAsHtml();
+        }
+        else if (response is PartialHttpResponse partialResponse)
+        {
+            output =
+                PocketViewTags.div[@class: ContainerClass](
```
jonsequitur, line 87: These can be more concise if you add this at the top:

```csharp
using static Microsoft.DotNet.Interactive.Formatting.PocketViewTags;
```

src/Microsoft.DotNet.Interactive.HttpRequest.Tests/HttpRequestKernelTests.cs
shyamnamboodiripad: The spinner is displayed on responses that take longer than a second. Yes glad to improve this if you have suggestions.

src/Microsoft.DotNet.Interactive.HttpRequest/Formatting/HttpResponseFormattingExtensions.cs
shyamnamboodiripad: Fixed in latest iteration.

jonsequitur (approved):

Issue #2855, opened 2023-03-21
HTML formatting of `IEnumerable` and `IEnumerable<T>` is different


jonsequitur:
<img width="695" alt="image" src="https://user-images.githubusercontent.com/547415/226713198-f0037325-4223-4ca4-9d18-d139075e0f23.png">

Issue #2854, opened 2023-03-21
HTML formatting of `IEnumerable` types skips properties


jonsequitur:
<img width="695" alt="image" src="https://user-images.githubusercontent.com/547415/226712993-bcdb40e9-bb6e-426e-b5b0-76a047e707be.png">

PR #3074, opened 2023-07-07
Show properties as well as values when formatting `IEnumerable` types

jonsequitur:
This fixes the following issues:

* #2854
* #2855


src/Microsoft.DotNet.Interactive.Formatting/HtmlFormatter{T}.cs
```diff
@@ -186,7 +192,17 @@ bool BuildTable(T source, FormatContext context, bool summarize)
                     }
                 }
 
-                rows.Add(tr(rowValues.Select(r => td(r))));
+                rows.Add(tr(rowValues.Select(value =>
```
colombod, line 195: do we need any change on the TabularDataResource Formatter?

jonsequitur: What kind of change?

src/Microsoft.DotNet.Interactive.Formatting/Formatter.cs
```diff
@@ -309,6 +309,7 @@ public static void Format(this ITypeFormatter formatter, object instance, TextWr
     public static void FormatTo<T>(
         this T obj,
         FormatContext context,
+        // FIX: (FormatTo) make this non-optional
```
shyamnamboodiripad, line 312: Did you intend to remove this before merge?

src/Microsoft.DotNet.Interactive.Formatting/Formatter.cs
```diff
@@ -766,4 +685,40 @@ private static bool TryGetPropertyValue<T>(object fromObject, string propertyNam
...
+        {
+            return false;
+        }
+
+        if (type.IsNestedPrivate) //e.g. RangeIterator
+        {
+            return false;
+        }
+
+        if (typeof(ICollection).IsAssignableFrom(type))
```
shyamnamboodiripad, line 701: Curious why properties are included for types that implement `IEnumerable` but not `ICollection`.

shyamnamboodiripad, line 701: This may be a micro-optimization, but considering that we are already looking at implemented interfaces in the loop below, would it be better to add an early out for `ICollection` there instead? (Taking a quick look at the implementation for `IsAssignableFrom` it seems to be doing a bunch morethings before checking if an implemented interface `IsSubClassOf(typeof(ICollection))` )

src/Microsoft.DotNet.Interactive.Formatting/HtmlFormatter.cs
```diff
@@ -43,13 +42,21 @@ void Initialize()
 
     internal static void FormatAndStyleAsPlainText(
         object value,
-        FormatContext context)
+        FormatContext context,
+        string mimeType = PlainTextFormatter.MimeType)
```
shyamnamboodiripad, line 46: It is a bit confusing to see the `PlainText` suffix in the name of the function when the mime type is also specified in this parameter. Could we do something to disambiguate?

src/Microsoft.DotNet.Interactive.Formatting/HtmlFormatter.cs
```diff
@@ -137,26 +144,36 @@ void Initialize()
...
+
             // This is approximate
             var isKnownDocType =
                 type.Namespace is not null &&
+                type.FullName is not null &&
                 (type.Namespace == "System" ||
                  type.Namespace.StartsWith("System.") ||
-                 type.Namespace.StartsWith("Microsoft."));
+                 type.Namespace.StartsWith("Microsoft.")) &&
+                !type.IsAnonymous();
```
shyamnamboodiripad, line 155: Consider moving the check for anonymous types earlier since its probably less expensive than the string comparisons above.

src/Microsoft.DotNet.Interactive.Formatting/HtmlFormatter{T}.cs
```diff
@@ -246,14 +292,16 @@ static bool BuildTreeView(T source, FormatContext context, MemberAccessor<T>[] m
...
             PocketView view = details[attributes](
                 summary(
-                    span[@class: "dni-code-hint"](code)),
+                    span[@class: "dni-code-hint"](summaryContent)),
                 div(
                     Html.Table(
                         headers: null,
-                        rows: memberAccessors.Select(
+                        rows: rows.Select(
                             a => (IHtmlContent)
```
shyamnamboodiripad, line 300: minor: Consider renaming `a` to `row`

shyamnamboodiripad: Never mind - looks like we also have other instances of `a` and `c` above and after staring at the code a bit longer it is clearer that they stand for `accessor` and `context` respectively :)

src/Microsoft.DotNet.Interactive.Formatting/HtmlFormatter{T}.cs
```diff
@@ -210,15 +226,45 @@ internal static HtmlFormatter<T> CreateTreeViewFormatterForAnyObject()
...
+            {
+                var enumerableFormatter = HtmlFormatter.GetDefaultFormatterForAnyEnumerable(typeof(T));
+
+                (HtmlTag propertyLabelTdTag, Func<T, HtmlTag> getPropertyValueTdTag) enumerableAccessor =
+                    (new HtmlTag("td", new HtmlTag("i", "(values)")), obj => new HtmlTag("td", c =>
+                        {
+                            enumerableFormatter.Format(obj, c);
+                        }));
+                Array.Resize(ref rows, rows.Length + 1);
+                rows[^1] = enumerableAccessor;
```
shyamnamboodiripad, line 257: minor: Would it be better to leave rows as an enumerable and use `Enumerable.Append` here (and call `ToArray` right before calling `BuildTreeView` below)?

src/Microsoft.DotNet.Interactive.Formatting/HtmlFormatter.cs
```diff
@@ -121,7 +128,7 @@ void Initialize()
 
         new HtmlFormatter<string>((s, context) =>
         {
-            // If PlainTextPreformat is true, then strings
+            // If PreformatPlainText is true, then strings
```
shyamnamboodiripad, line 131: Is `PreformatPlainText` referring to an option that is present inside `context`?

src/Microsoft.DotNet.Interactive.Formatting/PlainTextFormatter{T}.cs
```diff
@@ -107,10 +107,14 @@ internal static PlainTextFormatter<T> CreateForAnyEnumerable()
 
         return new((value, context) =>
         {
+            using var a = context.IncrementDepth();
+            using var b = context.IncrementDepth();
```
shyamnamboodiripad, line 111: The variables (`a` and `b`) seem to be unused - I am guessing you added them for testing and intended to remove them before merge.

src/Microsoft.DotNet.Interactive.Formatting/PlainTextSummaryFormatter.cs
```diff
@@ -8,6 +8,7 @@ namespace Microsoft.DotNet.Interactive.Formatting;
 
 public static class PlainTextSummaryFormatter
 {
+    // FIX: (PlainTextSummaryFormatter) rename this
     public const string MimeType = "text/plain+summary";
```
shyamnamboodiripad, line 12: Again, I am guessing you intend to fix this up before merge. If not, I wonder if it may be better to log an issue to track this...

jonsequitur: That's a bigger change because it affects the VS Code extension. I've opened #3078 to track it.

src/Microsoft.DotNet.Interactive.Formatting.Tests/Utility/StringExtensions.cs
```diff
@@ -25,7 +26,19 @@ public static string Crunch(this string s)
 
     public static string IndentHtml(this string html)
     {
-        return XElement.Parse(html).ToString();
+        try
+        {
+            return XElement.Parse(html).ToString();
+        }
+        catch (XmlException ex) when (ex.Message.Contains("multiple root elements"))
```
shyamnamboodiripad, line 33: Would we ever want to run tests on localized / non-ENU setups?

src/Microsoft.DotNet.Interactive.Formatting/Formatter.cs
```diff
@@ -766,4 +685,40 @@ private static bool TryGetPropertyValue<T>(object fromObject, string propertyNam
...
+            {
+                if (@interface.GetGenericTypeDefinition() == typeof(IOrderedEnumerable<>))
+                {
+                    return false;
+                }
+
+                if (@interface.GetGenericTypeDefinition() == typeof(IDictionary<,>))
+                {
+                    return false;
+                }
```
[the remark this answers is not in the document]
jonsequitur: These interfaces have properties that aren't very interesting, e.g. `Count` (since we already show it for all sequences.)

src/Microsoft.DotNet.Interactive.Formatting/Formatter.cs
jonsequitur: `ICollection` is a lot more specific and includes a `Count` property which is redundant with the existing sequence formatters, which show e.g. `(13 more)` after hitting the list expansion limit.

shyamnamboodiripad (approved):

src/Microsoft.DotNet.Interactive.Formatting/HtmlFormatter.cs
jonsequitur: We own the `IsAnonymous` method and it's likely the most expensive of the operations in this expression.

src/Microsoft.DotNet.Interactive.Formatting/HtmlFormatter{T}.cs
jonsequitur: I would assume that would be less performant than doing the array operations directly, though it would be far from the most critical piece of code to optimize here.

src/Microsoft.DotNet.Interactive.Formatting/HtmlFormatter.cs
jonsequitur: Deleting this comment. It was outdated and refers to something that no longer exists.

src/Microsoft.DotNet.Interactive.Formatting/PlainTextFormatter{T}.cs
jonsequitur: They are being used because we're `using` them. 😂 (Sorry, that pun was right there.) In other words, `Dispose` is being called on them.

The variable names are awful.

jonsequitur: I often use `_` as a variable name for these situations. I suppose `_` and `__` can work.

src/Microsoft.DotNet.Interactive.Formatting.Tests/Utility/StringExtensions.cs
jonsequitur: It's unlikely but if so, it will produce a false negative rather than a false positive, which is good, and it will likely be pretty clear what happened. I did look for a better programmatic way to determine the specific cause of the `XmlException` but I didn't find anything.

src/Microsoft.DotNet.Interactive.Formatting/PlainTextFormatter{T}.cs
shyamnamboodiripad: Ah I missed the `using` keyword 😄

src/Microsoft.DotNet.Interactive.Formatting/HtmlFormatter.cs
shyamnamboodiripad: Ah interesting - my bad I assumed it was a BCL API that would be checking some flag in the metadata.

PR #3090, opened 2023-07-14
Recognize VS as a well known front end

shyamnamboodiripad:
This will allow us to see usages that originate inside VS within our telemetry dashboards.

### DIFF
--- a/src/Microsoft.DotNet.Interactive.Telemetry/StartupTelemetryEventBuilder.cs
+++ b/src/Microsoft.DotNet.Interactive.Telemetry/StartupTelemetryEventBuilder.cs
@@ -277,7 +277,8 @@ public sealed class StartupTelemetryEventBuilder
                     return directive.Key;
             }
 
-            if (directive.Key.StartsWith("vs")) // VS also appends the process id to the string.
+            if (directive.Key.StartsWith("vs") &&
+                int.TryParse(directive.Key.Substring(2), out _)) // VS appends the process id after the "vs" prefix.
             {
                 return "vs";
             }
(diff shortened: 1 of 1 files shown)


src/Microsoft.DotNet.Interactive.Telemetry/StartupTelemetryEventBuilder.cs
```diff
@@ -276,6 +276,11 @@ private static CommandRuleItem Ignore(TokenType type, bool isOptional)
                 case "vscode":
                     return directive.Key;
             }
+
+            if (directive.Key.StartsWith("vs")) // VS also appends the process id to the string.
+            {
+                return "vs";
```
jonsequitur, line 282: Are we expecting other variations that start with "vs"? Why not just combine this into the `switch` above?

shyamnamboodiripad: Unlike the existing cases in the `switch` above, the vs case is not an exact match because the VS launcher currently includes the pid of the parent VS process as a suffix after "vs" (mainly because this has been useful for local testing / debugging). Is there some way to do a `StartsWith` check for a string literal inside a `case` statement?

jonsequitur: The code structure makes sense in this case, although I wonder if there's any risk of false positives here.

shyamnamboodiripad: I did consider that but assumed this may be ok for now given the limited number of use cases. I can certainly make the check stronger by recognizing the front end as VS only if the `vs` prefix is followed by an integer (for the pid).

shyamnamboodiripad: I've pushed a commit that does what I described above to reduce the chance of false positives.

AbhitejJohn (approved):

Issue #3086, opened 2023-07-13
add support for tabular display of `IEnumerable<T>`


jonsequitur:
Proposal: Add a `DisplayTable` extension method for `IEnumerable<T>` that will produce a tabular display instead of a treeview-style display.

A number of people have asked for the older table formatter to be made available again for sequences.

#3081
#3085

In #3081 I proposed a workaround by first calling `ToTabularDataResource`, which has a custom formatter that provides a tabular output:

<img width="678" alt="image" src="https://github.com/dotnet/interactive/assets/547415/5ad65849-c490-43a4-b7d7-68ce06ea3f2b">

The ergonomics could be improved though. 

In [this comment](https://github.com/dotnet/interactive/discussions/3081#discussioncomment-6431128), @scottstephens made a suggestion for an API that would sit alongside `Display` and not require a separate `using` using directive. This will make it more discoverable and decouple it from the `TabularDataResource` details, which are intended more for serialization than display.

PR #3098, opened 2023-07-20
add DisplayTable method

jonsequitur:
This addresses #3086.

It adds a new extension method, `IEnumerable<T>.DisplayTable`, which can be used to display objects in a tabular format with the property names as headers. This implementation is unlike the original tabular format replaced in #2671 in the following ways: 

* Nested objects are displayed as tree views rather than as nested tables. 
* Property headers are based on the type `T` rather than on the object instances, so instances of heterogeneous types in the sequence will not appear different from one another in the table.

I've also made some changes to the `DisplayedValue` class, removing the `MimeTypes` property and instead storing the complete `FormattedValue` instances, allowing inspection from the notebook, which was previously not possible:

image

Finally, I've removed a duplicate `Diagnostic` class.


src/Microsoft.DotNet.Interactive.Formatting/HtmlFormatter.cs
```diff
@@ -226,22 +226,7 @@ void Initialize()
             var formatter = GetDefaultFormatterForAnyObject(type);
             return formatter.Format(value, context);
         }),
-
-        // Final last resort is to convert to plain text
-        new HtmlFormatter<object>((value, context) =>
```
shyamnamboodiripad, line 231: Was this removed because it was redundant?

src/Microsoft.DotNet.Interactive.Formatting/HtmlFormatter.cs
```diff
@@ -226,22 +226,7 @@ void Initialize()
             var formatter = GetDefaultFormatterForAnyObject(type);
```
shyamnamboodiripad, line 226: Looks like most lambdas in this collection don't guard against the `value` being `null` (although the removed last resort one below was checking for null). Is this intentional or does some other piece of code else already filter out nulls before these lambdas are invoked?

src/Microsoft.DotNet.Interactive.Tests/Parsing/SubmissionParserTests.cs
```diff
@@ -134,15 +134,24 @@ public void Directive_parsing_errors_are_available_as_diagnostics()
...
+            .Be(node.Span.End);
+
+        diagnostics
             .Should()
-            .ContainSingle(d => d.Severity == DiagnosticSeverity.Error)
+            .ContainSingle(d => d.Severity == CodeAnalysis.DiagnosticSeverity.Error)
             .Which
-            .Location
-            .SourceSpan
+            .LinePositionSpan.Start.Character
```
shyamnamboodiripad, line 152: Considering creating a local for the LinePositionSpan so that the code to select the diagnostic need not be repeated across the 2 assertions.

src/Microsoft.DotNet.Interactive.Tests/LanguageKernelFormattingTests.cs
```diff
@@ -145,7 +146,69 @@ await kernel.FindKernelByName("csharp").As<CSharpKernel>()
     [Theory]
     [InlineData(Language.CSharp)]
     [InlineData(Language.FSharp)]
-    public async Task Display_helper_can_be_called_without_specifying_class_name(Language language)
+    public async Task DisplayTable_produces_tabular_HTML_output_for_IEnumerable_T(Language language)
```
shyamnamboodiripad, line 149: Should there be another test to validate what happens when the collection contains instances with heterogeneous types?

src/Microsoft.DotNet.Interactive/DisplayedValue.cs
```diff
@@ -6,50 +6,36 @@
...
-        }
-
-        if (mimeTypes is null || mimeTypes.Count(mimetype => !string.IsNullOrWhiteSpace(mimetype)) == 0)
-        {
-            throw new ArgumentException("Value cannot be null or empty.", nameof(mimeTypes));
-        }
-
-        _displayId = displayId;
-        _mimeTypes = new HashSet<string>(mimeTypes.Where(mimetype => !string.IsNullOrWhiteSpace(mimetype)));
+        FormattedValues = formattedValues ?? throw new ArgumentNullException(nameof(formattedValues));
```
shyamnamboodiripad, line 18: Should this also validate that there is at least one formatted value?

src/Microsoft.DotNet.Interactive/DisplayedValue.cs
```diff
@@ -6,50 +6,36 @@
...
         _context = context ?? throw new ArgumentNullException(nameof(context));
+
+        DisplayId = Guid.NewGuid().ToString();
     }
 
-    public IReadOnlyCollection<string> MimeTypes => _mimeTypes;
+    public string DisplayId { get; }
```
shyamnamboodiripad, line 24: Can this be private? Just checking since this seems like an implementation detail.

shyamnamboodiripad (approved):

shyamnamboodiripad:
@jonsequitur I have gone ahead and merged this since none of my comments above seemed blocking (and can be addressed in a follow up PR)

src/Microsoft.DotNet.Interactive.Formatting/HtmlFormatter.cs
jonsequitur: Yes. There were two different registrations for the same type. This one had no test coverage so it wasn't important.

src/Microsoft.DotNet.Interactive.Formatting/HtmlFormatter.cs
jonsequitur: Matching is done based on the runtime type and the formatter provides a universal early out for `null`.

PR #3099, opened 2023-07-21
Update PocketLogger version

shyamnamboodiripad:


ocallesp (approved):
It seems that a central package versioning is needed.

PR #3103, opened 2023-07-25
Addition of Http Parser 

bleaphar:
This is the beginning of a package to tokenize and parse htttp requests.


src/Microsoft.DotNet.Interactive.HttpRequest.Tests/LexerTests.cs
```diff
@@ -0,0 +1,10 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.DotNet.Interactive.HttpRequest.Tests;
+
+public class LexerTests
```
jonsequitur, line 6: We can leave out the empty test. It might not be necessary to directly test the lexer API if it's not public.

src/Microsoft.DotNet.Interactive.HttpRequest.Tests/ParserTests.cs
```diff
@@ -0,0 +1,614 @@
...
+    [Fact]
+    public void request_node_without_method_node_created()
+    {
+        var result = HttpRequestParser.Parse("https://example.com");
+
+        result.SyntaxTree.RootNode.ChildNodes.Should().ContainSingle<HttpRequestNode>().Which
+            .MethodNode.Should().BeNull();
+    }
+}
+// TODO: Test string with variable declarations but no requests
```
jonsequitur, line 350: Do we need the commented out bit below for anything?

src/Microsoft.DotNet.Interactive.HttpRequest.Tests/LexerTests.cs
shyamnamboodiripad: I think we should add tests for the tokenization i.e., how we split whitespace, new lines, words and punctuation into tokens.

src/Microsoft.DotNet.Interactive.HttpRequestParser/HttpRequestParser.cs
```diff
@@ -0,0 +1,418 @@
...
+
+            ParseLeadingTrivia(node);
+
+            if (MoreTokens() && CurrentToken.Kind is HttpTokenKind.Word)
+            {
+                if (CurrentToken.Kind is HttpTokenKind.Word && CurrentToken.Text.ToLower() is ("get" or "post" or "patch" or "put" or "delete" or "head" or "options" or "trace"))
+                {                 
+                    ConsumeCurrentTokenInto(node);
+                }
+                else if (CurrentToken.Kind is HttpTokenKind.Word && CurrentToken.Text.ToLower() is not ("get" or "post" or "patch" or "put" or "delete" or "head" or "options" or "trace" or "https"))
```
jonsequitur, line 168: We can simplify these `if` statements to reduce repetition.

src/Microsoft.DotNet.Interactive.HttpRequestParser/HttpRequestParser.cs
```diff
@@ -0,0 +1,418 @@
...
+                else if (CurrentToken.Kind is HttpTokenKind.Word && CurrentToken.Text.ToLower() is "https")
+                {
+                    return null;
+                }
+
+
+            }
+            else
+            {
+                /*var tokenSpan = _sourceText.GetSubText(CurrentToken.Span).Lines.GetLinePositionSpan(CurrentToken.Span);
```
jonsequitur, line 185: Any reason for this empty branch?

src/Microsoft.DotNet.Interactive.HttpRequest.Tests/Microsoft.DotNet.Interactive.HttpRequest.Tests.csproj
```diff
@@ -16,6 +16,18 @@
     <None Remove="TestResults\**" />
   </ItemGroup>
 
+  <ItemGroup>
+    <Compile Include="..\Microsoft.DotNet.Interactive.HttpRequestParser\*.cs" LinkBase="Parser" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <Compile Remove="..\Microsoft.DotNet.Interactive.HttpRequestParser\obj\Debug\netstandard2.0\**" />
```
shyamnamboodiripad, line 24: Do we need this line? I think it may no longer be needed since we changed the `Compile` item on line 20 to only include *.cs files that are directly in the project folder (as opposed to recursively).

src/Microsoft.DotNet.Interactive.HttpRequestParser/Reference/DynamicVariable.cs
```diff
@@ -0,0 +1,9 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Text.RegularExpressions;
+
+namespace Microsoft.DotNet.Interactive.HttpRequest.Reference;
+
+internal record DynamicVariable(Regex Expression, Func<HttpDocumentSnapshot, string, MatchCollection, string> InvokeHandler);
```
jonsequitur, line 9: Are the files under the `Reference` folder needed here?

If we can remove them and the tests still pass and things compile, now is probably a good time to delete them.

src/Microsoft.DotNet.Interactive.HttpRequestParser/TextSpan.cs
```diff
@@ -0,0 +1,259 @@
+﻿/*// Copyright (c) .NET Foundation and contributors. All rights reserved.
```
jonsequitur, line 1: Can this file be deleted?

src/Microsoft.DotNet.Interactive.HttpRequest.Tests/ParserTests.cs
```diff
@@ -0,0 +1,614 @@
...
+
+        var requestNode = result.SyntaxTree.RootNode
+              .ChildNodes.Should().ContainSingle<HttpRequestNode>().Which;
+
+        requestNode.HeadersNode.HeaderNodes.Count.Should().Be(0);
+    }
+
+    //TODO: Is it an error to include a body with no headers?
+    [Fact]
+    public void body_without_header_body_is_parsed_correctly()
```
shyamnamboodiripad, line 243: The names of this test and the one above are confusing. Could you rename them to match what is being asserted?

src/Microsoft.DotNet.Interactive.HttpRequest.Tests/ParserTests.cs
```diff
@@ -0,0 +1,614 @@
...
+    public void request_node_without_method_node_created_correctly()
+    {
+        var result = HttpRequestParser.Parse("https://example.com");
+
+        result.SyntaxTree.RootNode.ChildNodes.Should().ContainSingle<HttpRequestNode>().Which
+            .MethodNode.Should().BeNull();
+    }
+
+    [Fact]
+    public void request_node_without_method_node_created()
```
shyamnamboodiripad, line 342: This test seems identical to the above one.

jonsequitur, line 342: Can the test name be more explicit about the intent, e.g `when_the_verb_is_not_present_then_method_node_is_null`? 

Though I'm not sure this is the correct behavior. Did we discuss using an empty method node? (This wold also avoid some of the null warnings that failed the build.)

src/Microsoft.DotNet.Interactive.HttpRequest.Tests/ParserTests.cs
shyamnamboodiripad: The commented bit is the set of tests copied over from the old parser - we were cutting and pasting individual tests from below to above.

I think we may want to preserve this until we have tests that test the new parser's handling for the commented cases...

src/Microsoft.DotNet.Interactive.HttpRequest.Tests/LexerTests.cs
jonsequitur: We did discuss adding one to ensure the the same content (including different kinds of newlines) is parsed consistently so that we avoid OS-specific behaviors.

Beyond that, it's possible that the lexer tests will be redundant since this code is heavily exercised by the parser tests.

src/Microsoft.DotNet.Interactive.HttpRequest.Tests/ParserTests.cs
jonsequitur: Understood. We should at least delete the ones that represent unfinished work.

src/Microsoft.DotNet.Interactive.HttpRequest.Tests/LexerTests.cs
shyamnamboodiripad: They would still tightly pin-point the problem if something changed in how we tokenize text though, would it not (as opposed to tens of parser tests failing)?

src/Microsoft.DotNet.Interactive.HttpRequestParser/Reference/DynamicVariable.cs
shyamnamboodiripad: Yes, I agree we can remove them. They were added so that we could easily see what the old parser was doing with its regexes - but we could just as easily look at the original code in the other repo for this...

src/Microsoft.DotNet.Interactive.HttpRequest.Tests/LexerTests.cs
jonsequitur: Absolutely. But I think those tests are still parser tests because the parser and nodes are the public API, which also surfaces the tokens. The lexer should not be part of the public API. It's an implementation detail of the parser and we don't have a use case for people to call it directly from outside, hence testing it directly should not be a goal.

src/Microsoft.DotNet.Interactive.HttpRequestParser/HttpRequestParser.cs
shyamnamboodiripad: I think we have a pattern where we repeatedly check for certain special states (for example a common start state we check for is `MoreTokens() && CurrentToken.Kind is not (TokenKind.WhileSpace or TokenKind.NewLine)`.

As we go along, it may be a good idea to pull out helpers for repeated patterns that we check for.

src/Microsoft.DotNet.Interactive.HttpRequest.Tests/ParserTests.cs
shyamnamboodiripad: Yes, I think there may be some scenarios pertaining to features we have already implemented that would be nice to move above now as opposed to later.

src/Microsoft.DotNet.Interactive.HttpRequestParser/HttpTokenKind.cs
```diff
@@ -0,0 +1,20 @@
...
+    Word                      --- -----   ------- ---  ---------
+    Whitespace                   -
+    Punctuation                        ---       -   --         -
+    NewLine                                                      _
+    */
+    Word,
+    Whitespace,
+    Punctuation,
+    NewLine,
+    Missing
```
shyamnamboodiripad, line 19: Do we need this? Iirc we said we could infer missing syntax based on span (anything with 0 span is missing). Iirc, we also don't synthesize `SyntaxToken`s for missing syntax - we just create `SyntaxNode`s with 0 span for this - so I think we can remove this `TokenKind`...

src/Microsoft.DotNet.Interactive.HttpRequestParser/HttpLexer.cs
```diff
@@ -0,0 +1,97 @@
...
+                         ? HttpTokenKind.Punctuation
+                         : HttpTokenKind.Word,
+            };
+
+            if (previousTokenKind is { } previousTokenKindValue)
+            {
+                if (currentTokenKind is HttpTokenKind.NewLine && previousTokenKindValue is HttpTokenKind.NewLine 
+                    && previousCharacter is '\r' && currentCharacter is '\n')
+                {
+
```
shyamnamboodiripad, line 51: It would be nice to clean this up a bit to avoid the empty if block. Consider extracting a local helper (e.g., `IsCurrentTokenANewLinePreceededByCarriageReturn)` so that the resulting combined condition is a bit easier to understand.

src/Microsoft.DotNet.Interactive.HttpRequest.Tests/ParserTests.cs
```diff
@@ -0,0 +1,645 @@
...
+        </request>
+        """);
+
+            using var _ = new AssertionScope();
+
+            var requestNode = result.SyntaxTree.RootNode
+                  .ChildNodes.Should().ContainSingle<HttpRequestNode>().Which;
+
+            requestNode.HeadersNode.HeaderNodes.Count.Should().Be(2);
+            requestNode.BodySeparatorNode.TextWithTrivia.Should().Be("\r\n");
```
jonsequitur, line 264: This assertion should be removed. It's overspecifying the test.

src/Microsoft.DotNet.Interactive.HttpRequest.Tests/ParserTests.cs
```diff
@@ -0,0 +1,645 @@
...
+        [Fact]
+        public void newline_is_legal_at_the_beginning_of_a_request()
+        {
+            var result = HttpRequestParser.Parse(
+                """
+        
+        GET https://example.com
+        """);
+
+            result.SyntaxTree.RootNode
```
jonsequitur, line 97: I would assert that the leading trivia contains newline tokens rather than on the specific text of those tokens, which might vary (and is causing this test to fail on Linux).

src/Microsoft.DotNet.Interactive.HttpRequest.Tests/ParserTests.cs
```diff
@@ -0,0 +1,645 @@
...
+        [Fact]
+        public void newline_is_legal_at_the_end_of_a_request()
+        {
+            var result = HttpRequestParser.Parse(
+                """
+        GET https://example.com
+
+        """);
+
+            result.SyntaxTree.RootNode
```
jonsequitur, line 122: Same comment as above.

src/Microsoft.DotNet.Interactive.HttpRequest.Tests/ParserTests.cs
```diff
@@ -0,0 +1,645 @@
...
+            // TODO: Test error reporting.
+            // Regardless of number of newlines, there should be consistent nodes
+            // set of whitespace, newline, and punctuation
+            //each token kind
+            //combination of characters should be checked by the lexer
+        }
+    }
+    
+    
+    public class Lexer
```
jonsequitur, line 56: We should add a test that a specific set of newlines parses into a consistent number of newline tokens (e.g. `\n\n\r\n\n` -> 4 newline tokens). This will help ensure that that behavior doesn't vary accidentally by OS.

src/Microsoft.DotNet.Interactive.HttpRequest.Tests/ParserTests.cs
shyamnamboodiripad: I think it may be good to have a single test that validates the presence of body separator. 

Same also applies to the header separator - we currently assert the contents of the header separator (i.e., `:`) multiple times in the test that tests headers. It would be nice to have a single test that validates the presence of the header separator instead.

src/Microsoft.DotNet.Interactive.HttpRequest.Tests/ParserTests.cs
```diff
@@ -0,0 +1,645 @@
...
+        {
+            var result = HttpRequestParser.Parse(
+                """
+        GET https://example.com HTTP/1.1
+        Accept: */*
+        Accept-Encoding : gzip, deflate, br
+        Accept-Language : en-US,en;q=0.9
+        ContentLength:7060
+        Cookie: expor=;HSD=Ak_1ZasdqwASDASD;SSID=SASASSDFsdfsdf213123;APISID=WRQWRQWRQWRcc123123;
+        """);
```
shyamnamboodiripad, line 208: minor nit: Indentation seems off for many multi-line string literals in the file now. This probably happened when the code was copy pasted into nested classes...

src/Microsoft.DotNet.Interactive.HttpRequest.Tests/ParserTests.cs
```diff
@@ -0,0 +1,645 @@
...
+            var requestNode = result.SyntaxTree.RootNode
+                  .ChildNodes.Should().ContainSingle<HttpRequestNode>().Which;
+
+            requestNode.BodyNode.Text.Should().Be(
+                """
+        <request>
+            <name>sample</name>
+            <time>Wed, 21 Oct 2015 18:27:50 GMT</time>
+        </request>
+        """);
```
shyamnamboodiripad, line 320: minor nit: Same comment as above about indentation for the multi-line string literals being off...

src/Microsoft.DotNet.Interactive.HttpRequestParser/HttpSyntaxNodeOrToken.cs
```diff
@@ -0,0 +1,51 @@
...
+    {
+        SourceText = sourceText;
+        SyntaxTree = syntaxTree;
+    }
+
+    protected SourceText SourceText { get; }
+
+    public HttpSyntaxNode? Parent { get; internal set; }
+
+    public abstract TextSpan Span { get; }
```
shyamnamboodiripad, line 24: From our discussions yesterday, I think it may be useful to have a `IsMissing` property (that returns `true` if the span has `0` length. I think we'll need it pretty quickly once we start consuming the parser - but we can probably wait and add it later when we first need it...

src/Microsoft.DotNet.Interactive.HttpRequestParser/HttpSyntaxNodeOrToken.cs
```diff
@@ -0,0 +1,51 @@
...
+    public string Text => TextWithTrivia.Trim();
+
+    /// <summary>
+    /// Gets the text of the current node or token, including trivia.
+    /// </summary>
+    public string TextWithTrivia => SourceText.ToString(Span); 
+
+    public override string ToString() => $"{GetType().Name}: {Text}";
+
+    public abstract IEnumerable<Diagnostic> GetDiagnostics();
```
shyamnamboodiripad, line 42: It may be useful to have a `ContainsDiagnostics` convenience API. Roslyn has a similar API. But again, we can wait and add it later when we first need it in the consuming code.

src/Microsoft.DotNet.Interactive.HttpRequestParser/HttpSyntaxTree.cs
```diff
@@ -0,0 +1,38 @@
...
+
+#nullable enable
+
+using Microsoft.CodeAnalysis.Text;
+
+namespace Microsoft.DotNet.Interactive.HttpRequest;
+
+/*
+
+Rough sketch, probably out of date:
```
shyamnamboodiripad, line 12: This is already out of date :) Perhaps now would be a good time to delete it since we have a shared understanding of what the tree looks like.

src/Microsoft.DotNet.Interactive.HttpRequestParser/HttpSyntaxNodeOrToken.cs
jonsequitur: +1 to adding this only when we decide we need it.

shyamnamboodiripad: One other thing I realized is that Roslyn has both `Span` and `FullSpan` (in parallel to `Text` and `FullText`). `Span` excludes (leading and trailing) trivia while `FullSpan` does not.

I think we may also end up needing similar APIs as otherwise it can get cumbersome to walk the tree and figure out what the first and last significant token is each time...

jonsequitur: Makes sense.

src/Microsoft.DotNet.Interactive.HttpRequestParser/HttpRequestParser.cs
```diff
@@ -0,0 +1,411 @@
...
+
+            ParseLeadingTrivia(node);
+
+            if (MoreTokens() && CurrentToken.Kind is HttpTokenKind.Word)
+            {
+                if (CurrentToken.Kind is HttpTokenKind.Word && CurrentToken.Text.ToLower() is ("get" or "post" or "patch" or "put" or "delete" or "head" or "options" or "trace"))
+                {                 
+                    ConsumeCurrentTokenInto(node);
+                }
+                else if (CurrentToken.Kind is HttpTokenKind.Word && CurrentToken.Text.ToLower() is not ("get" or "post" or "patch" or "put" or "delete" or "head" or "options" or "trace" or "https"))
```
shyamnamboodiripad, line 168: Is `https` the only valid URL scheme / protocol prefix? I think we may need to support `http` as well. 

Not sure if REST urls can also have other protocol prefixes (besides http and https). @jonsequitur do you know?

src/Microsoft.DotNet.Interactive.HttpRequestParser/HttpRequestParser.cs
```diff
@@ -0,0 +1,411 @@
...
+
+        private HttpMethodNode? ParseMethod()
+        {
+            var node = new HttpMethodNode(_sourceText, _syntaxTree);
+
+            ParseLeadingTrivia(node);
+
+            if (MoreTokens() && CurrentToken.Kind is HttpTokenKind.Word)
+            {
+                if (CurrentToken.Kind is HttpTokenKind.Word && CurrentToken.Text.ToLower() is ("get" or "post" or "patch" or "put" or "delete" or "head" or "options" or "trace"))
```
shyamnamboodiripad, line 164: We can omit `CurrentToken.Kind is HttpTokenKind.Word` in the 3 inner if conditions since the outer if condition already checks this.

src/Microsoft.DotNet.Interactive.HttpRequestParser/HttpRequestParser.cs
jonsequitur: This has got little to do with REST, really. I think we only need to support `http` and `https` for now.

src/Microsoft.DotNet.Interactive.HttpRequestParser/HttpRequestParser.cs
```diff
@@ -0,0 +1,411 @@
...
+                {        
+                    var tokenSpan = _sourceText.GetSubText(CurrentToken.Span).Lines.GetLinePositionSpan(CurrentToken.Span);                
+
+                    var diagnostic = new Diagnostic(LinePositionSpan.FromCodeAnalysisLinePositionSpan(tokenSpan), DiagnosticSeverity.Warning, CurrentToken.Text.ToLower(), $"Unrecognized HTTP verb {CurrentToken.Text}");
+                    node.AddDiagnostic(diagnostic);
+                    ConsumeCurrentTokenInto(node);
+                }
+                else if (CurrentToken.Kind is HttpTokenKind.Word && CurrentToken.Text.ToLower() is "https")
+                {
+                    return null;
```
shyamnamboodiripad, line 178: `null` returns should always happen in an early out block at the very top of the function (before we even construct the node). Otherwise, we will end up parsing the leading trivia onto the `node` variable which is discarded and never attached to the tree. The resulting tree would then fail to round trip (because it would no longer preserve the complete text in the original string).

```
private HttpMethodNode? ParseMethod()
{
    if (MoreTokens() && CurrentToken.Kind is HttpTokenKind.Word && CurrentToken.Text.ToLower() is "https")
    {
        return null;
    }
    
    var node = new HttpMethodNode(_sourceText, _syntaxTree);

    ParseLeadingTrivia(node);

...
}

File: src/Microsoft.DotNet.Interactive.HttpRequest.Tests/ParserTests.cs [hunk 24]
```diff
@@ -0,0 +1,645 @@
...
+            //combination of characters should be checked by the lexer
+        }
+    }
+    
+    
+    public class Lexer
+    {
+        [Fact]
+
+        public void multiple_whitespaces_are_treated_as_a_single_token()
```
shyamnamboodiripad, line 60: Please add similar tokenization tests for the other token kinds listed in the TODO comment above.

For example, punctuation should always be split into separate tokens, newline chars should always be split into separate tokens (except when it is `\r` followed by `\n` and words should always be combined into a single token.

The tokenization test for newlines would also address Jon's comment above.

src/Microsoft.DotNet.Interactive.HttpRequest.Tests/ParserTests.cs
```diff
@@ -0,0 +1,645 @@
...
+
+using System.Linq;
+using FluentAssertions;
+using FluentAssertions.Execution;
+using Microsoft.DotNet.Interactive.HttpRequest.Tests.Utility;
+using Xunit;
+
+namespace Microsoft.DotNet.Interactive.HttpRequest.Tests;
+
+public class ParserTests
```
shyamnamboodiripad, line 12: (Possibly something to tackle in a follow up PR)

I just realized that we don't have any test that validates roundtripping for syntax trees produced by the parser. Roundtripping is an invariant property of the trees - in other words, every tree generated by the parser should preserve all the text present in the original input string and calling `tree.RootNode.TextWithTrivia` should always produce a string that is identical to the original input string.

It may be useful to validate roundtripping for some of the interesting cases being tested below (especially the ones related to leading and trailing trivia, optional syntax elements etc.) as that could alert us to future bugs in the implementation. However, adding asserts for this in the corresponding tests would violate the one-assertion-per-test guideline.

Perhaps we should add some dedicated roundtripping tests? For example, one such test could parse a string that includes all types of tokens, trivia and nodes in the grammar.

Thoughts?

jonsequitur: Sounds like a very good test. It might be a good candidate for a reusable FluentAssertions extension method.

src/Microsoft.DotNet.Interactive.HttpRequest.Tests/ParserTests.cs
```diff
@@ -91,12 +109,12 @@ public void newline_is_legal_at_the_beginning_of_a_request()
...
         
-        GET https://example.com
-        """);
+                GET https://example.com
+                """);
 
             result.SyntaxTree.RootNode
                   .ChildNodes.Should().ContainSingle<HttpRequestNode>().Which
-                  .MethodNode.ChildTokens.First().TextWithTrivia.Should().Be("\r\n");
+                  .MethodNode.ChildTokens.First().TextWithTrivia.Should().Contain("\r\n");
```
shyamnamboodiripad, line 117: Can we assert that there are 3 tokens with `token.Kind is TokenKind.Newline` instead?

src/Microsoft.DotNet.Interactive.HttpRequest.Tests/ParserTests.cs
```diff
@@ -115,13 +133,13 @@ public void newline_is_legal_at_the_end_of_a_request()
...
-        GET https://example.com
+                GET https://example.com
 
-        """);
+                """);
 
             result.SyntaxTree.RootNode
                   .ChildNodes.Should().ContainSingle<HttpRequestNode>().Which
-                  .UrlNode.ChildTokens.Last().TextWithTrivia.Should().Be("\r\n");
+                  .UrlNode.ChildTokens.Last().TextWithTrivia.Should().Contain("\r\n");
```
shyamnamboodiripad, line 142: Same here - it would be betterr to assert that there are 3 tokens with token.Kind is TokenKind.Newline instead. That would eliminate the need to check the exact text here - exact text for newline tokens is already validated by the lexer test for newlines above.

src/Microsoft.DotNet.Interactive.HttpRequest.Tests/ParserTests.cs
```diff
@@ -71,8 +71,26 @@ public void multiple_whitespaces_are_treated_as_a_single_token()
...
+                  .MethodNode.ChildTokens.Count().Should().Be(4);
+        }
+
+        [Fact]
+        public void multiple_punctuations_are_parsed_into_different_tokens()
+        {
+            var result = HttpRequestParser.Parse(".!?.");
+            result.SyntaxTree.RootNode
+                  .ChildNodes.Should().ContainSingle<HttpRequestNode>().Which
+                  .UrlNode.ChildTokens.Count().Should().Be(4);
```
shyamnamboodiripad, line 90: This test should ideally check that there are 4 tokens and that their `token.Kind is TokenKind.Punctuation`.

src/Microsoft.DotNet.Interactive.HttpRequest.Tests/ParserTests.cs
shyamnamboodiripad: Looks like the tests for whitespace, newline and punctuation are now in place. It would be great to add one for "words should always be combined into a single token" so that we round out all the kinds of tokens that the lexer produces.

src/Microsoft.DotNet.Interactive.HttpRequest.Tests/ParserTests.cs
```diff
@@ -0,0 +1,595 @@
...
+    }
+
+
+    public class Lexer
+    {
+        [Fact]
+
+        public void multiple_whitespaces_are_treated_as_a_single_token()
+        {
+            var result = HttpRequestParser.Parse("     ");
```
shyamnamboodiripad, line 62: Could you please also include a tab character in the middle? The tab character should also be lexed into to the same whitespace token.

src/Microsoft.DotNet.Interactive.HttpRequest.Tests/ParserTests.cs
```diff
@@ -0,0 +1,595 @@
...
+
+            using var _ = new AssertionScope();
+
+            result.SyntaxTree.RootNode
+            .ChildNodes.Should().ContainSingle<HttpRequestNode>().Which
+            .MethodNode.ChildTokens.First().Should().BeOfType<HttpSyntaxToken>();
+
+            result.SyntaxTree.RootNode
+            .ChildNodes.Should().ContainSingle<HttpRequestNode>().Which
+            .MethodNode.ChildTokens.First().TextWithTrivia.Should().Be("     ");
```
shyamnamboodiripad, line 72:
```suggestion
            .MethodNode.ChildTokens.Single().TextWithTrivia.Should().Be("     ");
```

src/Microsoft.DotNet.Interactive.HttpRequest.Tests/ParserTests.cs
shyamnamboodiripad:
```suggestion
            var result = HttpRequestParser.Parse("  \t   ");
```

src/Microsoft.DotNet.Interactive.HttpRequest.Tests/ParserTests.cs
```diff
@@ -0,0 +1,595 @@
...
+                """
+                <request>
+                    <name>sample</name>
+                    <time>Wed, 21 Oct 2015 18:27:50 GMT</time>
+                </request>
+                """);
+        }
+
+        [Fact]
+        public void header_separator_is_present()
```
shyamnamboodiripad, line 292: Move this test to the Header nested class above.

src/Microsoft.DotNet.Interactive.HttpRequest.Tests/ParserTests.cs
```diff
@@ -0,0 +1,595 @@
...
+            headerNodes[2].NameNode.Text.Should().Be("Accept-Language");
+            headerNodes[2].SeparatorNode.Text.Should().Be(":");
+            headerNodes[2].ValueNode.Text.Should().Be("en-US,en;q=0.9");
+
+            headerNodes[3].NameNode.Text.Should().Be("ContentLength");
+            headerNodes[3].SeparatorNode.Text.Should().Be(":");
+            headerNodes[3].ValueNode.Text.Should().Be("7060");
+
+            headerNodes[4].NameNode.Text.Should().Be("Cookie");
+            headerNodes[4].SeparatorNode.Text.Should().Be(":");
```
shyamnamboodiripad, line 254: Now that we have a `header_separator_is_present` test, we can remove the 4 assertions that validate the separator node's text from this test.

src/Microsoft.DotNet.Interactive.HttpRequest.Tests/ParserTests.cs
```diff
@@ -0,0 +1,595 @@
...
+            """);         
+
+            var requestNode = result.SyntaxTree.RootNode
+                    .ChildNodes.Should().ContainSingle<HttpRequestNode>().Which;
+
+            requestNode.BodySeparatorNode.TextWithTrivia.Should().Be("\r\n");          
+        }
+
+        [Fact]
+        public void body_without_header_header_is_parsed_correctly()
```
shyamnamboodiripad, line 329:
```suggestion
        public void when_headers_are_not_present_there_should_be_no_header_nodes
```

src/Microsoft.DotNet.Interactive.HttpRequest.Tests/ParserTests.cs
```diff
@@ -0,0 +1,595 @@
...
+
+            using var _ = new AssertionScope();
+
+            var requestNode = result.SyntaxTree.RootNode
+                  .ChildNodes.Should().ContainSingle<HttpRequestNode>().Which;
+
+            requestNode.HeadersNode.HeaderNodes.Count.Should().Be(0);
+        }
+
+        //TODO: Is it an error to include a body with no headers?
```
shyamnamboodiripad, line 349: Let's remove this TODO - I think this will bubble naturally if it is required when we implement all the error reporting.

src/Microsoft.DotNet.Interactive.HttpRequest.Tests/ParserTests.cs
```diff
@@ -0,0 +1,595 @@
...
+        [Fact]
+        public void body_without_header_body_is_parsed_correctly()
```
shyamnamboodiripad, line 351:
```suggestion
        public void body_is_parsed_correctly_when_headers_are_not_present()
```

src/Microsoft.DotNet.Interactive.HttpRequest.Tests/ParserTests.cs
```diff
@@ -0,0 +1,595 @@
...
+            <request>
+                <name>sample</name>
+                <time>Wed, 21 Oct 2015 18:27:50 GMT</time>
+            </request>
+            """);         
+
+            var requestNode = result.SyntaxTree.RootNode
+                    .ChildNodes.Should().ContainSingle<HttpRequestNode>().Which;
+
+            requestNode.BodySeparatorNode.TextWithTrivia.Should().Be("\r\n");
```
shyamnamboodiripad, line 325: Again, I think we should validate the token kind as opposed to the actual text for newlines. This would also address the test failure on Linux.

src/Microsoft.DotNet.Interactive.HttpRequest.Tests/ParserTests.cs
```diff
@@ -0,0 +1,595 @@
...
+
+    [Fact]
+    public async Task CommentInBetweenHeadersAsync()
+    {
+        string text = """
+            POST https://example.com
+            Content-Type:application/json
+            #comment
+            Accept: gzip
+            """;
```
shyamnamboodiripad, line 551: This test should work now, should it not? The comment will be added as leading trivia for the header.

src/Microsoft.DotNet.Interactive.HttpRequest.Tests/ParserTests.cs
```diff
@@ -0,0 +1,595 @@
...
+        Assert.AreEqual(ItemType.HeaderName, first.Children[2].Type);
+        Assert.AreEqual(ItemType.HeaderValue, first.Children[3].Type);
+        Assert.AreEqual(ItemType.Comment, first.Children[4].Type);
+        Assert.AreEqual(ItemType.EmptyLine, first.Children[5].Type);
+        Assert.AreEqual(ItemType.Body, first.Children[6].Type);
+        Assert.AreEqual("{\r\n    \"enabled\": true\r\n}\r\n", first.Body);
+    }
+
+    [Fact]
+    public async Task BodyAfterCommentAsync()
```
shyamnamboodiripad, line 502: This test should work now, should it not? The comment should be added as leading trivia for the body.

src/Microsoft.DotNet.Interactive.HttpRequest.Tests/ParserTests.cs
jonsequitur: We have a few FluentAssertions tricks for this. Easier to pair on it.

src/Microsoft.DotNet.Interactive.HttpRequestParser/HttpLexer.cs
```diff
@@ -0,0 +1,103 @@
...
+#nullable enable
+
+using System;
+using System.Collections.Generic;
+using Microsoft.CodeAnalysis.Text;
+
+namespace Microsoft.DotNet.Interactive.HttpRequest;
+
+internal class HttpLexer
```
shyamnamboodiripad, line 12: Per our discussion on Tuesday, we decided to make the Lexer a private class inside the Parser (and always test the tokenization via the parser's public API).

src/Microsoft.DotNet.Interactive.HttpRequestParser/HttpLexer.cs
```diff
@@ -0,0 +1,103 @@
...
+        }
+
+        return _textWindow.End < _sourceText.Length;
+    }
+
+    private bool IsCurrentTokenANewLinePrecededByACarriageReturn(HttpTokenKind previousTokenKindValue, char previousCharacter, HttpTokenKind currentTokenKind, char currentCharacter)
+    {
+        return (currentTokenKind is HttpTokenKind.NewLine && previousTokenKindValue is HttpTokenKind.NewLine 
+            && previousCharacter is '\r' && currentCharacter is '\n');
+        {
```
shyamnamboodiripad, line 99: Empty block can be removed.

src/Microsoft.DotNet.Interactive.HttpRequestParser/HttpRequestParser.cs
```diff
@@ -0,0 +1,412 @@
...
+            {
+                if (CurrentToken.Text.ToLower() is ("get" or "post" or "patch" or "put" or "delete" or "head" or "options" or "trace"))
+                {                 
+                    ConsumeCurrentTokenInto(node);
+                }
+                else
+                {        
+                    var tokenSpan = _sourceText.GetSubText(CurrentToken.Span).Lines.GetLinePositionSpan(CurrentToken.Span);                
+
+                    var diagnostic = new Diagnostic(LinePositionSpan.FromCodeAnalysisLinePositionSpan(tokenSpan), DiagnosticSeverity.Warning, CurrentToken.Text.ToLower(), $"Unrecognized HTTP verb {CurrentToken.Text}");
```
shyamnamboodiripad, line 177: minor: Is the ToLower() call needed?

src/Microsoft.DotNet.Interactive.HttpRequest.Tests/ParserTests.cs
```diff
@@ -0,0 +1,595 @@
...
+        [Fact]
+        public void it_can_parse_a_string_with_only_newlines()
+        {
+            var result = HttpRequestParser.Parse("\r\n\n\r\n");
+
+            result.SyntaxTree.RootNode
+                  .ChildNodes.Should().ContainSingle<HttpRequestNode>().Which
+                  .MethodNode.TextWithTrivia.Should().Be("\r\n\n\r\n");
+
+            // TODO: Test error reporting.
```
shyamnamboodiripad, line 47: I think this TODO can now be deleted.

src/Microsoft.DotNet.Interactive.HttpRequest.Tests/ParserTests.cs
shyamnamboodiripad: Resolving since we have now added this (in the new Parser() method).

src/Microsoft.DotNet.Interactive.HttpRequestParser/HttpLexer.cs
shyamnamboodiripad: Resolving since this was done offline.

src/Microsoft.DotNet.Interactive.HttpRequest.Tests/ParserTests.cs
shyamnamboodiripad: The `Single()` part was an incorrect suggestion - resolving.

src/Microsoft.DotNet.Interactive.HttpRequestParser/HttpRequestParser.cs
bleaphar: I think it is because then GET wouldn't be recognized correct?

shyamnamboodiripad: ToLower makes sense in the conditions in the if block above. I am not sure it is needed when we construct the diagnostic though, is it?

shyamnamboodiripad (approved):
LGTM modulo comments 👍🏾

src/Microsoft.DotNet.Interactive.HttpRequest.Tests/ParserTests.cs
shyamnamboodiripad: Looks like this is now fixed.

src/Microsoft.DotNet.Interactive.HttpRequest.Tests/ParserTests.cs
shyamnamboodiripad: Also fixed.

src/Microsoft.DotNet.Interactive.HttpRequest.Tests/ParserTests.cs
shyamnamboodiripad: Resolving since this is fixed.

bleaphar:
Merging

src/Microsoft.DotNet.Interactive.HttpRequestParser/HttpSyntaxNodeOrToken.cs
shyamnamboodiripad: This was addressed in #3122

PR #3104, opened 2023-07-25
Remove buildalyzer

ocallesp:


ocallesp:
Abandon this PR because I took a different approach in https://github.com/dotnet/interactive/pull/3112

PR #3106, opened 2023-07-26
use proposed api for autoclosing

colombod:
adopt new proposed api for autoclosing pairs


jonsequitur (approved):

colombod:
@rebornix is this enough?

PR #3108, opened 2023-07-26
remove IKernelConnector

jonsequitur:


colombod (approved):

src/Microsoft.DotNet.Interactive.SQLite/ConnectSQLiteCommand.cs
```diff
@@ -25,9 +25,9 @@ public ConnectSQLiteCommand()
         InvocationContext commandLineContext)
     {
         var connectionString = commandLineContext.ParseResult.GetValueForArgument(ConnectionStringArgument);
-        var connector = new SQLiteKernelConnector(connectionString);
         var localName = commandLineContext.ParseResult.GetValueForOption(KernelNameOption);
-        var kernel = await connector.CreateKernelAsync(localName);
+        var kernel1 = new SQLiteKernel($"sql-{localName}", connectionString);
+        var kernel = await Task.FromResult<Kernel>(kernel1);
         return new[] { kernel };
```
shyamnamboodiripad, line 31: The await is a bit confusing. Why not remove the `async` keyword and say something like `return Task.FromResult(new [] { new SQLiteKernel($"sql-{localName}", connectionString) });`?

src/Microsoft.DotNet.Interactive.Tests/NamedPipeConnectionTests.cs
```diff
@@ -23,15 +23,15 @@ public NamedPipeConnectionTests(ITestOutputHelper output) : base(output)
...
+    protected override Func<string, Task<ProxyKernel>> CreateConnector()
     {
         CreateRemoteKernelTopology(_pipeName);
 
         var connector = new NamedPipeKernelConnector(_pipeName);
 
         RegisterForDisposal(connector);
 
-        return connector;
+        return name => connector.CreateKernelAsync(name);
```
shyamnamboodiripad, line 34: Can this be simplified to `return connector.CreateKernelAsync`?

src/Microsoft.DotNet.Interactive.Tests/ProxyKernelConnectionTestsBase.cs
```diff
@@ -133,7 +133,7 @@ public async Task fast_path_commands_over_proxy_can_be_handled()
     {
         var connector = CreateConnector();
```
shyamnamboodiripad, line 134: minor nit: Also rename this variable from `connector` to `createKernel` to match the above test?

src/dotnet-interactive/Connection/StdIoKernelConnector.cs
```diff
@@ -60,17 +60,6 @@ public class StdIoKernelConnector : IKernelConnector
         _remoteKernelInfoCache = new ConcurrentDictionary<string, KernelInfo>();
     }
 
-    /// <remarks>
-    /// TODO: Does it even make sense to implement <see cref="IKernelConnector"/> here considering that we have
-    /// concepts (such as a single root <see cref="ProxyKernel"/>, and zero or more (child) <see cref="ProxyKernel"/>s)
-    /// that the <see cref="IKernelConnector"/> abstraction does not understand / support? Should the '#!connect stdio'
```
shyamnamboodiripad, line 66: Just curious - any changes in our thoughts around the `#!connect stdio`? Can we remove that too?

shyamnamboodiripad (approved):

src/Microsoft.DotNet.Interactive.SQLite/ConnectSQLiteCommand.cs
jonsequitur: That was a weird artifact of the inlining refactoring. Fixed.

src/dotnet-interactive/Connection/StdIoKernelConnector.cs
jonsequitur: I think we might still want this, especially as project integration comes along and we think about .NET Interactive instances for other TFMs, different dependency graphs, etc.

PR #3112, opened 2023-07-31
Replace Buildalyzer from CSharpProject

ocallesp:
This Pull Request seeks to eliminate all dependencies on Buildalyzer, due to its frequent breakdown during runtime. The instability can be attributed to frequent changes in the binlog format, necessitating either a regular update of Buildalyzer to the latest version supporting the new format, or the submission of respective updates to the Buildalyzer project repository.

This update involves numerous changes, the crux of which is a revision to the build process. During the build, a temporary file named Directory.Build.Targets is generated. This file initiates the creation of a target called DirectoryBuildTargetsContent which, in turn, generates a cache file containing all the necessary data to create a Roslyn Workspace object.

Subsequent to this update, instead of analyzing a binlog, the system reads from the cache file to create a Workspace, providing a more stable and efficient solution.


ocallesp:
@colombod 

These are remaining unit-tests that are failing

image

Some of them seems not related to my changes

src/Microsoft.DotNet.Interactive.CSharpProject/BuildCacheFileUtilities.cs
```diff
@@ -0,0 +1,146 @@
...
+using System.Threading.Tasks;
+
+namespace Microsoft.DotNet.Interactive.CSharpProject;
+
+public static class BuildCacheFileUtilities
+{
+    internal static string cacheFilenameSuffix = ".interactive.workspaceData.cache";
+    internal static string DirectoryBuildTargetFilename = "Directory.Build.Targets";
+
+    internal static string DirectoryBuildTargetsContent = $@"
```
jonsequitur, line 16: A raw string rather than a verbatim string would be nice here.

src/Microsoft.DotNet.Interactive.CSharpProject/BuildCacheFileUtilities.cs
```diff
@@ -0,0 +1,146 @@
...
+        public string PropertyOutputType { get; init; }
+        // TODO : Modify the target to get this value
+        public string PropertyDefineConstants { get; init; }
+        public string[] CompileInputs { get; init; }
+    }
+
+    internal static async Task<string> BuildAndCreateCacheFileAsync(string csprojFilePath)
+    {
+        DirectoryInfo directoryInfo = new DirectoryInfo(Path.GetDirectoryName(csprojFilePath));
+        FileInfo lastBuildErrorLogFile = new FileInfo(Path.Combine(directoryInfo.FullName, ".trydotnet-builderror"));
```
jonsequitur, line 101:
```suggestion
        FileInfo lastBuildErrorLogFile = new FileInfo(Path.Combine(directoryInfo.FullName, ".net-interactive-builderror"));
```

We should generally remove the Try .NET branding from the code because it has wider uses than just the Try .NET experience.

src/Microsoft.DotNet.Interactive.CSharpProject/BuildCacheFileUtilities.cs
```diff
@@ -0,0 +1,146 @@
...
+
+        CleanObjFolder(directoryInfo);
+
+        string tempDirectoryBuildTarget =
+            Path.Combine(Path.GetDirectoryName(csprojFilePath), BuildCacheFileUtilities.DirectoryBuildTargetFilename);
+        File.WriteAllText(tempDirectoryBuildTarget, BuildCacheFileUtilities.DirectoryBuildTargetsContent);
+
+        var args = "";
+        if (Path.Exists(csprojFilePath) == true)
```
jonsequitur, line 110:
```suggestion
        if (Path.Exists(csprojFilePath))
```

src/Microsoft.DotNet.Interactive.CSharpProject/BuildCacheFileUtilities.cs
```diff
@@ -0,0 +1,146 @@
...
+        public Guid ProjectGuid { get; init; }
+        public IEnumerable<string> ProjectReferences { get; init; }
+        public string ProjectFilePath { get; init; }
+        public string LanguageName { get; init; }
+        public string PropertyTargetPath { get; init; }
+        public string[] SourceFiles { get; init; }
+        public string[] References { get; init; }
+        public string[] AnalyzerReferences { get; init; }
+        public string[] PreprocessorSymbols { get; init; }
+        public string PropertyLangVersion { get; init; }
```
jonsequitur, line 91:
```suggestion
        public string LangVersion { get; init; }
```

src/Microsoft.DotNet.Interactive.CSharpProject/BuildCacheFileUtilities.cs
```diff
@@ -0,0 +1,146 @@
...
+        public string PropertyOutputType { get; init; }
```
jonsequitur, line 92:
```suggestion
        public string OutputType { get; init; }
```

src/Microsoft.DotNet.Interactive.CSharpProject/BuildCacheFileUtilities.cs
```diff
@@ -0,0 +1,146 @@
...
+        // TODO : Modify the target to get this value
+        public string PropertyDefineConstants { get; init; }
```
jonsequitur, line 94:
```suggestion
        public string DefineConstants { get; init; }
```

src/Microsoft.DotNet.Interactive.CSharpProject/BuildCacheFileUtilities.cs
```diff
@@ -0,0 +1,146 @@
...
+                              File=""$(MSBuildProjectFullPath).interactive.workspaceData.cache""
+                              Overwrite=""True""
+                              WriteOnlyWhenDifferent=""True"" />
+          </Target>
+        </Project>";
+
+    public record BuildProjectData
+    {
+        public Guid ProjectGuid { get; init; }
+        public IEnumerable<string> ProjectReferences { get; init; }
```
jonsequitur, line 83: This should be a fixed-size type, e.g. `IReadOnlyList<string>`

src/Microsoft.DotNet.Interactive.CSharpProject/RoslynWorkspaceUtilities.cs
```diff
@@ -0,0 +1,389 @@
...
+        public IReadOnlyDictionary<string, string> Metadata { get; }
+
+        internal ProjectItem()
+        {
+//            ItemSpec = taskItem.ItemSpec;
+//            Metadata = taskItem.MetadataNames.Cast<string>().ToDictionary(x => x, x => taskItem.GetMetadata(x));
+        }
+    }
+
+    public class BuildDataResults
```
jonsequitur, line 30: These probably shouldn't be nested classes and by convention they should each be in separate files.

src/Microsoft.DotNet.Interactive.CSharpProject/RoslynWorkspaceUtilities.cs
```diff
@@ -0,0 +1,389 @@
...
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Text;
+using static Microsoft.DotNet.Interactive.CSharpProject.BuildCacheFileUtilities;
+
+namespace Microsoft.DotNet.Interactive.CSharpProject;
+
+public static class RoslynWorkspaceUtilities
```
jonsequitur, line 16: Can this be made internal?

src/Microsoft.DotNet.Interactive.CSharpProject/RoslynWorkspaceUtilities.cs
```diff
@@ -0,0 +1,389 @@
...
+
+    public class BuildDataResults
+    {
+        public BuildProjectData Data { get; set; }
+
+        public Microsoft.CodeAnalysis.Workspace Workspace { get; set; }
+
+        public bool Succeeded { get; set; }
+
+        public string cacheFile { get; set; }
```
jonsequitur, line 38:
```suggestion
        public string CacheFile { get; set; }
```

jonsequitur: This property name is a little unclear. Is intended to be this a file path? Fully-qualified or not? Consider using `FileInfo` if so.

src/Microsoft.DotNet.Interactive.CSharpProject/RoslynWorkspaceUtilities.cs
```diff
@@ -0,0 +1,389 @@
...
+        internal ProjectItem()
+        {
+//            ItemSpec = taskItem.ItemSpec;
+//            Metadata = taskItem.MetadataNames.Cast<string>().ToDictionary(x => x, x => taskItem.GetMetadata(x));
+        }
+    }
+
+    public class BuildDataResults
+    {
+        public BuildProjectData Data { get; set; }
```
jonsequitur, line 32: I think it's a little clearer if the property name matches the type name.

src/Microsoft.DotNet.Interactive.CSharpProject/RoslynWorkspaceUtilities.cs
```diff
@@ -0,0 +1,389 @@
...
+        }
+
+        var cacheFilePath = csprojFilePath + BuildCacheFileUtilities.cacheFilenameSuffix;
+
+        return GetResultsFromCacheFile(cacheFilePath);
+    }
+
+    internal static void PopulateBuildProjectData(string fileContent, out BuildProjectData buildProjectData)
+    {
+        buildProjectData = new BuildProjectData
```
jonsequitur, line 98: Typically it's preferable for these property initializations to be the defaults for the class itself.

src/Microsoft.DotNet.Interactive.CSharpProject/BuildCacheFileUtilities.cs
```diff
@@ -0,0 +1,146 @@
...
+        
+            <!-- Write collected project data to a file -->
+            <WriteLinesToFile Lines=""@(ProjectDataLines)""
+                              File=""$(MSBuildProjectFullPath).interactive.workspaceData.cache""
+                              Overwrite=""True""
+                              WriteOnlyWhenDifferent=""True"" />
+          </Target>
+        </Project>";
+
+    public record BuildProjectData
```
jonsequitur, line 80: Why `record`?

src/Microsoft.DotNet.Interactive.CSharpProject/RoslynWorkspaceUtilities.cs
```diff
@@ -0,0 +1,389 @@
...
+            case ".csproj":
+                languageName = LanguageNames.CSharp;
+                return true;
+            default:
+                languageName = null;
+                return false;
+        }
+    }
+
+    private static IEnumerable<DocumentInfo> GetDocuments(BuildProjectData buildProjectData, ProjectId projectId) =>
```
jonsequitur, line 291: I would return `IReadOnlyList<DocumentInfo>` here. 

The current implementation will redo the file reads on multiple enumerations including trying to get a count, so this change will then force the use of an `ToArray` or `ToList` call, which will prevent that.

src/Microsoft.DotNet.Interactive.CSharpProject/RoslynWorkspaceUtilities.cs
```diff
@@ -0,0 +1,389 @@
...
+    .Select(x => DocumentInfo.Create(
+                DocumentId.CreateNewId(projectId),
+    Path.GetFileName(x),
+                loader: TextLoader.From(
+                    TextAndVersion.Create(
+                        SourceText.From(File.ReadAllText(x), Encoding.Unicode), VersionStamp.Create())),
+                filePath: x))
+        ?? Array.Empty<DocumentInfo>();
+
+    private static IEnumerable<ProjectReference> GetExistingProjectReferences(BuildProjectData buildProjectData, Microsoft.CodeAnalysis.Workspace workspace) =>
```
jonsequitur, line 303: Ditto.

src/Microsoft.DotNet.Interactive.CSharpProject/RoslynWorkspaceUtilities.cs
```diff
@@ -0,0 +1,389 @@
...
+        ?? Array.Empty<ProjectReference>();
+
+    private static IEnumerable<MetadataReference> GetMetadataReferences(BuildProjectData buildProjectData) =>
+        buildProjectData
+            .References?.Where(File.Exists)
+            .Select(x => MetadataReference.CreateFromFile(x))
+        ?? (IEnumerable<MetadataReference>)Array.Empty<MetadataReference>();
+
+
+    private static IEnumerable<AnalyzerReference> GetAnalyzerReferences(BuildProjectData buildProjectData, Microsoft.CodeAnalysis.Workspace workspace)
```
jonsequitur, line 318: Ditto,

src/Microsoft.DotNet.Interactive.CSharpProject/BuildCacheFileUtilities.cs
ocallesp: done.

Replaced  ".trydotnet-builderror" with ".net-interactive-builderror" in all the codebase

src/Microsoft.DotNet.Interactive.CSharpProject/BuildCacheFileUtilities.cs
ocallesp: Replacing '==' with 'is' for null check in the codebase for this project

src/Microsoft.DotNet.Interactive.CSharpProject/BuildCacheFileUtilities.cs
ocallesp: Done.

Renamed

src/Microsoft.DotNet.Interactive.CSharpProject/BuildCacheFileUtilities.cs
ocallesp: Done.

Renamed

src/Microsoft.DotNet.Interactive.CSharpProject/BuildCacheFileUtilities.cs
ocallesp: Done.

Renamed

src/Microsoft.DotNet.Interactive.CSharpProject/RoslynWorkspaceUtilities.cs
ocallesp: Done.

Renamed 'Data' with 'BuildProjectData'

src/Microsoft.DotNet.Interactive.CSharpProject/RoslynWorkspaceUtilities.cs
```diff
@@ -0,0 +1,389 @@
...
+            ProjectReferences = projectReferencesList.ToArray(),
+            AnalyzerReferences = analyzerReferencesList.ToArray(),
+            PreprocessorSymbols = preprocessorSymbolsList.ToArray(),
+            SourceFiles = sourceFilesList.ToArray()
+        };
+    }
+
+    public static AdhocWorkspace GetWorkspace(BuildProjectData buildProjectData, bool addProjectReferences = false)
+    {
+        if (buildProjectData == null)
```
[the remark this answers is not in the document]
ocallesp: done

src/Microsoft.DotNet.Interactive.CSharpProject/RoslynWorkspaceUtilities.cs
ocallesp: Done.

Now this method returns IReadOnlyList

src/Microsoft.DotNet.Interactive.CSharpProject/RoslynWorkspaceUtilities.cs
ocallesp: Done.

Now this method returns IReadOnlyList

src/Microsoft.DotNet.Interactive.CSharpProject/RoslynWorkspaceUtilities.cs
ocallesp: Done.

Now this method returns IReadOnlyList

src/Microsoft.DotNet.Interactive.CSharpProject/BuildCacheFileUtilities.cs
ocallesp: changed

src/Microsoft.DotNet.Interactive.CSharpProject/RoslynWorkspaceUtilities.cs
ocallesp: Renamed to CacheFilePath since this contains the full path to the file.

I will change this to FileInfo if needed.

jonsequitur: `FileInfo` is better because it's more specific and safer. `string` won't enforce anything about the content, and also brings the `null` versus `string.Empty` confusion.

src/Microsoft.DotNet.Interactive.CSharpProject/RoslynWorkspaceUtilities.cs
ocallesp: Done.

Did a refactor to move this class to its own file

src/Microsoft.DotNet.Interactive.CSharpProject/RoslynWorkspaceUtilities.cs
ocallesp: Done

src/Microsoft.DotNet.Interactive.CSharpProject/BuildCacheFileUtilities.cs
ocallesp: changed to class.

I was using record because this is read-only data

ocallesp:
@colombod Fixed the issue with Document Ids in the last commit.

src/Microsoft.DotNet.Interactive.CSharpProject/RoslynWorkspaceUtilities.cs
ocallesp: done in the last commit

src/Microsoft.DotNet.Interactive.CSharpProject/BuildCacheFileUtilities.cs
ocallesp: done. In the latest commit

src/Microsoft.DotNet.Interactive.CSharpProject/Packaging/AnalyzerResultExtensions.cs
```diff
@@ -3,35 +3,35 @@
...
 using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.DotNet.Interactive.CSharpProject.RoslynWorkspaceUtilities;
 using Microsoft.DotNet.Interactive.CSharpProject.Servers.Roslyn;
+using static Microsoft.DotNet.Interactive.CSharpProject.RoslynWorkspaceUtilities.RoslynWorkspaceUtilities;
 
 //adpated from https://github.com/daveaglick/Buildalyzer/blob/master/src/Buildalyzer.Workspaces/AnalyzerResultExtensions.cs
 
 namespace Microsoft.DotNet.Interactive.CSharpProject.Packaging;
 
 public static class AnalyzerResultExtensions
```
jonsequitur, line 17: Do all of these need to be public?

src/Microsoft.DotNet.Interactive.CSharpProject/Packaging/PackageBase.cs
```diff
@@ -89,13 +90,17 @@ protected async Task DotnetBuildAsync()
         {
             this.CleanObjFolder();
             var projectFile = this.GetProjectFile();
-            var args = $"/bl:{FullBuildBinlogFileName}";
+
+            string tempDirectoryBuildTarget =
```
jonsequitur, line 94: It's not clear from the name (and because the type is `string`) that this intended to be a file path, not a directory path.

src/Microsoft.DotNet.Interactive.CSharpProject/Packaging/PackageBase.cs
```diff
@@ -109,6 +114,9 @@ protected async Task DotnetBuildAsync()
                 LastBuildErrorLogFile.Delete();
             }
 
+            // Clean up the temp project file
+            File.Delete(tempDirectoryBuildTarget);
```
jonsequitur, line 118: What is the purpose of the temp file in this method? It looks like we create it and delete it but never use it for anything.

ocallesp: This is the Directory.Build.Targets that contains the target to generate the Cache.

We only need this file during the build.

src/Microsoft.DotNet.Interactive.CSharpProject/RoslynWorkspaceUtilities/BuildDataResults.cs
```diff
@@ -0,0 +1,27 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using static Microsoft.DotNet.Interactive.CSharpProject.RoslynWorkspaceUtilities.RoslynWorkspaceUtilities;
+
+namespace Microsoft.DotNet.Interactive.CSharpProject.RoslynWorkspaceUtilities;
+
+public class BuildDataResults
```
jonsequitur, line 11: Does this need to be public?

For every type introduced, we should make it internal if it's simple to do so. For the ones where it's less clear (for example you're using them in unit tests), let's have a discussion when we meet.

src/Microsoft.DotNet.Interactive.CSharpProject/RoslynWorkspaceUtilities/RoslynWorkspaceUtilities.cs
```diff
@@ -0,0 +1,345 @@
...
+                        SourceText.From(File.ReadAllText(x), Encoding.Unicode), VersionStamp.Create())),
+                    filePath: x))
+            ?? Array.Empty<DocumentInfo>()).ToList();
+
+    private static IReadOnlyList<ProjectReference> GetExistingProjectReferences(BuildProjectData buildProjectData, CodeAnalysis.Workspace workspace) =>
+        (buildProjectData.ProjectReferences
+            .Select(x => workspace.CurrentSolution.Projects.FirstOrDefault(y => y.FilePath.Equals(x, StringComparison.OrdinalIgnoreCase)))
+            .Where(x => x != null)
+            .Select(x => new ProjectReference(x.Id))
+        ?? Array.Empty<ProjectReference>()).ToList();
```
jonsequitur, line 264: Calling `ToList` on the return value of `Array.Empty` is redundant. You only need to do that on the result of the expression before the `??`.

src/Microsoft.DotNet.Interactive.CSharpProject/RoslynWorkspaceUtilities/RoslynWorkspaceUtilities.cs
```diff
@@ -0,0 +1,345 @@
...
+
+    private static IEnumerable<MetadataReference> GetMetadataReferences(BuildProjectData buildProjectData) =>
```
jonsequitur, line 266: It's more appropriate here to return `IReadOnlyList<MetadataReference>`, for the same reasons as the others. Otherwise, repeat evaluations will redo expensive work.

src/Microsoft.DotNet.Interactive.CSharpProject/RoslynWorkspaceUtilities/RoslynWorkspaceUtilities.cs
```diff
@@ -0,0 +1,345 @@
...
+        }
+
+        return null;
+    }
+
+    private static CSharpParseOptions CreateCSharpParseOptions(BuildProjectData buildProjectData)
+    {
+        CSharpParseOptions parseOptions = new CSharpParseOptions();
+
+        // Add any constants
```
jonsequitur, line 299: Feel free to remove comments like this. They don't add useful information.

src/Microsoft.DotNet.Interactive.CSharpProject/Packaging/ProjectAsset.cs
```diff
@@ -84,35 +84,33 @@ private async Task<IAnalyzerResult> LoadResultOrCleanAsync()
...
-        }
-        throw new InvalidOperationException("Failed creating workspace");
+        var ws = result.Workspace;
+        var projectId = ws.CurrentSolution.ProjectIds.FirstOrDefault();
+        var references = result.BuildProjectData.References;
+        var metadataReferences = references.GetMetadataReferences();
+        var solution = ws.CurrentSolution;
+        solution = solution.WithProjectMetadataReferences(projectId, metadataReferences);
+        ws.TryApplyChanges(solution);
+        return Task.FromResult(ws);
```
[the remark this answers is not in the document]
ocallesp: It appears that this method was deliberately designed to return a Task because it is passed as an argument to a method that takes a Func<T, Task<U>> nextStep

We want to avoid using Task, then we will have to create another method Then<U>(Func<T, U> nextStep)

We can do this change in another commit.

src/Microsoft.DotNet.Interactive.CSharpProject/Packaging/PackageBase.cs
ocallesp: changed to directoryBuildTargetsFile

src/Microsoft.DotNet.Interactive.CSharpProject/RoslynWorkspaceUtilities/BuildCacheFileUtilities.cs
```diff
@@ -0,0 +1,128 @@
...
+using System.Threading.Tasks;
+
+namespace Microsoft.DotNet.Interactive.CSharpProject.RoslynWorkspaceUtilities;
+
+internal static class BuildCacheFileUtilities
+{
+    internal static string cacheFilenameSuffix = ".interactive.workspaceData.cache";
+    internal static string DirectoryBuildTargetFilename = "Directory.Build.Targets";
+
+    internal static string DirectoryBuildTargetsContent =
```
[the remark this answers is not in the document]
ocallesp: changed to use raw string

src/Microsoft.DotNet.Interactive.CSharpProject/Packaging/PackageBase.cs
jonsequitur: Ah, I see. The build reads the file. 👍

Maybe rename it to something that says exactly what it is, e.g. `tempTargetFilePath`

src/Microsoft.DotNet.Interactive.CSharpProject/RoslynWorkspaceUtilities/RoslynWorkspaceUtilities.cs
ocallesp: removed most of the comments since the look redundant.

src/Microsoft.DotNet.Interactive.CSharpProject/RoslynWorkspaceUtilities/RoslynWorkspaceUtilities.cs
ocallesp: changed

src/Microsoft.DotNet.Interactive.CSharpProject/Packaging/AnalyzerResultExtensions.cs
ocallesp: If this is only used internally, then we can delete the complete file. There are zero references to it.

ocallesp: removed

src/Microsoft.DotNet.Interactive.CSharpProject/Packaging/ProjectAsset.cs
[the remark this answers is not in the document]
ocallesp: Done. This method is now sync and a new implementation of Then was added

src/Microsoft.DotNet.Interactive.CSharpProject/Packaging/PackageBase.cs
ocallesp: done

src/Microsoft.DotNet.Interactive.CSharpProject/RoslynWorkspaceUtilities/BuildDataResults.cs
ocallesp: In order to accomplish this, I will need to make other classes 'internal'.

I can implement these changes once I gain a clearer understanding of which APIs the service will utilize

jonsequitur: It's better to make them internal until a clear need for them to be public emerges. 

The main public methods that the Try .NET web service will need to call will be the kernel commands and events, and this infrastructure should be an implementation detail.

src/Microsoft.DotNet.Interactive.CSharpProject/RoslynWorkspaceUtilities/RoslynWorkspaceUtilities.cs
ocallesp: done.

Removed redundant ToList() for the empty array

ocallesp:
These unit-tests are fail when using Net 8.0.
It seems that there are some differences between Net 7.0 and Net 8.0 that are affecting the behavior's.

image

To fix When_diagnostics_are_outside_of_active_file_then_they_are_omitted we can change the unit-test to add these using to program:
```
using System;
using System.Linq;
```

To fix When_package_contains_aspnet_project_then_entry_point_dll_is_in_the_publish_directory probably we need to add `global.json`

jonsequitur:
We should probably always have a global.json.

src/Microsoft.DotNet.Interactive/Utility/Dotnet.cs
```diff
@@ -1,34 +1,57 @@
...
-    public Task<CommandLineResult> New(string templateName, string args = null)
+    public async Task<CommandLineResult> New(string templateName, string args = null)
     {
         if (string.IsNullOrWhiteSpace(templateName))
         {
             throw new ArgumentException("Value cannot be null or whitespace.", nameof(templateName));
         }
 
-        return Execute($@"new ""{templateName}"" {args}");
+        if (_sdkVersion is null)
```
colombod, line 34: sdk version should be driven by the package, not the dotnet class.

ocallesp: I'll move the code that creates Global.json to the abstract class Package.

ocallesp: done in last commit.

src/Microsoft.DotNet.Interactive.CSharpProject/Packaging/PackageInitializer.cs
```diff
@@ -67,4 +73,32 @@ private static string GetLanguageFromProjectName(string projectName)
...
+
+            sdkVersion = _defaultSdkVersion;
+        }
+
+        await dotnet.NewGlobalJson(sdkVersion);
+    }
+
+    private string GetLatestVersionBeforeEight(IEnumerable<string> versions)
+    {
+        var maxVersion = new Version(8, 0);  // specify the upper limit version
```
jonsequitur, line 95: Is this going to do something different from what we can do with `dotnet new globaljson`?

ocallesp: Removed.

Added the suggested change to create a global.json with latest sdk

ocallesp:
Git squash

ocallesp:
I removed the last commit and did a git rebase onto main

src/Microsoft.DotNet.Interactive/Utility/Dotnet.cs
```diff
@@ -21,14 +23,19 @@ public Dotnet(DirectoryInfo workingDirectory = null)
...
         if (string.IsNullOrWhiteSpace(templateName))
         {
             throw new ArgumentException("Value cannot be null or whitespace.", nameof(templateName));
         }
 
-        return Execute($@"new ""{templateName}"" {args}");
+        return await Execute($@"new ""{templateName}"" {args}");
+    }
+
+    public Task<CommandLineResult> NewGlobalJson()
```
jonsequitur, line 36: As long as this class is public, we should try to avoid adding new public methods.

src/Microsoft.DotNet.Interactive.CSharpProject/RoslynWorkspaceUtilities/RoslynWorkspaceUtilities.cs
```diff
@@ -0,0 +1,332 @@
...
+        {
+            return null;
+        }
+    }
+
+    internal static BuildDataResults GetResultsFromProjectFile(string csprojFilePath)
+    {
+        if (!File.Exists(csprojFilePath))
+        {
+            return null;
```
jonsequitur, line 59: This looks like it should throw rather than return null. If this happens, it's likely due to a code bug.

src/Microsoft.DotNet.Interactive.CSharpProject/RoslynWorkspaceUtilities/RoslynWorkspaceUtilities.cs
```diff
@@ -0,0 +1,332 @@
...
+                CSharpParseOptions = cSharpParseOptions
+            };
+        }
+        catch (Exception)
+        {
+            return null;
+        }
+    }
+
+    internal static BuildDataResults GetResultsFromProjectFile(string csprojFilePath)
```
jonsequitur, line 55: This method name is misleading, since it's actually using the cache file, not the project file.

src/Microsoft.DotNet.Interactive.CSharpProject/RoslynWorkspaceUtilities/RoslynWorkspaceUtilities.cs
```diff
@@ -0,0 +1,332 @@
...
+    {
+        buildProjectData = new BuildProjectData();
+
+        string[] lines = fileContent.Split(Environment.NewLine, StringSplitOptions.RemoveEmptyEntries);
+
+        List<string> referencesList = new List<string>();
+        List<string> projectReferencesList = new List<string>();
+        List<string> analyzerReferencesList = new List<string>();
+        List<string> preprocessorSymbolsList = new List<string>();
+        List<string> sourceFilesList = new List<string>();
```
jonsequitur, line 77: Nitpick: Please use `var` for these.

src/Microsoft.DotNet.Interactive.CSharpProject/RoslynWorkspaceUtilities/RoslynWorkspaceUtilities.cs
```diff
@@ -0,0 +1,332 @@
...
+            outputFilePath: buildProjectData.TargetPath,
+            documents: GetDocuments(buildProjectData, projectId),
+            projectReferences: GetExistingProjectReferences(buildProjectData, workspace),
+            metadataReferences: GetMetadataReferences(buildProjectData),
+            analyzerReferences: GetAnalyzerReferences(buildProjectData, workspace),
+            parseOptions: CreateParseOptions(buildProjectData, languageName),
+            compilationOptions: CreateCompilationOptions(buildProjectData, languageName));
+    }
+
+    private static bool TryGetSupportedLanguageName(string projectPath, out string languageName)
```
jonsequitur, line 228: Is there a reason for this method to exist given that C# is the only language this project supports?

src/Microsoft.DotNet.Interactive.CSharpProject/RoslynWorkspaceUtilities/RoslynWorkspaceUtilities.cs
```diff
@@ -0,0 +1,332 @@
...
+                return true;
+            default:
+                languageName = null;
+                return false;
+        }
+    }
+
+    private static IReadOnlyList<DocumentInfo> GetDocuments(BuildProjectData buildProjectData, ProjectId projectId) =>
+        buildProjectData
+            .SourceFiles?
```
jonsequitur, line 243: This method should not be allowed to return null. If there are no source files, we should return an empty list.

src/Microsoft.DotNet.Interactive.CSharpProject/RoslynWorkspaceUtilities/RoslynWorkspaceUtilities.cs
```diff
@@ -0,0 +1,332 @@
...
+        buildProjectData
+            .References?.Where(File.Exists)
+            .Select(x => MetadataReference.CreateFromFile(x)).ToList();
+
+    private static IReadOnlyList<AnalyzerReference> GetAnalyzerReferences(BuildProjectData buildProjectData, CodeAnalysis.Workspace workspace)
+    {
+        IAnalyzerAssemblyLoader loader = workspace.Services.GetRequiredService<IAnalyzerService>().GetLoader();
+
+        return buildProjectData
+                .AnalyzerReferences?
```
jonsequitur, line 269: This method should not be allowed to return null. If there are no source files, we should return an empty list.

src/Microsoft.DotNet.Interactive.CSharpProject/RoslynWorkspaceUtilities/RoslynWorkspaceUtilities.cs
```diff
@@ -0,0 +1,332 @@
...
+                    .Where(File.Exists)
+                    .Select(x => new AnalyzerFileReference(x, loader)).ToList();
+    }
+
+    private static ParseOptions CreateParseOptions(BuildProjectData buildProjectData, string languageName)
+    {
+        if (languageName == LanguageNames.CSharp)
```
jonsequitur, line 276: Same comment as above. This project doesn't currently need to try to generalize to other languages.

src/Microsoft.DotNet.Interactive.CSharpProject/RoslynWorkspaceUtilities/RoslynWorkspaceUtilities.cs
```diff
@@ -0,0 +1,332 @@
...
+                kind = OutputKind.NetModule;
+                break;
+            case "Winexe":
+                kind = OutputKind.WindowsApplication;
+                break;
+        }
+
+        if (kind.HasValue)
+        {
+            if (languageName == LanguageNames.CSharp)
```
jonsequitur, line 322: Same comment. And removing this generalization here means we avoid a code branch where this can return null.

ocallesp: done

src/Microsoft.DotNet.Interactive.CSharpProject/RoslynWorkspaceUtilities/RoslynWorkspaceUtilities.cs
ocallesp: fixed

src/Microsoft.DotNet.Interactive.CSharpProject/RoslynWorkspaceUtilities/RoslynWorkspaceUtilities.cs
ocallesp: fixed

src/Microsoft.DotNet.Interactive.CSharpProject/RoslynWorkspaceUtilities/RoslynWorkspaceUtilities.cs
ocallesp: fixed

src/Microsoft.DotNet.Interactive.CSharpProject/RoslynWorkspaceUtilities/RoslynWorkspaceUtilities.cs
ocallesp: fixed

src/Microsoft.DotNet.Interactive.CSharpProject/RoslynWorkspaceUtilities/RoslynWorkspaceUtilities.cs
ocallesp: fixed

src/Microsoft.DotNet.Interactive.CSharpProject/RoslynWorkspaceUtilities/RoslynWorkspaceUtilities.cs
ocallesp: removed

src/Microsoft.DotNet.Interactive.CSharpProject/RoslynWorkspaceUtilities/RoslynWorkspaceUtilities.cs
ocallesp: changed

src/Microsoft.DotNet.Interactive.CSharpProject/RoslynWorkspaceUtilities/RoslynWorkspaceUtilities.cs
ocallesp: renamed

src/Microsoft.DotNet.Interactive.CSharpProject/RoslynWorkspaceUtilities/RoslynWorkspaceUtilities.cs
ocallesp: fixed

src/Microsoft.DotNet.Interactive/Utility/Dotnet.cs
ocallesp: removed

src/Microsoft.DotNet.Interactive.CSharpProject/RoslynWorkspaceUtilities/RoslynWorkspaceUtilities.cs
```diff
@@ -0,0 +1,302 @@
...
+            {
+                BuildProjectData = buildProjectData,
+                ProjectFilePath = cacheFilePath,
+                CacheFilePath = cacheFilePath + cacheFilenameSuffix,
+                Workspace = ws,
+                Succeeded = true,
+                CSharpParseOptions = cSharpParseOptions
+            };
+        }
+        catch (Exception)
```
jonsequitur, line 50: We shouldn't ever catch `Exception`. Is there a specific exception you're expecting here?

src/Microsoft.DotNet.Interactive.CSharpProject/RoslynWorkspaceUtilities/RoslynWorkspaceUtilities.cs
```diff
@@ -0,0 +1,302 @@
...
+        var referencesList = new List<string>();
+        var projectReferencesList = new List<string>();
+        var analyzerReferencesList = new List<string>();
+        var preprocessorSymbolsList = new List<string>();
+        var sourceFilesList = new List<string>();
+
+        foreach (string line in lines)
+        {
+            string[] parts = line.Split('=');
+            if (parts.Length == 2)
```
jonsequitur, line 83: What if the value contains an `=`?

ocallesp: This parser is reading values from the auto-generated cache file. This cache file contains values from msbuild properties, and if one of this values are corrupted, then the build will fail.

src/Microsoft.DotNet.Interactive.CSharpProject/RoslynWorkspaceUtilities/RoslynWorkspaceUtilities.cs
ocallesp: Fixed. This will only catch the exception from the parser

src/Microsoft.DotNet.Interactive.CSharpProject/RoslynWorkspaceUtilities/RoslynWorkspaceUtilities.cs
```diff
@@ -0,0 +1,302 @@
...
+        ProjectId projectId = ProjectId.CreateFromSerialized(analyzerResult.ProjectGuid);
+
+        ConcurrentDictionary<Guid, string[]> WorkspaceProjectReferences = new ConcurrentDictionary<Guid, string[]>();
+        WorkspaceProjectReferences[projectId.Id] = analyzerResult.ProjectReferences.ToArray();
+
+        ProjectInfo projectInfo = GetProjectInfo(analyzerResult, workspace, projectId);
+        if (projectInfo is null)
+        {
+            // Error
+            return null;
```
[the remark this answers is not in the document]
ocallesp: Fixed.
This now returns an exception

src/Microsoft.DotNet.Interactive.CSharpProject/RoslynWorkspaceUtilities/RoslynWorkspaceUtilities.cs
```diff
@@ -47,7 +47,7 @@ internal static BuildDataResults GetResultsFromCacheFile(string cacheFilePath)
                 CSharpParseOptions = cSharpParseOptions
             };
         }
-        catch (Exception)
+        catch (ArgumentNullException)
         {
             return null;
```
jonsequitur, line 52: Why are we hiding an exception here at all? This is presumably happening due to a bug in our code.

src/Microsoft.DotNet.Interactive.CSharpProject/RoslynWorkspaceUtilities/RoslynWorkspaceUtilities.cs
```diff
@@ -172,8 +172,7 @@ public static CodeAnalysis.Project AddToWorkspace(BuildProjectData analyzerResul
         ProjectInfo projectInfo = GetProjectInfo(analyzerResult, workspace, projectId);
         if (projectInfo is null)
         {
-            // Error
-            return null;
+            throw new ArgumentNullException(nameof(projectInfo));
```
jonsequitur, line 175: This isn't the correct exception type, since it's not an argument.

What causes `projectInfo` to be null? Is it due to a bug somewhere else?

jonsequitur (approved):

src/Microsoft.DotNet.Interactive.CSharpProject/RoslynWorkspaceUtilities/RoslynWorkspaceUtilities.cs
ocallesp: I'll push a cleanup pr to fix this

PR #3122, opened 2023-08-08
update HttpRequestKernel to use new parser; support variable expressions in request body

jonsequitur:


src/Microsoft.DotNet.Interactive.HttpRequestParser/HttpSyntaxNode.cs
```diff
@@ -21,22 +21,67 @@ internal abstract class HttpSyntaxNode : HttpSyntaxNodeOrToken
...
+    {
+        get
+        {
+            var firstSignificantNodeOrToken = ChildNodesAndTokens
+                .FirstOrDefault(n => n.IsSignificant);
+
+            var lastSignificantNodeOrToken = ChildNodesAndTokens
+                .LastOrDefault(n => n.IsSignificant);
+
+            var startOfNonTriviaAndComments =
```
shyamnamboodiripad, line 54: I think something like `startOfSignificantText` may be less confusing - since comment and trivia tokens in the middle (after the first significant token and before the last significant token) are still considered part of the significant text...

src/Microsoft.DotNet.Interactive.HttpRequestParser/HttpSyntaxNode.cs
```diff
@@ -21,22 +21,67 @@ internal abstract class HttpSyntaxNode : HttpSyntaxNodeOrToken
...
+                firstSignificantNodeOrToken?.Span.Start ??
+                FullSpan.Start;
+
+            var endOfNonTriviaAndComments =
+                lastSignificantNodeOrToken?.Span.End ??
+                FullSpan.End;
```
shyamnamboodiripad, line 60: You probably fixed this already, but in the case where a `HttpSyntaxNode` only contains insignificant text (e.g., comments and whitespace token), looks like `Span` would return the same value as `FullSpan` (and include the insignificant text).

We probably need an early out to return a 0-width span that starts and ends at the start position of the first (insignificant) child in this case...

Of course, if we move this computation to the point when nodes are constructed, we may be able to handle this case more easily over there...

jonsequitur: I've moved span calculations to `GrowSpan`.

src/Microsoft.DotNet.Interactive.HttpRequest.Tests/Utility/AssertionExtensions.cs
```diff
@@ -22,4 +22,17 @@ internal static class AssertionExtensions
 
         return new AndWhichConstraint<ObjectAssertions, T>(subject.Should(), subject);
     }
+    
+    public static AndWhichConstraint<ObjectAssertions, T> ContainSingle<T>(
+        this GenericCollectionAssertions<HttpSyntaxNodeOrToken> should)
+        where T : HttpSyntaxNode
```
shyamnamboodiripad, line 28: Can we remove the overload on line 13 since `HttpSyntaxNodeOrToken` is a base type of `HttpSyntaxNode` (and since `IReadOnlyList` supports covariance)?

src/Microsoft.DotNet.Interactive.HttpRequestParser/HttpSyntaxNodeOrToken.cs
```diff
@@ -23,19 +23,18 @@ private protected HttpSyntaxNodeOrToken(SourceText sourceText, HttpSyntaxTree? s
...
     /// <summary>
     /// Gets the significant text of the current node or token, without trivia.
     /// </summary>
-    public string Text => TextWithTrivia.Trim();
-
-    /// <summary>
-    /// Gets the text of the current node or token, including trivia.
-    /// </summary>
-    public string TextWithTrivia => SourceText.ToString(Span);
+    public abstract string Text { get; }
```
shyamnamboodiripad, line 37: I don't think this needs to be abstract since the implementation can be identical for all derived types.

```suggestion
    public string Text => SourceText.ToString(Span);
    public string FullText => SourceText.ToString(FullSpan);
```

src/Microsoft.DotNet.Interactive.HttpRequestParser/HttpSyntaxNode.cs
```diff
@@ -12,31 +12,44 @@ namespace Microsoft.DotNet.Interactive.HttpRequest;
...
+    public override bool IsSignificant => _isSignificant;
+
+    public override string Text => SourceText.ToString(Span);
+
     public override TextSpan Span => _span;
 
+    /// <summary>
+    /// Gets the text of the current node, including trivia.
+    /// </summary>
+    public string FullText => SourceText.ToString(FullSpan);
```
shyamnamboodiripad, line 37: It is a bit strange that `FullSpan` is available on the base `HttpSyntaxNodeOrToken` but `FullText` is not.

If we think that `FullSpan` and `FullText` are only meaningful for `HttpSyntaxNode`, perhaps we should also move `FullSpan` down to `HttpSyntaxNode`. Alternatively, we can move `FullText` up to the base type so that caller does not have to think about whether they are dealing a node or a token when they want to get its full text.

src/Microsoft.DotNet.Interactive.HttpRequest.Tests/ParserTests.Comments.cs
```diff
@@ -12,13 +12,14 @@ public partial class ParserTests
...
-        public void comments_are_parsed_correctly()
+        public void line_comments_before_method_and_url_is_parsed_correctly()
         {
-            var result = Parse(
-                """
+            var code = """
                 # This is a comment
-                GET https://example.com HTTP/1.1"
-                """);
+                GET https://example.com HTTP/1.1
```
shyamnamboodiripad, line 19: The version part can probably be omitted.

src/Microsoft.DotNet.Interactive.HttpRequest.Tests/ParserTests.Comments.cs
```diff
@@ -12,13 +12,14 @@ public partial class ParserTests
     public class Comments
     {
         [Fact]
-        public void comments_are_parsed_correctly()
+        public void line_comments_before_method_and_url_is_parsed_correctly()
```
shyamnamboodiripad, line 15: nit: 
```suggestion
        public void line_comment_before_method_and_url_is_parsed_correctly()
```

src/Microsoft.DotNet.Interactive.HttpRequest.Tests/ParserTests.Method.cs
```diff
@@ -32,9 +32,35 @@ public void newline_is_legal_at_the_beginning_of_a_request()
...
+                """);
+
+            var methodNode = result.SyntaxTree
+                                   .RootNode
+                                   .ChildNodes
+                                   .Should()
+                                   .ContainSingle<HttpRequestNode>().Which
+                                   .MethodNode;
+
+            methodNode.ChildTokens.First().Kind.Should().Be(HttpTokenKind.NewLine);
```
shyamnamboodiripad, line 62: This assert seems less relevant for this test (and probably duplicates coverage from the previous test on line 42). Perhaps the assert should be changed to check for the comment node instead?

```suggestion
            methodNode.ChildNodes.First().Should().BeOfType(typeof(HttpCommentNode));
```

src/Microsoft.DotNet.Interactive.HttpRequest.Tests/ParserTests.Request.cs
```diff
@@ -34,6 +33,9 @@ public void multiple_request_are_parsed_correctly()
             var requestNodes = result.SyntaxTree.RootNode
                                      .ChildNodes.OfType<HttpRequestNode>();
 
+            var request = requestNodes.First();
+            var span = request.Span;
```
shyamnamboodiripad, line 37: Looks like dead code. Lines 36 and 37 should probably be removed.

src/Microsoft.DotNet.Interactive.HttpRequest.Tests/ParserTests.Trivia.cs
```diff
@@ -28,9 +26,7 @@ public void it_can_parse_a_string_with_only_whitespace()
 
             result.SyntaxTree.RootNode
                   .ChildNodes.Should().ContainSingle<HttpRequestNode>().Which
-                  .MethodNode.ChildTokens.First().TextWithTrivia.Should().Be(" \t ");
-
-            // TODO: Test error reporting.
+                  .MethodNode.ChildTokens.First().Text.Should().Be(" \t ");
```
shyamnamboodiripad, line 29: In this case I think we'll need to report an error for missing URL. This is probably a common state that the parser will find itself in as the user edits a file / cell.

jonsequitur: We can go deeper into error reporting in a future PR.

src/Microsoft.DotNet.Interactive.HttpRequestParser/HttpSyntaxNodeOrToken.cs
jonsequitur: Good catch.

src/Microsoft.DotNet.Interactive.HttpRequestParser/HttpSyntaxNode.cs
jonsequitur: This makes sense, but the meaning of the two is also different. On a token, `Span` really means the full span. `FullSpan` is used for things like `FindNode` and `GrowSpan`. Removing `HttpSyntaxNodeOrToken.FullSpan` results in needing to switch on the two types in a number of places.

src/Microsoft.DotNet.Interactive.HttpRequest.Tests/ParserTests.Method.cs
jonsequitur: Nice catch.

colombod (approved):

src/Microsoft.DotNet.Interactive.HttpRequestParser/HttpRequestNode.cs
```diff
@@ -70,10 +74,9 @@ internal class HttpRequestNode : HttpSyntaxNode
     public HttpBindingResult<HttpRequestMessage> TryGetHttpRequestMessage(HttpBindingDelegate bind)
     {
         var request = new HttpRequestMessage();
-        var diagnostics = new List<Diagnostic>();
-        var success = true;
+        var diagnostics = new List<Diagnostic>(base.GetDiagnostics());
```
shyamnamboodiripad, line 77: The binding methods for the child nodes (such as TryGetUri and TryGetBody) should also call base.GetDiagnostics() to include syntax errors (and return false if there are any diagnostics with severity Error).

shyamnamboodiripad, line 77: Should we early out and avoid trying to bind here if `diagnostics` contains (syntax) errors?

src/Microsoft.DotNet.Interactive.HttpRequestParser/HttpRequestNode.cs
```diff
@@ -70,10 +74,9 @@ internal class HttpRequestNode : HttpSyntaxNode
...
 
-        if (MethodNode is { Span.IsEmpty: false })
+        if (MethodNode is { FullSpan.IsEmpty: false })
```
shyamnamboodiripad, line 79: minor: Shouldn't this still be `.Span`? This condition will currently return `true` if the node happens to contain only whitespace. Is the `HttpMethod` constructor tolerant of empty string it will be passed to it in such cases?

src/Microsoft.DotNet.Interactive.HttpRequestParser/HttpRequestNode.cs
```diff
@@ -85,11 +88,57 @@ public HttpBindingResult<HttpRequestMessage> TryGetHttpRequestMessage(HttpBindin
         }
         else
         {
-            success = false;
             diagnostics.AddRange(uriBindingResult.Diagnostics);
         }
 
-        if (success)
+        var headers =
+            HeadersNode?.HeaderNodes.Select(h => new KeyValuePair<string, string>(h.NameNode.Text, h.ValueNode.Text)).ToArray()
```
shyamnamboodiripad, line 95: The headers could also contain expressions. But I guess we will flesh that out in future PRs cc @bleaphar

src/Microsoft.DotNet.Interactive.HttpRequestParser/HttpRequestNode.cs
```diff
@@ -85,11 +88,57 @@ public HttpBindingResult<HttpRequestMessage> TryGetHttpRequestMessage(HttpBindin
...
+                    request.Content.Headers.ContentType = MediaTypeHeaderValue.Parse(kvp.Value);
+                    break;
+                case "accept":
+                    request.Headers.Accept.Add(MediaTypeWithQualityHeaderValue.Parse(kvp.Value));
+                    break;
+                case "user-agent":
+                    request.Headers.UserAgent.Add(ProductInfoHeaderValue.Parse(kvp.Value));
+                    break;
+                default:
+                    request.Headers.Add(kvp.Key, kvp.Value);
```
shyamnamboodiripad, line 118: minor: It is surprising that this call would not internally handle `content-type`, `accept` and `user-agent` cases above. Are these the only three header kinds that callers are expected to handle separately?

src/Microsoft.DotNet.Interactive.HttpRequestParser/HttpRequestNode.cs
shyamnamboodiripad: I guess that could lead to some duplication at higher levels (unless we change the list to a hashset)...

src/Microsoft.DotNet.Interactive.HttpRequestParser/HttpRequestNode.cs
```diff
@@ -85,11 +88,57 @@ public HttpBindingResult<HttpRequestMessage> TryGetHttpRequestMessage(HttpBindin
...
+            Array.Empty<KeyValuePair<string, string>>();
+
+        foreach (var kvp in headers)
+        {
+            switch (kvp.Key.ToLowerInvariant())
+            {
+                case "content-type":
+                    if (request.Content is null)
+                    {
+                        request.Content = new StringContent("");
```
shyamnamboodiripad, line 106: Should this be handled globally in the `else` for the `if` block on line 136 below? Looks like Content being null is handled here - but if content is already an empty string we won't set `request.Content`.

shyamnamboodiripad, line 106: Can this be done globally in the `else` clause for the `if` block on line 136 below instead (i.e. always set to empty string if a parsed body is empty)?

The null check for `request.Content` is also confusing since we have not touched `request.Content` above this line. Did you mean to check something else perhaps?

src/Microsoft.DotNet.Interactive.HttpRequestParser/HttpSyntaxNode.cs
shyamnamboodiripad: I see... Should we also move `FullText` to the base type then so that the API is consistent?

src/Microsoft.DotNet.Interactive.HttpRequestParser/HttpSyntaxNode.cs
```diff
@@ -48,15 +59,34 @@ internal abstract class HttpSyntaxNode : HttpSyntaxNodeOrToken
...
-            _span = new TextSpan(_spanStart, _spanEnd - _span.Start);
+            var fullSpanStart = Math.Min(_fullSpan.Start, child.FullSpan.Start);
+            var fullSpanEnd = Math.Max(_fullSpan.End, child.FullSpan.End);
+            _fullSpan = new TextSpan(fullSpanStart, fullSpanEnd - _fullSpan.Start);
+
+            var firstSignificantNodeOrToken = ChildNodesAndTokens
+                .FirstOrDefault(n => n.IsSignificant);
+
+            var lastSignificantNodeOrToken = ChildNodesAndTokens
+                .LastOrDefault(n => n.IsSignificant);
```
shyamnamboodiripad, line 77: I think we can improve this part by remembering the first and last significant children in fields as children are added inside `Add` below. That way we won't need to traverse all the child nodes here each time a child is added.

src/Microsoft.DotNet.Interactive.HttpRequestParser/HttpRequestParser.cs
```diff
@@ -445,15 +441,22 @@ private HttpHeaderValueNode ParseHeaderValue()
...
+                CurrentToken.Kind is not (HttpTokenKind.Whitespace or HttpTokenKind.NewLine) &&
+                !IsRequestSeparator())
             {
                 ConsumeCurrentTokenInto(node);
 
                 while (MoreTokens() && !IsRequestSeparator())
                 {
-
-                    ConsumeCurrentTokenInto(node);
+                    if (!IsAtStartOfEmbeddedExpression())
```
shyamnamboodiripad, line 452: minor nit: Invert the condition

src/Microsoft.DotNet.Interactive.HttpRequest/HttpRequestKernel.cs
```diff
@@ -106,25 +107,23 @@ async Task IKernelCommandHandler<SubmitCode>.HandleAsync(SubmitCode command, Ker
 
         try
         {
-            foreach (var parsedRequest in parsedRequests)
+            foreach (var requestMessage in requestMessages)
             {
-                await HandleRequestAsync(parsedRequest, command, context);
+                if (requestMessage is not null)
```
shyamnamboodiripad, line 112: consider: Move this into the `Select` lambda on line 97 (and only select results that have `BindingResult.IsSuccessful` set to `true`)

src/Microsoft.DotNet.Interactive.HttpRequest/HttpRequestKernel.cs
```diff
@@ -269,192 +234,61 @@ private static HttpRequestMessage GetRequestMessage(ParsedHttpRequest parsedRequ
...
             context.Publish(new DiagnosticsProduced(diagnostics, command, formattedDiagnostics));
         }
     }
 
     Task IKernelCommandHandler<RequestDiagnostics>.HandleAsync(RequestDiagnostics command, KernelInvocationContext context)
     {
-        var requestsAndDiagnostics = InterpolateAndGetDiagnostics(command.Code);
-        var diagnostics = requestsAndDiagnostics.SelectMany(r => r.Diagnostics);
+        var parseResult = HttpRequestParser.Parse(command.Code);
+        var requestsAndDiagnostics = GetAllDiagnostics(parseResult);
```
shyamnamboodiripad, line 254: nit: The variable name is confusing.

shyamnamboodiripad (commented):
@jonsequitur The changes mostly look great 👍🏾

I've added a few comments that may be worth a look. Some of those (such as a potential perf improvement for `GrowSpan()`) may be worth fixing up in follow up PRs.

src/Microsoft.DotNet.Interactive.HttpRequestParser/HttpRequestNode.cs
jonsequitur: Yeah, I deliberately kept these concerns separate.

jonsequitur: My thinking was that binding can provide additional and potentially orthogonal errors. It might be reasonable to see errors for both, e.g. a misspelled HTTP verb **_and_** an undeclared variable.

src/Microsoft.DotNet.Interactive.HttpRequestParser/HttpRequestNode.cs
jonsequitur: The headers grammar is conceptually flat but the `HttpRequestMessage` object model spreads them across these different child objects. There might be other cases that require special treatment. This is probably also something we'll need to take a systematic approach to in the next rounds of testing, like the grammar combinations.

src/Microsoft.DotNet.Interactive.HttpRequestParser/HttpRequestNode.cs
shyamnamboodiripad: I see - that sounds reasonable.

src/Microsoft.DotNet.Interactive.HttpRequest/HttpRequestKernel.cs
```diff
@@ -269,192 +234,61 @@ private static HttpRequestMessage GetRequestMessage(ParsedHttpRequest parsedRequ
...
-        return lines.Any(line => IsRequest.IsMatch(line));
-        //return lines.Any() && lines.Any(line => !string.IsNullOrWhiteSpace(line));
-    }
-
-    private IEnumerable<ParsedHttpRequest> ParseRequests(string requests)
-    {
-        var parsedRequests = new List<ParsedHttpRequest>();
-
-        foreach (var (request, diagnostics) in InterpolateAndGetDiagnostics(requests))
+        foreach (var expressionNode in parseResult.SyntaxTree.RootNode.DescendantNodesAndTokensAndSelf().OfType<HttpExpressionNode>())
```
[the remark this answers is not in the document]
jonsequitur: > Some of those diagnostics may be reported in URIs / headers with no embedded expressions (but would still be diagnostics that are reported at bind-time).

The block before this (`parseResult.GetDiagnostics()`) is getting the syntactic diagnostics and this code is getting the symbolic diagnostics.  My thinking was that keeping them as separate steps is useful in this code path (i.e. in the context of `RequestDiagnostics`) because the latter will sometimes include side effects such as UI prompts which we shouldn't trigger here.

src/Microsoft.DotNet.Interactive.HttpRequest/HttpRequestKernel.cs
jonsequitur: This was a refactoring artifact. Inlined.

src/Microsoft.DotNet.Interactive.HttpRequest/HttpRequestKernel.cs
jonsequitur: The main reason for the `if not null` check was to avoid the nullability warning from the compiler. I agree that moving the check into LINQ code seems cleaner. Doing that in a way that avoids the warning seems a bit uglier. All highly subjective and I don't have a strong preference beyond correctness here.

Here's what I've got. (Not the `r.Value!` is needed).

```csharp

        var requestMessages = httpRequestResults
                              .Where(r => r is { IsSuccessful: true, Value: not null })
                              .Select(r => r.Value!).ToArray();

        try
        {
            foreach (var requestMessage in requestMessages)
            {
                await SendRequestAsync(requestMessage, command, context);
            }
        }

```

src/Microsoft.DotNet.Interactive.HttpRequestParser/HttpRequestNode.cs
jonsequitur: No, I meant to check this. Coverage indicates that `request.Content` is null 100% of the time here. But I realized there's a bug in that this header can be get overridden when we set the `Content` again for the body below, so some refactoring is needed.

src/Microsoft.DotNet.Interactive.HttpRequest/HttpRequestKernel.cs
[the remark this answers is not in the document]
shyamnamboodiripad: What I meant is that there could be more to the symbolic diagnostics beyond variable resolution and this loop only seems to be handling variable resolution.

The `TryGetRequest` API (and other APIs it calls such as `TryGetUri`) would already be considering all such diagnostics - so my thinking was that it may be better to rely exclusively on that API and centralize all validation / binding logic over there.

src/Microsoft.DotNet.Interactive.HttpRequestParser/HttpRequestNode.cs
shyamnamboodiripad: Ah does the System.Net.Http API try to deduce and set the content-type header implicitly when `Content` is set below? That is certainly not very intuitive 😄

I agree this needs some refactoring... I think we can delete this if block for now as it is dead code.

src/Microsoft.DotNet.Interactive.HttpRequestParser/HttpRequestNode.cs
shyamnamboodiripad: Thinking more perhaps it may be better for the binding API calls to only return binding errors. The syntax errors are already available on the node which the caller is trying to bind.

This may be more flexible from an API standpoint and would also keep the behavior consistent regardless of which construct is being bound... Otherwise, it forces callers to reason about which construct they are binding and whether or not they should fetch the syntax diagnostics separately.

src/Microsoft.DotNet.Interactive.HttpRequest/HttpRequestKernel.cs
shyamnamboodiripad: We could use the nullable annotation on `.Value` to indicate that it can never be `null` when `IsSuccessful` is `true`. But I think the annotations are not universally available so we would need to add a source file that declares those attributes in the netstandard2.0 case (and make sure that similar code is also available in the other repos where the source package is consumed).

src/Microsoft.DotNet.Interactive.HttpRequest/HttpRequestKernel.cs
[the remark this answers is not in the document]
jonsequitur: > What I meant is that there could be more to the symbolic diagnostics beyond variable resolution and this loop only seems to be handling variable resolution.

Do we have an example of this?

src/Microsoft.DotNet.Interactive.HttpRequestParser/HttpRequestNode.cs
jonsequitur: The code isn't dead code. 100% of tests enter the `if` block.

> Ah does the System.Net.Http API try to deduce and set the content-type header implicitly when Content is set below?

It doesn't and there's no way it reliably could.

src/Microsoft.DotNet.Interactive.HttpRequest/HttpRequestKernel.cs
jonsequitur: > We could use the nullable annotation on .Value to indicate that it can never be null when IsSuccessful is true.

`null` is a valid value.

src/Microsoft.DotNet.Interactive.HttpRequest/HttpRequestKernel.cs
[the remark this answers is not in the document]
shyamnamboodiripad: One example may be if we add validation for supported version numbers for `HttpVersionNode`. Another may be if we need to add validation for specific values that are supported / unsupported for `HeaderValueNode`.

Wonder if we should try and create a more general `TryBind()` API which performs the same binding steps that `TryGetHttpRequest()` does without creating / returning the `System.Net.Http.*` objects. However, I feel some of these validations may require calling into `System.Net.Http` regardless so we may not be saving much...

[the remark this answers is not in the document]
jonsequitur: I would expect these to be directly handled by the parser, since they don't change often and they don't vary based on the host environment.

PR #3129, opened 2023-08-09
Refactor build and package scripts, and improve cleanup

ocallesp:
This change:

- Creation of a destination directory for NuGet packages if it doesn't exist.
- check for folders before trying to delete them

### DIFF
--- a/repack.ps1
+++ b/repack.ps1
@@ -8,8 +8,17 @@ dotnet build
 dotnet pack  /p:PackageVersion=2.0.0
 
 # copy the dotnet-interactive packages to the temp directory
-Get-ChildItem -Recurse -Filter *.nupkg | Move-Item -Destination c:\temp\packages -Force
+$destinationPath = "C:\temp\packages"
+if (-not (Test-Path -Path $destinationPath -PathType Container)) {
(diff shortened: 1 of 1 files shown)


repack.ps1
```diff
@@ -8,8 +8,17 @@ dotnet build
 dotnet pack  /p:PackageVersion=2.0.0
 
 # copy the dotnet-interactive packages to the temp directory
-Get-ChildItem -Recurse -Filter *.nupkg | Move-Item -Destination c:\temp\packages -Force
+$destinationPath = "C:\temp\packages"
+if (-not (Test-Path -Path $destinationPath -PathType Container)) {
+    New-Item -Path $destinationPath -ItemType Directory -Force
+}
+Get-ChildItem -Path ".\artifacts" -Recurse -Filter *.nupkg | Move-Item -Destination $destinationPath -Force
```
jonsequitur, line 15: The artifacts path is only present when using the Arcade build, which we typically disable for local development. Limiting the search to this directory will break that.

ocallesp: let me remove ".\artifacts"

ocallesp: removed in last commit

jonsequitur (approved):

PR #3162, opened 2023-09-05
use PasswordString object

colombod:
using PasswordString object in GetPAsswordGesture to prevent printing password values in clear

image

### DIFF
--- a/src/Microsoft.DotNet.Interactive/PasswordString.cs
+++ b/src/Microsoft.DotNet.Interactive/PasswordString.cs
@@ -1,8 +1,12 @@
 ﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using Microsoft.DotNet.Interactive.Formatting;
+using System.Text.Json.Serialization;
+
 namespace Microsoft.DotNet.Interactive;
 
+[JsonConverter(typeof(PasswordStringJsonConverter))]
 public class PasswordString
 {
     private readonly string _clearTextPassword;
@@ -13,4 +17,9 @@ public class PasswordString
     }
 
     public string GetClearTextPassword() => _clearTextPassword;
+
+    public override string ToString()
+    {
+        return "************";
+    }
 }
--- a/src/Microsoft.DotNet.Interactive/PasswordStringJsonConverter.cs
+++ b/src/Microsoft.DotNet.Interactive/PasswordStringJsonConverter.cs
@@ -1,0 +1,52 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Text.Json;
+
+namespace Microsoft.DotNet.Interactive;
+
+internal class PasswordStringJsonConverter : JsonConverter<PasswordString>
+{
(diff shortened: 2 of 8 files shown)


src/Microsoft.DotNet.Interactive/PasswordString.cs
```diff
@@ -1,8 +1,13 @@
 ﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using Microsoft.DotNet.Interactive.Formatting;
```
colombod, line 4: This is not needed

src/Microsoft.DotNet.Interactive/PasswordStringFormatterSource.cs
```diff
@@ -0,0 +1,31 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Collections.Generic;
+using System.Text.Json;
+using Microsoft.DotNet.Interactive.Formatting;
+
+namespace Microsoft.DotNet.Interactive;
+
+internal class PasswordStringFormatterSource : ITypeFormatterSource
```
jonsequitur, line 10: This seems more complex than needed. A custom `ToString` should work fine.

colombod, line 10: not needed

src/Microsoft.DotNet.Interactive/PasswordStringJsonConverter.cs
```diff
@@ -0,0 +1,52 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Text.Json;
+
+namespace Microsoft.DotNet.Interactive;
+
+internal class PasswordStringJsonConverter : JsonConverter<PasswordString>
+{
+    public override PasswordString Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
```
colombod, line 11: remove this

src/Microsoft.DotNet.Interactive/PasswordStringJsonConverter.cs
```diff
@@ -0,0 +1,52 @@
...
+            }
+        }
+
+        throw new JsonException($"Cannot deserialize {typeToConvert.Name}");
+    }
+
+    public override void Write(Utf8JsonWriter writer, PasswordString value, JsonSerializerOptions options)
+    {
+        writer.WriteStartObject();
+        writer.WriteString("password", value.GetClearTextPassword());
```
shyamnamboodiripad, line 49: Is it by design that passwords flow through the system as clear text (across process boundary)? Or is this a known point in time issue that we plan to address later.

shyamnamboodiripad (approved):

src/Microsoft.DotNet.Interactive/PasswordStringJsonConverter.cs
jonsequitur: It's by design. Encrypting data in transit is the responsibility of transport-level security as passwords aren't the only sensitive information we can assume people are sending and receiving.

With than in mind, this PR is intended only to obfuscate certain pieces of information from display in the UI.

PR #3169, opened 2023-09-11
updates for Microsoft.DotNet.Interactive.Documents API stabilization

jonsequitur:


src/Microsoft.DotNet.Interactive.Documents/InteractiveDocument.cs
```diff
@@ -264,19 +259,23 @@ internal static void MergeKernelInfos(KernelInfoCollection destination, KernelIn
...
 
-    internal static bool TryGetKernelInfoFromMetadata(
+    public bool TryGetKernelInfosFromMetadata(
+        [NotNullWhen(true)] out KernelInfoCollection? kernelInfos) =>
+        TryGetKernelInfosFromMetadata(Metadata, out kernelInfos);
+
+    internal static bool TryGetKernelInfosFromMetadata(
         IDictionary<string, object>? metadata,
-        [NotNullWhen(true)] out KernelInfoCollection? kernelInfo)
+        [NotNullWhen(true)] out KernelInfoCollection? kernelInfos)
```
shyamnamboodiripad, line 268: Can this be inlined into the above non-static overload (with all callers switched to calling the non-static overload)? Or can this be made private?

Please ignore the above if we have more callers for the static overload outside the current class.

src/Microsoft.DotNet.Interactive.Documents/InteractiveDocument.cs
```diff
@@ -65,7 +65,7 @@ public async IAsyncEnumerable<InteractiveDocument> GetImportsAsync(bool recursiv
     {
         EnsureImportFieldParserIsInitialized();
 
-        if (!TryGetKernelInfoFromMetadata(Metadata, out var kernelInfos))
+        if (!TryGetKernelInfosFromMetadata(out var kernelInfos))
```
shyamnamboodiripad, line 68: minor nit: This call uses the public instance overload while others below use the static overload. Could we standardize to the same call?

src/Microsoft.DotNet.Interactive.Documents/InteractiveDocument.cs
```diff
@@ -198,7 +198,7 @@ internal void NormalizeElementKernelNames(KernelInfoCollection kernelInfos)
 
     public string? GetDefaultKernelName()
     {
-        if (TryGetKernelInfoFromMetadata(Metadata, out var kernelInfo))
+        if (TryGetKernelInfosFromMetadata(Metadata, out var kernelInfo))
         {
             return kernelInfo.DefaultKernelName;
         }
```
shyamnamboodiripad, line 204: Why is the public API missing the remaining heuristics (like checking for` kernelspec` metadata)

shyamnamboodiripad (approved):

src/Microsoft.DotNet.Interactive.Documents/InteractiveDocument.cs
```diff
@@ -264,19 +259,23 @@ internal static void MergeKernelInfos(KernelInfoCollection destination, KernelIn
         destination.AddRange(source.Where(ki => added.Add(ki.Name)));
     }
 
-    internal static bool TryGetKernelInfoFromMetadata(
+    public bool TryGetKernelInfosFromMetadata(
```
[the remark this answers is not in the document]
jonsequitur: We discussed this during the sync, that there's currently no way to get the kernel info using the public API, and that it's very useful to have.

PR #3171, opened 2023-09-12
Break dependency of HttpRequestParser on Microsoft.DotNet.Interactive.VisualStudio.dll

shyamnamboodiripad:
This ensures that consumers can update the source-only `Microsoft.DotNet.Interactive.HttpRequestParser` package and the `Microsoft.DotNet.Interactive.VisualStudio` package independently.

To break the dependency, `HttpRequestParser` now uses Roslyn's `Microsoft.CodeAnalysis.Diagnostic` type to report diagnostics (as opposed to using the `Microsoft.DotNet.Interactive.Diagnostic` type).


src/Microsoft.DotNet.Interactive.HttpRequestParser/Parser/HttpDiagnostics.cs
```diff
@@ -0,0 +1,74 @@
...
+        params object[] messageArguments)
+    {
+        var id = $"HTTP{(int)diagnosticCode:0000}";
+        string messageFormat;
+        var severity = DiagnosticSeverity.Error;
+
+        switch (diagnosticCode)
+        {
+            case WellKnownHttpDiagnostics.UnrecognizedVerb:
+                messageFormat = "Unrecognized HTTP verb '{0}'.";
```
jonsequitur, line 24: I would strongly recommend creating a class that exposes each message as a method so that understanding the number of tokens that need to be passed in the array isn't an issue.

src/Microsoft.DotNet.Interactive.HttpRequestParser/Parser/WellKnownHttpDiagnostics.cs
```diff
@@ -0,0 +1,19 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.DotNet.Interactive.HttpRequest;
+
+internal enum WellKnownHttpDiagnostics
```
jonsequitur, line 6: This shouldn't be needed if we use methods instead.

jonsequitur (approved):

PR #3174, opened 2023-09-14
fix #2095: remove SubmitCode.SubmissionType

### LINKED ISSUE #2095: remove SubmitCode.SubmissionType

jonsequitur:
Fixes #2095.


src/Microsoft.DotNet.Interactive.Jupyter/IsCompleteRequestHandler.cs
```diff
@@ -32,11 +34,15 @@ public async Task Handle(JupyterRequestContext context)
     {
         switch (@event)
         {
-            case CompleteCodeSubmissionReceived _:
-                Reply(true, context.JupyterRequestMessageEnvelope, context.JupyterMessageSender);
+            case DiagnosticsProduced diagnosticsProduced:
```
shyamnamboodiripad, line 37: Should `CommandFailed` also be handled?

src/Microsoft.DotNet.Interactive.Jupyter/IsCompleteRequestHandler.cs
```diff
@@ -32,11 +34,15 @@ public async Task Handle(JupyterRequestContext context)
...
+                if (diagnosticsProduced.Diagnostics.Any(d => d.Severity == DiagnosticSeverity.Error))
+                {
+                    Reply(false, context.JupyterRequestMessageEnvelope, context.JupyterMessageSender);
+                }
                 break;
-            case IncompleteCodeSubmissionReceived _:
-                Reply(false, context.JupyterRequestMessageEnvelope, context.JupyterMessageSender);
```
shyamnamboodiripad, line 39: Were the removed code paths pretty much unused?

shyamnamboodiripad (approved):

src/Microsoft.DotNet.Interactive.Jupyter/IsCompleteRequestHandler.cs
jonsequitur: No. If `RequestDiagnostics` fails that's a bug.

src/Microsoft.DotNet.Interactive.Jupyter/IsCompleteRequestHandler.cs
jonsequitur: They're unused after this code change.

shyamnamboodiripad: I am not very familiar with the code, but the current class is called `IsCompleteRequestHandler` and I was a bit confused since it seemed like it was meant to check whether the complete submission has been received while now it checks whether or not the submission contains errors. Do we need to rename this for clarity?

jonsequitur: Checking for errors is an implementation detail. This handler's purpose is the same as before, which is to inform the Jupyter console as to whether a code submission is "complete". This isn't a common use case for us in any event.

PR #3175, opened 2023-09-14
fix parsing of multiple comment lines

jonsequitur:
This PR adds combinatorial testing to produce test cases with multiple consecutive line comments in all of the valid positions, and fixes the previous issue with parsing them.

### DIFF
--- a/src/Microsoft.DotNet.Interactive.HttpRequestParser/Parser/HttpRequestParser.cs
+++ b/src/Microsoft.DotNet.Interactive.HttpRequestParser/Parser/HttpRequestParser.cs
@@ -46,10 +46,7 @@ internal class HttpRequestParser
 
             while (MoreTokens())
             {
-                if (ParseComment() is { } commentNode)
-                {
-                    commentsToPrepend.Add(commentNode);
-                }
+                commentsToPrepend.AddRange(ParseComments());
 
                 if (ParseVariableDeclarations() is { } variableNodes)
                 {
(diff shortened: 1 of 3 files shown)


src/Microsoft.DotNet.Interactive.HttpRequestParser/Parser/HttpRequestParser.cs
```diff
@@ -697,9 +694,9 @@ private HttpHeaderValueNode ParseHeaderValue()
             return node;
         }
 
-        private HttpCommentNode? ParseComment()
+        private IEnumerable<HttpCommentNode> ParseComments()
         {
-            if (IsComment())
+            while (IsComment())
```
shyamnamboodiripad, line 699: We may have another bug if there is a blank line (or a line with only whitespaces) in between two comments.

src/Microsoft.DotNet.Interactive.HttpRequestParser/Parser/HttpRequestParser.cs
```diff
@@ -356,9 +353,9 @@ private void ConsumeCurrentTokenInto(HttpSyntaxNode node)
                 AddCommentsIfAny(comments, requestNode, addBefore: false);
             }
 
-            if (ParseComment() is { } commentAfterHeaders)
+            foreach (var comment in ParseComments())
             {
-                requestNode.Add(commentAfterHeaders);
+                requestNode.Add(comment);
```
shyamnamboodiripad, line 358: Consider: It may be useful to introduce an `HttpSyntaxNode.AddRange` method for comments.

shyamnamboodiripad (approved):

PR #3213, opened 2023-10-02
Create a dedicated script for the dotnet portion of the build

shyamnamboodiripad:
Adds a new `build-dotnet.cmd` next to the existing `build-js.cmd`.

This helps save some time when I am testing changes that only impact dotnet side (such as updating dotnet sdk or other arcade settings, or when trying to build NuGet packages locally by passing the `-pack` argument).


shyamnamboodiripad:
Sounds like `dotnet build dotnet-interactive.sln` should already handle most of the scenarios served by the above script.

PR #3216, opened 2023-10-02
Add info about adding new languages

ShalokShalom:
Add info about adding new languages


jonsequitur:
There's a bit more to this. I'm going to close this PR and I'll make a docs update for it. 

The information you've added probably belongs in the developer documentation (since it requires a PR to this repo per language), and needs additional context (since Jupyter subkernels aren't the only way to add support for additional languages).

ShalokShalom:
Yes, I was tempted to add it to the docs as well, and I thought some place that is a bit more prominent would be good also. 

What do you think about the FAQ?

PR #3246, opened 2023-10-11
Multi-target Microsoft.DotNet.Interactive.csproj to support both netstandard2.0 as well as net8.0

shyamnamboodiripad:
Also includes the following related changes:

* Remove Microsoft.DotNet.Interactive.VisualStudio.csproj now that Microsoft.DotNet.Interactive.csproj itself supports netstandard2.0.
* Rename Microsoft.DotNet.Interactive.VisualStudio.Tests.csproj to Microsoft.DotNet.Interactive.NetFramework.Tests.csproj.
* Move StdIoKernelConnector from dotnet-interactive.csproj into Microsoft.DotNet.Interactive.csproj.
* Include some additional tests related to ProxyKernel in Microsoft.DotNet.Interactive.NetFramework.Tests.csproj.


src/Microsoft.DotNet.Interactive/Microsoft.DotNet.Interactive.csproj
```diff
@@ -12,6 +12,19 @@
     <PackageDescription Condition="'$(PackageDescription)'==''">Core types for building applications providing interactive programming for .NET.</PackageDescription>
     <PackageTags Condition="'$(PackageTags)'==''">polyglot notebook dotnet interactive</PackageTags>
   </PropertyGroup>
+  
+  <ItemGroup Label="Resources">
+    <!-- JavaScript kernel -->
+    <EmbeddedResource Include="..\polyglot-notebooks\lib\polyglot-notebooks.js" />
```
jonsequitur, line 18: Embedding this in this library seems odd. What pulled this over here?

shyamnamboodiripad: Ah this is needed on the VS side and was originally part of `Microsoft.DotNet.Interactive.VisualStudio.dll`. I moved it here as part of deleting `Microsoft.DotNet.Interactive.VisualStudio.csproj`. I just remembered that we had said we could put this resource in the `BrowserKernels` package, multi-target that to support `netstandard2.0` and fetch it from there instead. Let me look into that...

colombod (approved):

PR #3258, opened 2023-10-13
gpt function support

colombod:


src/Microsoft.DotNet.Interactive.AIUtilities.Tests/FunctionTests.cs
```diff
@@ -0,0 +1,250 @@
...
+                                                  "a",
+                                                  "b",
+                                                  "c"
+                                                ]
+                                              }
+                                              """);
+    }
+
+    [Fact]
+    public void can_create_function_from_delegate()
```
jonsequitur, line 93: This one seems less specific than the others and probably not needed.

src/Microsoft.DotNet.Interactive.AIUtilities.Tests/FunctionTests.cs
```diff
@@ -0,0 +1,250 @@
...
+        var function = GPTFunctionDefinition.Create((string a, double b) => $"{a} {b}", "concatString");
+
+        var jsonArgs = """
+                        {
+                            "name": "concatString",
+                            "arguments": "{ \"a\": \"Diego\", \"b\":123.0}"
+                        }
+                       """;
+        
+        function.Execute<string>(jsonArgs, out var result);
```
jonsequitur, line 230: The return type can be inferred if the function definition type is generic.

src/Microsoft.DotNet.Interactive.AIUtilities/Function.cs
```diff
@@ -0,0 +1,202 @@
...
+    internal GptFunction(Delegate function, string jsonSignature, string name)
+    {
+        Name = name;
+        _function = function;
+        JsonSignature = jsonSignature;
+    }
+
+    public object? Execute(string parameterJson)
+    {
+        // parameters extraction
```
jonsequitur, line 26: Comment is unnecessary.

src/Microsoft.DotNet.Interactive.AIUtilities.Tests/FunctionTests.cs
```diff
@@ -0,0 +1,221 @@
...
+                                                               "c"
+                                                             ]
+                                                           }
+                                                           """);
+    }
+
+    [Fact]
+    public void can_create_function_from_delegate_with_enums_as_parameters()
+    {
+        var declaration = GptFunction.Create((int a, double b, EnumType c) => $"{a} {b} {c}", "DoCompute");
```
jonsequitur, line 96: I would either put the name property first or give it a default value. Why does the caller care about the function name?

colombod: done

jonsequitur (approved):

PR #3275, opened 2023-10-19
return false on command and restrict cursor position >= 0

cartercanedy:


cartercanedy:
This fixes a bug that caused crashes in dotnet-repl. I used the NuGet pkgs containing the bug fixes in a test build of dotnet-repl, changing API consumption as necessary to correct errors introduced by the interface changes between the current state of main and the current build referenced by the project, and the crashes were remedied.

colombod (approved):

colombod:
I merged but I noticed no test has been added. Can you add the test case or describe the repro?

cartercanedy:
@colombod created new pr with the testing and made another change to

PR #3280, opened 2023-10-19
corrects buffer position logic & create tests

cartercanedy:
### Changes:
* buffer position logic only applied to first conditional, but position could still be decremented if second conditional was true
* added a test to verify that unhandled exceptions aren't being emitted from language service commands

The test case that I created replicates the failing condition that I observed while debugging dotnet-repl.


cartercanedy:
@dotnet-policy-service agree

cartercanedy:
Just updated the test so that it is excluded from netfx testing. Also made sure that the testing kernel being created was getting disposed on test exit

PR #3281, opened 2023-10-19
amend cursor position decrement logic & create language service command kernel test

cartercanedy:
### Changes:
* Refactored the logic around cursor position decrement to skip conditionals altogether if cursor position is at pos 0
* Added a test to catch a kernel crash on submission of a language service command where the cursor position is set to 0 in a 0-length buffer

### DIFF
--- a/src/Microsoft.DotNet.Interactive.Tests/KernelTests.cs
+++ b/src/Microsoft.DotNet.Interactive.Tests/KernelTests.cs
@@ -14,7 +14,6 @@ using Microsoft.DotNet.Interactive.Commands;
 using Microsoft.DotNet.Interactive.Events;
 using Microsoft.DotNet.Interactive.Tests.Utility;
 using Xunit;
-using Microsoft.DotNet.Interactive.App;
 
 #if !NETFRAMEWORK
 using Microsoft.DotNet.Interactive.CSharp;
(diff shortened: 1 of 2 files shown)


colombod:
Thank you for this contribution

colombod (approved):

shyamnamboodiripad:
Thanks for the contribution @cartercanedy! Could you please clarify why the test was excluded from netfx? I see that some other tests in the file were already excluded - but I don't remember why.

cartercanedy:
It looks like the test case is available for both netcore and netfx through Microsoft.DotNet.Interactive.Tests Microsoft.DotNet.Interactive.VisualStudio.Tests. The test uses a CSharpKernel nested in a CompositeKernel, which is unavailable in the latter.

shyamnamboodiripad:
Ah indeed - that makes sense. I was wondering if there was some other reason why the test fails on netfx. Thanks for clarifying.

cartercanedy:
> It looks like the test case is available for both netcore and netfx through Microsoft.DotNet.Interactive.Tests Microsoft.DotNet.Interactive.VisualStudio.Tests. The test uses a CSharpKernel nested in a CompositeKernel, which is unavailable in the latter, so the test project wouldn't compile otherwise

Issue #3286, opened 2023-10-23
KeyNotFoundException when notebook imports another notebook that has magic commands


tparviainen:
### Describe the bug
1. Create a new blank notebook (dib file, C# language), add next line to that file
```
#r "nuget: Newtonsoft.Json, 13.0.3"
```
2. Save the file to a name `Notebook1.dib`
3. Create a new blank notebook (dib file, C# language), add next line to that file
```
#!import "./Notebook1.dib"
```
4. Save the file to a name `Notebook2.dib` and to the same folder where `Notebook1.dib` file is.
5. Run the cell from the `Notebook2.dib` that that imports the another notebook
6. The output is:

```
{
	"name": "Error",
	"message": "System.Collections.Generic.KeyNotFoundException: The given key 'SubmitCode: #r \"nuget: Newtonsoft.Json, 13.0.3\"' was not present in the dictionary.
   at System.Collections.Concurrent.ConcurrentDictionary`2.ThrowKeyNotFoundException(TKey key)
   at Microsoft.DotNet.Interactive.KernelInvocationContext.ResultFor(KernelCommand command) in D:\\a\\_work\\1\\s\\src\\Microsoft.DotNet.Interactive\\KernelInvocationContext.cs:line 291
   at Microsoft.DotNet.Interactive.Kernel.SendAsync(KernelCommand command, CancellationToken cancellationToken) in D:\\a\\_work\\1\\s\\src\\Microsoft.DotNet.Interactive\\Kernel.cs:line 460
   at Microsoft.DotNet.Interactive.KernelExtensions.LoadAndRunInteractiveDocument(Kernel kernel, FileInfo file) in D:\\a\\_work\\1\\s\\src\\Microsoft.DotNet.Interactive\\KernelExtensions.cs:line 131
   at Microsoft.DotNet.Interactive.KernelExtensions.<>c__DisplayClass4_0`1.<<UseImportMagicCommand>b__0>d.MoveNext() in D:\\a\\_work\\1\\s\\src\\Microsoft.DotNet.Interactive\\KernelExtensions.cs:line 103
--- End of stack trace from previous location ---
   at System.CommandLine.Invocation.AnonymousCommandHandler.InvokeAsync(InvocationContext context)
   at System.CommandLine.Invocation.InvocationPipeline.<>c__DisplayClass4_0.<<BuildInvocationChain>b__0>d.MoveNext()
--- End of stack trace from previous location ---
[...]

PR #3292, opened 2023-10-30
#import bug

colombod:
Addressing issue #3286


src/Microsoft.DotNet.Interactive/KernelExtensions.cs
```diff
@@ -110,7 +111,8 @@ public static TKernel UseImportMagicCommand<TKernel>(this TKernel kernel)
 
     public static async Task LoadAndRunInteractiveDocument(
         this Kernel kernel,
-        FileInfo file)
+        FileInfo file,
+        KernelCommand parentCommand)
```
jonsequitur, line 115: This should probably be optional. There won't always be a parent command for this method, e.g. in automation scenarios.

jonsequitur (approved):

PR #3305, opened 2023-11-08
address comment  to remove BlobFeed override

colombod:


shyamnamboodiripad (approved):

azure-pipelines.yml
```diff
@@ -125,12 +125,11 @@ stages:
         - ${{ if and(ne(variables['System.TeamProject'], 'public'), notin(variables['Build.Reason'], 'PullRequest')) }}:
           - group: DotNet-Symbol-Server-Pats
           - name: _SignType
-            value: Real
+            value: Realb
```
shyamnamboodiripad, line 128: Was this a typo?

colombod: yes, thank you for spotting this

colombod, line 128:
```suggestion
            value: Real
```

PR #3326, opened 2023-11-16
PowerShellKernel sends CommandFailed on unrecognized command

jonsequitur:
This provides a workaround for https://github.com/PowerShell/PowerShell/issues/20079.

### DIFF
--- a/src/Microsoft.DotNet.Interactive.PowerShell/PowerShellKernel.cs
+++ b/src/Microsoft.DotNet.Interactive.PowerShell/PowerShellKernel.cs
@@ -49,6 +49,7 @@ public class PowerShellKernel :
 
     private readonly PSKernelHost _psHost;
     private readonly Lazy<PowerShell> _lazyPwsh;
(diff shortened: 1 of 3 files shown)


src/Microsoft.DotNet.Interactive.PowerShell/PowerShellKernel.cs
```diff
@@ -277,9 +278,9 @@ Task SetAsync(string name, object value, Type declaredType)
         }
         else
         {
-            RunSubmitCodeLocally(code);
+            var success = RunSubmitCodeLocally(code);
 
-            if (pwsh.HadErrors)
+            if (!success || pwsh.HadErrors)
```
shyamnamboodiripad, line 283: Consider returning `false` from `RunSubmitCodeLocally` when `pwsh.HadErrors` is `true` so that we can just check for `!success` over here.

shyamnamboodiripad (approved):

PR #3338, opened 2023-11-23
Enable C#12

colombod:
Complete .NET8 release with C#12 support


johngardner58work:
Error: (4,22): error CS1002: ; expected

colombod:
Using statement are not supported in csharp scripting, they won't work

jonsequitur:
@johngardner58work Is it your expectation that disposal would happen when the kernel is shut down (i.e. when you restart it or close the notebook)?

If so you can do this:

```csharp
Microsoft.DotNet.Interactive.Kernel.Root.RegisterForDisposal(factory);
```

jonsequitur:
> Perhaps I should repost for Csharp scripting?

The place to open this issue would be in [dotnet/roslyn](https://github.com/dotnet/roslyn/issues).

PR #3344, opened 2023-11-27
A few miscellaneous docs updates

jonsequitur:


src/polyglot-notebooks-vscode-insiders/README.md
```diff
@@ -10,18 +10,18 @@ Polyglot Notebooks are fully interoperable with Jupyter and support the `.ipynb`
...
-| F#                                          |        ✅                 |   
-| PowerShell                                  |        ✅                 |          
-| JavaScript                                  |        ✅                 |          
-| SQL                                         |        ✅                 |   
-| KQL ([Kusto Query Language](https://learn.microsoft.com/en-us/azure/data-explorer/kusto/query/))    |        ✅       |       
-| HTML                                        |        ⛔                 |     
-| [Mermaid](https://mermaid.js.org/intro/)    |        ⛔                 |        
-| Python (Preview)                            |        ✅                 |          
-| R (Preview)                                 |        ✅                 |          
+| Language                      | Variable sharing |
```
AbhitejJohn, line 13: Do we want to add http to this list as well?

jonsequitur: It's not available yet by default.

shyamnamboodiripad: I was planning a separate PR for that - that should also simplify the setup on the VS-side a bit. I will also update the readmes in that PR.

AbhitejJohn: I see, I thought we added it to the default list of kernels - maybe it was a conversation but we haven't really updated the list.

src/polyglot-notebooks-vscode-insiders/README.md
```diff
@@ -10,18 +10,18 @@ Polyglot Notebooks are fully interoperable with Jupyter and support the `.ipynb`
...
+|-------------------------------|------------------|
+| C#                            |        ✅       |
+| F#                            |        ✅       |   
+| PowerShell                    |        ✅       |          
+| JavaScript                    |        ✅       |          
+| SQL                           |        ✅       |   
+| KQL ([Kusto Query Language](https://learn.microsoft.com/en-us/azure/data-explorer/kusto/query/))    |        ✅       |  
+| [Python](docs/jupyter-in-polyglot-notebooks.md)  |        ✅       |
```
AbhitejJohn, line 21: Should we be marking these as preview features still? I understand they are shipping inbox but we've added a preview tag on them in the product, so it would be nice for these to be consistent.

src/polyglot-notebooks-vscode-insiders/README.md
shyamnamboodiripad: The assembly containing the http kernel is included by default as part of the dotnet-interactive tool - this avoids the need for a separate nuget package install for the http functionality (which was required to simplify the VS side setup). However, we haven't yet included the http kernel in the default set of kernels that the dotnet-interactive tool instantiates. As we discussed last week, we should be ready to include this by default now. I'll create a PR for this soon...

src/polyglot-notebooks-vscode-insiders/README.md
jonsequitur: There's no preview tag in the product for individual kernels.

PR #3375, opened 2023-12-06
Support for an embedded expression in variable value node

bleaphar:
In order to accommodate for some failing tests, add a test and support for the variable value node class to have the ability to parse embedded expressions


src/Microsoft.DotNet.Interactive.Http.Parsing/Parsing/HttpRequestParser.cs
```diff
@@ -131,6 +131,15 @@ public HttpSyntaxTree Parse()
                         node = new HttpVariableValueNode(_sourceText, _syntaxTree);
 
                         ParseLeadingWhitespaceAndComments(node);
+                    } 
+                    else if (IsAtStartOfEmbeddedExpression())
+                    {
+                        node = new HttpVariableValueNode(_sourceText, _syntaxTree);
+                        node.Add(ParseEmbeddedExpression());
+                        if(CurrentToken is { Kind: HttpTokenKind.NewLine })
```
shyamnamboodiripad, line 139: Is this check needed considering that the while loop condition above on line 125 already checks for this?

bleaphar: This check is important because we only want to break after parsing the embedded expression if there's a newline because {{host}}/anything should be considered valid and we'll want to consume the slash and following tokens through the while loop

shyamnamboodiripad: Won't that happen automatically even without this condition and break though? Looks like if we don't add this condition and we are at a newline, we will simply continue to the while loop condition above and break out of the loop over there. Am I missing something?

bleaphar: So it will actually be consumed into the node because the end of the is node null check it sees if there's a start of an embedded expression and if not it's consumed into the node

shyamnamboodiripad (approved):

src/Microsoft.DotNet.Interactive.Http.Tests/ParserTests.Request.cs
```diff
@@ -64,6 +64,32 @@ public void request_node_containing_only_url_and_no_variable_expressions_returns
...
+                @fakeuser=fakeuser
+                @fakepwd=fakepwd
+
+                https://httpbin.org/get
+                """);
+
+            var requestNode = result.SyntaxTree.RootNode.ChildNodes.Should().ContainSingle<HttpRequestNode>().Which;
+
+            var variableNodes = result.SyntaxTree.RootNode.ChildNodes.OfType<HttpVariableDeclarationAndAssignmentNode>();
+            variableNodes.Count().Should().Be(7);
```
shyamnamboodiripad, line 88: Should we validate the name and value of the variables?

src/Microsoft.DotNet.Interactive.Http.Tests/ParserTests.Request.cs
```diff
@@ -64,6 +64,32 @@ public void request_node_containing_only_url_and_no_variable_expressions_returns
...
+
+            requestNode.UrlNode.Text.Should().Be("https://httpbin.org/get");
```
shyamnamboodiripad, line 90: This seems redundant considering this test is mainly about variable values.

PR #3389, opened 2023-12-20
Disallow token change via SetParent after a command has been deserialized

jonsequitur:
This PR fixes an issue where a command, after deserialization, can have its parent set again, overwriting the original token and breaking scheduler correlation.


colombod:
Should we have changes in typescript too?

colombod:
#3396 will align js to this

colombod (approved):

Issue #1968, opened 2022-04-15
PostgreSQL support in SQL cells


jcmrva:
I'd like to be able to run queries against a PostgreSQL database from a SQL cell. Will that be considered?

Having found [this Stackoverflow post](https://stackoverflow.com/questions/67907347/query-postgres-database-from-a-sql-cell-in-a-net-interactive-notebook) and looked at the [SQL Server code](https://github.com/dotnet/interactive/tree/main/src/Microsoft.DotNet.Interactive.SqlServer), could `pgsql` support be bootstrapped with that code?

jonsequitur:
Yes, we are planning on adding support for PostgreSQL.

RogerDerekWeiss:
Hi there, any update on this?

IntegerMan:
This is going to be a very popular request for folks. It's understandable for this to be absent or in beta until the SQL Server support leaves beta, but I strongly expect community demand for this feature should Polyglot Notebook adoption increase significantly.

thangchung:
Is there any progress on it?

jonsequitur:
This is not currently being worked on. I would be happy to help guide a PR if someone wanted to contribute it, though.

PR #3390, opened 2023-12-21
Fix devcontainer dotnet image tag

kojo12228:
I want to pick up two mini-projects with dotnet interactive, namely support for F# 8.0 and Postgresql support #1968. My first test was to start up a the devcontainer and build the project, which failed because [the registry doesn't contain a `devcontainer/dotnet` image with tag `0-8.0`](https://mcr.microsoft.com/en-us/product/devcontainers/dotnet/tags).

This PR changes both default tag in the Dockerfile and `devcontainer.json` file to `1-8.0-bookworm-slim`, and for the `VARIANT` arg in the Dockerfile not to have the implicit prefix `0-.*`.

I had two other problems in my test, one related to VSCode not correctly handling the `.devcontainer.json` file which I was able to fix by adding an `initializeCommand` (see https://github.com/microsoft/vscode-remote-release/issues/9302#issuecomment-1859139566) and the FSharp.Core.xml file couldn't be found in the NuGet cache (cleaning the cache didn't fix the issue but manually writing the file did), I don't think either issue requires a fix in the repo.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,8 +1,8 @@
 # See here for image contents: https://github.com/microsoft/vscode-dev-containers/tree/v0.205.2/containers/dotnet/.devcontainer/base.Dockerfile
 
 # [Choice] .NET version: 6.0, 5.0, 3.1, 6.0-bullseye, 5.0-bullseye, 3.1-bullseye, 6.0-focal, 5.0-focal, 3.1-focal
-ARG VARIANT="6.0-bullseye-slim"
-FROM mcr.microsoft.com/vscode/devcontainers/dotnet:0-${VARIANT}
+ARG VARIANT="1-8.0-bookworm-slim"
+FROM mcr.microsoft.com/vscode/devcontainers/dotnet:${VARIANT}
 
 # [Choice] Node.js version: none, lts/*, 16, 14, 12, 10
 ARG NODE_VERSION="none"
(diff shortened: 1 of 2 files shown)


kojo12228:
@dotnet-policy-service agree

.devcontainer/Dockerfile
```diff
@@ -1,8 +1,8 @@
 # See here for image contents: https://github.com/microsoft/vscode-dev-containers/tree/v0.205.2/containers/dotnet/.devcontainer/base.Dockerfile
 
 # [Choice] .NET version: 6.0, 5.0, 3.1, 6.0-bullseye, 5.0-bullseye, 3.1-bullseye, 6.0-focal, 5.0-focal, 3.1-focal
-ARG VARIANT="6.0-bullseye-slim"
-FROM mcr.microsoft.com/vscode/devcontainers/dotnet:0-${VARIANT}
+ARG VARIANT="1-8.0-bookworm-slim"
+FROM mcr.microsoft.com/vscode/devcontainers/dotnet:${VARIANT}
```
WeihanLi, line 5: The image moved to `mcr.microsoft.com/devcontainers/dotnet`
```suggestion
FROM mcr.microsoft.com/devcontainers/dotnet:${VARIANT}
```

https://mcr.microsoft.com/en-us/product/devcontainers/dotnet/tags

colombod (approved):

kojo12228:
Thanks for the feedback @WeihanLi and thanks for addressing it @colombod, I forgot to come back to this!

PR #3392, opened 2023-12-22
Upgrade F# language version for 8.0

kojo12228:
Also update another reference to C# language to 12 (the other one being amended in #3338).

Because the language version in compiler args is `preview` anyway, this doesn't matter too much (although it did surprise me that it was set to preview).


jonsequitur:
Thanks, @kojo12228! This is a good catch.

The change you made doesn't affect the actual language version but it affects the language version that these two kernels advertise. There's probably a way (such as a test) to better keep these in sync in the future. 

It incidentally caused one of the contract tests to fail. 

https://github.com/dotnet/interactive/blob/e89e1d0141b775eba3036b053e3ecacbdb8b77e9/src/Microsoft.DotNet.Interactive.ApiCompatibility.Tests/ApiCompatibilityTests.jupyter_api_is_not_changed.approved.txt#L241

This value should probably not be in the constructor for `CSharpLanguageInfo` and `FSharpLanguageInfo` since it's very much a moving target. But for now I'll hand-edit the approval file which should unblock this PR, and I'll open an issue for that afterwards.

jonsequitur (approved):

PR #3399, opened 2024-01-04
Add description field to kernelInfo

colombod:
This Pr adds support for description text on `KernelInfo`.
The description property is forwarded to `ProxyKernels` too.

Removed unused code with mermaid extension.


src/Microsoft.DotNet.Interactive.CSharp/CSharpKernel.cs
```diff
@@ -61,6 +61,10 @@ public CSharpKernel(string name) : base(name)
         KernelInfo.LanguageName = "C#";
         KernelInfo.LanguageVersion = "12.0";
         KernelInfo.DisplayName = $"{KernelInfo.LocalName} - C# Script";
+        KernelInfo.Description = """
+                                 This Kernel can compile and execute C# code and display the results.
+                                 The language is C# Scripting, a dialect of C# that is used for interactive programming.
+                                 """;
```
shyamnamboodiripad, line 67: Is this displayed anywhere in the UI? Will we need to localize this description for users who are on other languages?

shyamnamboodiripad:
Should we add a description for the composite kernel created by the dotnet-interactive tool?

colombod:
No it is not displayed at all

Dr Diego Colombo
(he/him)
Principal software engineer
.NET Interactive <https://github.com/dotnet/interactive>

shyamnamboodiripad (approved):

shyamnamboodiripad:
> ... I added already the description to the composite kernel in the extension and on js and html kernels too

Ah I was referring to the composite kernel that is created in [this method](https://github.com/dotnet/interactive/blob/main/src/dotnet-interactive/CommandLine/CommandLineParser.cs#L454C3-L454C49). Not super important though - we can always add it later.

colombod:
That composite kernel (like the one in the dotnet tool) doesn't do anything special besides routing, there is no need for custom description.

PR #3407, opened 2024-01-11
Skip function types in sharing variables with Python

colombod:


colombod:
@shibbas this seems to do the trick

PR #3412, opened 2024-01-12
Restructure Variable Parsing to support underscores

bleaphar:
Made a few changes around the parser design so that underscores are supported in the name of a variable and added two tests for binding and the underscores in the variable respectively.

### DIFF
--- a/src/Microsoft.DotNet.Interactive.Http.Parsing/Parsing/HttpRequestParser.cs
+++ b/src/Microsoft.DotNet.Interactive.Http.Parsing/Parsing/HttpRequestParser.cs
@@ -186,15 +186,7 @@ internal class HttpRequestParser
 
                 while (MoreTokens())
                 {
-                    if (CurrentToken is { Text: "@" })
-                    {
-                        ConsumeCurrentTokenInto(node);
-                    }
-                    else if (CurrentToken is { Kind: HttpTokenKind.Word })
-                    {
-                        ConsumeCurrentTokenInto(node);
-                    }
-                    else if(CurrentToken is { Text: "_" })
+                    if (CurrentToken is { Text: "@" } or { Kind: HttpTokenKind.Word } or { Text: "_" })
                     {
                         ConsumeCurrentTokenInto(node);
                     }
--- a/src/Microsoft.DotNet.Interactive.Http.Parsing/Parsing/HttpRequestParser.cs
+++ b/src/Microsoft.DotNet.Interactive.Http.Parsing/Parsing/HttpRequestParser.cs
@@ -194,6 +194,10 @@ internal class HttpRequestParser
                     {
                         ConsumeCurrentTokenInto(node);
                     }
+                    else if(CurrentToken is { Text: "_" })
+                    {
+                        ConsumeCurrentTokenInto(node);
+                    }
                     else
                     {
                         break;
--- a/src/Microsoft.DotNet.Interactive.Http.Parsing/Parsing/HttpVariableDeclarationNode.cs
+++ b/src/Microsoft.DotNet.Interactive.Http.Parsing/Parsing/HttpVariableDeclarationNode.cs
@@ -17,7 +17,7 @@ internal class HttpVariableDeclarationNode : HttpSyntaxNode
     {
     }
 
-    public string VariableName => Text[1..];
+    public string VariableName => Text.Substring(1);
 
     public override IEnumerable<Diagnostic> GetDiagnostics()
     {
--- a/src/Microsoft.DotNet.Interactive.Http.Parsing/Parsing/HttpVariableDeclarationNode.cs
+++ b/src/Microsoft.DotNet.Interactive.Http.Parsing/Parsing/HttpVariableDeclarationNode.cs
@@ -17,7 +17,7 @@ internal class HttpVariableDeclarationNode : HttpSyntaxNode
     {
     }
 
-    public string VariableName => ChildTokens.Where(t => t.Kind == HttpTokenKind.Word).Select(t => t.Text).FirstOrDefault() ?? string.Empty;
+    public string VariableName => string.Join("", ChildTokens.Where(t => t.Kind == HttpTokenKind.Word || (t.Kind == HttpTokenKind.Punctuation && t.Text is "_")).Select(t => t.Text)) ?? string.Empty;
 
     public override IEnumerable<Diagnostic> GetDiagnostics()
     {
--- a/src/Microsoft.DotNet.Interactive.Http.Parsing/Parsing/HttpVariableDeclarationNode.cs
+++ b/src/Microsoft.DotNet.Interactive.Http.Parsing/Parsing/HttpVariableDeclarationNode.cs
@@ -17,7 +17,7 @@ internal class HttpVariableDeclarationNode : HttpSyntaxNode
     {
     }
 
-    public string VariableName => string.Join("", ChildTokens.Where(t => t.Kind == HttpTokenKind.Word || (t.Kind == HttpTokenKind.Punctuation && t.Text is "_")).Select(t => t.Text)) ?? string.Empty;
+    public string VariableName => Text[1..];
 
     public override IEnumerable<Diagnostic> GetDiagnostics()
     {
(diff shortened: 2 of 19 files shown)


src/Microsoft.DotNet.Interactive.Http.Parsing/Parsing/HttpVariableDeclarationNode.cs
```diff
@@ -17,7 +17,7 @@ internal HttpVariableDeclarationNode(SourceText sourceText, HttpSyntaxTree? synt
     {
     }
 
-    public string VariableName => ChildTokens.Where(t => t.Kind == HttpTokenKind.Word).Select(t => t.Text).FirstOrDefault() ?? string.Empty;
+    public string VariableName => string.Join("", ChildTokens.Where(t => t.Kind == HttpTokenKind.Word || (t.Kind == HttpTokenKind.Punctuation && t.Text is "_")).Select(t => t.Text)) ?? string.Empty;
```
shyamnamboodiripad, line 20: Isn't this just `this.Text`?

https://github.com/dotnet/interactive/blob/main/src/Microsoft.DotNet.Interactive.Http.Parsing/Parsing/HttpSyntaxNodeOrToken.cs#L39

shyamnamboodiripad: Ah looks like the above node is for the entire declaration and not just the variable name. I think it may be better to have a dedicated sub-node for the variable name (which could potentially also be used to parse the variable references that appear inside `{{}}` expressions).

src/Microsoft.DotNet.Interactive.Http.Parsing/Parsing/HttpRequestParser.cs
```diff
@@ -194,6 +194,10 @@ private HttpVariableDeclarationNode ParseVariableDeclaration()
                     {
                         ConsumeCurrentTokenInto(node);
                     }
+                    else if(CurrentToken is { Text: "_" })
+                    {
+                        ConsumeCurrentTokenInto(node);
```
shyamnamboodiripad, line 199: This addresses the variable declarations, but don't we also need to handle this inside embedded expressions (i.e. when parsing variable names that appear as part of expressions)?

src/Microsoft.DotNet.Interactive.Http.Parsing/Parsing/HttpVariableDeclarationNode.cs
bleaphar: We could create a sub-node but I think the main utility would just be getting the value of the text and reconfiguring the parser to use this new node and I'm not sure that it is necessarily cleaner and worth the rework

src/Microsoft.DotNet.Interactive.Http.Parsing/Parsing/HttpRequestParser.cs
bleaphar: Just created a test for this but we don't have to check in embedded expressions because the way it set up currently is that anything inside of the curly braces will be parsed as the expression.

src/Microsoft.DotNet.Interactive.Http.Parsing/Parsing/HttpVariableDeclarationNode.cs
shyamnamboodiripad: I may be wrong, but I think we'll need to parse variable names once we start interpreting expressions and I think we may need to start doing that for the named requests feature (to implement support for [this syntax](https://marketplace.visualstudio.com/items?itemName=humao.rest-client#:~:text=%7B%7BrequestName.(response%7Crequest).(body%7Cheaders).(*%7CJSONPath%7CXPath%7CHeader%20Name)%7D%7D) for dotting into the named requests).

We may be able to sidestep this by using a separate / simpler (regex-based?) parser just for parsing the expression contents. Not sure if that is the path we will decide to go down though...

In any case, your call to keep this simple for now seems reasonable considering that we don't need to parse variable names in other places yet.

shyamnamboodiripad, line 20: Should we report an error if a variable name contains characters like `-` that it should not contain?

shyamnamboodiripad, line 20: Looks like this could return the wrong thing if there happened to be an `_` in the value portion of a variable declaration. Could you please add a test for that? (Again, I think having a dedicated sub node for the variable name may make things easier.)

jonsequitur: I don't think a regex is the right path. The different approaches within the same parser will make it harder to understand, and the regex isn't going to support language services as well.

src/Microsoft.DotNet.Interactive.Http.Parsing/Parsing/HttpRequestParser.cs
```diff
@@ -194,6 +194,10 @@ private HttpVariableDeclarationNode ParseVariableDeclaration()
                     {
                         ConsumeCurrentTokenInto(node);
                     }
+                    else if(CurrentToken is { Text: "_" })
```
jonsequitur, line 197: Since these three branches perform the same action, you can combine them into this:

```csharp
if (CurrentToken is { Text: "@" } or { Kind: HttpTokenKind.Word } or { Text: "_" })
{
    ConsumeCurrentTokenInto(node);
}
```

src/Microsoft.DotNet.Interactive.Http.Parsing/Parsing/HttpVariableDeclarationNode.cs
jonsequitur: This suggests an alternative approach, which is to parse everything but whitespace into the variable declaration node, and then report diagnostics for invalid characters.

shyamnamboodiripad: +1 I had the same thoughts. Having a dedicated sub node could also make the implementation of features like go to definition / find all references simpler / more natural.

shyamnamboodiripad: Note that characters like `-` will be allowed on the RHS of the variable declaration (if they appear inside a string literal or embedded expression for example) which is why I think having dedicated parsing for the name portion may be helpful.

shyamnamboodiripad: Ah never mind, I was confused by the naming. I assumed the `HttpVariableDeclarationNode` encompasses the entire declaration (inlcuding the `=` and the expression on the RHS). Looks like that is the `HttpVariableDeclarationAndAssignmentNode` :)

Since `HttpVariableDeclarationNode` only includes the name portion in addition to the `@` prefix, perhaps we could just take `this.Text` (which will strip away any leading and trailing trivia) and then strip out the first `@` character.

In other words, would something like below work?

```suggestion
    public string VariableName => this.Text.SubString(1)
```

shyamnamboodiripad: Ah never mind, I was confused by the naming. I assumed the HttpVariableDeclarationNode encompasses the entire declaration (inlcuding the = and the expression on the RHS). Looks like that is the HttpVariableDeclarationAndAssignmentNode :) So please ignore my last comment above.

shyamnamboodiripad: (I also wonder if the naming could be made less confusing - but perhaps its ok since it was not as confusing once I looked at the names of all the involved node types :))

shyamnamboodiripad (approved):

PR #3418, opened 2024-01-17
SubmitCode can specify mimetype to use in trainiling expression formatting

colombod:
SubmitCode can specify a list of mimetypes to be used when formatting trailing expressions for the 'ReturnValueProduced' event

### DIFF
--- a/src/Microsoft.DotNet.Interactive.Tests/LanguageKernelTests.cs
+++ b/src/Microsoft.DotNet.Interactive.Tests/LanguageKernelTests.cs
@@ -865,6 +865,23 @@ $${languageSpecificCode}
             .Be(25);
     }
 
+    [Theory]
+    [InlineData(Language.CSharp, "20")]
+    [InlineData(Language.FSharp, "20")]
+    [InlineData(Language.CSharp, "new int[] {1,2,3}")]
+    [InlineData(Language.FSharp, "[|1;2;3|]")]
+    public async Task it_returns_a_result_with_specified_mimetype(Language language, string expression)
+    {
+        var kernel = CreateKernel(language);
+
+        var command = new SubmitCode(expression, mimeTypes: new []{JsonFormatter.MimeType});
+        await kernel.SendAsync(command);
+
+        var returnedValue =  KernelEvents.Should().ContainSingle<ReturnValueProduced>().Which;
+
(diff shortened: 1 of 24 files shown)


src/Microsoft.DotNet.Interactive.Tests/LanguageKernelTests.cs
```diff
@@ -865,6 +865,23 @@ public async Task it_returns_a_result_for_a_if_expressions(Language language, st
...
+    [Theory]
+    [InlineData(Language.CSharp, "20")]
+    [InlineData(Language.FSharp, "20")]
+    [InlineData(Language.CSharp, "new int[] {1,2,3}")]
+    [InlineData(Language.FSharp, "[|1;2;3|]")]
+    public async Task it_returns_a_result_with_specified_mimetype(Language language, string expression)
+    {
+        var kernel = CreateKernel(language);
+
+        var command = new SubmitCode(expression, mimeTypes: new []{JsonFormatter.MimeType});
```
shyamnamboodiripad, line 877: I find the API a little strange - for example the specified mime types apply only to `ReturnValueProduced` but not to other events...

colombod:
That is the point of the pr. Any other event is more explicit . This is handling the automatic formatting for return values

Dr Diego Colombo
(he/him)
Principal software engineer
.NET Interactive <https://github.com/dotnet/interactive>

shyamnamboodiripad:
> That is the point of the pr. Any other event is more explicit . This is handling the automatic formatting for return values

I am not sure I understand fully. I am probably missing something.

Currently, a `SubmitCode` command can also produce `DisplayedValueProduced` events and each such event can include multiple `FormattedValues` with different mime types. Looking at the above constructor for `SubmitCode`, it is possible to assume that the supplied `mimeTypes` argument will impact the mime types used in any resulting `DisplayedValueProduced.FormattedValues`. But that is not what will happen, and I think that can be confusing...

If the specified mime type only applies for `ReturnValueProduced` event, at the very least the parameter name should make this more explicit (e.g. `mimeTypesForReturnValue`). This could reduce the confusion, but I still find it strange that this ability to specify mime types exists for `ReturnValueProduced` but not for other events produced by the same command which can include one or more `FormattedValues`.

Also tagging @jonsequitur since you mentioned you had thoughts around the API... This seems worthy of a quick offline design discussion :)

PR #3424, opened 2024-01-23
Update kernel to support dynamic variables

bleaphar:
Added support to the parser to compute several types of additional variables in request

### DIFF
--- a/src/Microsoft.DotNet.Interactive.Http.Parsing/Parsing/HttpDiagnostics.cs
+++ b/src/Microsoft.DotNet.Interactive.Http.Parsing/Parsing/HttpDiagnostics.cs
@@ -110,7 +110,7 @@ internal static class HttpDiagnostics
         var id = $"HTTP0013";
         var severity = DiagnosticSeverity.Error;
         var messageFormat =
-            """The supplied expression '{0}' does not follow the correct pattern. The expression should adhere to the following pattern: '{{{1} |iso8601|"custom format" [offset option]}}' where offset is a valid integer and option is one of the following: ms, s, m, h, d, w, M, Q, y. See https://aka.ms/http-date-time-format for more details.""";
+            """The supplied expression '{0}' does not follow the correct pattern. The expression should adhere to the following pattern: '{{{1} [rfc1123|iso8601|"custom format"] [offset option]}}' where offset (if specified) must be a valid integer and option must be one of the following: ms, s, m, h, d, w, M, Q, y. See https://aka.ms/http-date-time-format for more details.""";
         return new HttpDiagnosticInfo(id, messageFormat, severity, expression, dateTimeType);
     }
 
--- a/src/Microsoft.DotNet.Interactive.Http.Parsing/Parsing/HttpDiagnostics.cs
+++ b/src/Microsoft.DotNet.Interactive.Http.Parsing/Parsing/HttpDiagnostics.cs
@@ -105,13 +105,22 @@ internal static class HttpDiagnostics
         return new HttpDiagnosticInfo(id, messageFormat, severity, symbol);
     }
 
-    internal static HttpDiagnosticInfo DateTimeFormatError(string expression)
+    internal static HttpDiagnosticInfo IncorrectDateTimeFormat(string expression)
     {
         var id = $"HTTP0013";
         var severity = DiagnosticSeverity.Error;
         var messageFormat =
-            """The supplied expression '{0}' does not follow the correct pattern. The expression should adhere to the following pattern: '{{$datetime |iso8601|"custom format" [offset option]}}'. See https://aka.ms/http-date-time-format for more details.""";
+            """The supplied expression '{0}' does not follow the correct pattern. The expression should adhere to the following pattern: '{{$datetime |iso8601|"custom format" [offset option]}}' where offset is a valid integer and option is one of the following: ms, s, m, h, d, w, M, Q, y. See https://aka.ms/http-date-time-format for more details.""";
         return new HttpDiagnosticInfo(id, messageFormat, severity, expression);
+    }
+
+    internal static HttpDiagnosticInfo IncorrectTimestampFormat(string timestamp)
+    {
+        var id = $"HTTP0014";
+        var severity = DiagnosticSeverity.Error;
+        var messageFormat =
+            "The supplied expression '{0}' does not follow the correct pattern. The expression should adhere to the following pattern: '{{$timestamp [offset option]}}' where offset is a valid integer and option is one of the following: ms, s, m, h, d, w, M, Q, y. See https://aka.ms/http-date-time-format for more details.";
+        return new HttpDiagnosticInfo(id, messageFormat, severity, timestamp);
     }
 
     internal static HttpDiagnosticInfo InvalidOffset(string expression, string offset)
@@ -122,30 +131,30 @@ internal static class HttpDiagnostics
         return new HttpDiagnosticInfo(id, messageFormat, severity, expression, offset);
     }
 
-    internal static HttpDiagnosticInfo TimestampFormatError(string timestamp)
+    internal static HttpDiagnosticInfo InvalidOption(string expression, string option)
     {
         var id = $"HTTP0016";
         var severity = DiagnosticSeverity.Error;
-        var messageFormat = 
-            "The supplied expression '{0}' does not follow the correct pattern. The expression should adhere to the following pattern: '{{$timestamp [offset option]}}'. See https://aka.ms/http-date-time-format for more details.";
-        return new HttpDiagnosticInfo(id, messageFormat, severity, timestamp);
+        var messageFormat = "The supplied option '{1}' in the expression '{0}' is not supported.";
+        return new HttpDiagnosticInfo(id, messageFormat, severity, expression, option);
     }
 
-    internal static HttpDiagnosticInfo InvalidOption(string expression, string option)
+    internal static HttpDiagnosticInfo IncorrectRandomIntFormat(string expression)
     {
         var id = $"HTTP0017";
         var severity = DiagnosticSeverity.Error;
-        var messageFormat = "The supplied option '{1}' in the expression '{0}' is not supported.";
-        return new HttpDiagnosticInfo(id, messageFormat, severity, expression, option);
+        var messageFormat =
+            """The supplied expression '{0}' does not follow the correct pattern. The expression should adhere to the following pattern: '{{$randomInt min max]}}' where min and max are valid integers.""";
+        return new HttpDiagnosticInfo(id, messageFormat, severity, expression);
     }
 
     internal static HttpDiagnosticInfo RandomIntMinMustNotBeGreaterThanMax(string expression, string min, string max)
     {
         var id = $"HTTP0018";
         var severity = DiagnosticSeverity.Error;
-        var messageFormat = 
+        var messageFormat =
             """The supplied argument '{1}' in the expression '{0}' must not be greater than the supplied argument '{2}'.""";
-        return new (id, messageFormat, severity, expression, min, max);
+        return new(id, messageFormat, severity, expression, min, max);
     }
 
     internal static HttpDiagnosticInfo InvalidRandomIntArgument(string expression, string argument)
@@ -154,15 +163,6 @@ internal static class HttpDiagnostics
         var severity = DiagnosticSeverity.Error;
         var messageFormat = "The supplied argument '{1}' in the expression '{0}' is not a valid integer.";
         return new HttpDiagnosticInfo(id, messageFormat, severity, expression, argument);
-    }
-
-    internal static HttpDiagnosticInfo RandomIntFormatError(string expression)
-    {
-        var id = $"HTTP0020";
-        var severity = DiagnosticSeverity.Error;
-        var messageFormat =
-            """The supplied expression '{0}' does not follow the correct pattern. The expression should adhere to the following pattern: '{{$randomInt minValue maxValue]}}'. See https://aka.ms/http-date-time-format for more details.""";
-        return new HttpDiagnosticInfo(id, messageFormat, severity, expression);
     }
 
 }
--- a/src/Microsoft.DotNet.Interactive.Http.Parsing/Parsing/HttpDiagnostics.cs
+++ b/src/Microsoft.DotNet.Interactive.Http.Parsing/Parsing/HttpDiagnostics.cs
@@ -105,13 +105,13 @@ internal static class HttpDiagnostics
         return new HttpDiagnosticInfo(id, messageFormat, severity, symbol);
     }
 
-    internal static HttpDiagnosticInfo IncorrectDateTimeFormat(string expression)
+    internal static HttpDiagnosticInfo IncorrectDateTimeFormat(string expression, string dateTimeType)
     {
         var id = $"HTTP0013";
         var severity = DiagnosticSeverity.Error;
         var messageFormat =
-            """The supplied expression '{0}' does not follow the correct pattern. The expression should adhere to the following pattern: '{{$datetime |iso8601|"custom format" [offset option]}}' where offset is a valid integer and option is one of the following: ms, s, m, h, d, w, M, Q, y. See https://aka.ms/http-date-time-format for more details.""";
-        return new HttpDiagnosticInfo(id, messageFormat, severity, expression);
+            """The supplied expression '{0}' does not follow the correct pattern. The expression should adhere to the following pattern: '{{{1} |iso8601|"custom format" [offset option]}}' where offset is a valid integer and option is one of the following: ms, s, m, h, d, w, M, Q, y. See https://aka.ms/http-date-time-format for more details.""";
+        return new HttpDiagnosticInfo(id, messageFormat, severity, expression, dateTimeType);
     }
 
     internal static HttpDiagnosticInfo IncorrectTimestampFormat(string timestamp)
@@ -119,7 +119,7 @@ internal static class HttpDiagnostics
         var id = $"HTTP0014";
         var severity = DiagnosticSeverity.Error;
         var messageFormat =
-            "The supplied expression '{0}' does not follow the correct pattern. The expression should adhere to the following pattern: '{{$timestamp [offset option]}}' where offset is a valid integer and option is one of the following: ms, s, m, h, d, w, M, Q, y. See https://aka.ms/http-date-time-format for more details.";
+            "The supplied expression '{0}' does not follow the correct pattern. The expression should adhere to the following pattern: '{{$timestamp [offset option]}}' where offset (if specified) must be a valid integer and option must be one of the following: ms, s, m, h, d, w, M, Q, y. See https://aka.ms/http-date-time-format for more details.";
         return new HttpDiagnosticInfo(id, messageFormat, severity, timestamp);
     }
 
@@ -144,7 +144,7 @@ internal static class HttpDiagnostics
         var id = $"HTTP0017";
         var severity = DiagnosticSeverity.Error;
         var messageFormat =
-            """The supplied expression '{0}' does not follow the correct pattern. The expression should adhere to the following pattern: '{{$randomInt min max]}}' where min and max are valid integers.""";
+            """The supplied expression '{0}' does not follow the correct pattern. The expression should adhere to the following pattern: '{{$randomInt [min] [max]]}}' where min and max (if specified) must be valid integers.""";
         return new HttpDiagnosticInfo(id, messageFormat, severity, expression);
     }
 
--- a/src/Microsoft.DotNet.Interactive.Http.Parsing/Parsing/HttpDiagnostics.cs
+++ b/src/Microsoft.DotNet.Interactive.Http.Parsing/Parsing/HttpDiagnostics.cs
@@ -97,11 +97,11 @@ internal static class HttpDiagnostics
         return new HttpDiagnosticInfo(id, messageFormat, severity);
     }
 
-    internal static HttpDiagnosticInfo CannotResolveSymbol(string symbol)
+    internal static HttpDiagnosticInfo UnableToEvaluateExpression(string symbol)
     {
         var id = $"HTTP0012";
         var severity = DiagnosticSeverity.Error;
-        var messageFormat = "Cannot resolve symbol '{0}'.";
+        var messageFormat = "Unable to evaluate expression '{0}'.";
         return new HttpDiagnosticInfo(id, messageFormat, severity, symbol);
     }
 
@@ -109,13 +109,9 @@ internal static class HttpDiagnostics
     {
         var id = $"HTTP0013";
         var severity = DiagnosticSeverity.Error;
-        var messageFormat = 
+        var messageFormat =
             """
-            Error in pattern.
-
-            Usage: {{{0} rfc1123|iso8601|"custom format" [offset option]}}
-
-            You can specify a date time relative to the current date like: {{{0} rfc1123 3 M}} to represent 3 months later in RFC1123 format.
+            The supplied argument "xyz" in the expression '{0}' does not follow the correct pattern. The argument should adhere to the following pattern: '{{0} |iso8601|"custom format" [offset option]}}'. See https://aka.ms/http-date-time-format for more details.
             """;
         return new HttpDiagnosticInfo(id, messageFormat, severity, datetime);
     }
@@ -140,25 +136,18 @@ internal static class HttpDiagnostics
     {
         var id = $"HTTP0016";
         var severity = DiagnosticSeverity.Error;
-        var messageFormat = """
-            Error in pattern.
-
-            Usage: {{$timestamp [offset option]}}
-
-            You can specify an offset relative to the current timestamp like: {{$timestamp 3 M}} to represent a timestamp 3 months later.
-            """;
+        var messageFormat = 
+            "The supplied argument in the expression '{0}' does not follow the correct pattern. The argument should adhere to the following pattern '{{$timestamp [offset option]}}'. See https://aka.ms/http-date-time-format for more details.";
         return new HttpDiagnosticInfo(id, messageFormat, severity, timestamp);
     }
 
-    public static HttpDiagnosticInfo RandomIntMustBeGreaterThanOrEqualToZero() {
+    internal static HttpDiagnosticInfo RandomIntMustBeGreaterThanOrEqualToZero() {
         var id = $"HTTP0017";
         var severity = DiagnosticSeverity.Error;
-        var messageFormat = """
-            Parameter must be greater than or equal to 0.
-            """;
+        var messageFormat = "Parameter must be greater than or equal to 0.";
         return new HttpDiagnosticInfo(id, messageFormat, severity);
     }
-    public static HttpDiagnosticInfo RandomIntMustBeIntegerArgument(string value)
+    internal static HttpDiagnosticInfo RandomIntMustBeIntegerArgument(string value)
     {
         var id = $"HTTP0018";
         var severity = DiagnosticSeverity.Error;
@@ -167,19 +156,21 @@ internal static class HttpDiagnostics
             """;
         return new (id, messageFormat, severity, value);
     } 
-    public static HttpDiagnosticInfo RandomIntInvalidArguments(string max, string min)
+    internal static HttpDiagnosticInfo RandomIntInvalidArguments(string max, string min)
     {
         var id = $"HTTP0019";
         var severity = DiagnosticSeverity.Error;
-        var messageFormat = """
+        var messageFormat = 
+            """
             The parameter "{0}" must not be greater than the parameter "{1}".
             """;
         return new (id, messageFormat, severity, [max, min]);
     } 
-    public static HttpDiagnosticInfo NoPatternMatch(string text, string pattern) {
+    internal static HttpDiagnosticInfo NoPatternMatch(string text, string pattern) {
         var id = $"HTTP0020";
         var severity = DiagnosticSeverity.Error;
-        var messageFormat = """
+        var messageFormat = 
+            """
             The text "{0}" did not match the pattern, "{1}".
             """;
         return new(id,messageFormat, severity, [text, pattern]);
--- a/src/Microsoft.DotNet.Interactive.Http.Parsing/Parsing/HttpDiagnostics.cs
+++ b/src/Microsoft.DotNet.Interactive.Http.Parsing/Parsing/HttpDiagnostics.cs
@@ -4,6 +4,8 @@
 #nullable enable
 
 using Microsoft.CodeAnalysis;
+using Microsoft.VisualBasic;
+using System.CommandLine;
 
 namespace Microsoft.DotNet.Interactive.Http.Parsing;
 
@@ -104,4 +106,85 @@ internal static class HttpDiagnostics
         var messageFormat = "Cannot resolve symbol '{0}'.";
         return new HttpDiagnosticInfo(id, messageFormat, severity, symbol);
     }
+
+    internal static HttpDiagnosticInfo DateTimePatternMatchError(string datetime)
+    {
+        var id = $"HTTP0013";
+        var severity = DiagnosticSeverity.Error;
+        var messageFormat = 
+            """
+            Error in pattern.
+
+            Usage: {{{0} rfc1123|iso8601|"custom format" [offset option]}}
+
+            You can specify a date time relative to the current date like: {{{0} rfc1123 3 M}} to represent 3 months later in RFC1123 format.
+            """;
+        return new HttpDiagnosticInfo(id, messageFormat, severity, datetime);
+    }
+
+    internal static HttpDiagnosticInfo InvalidFormat(string format)
+    {
+        var id = $"HTTP0014";
+        var severity = DiagnosticSeverity.Error;
+        var messageFormat = "The format string '{0}' is invalid.";
+        return new HttpDiagnosticInfo(id, messageFormat, severity, format);
+    }
+
+    internal static HttpDiagnosticInfo InvalidOffset(string offset)
+    {
+        var id = $"HTTP0015";
+        var severity = DiagnosticSeverity.Error;
+        var messageFormat = "An error occurred while applying the offset '{0}'.";
+        return new HttpDiagnosticInfo(id, messageFormat, severity, offset);
+    }
+
+    internal static HttpDiagnosticInfo TimestampFormatError(string timestamp)
+    {
+        var id = $"HTTP0016";
+        var severity = DiagnosticSeverity.Error;
+        var messageFormat = """
+            Error in pattern.
+
+            Usage: {{$timestamp [offset option]}}
+
+            You can specify an offset relative to the current timestamp like: {{$timestamp 3 M}} to represent a timestamp 3 months later.
+            """;
+        return new HttpDiagnosticInfo(id, messageFormat, severity, timestamp);
+    }
+
+    public static HttpDiagnosticInfo RandomIntMustBeGreaterThanOrEqualToZero() {
+        var id = $"HTTP0017";
+        var severity = DiagnosticSeverity.Error;
+        var messageFormat = """
+            Parameter must be greater than or equal to 0.
+            """;
+        return new HttpDiagnosticInfo(id, messageFormat, severity);
+    }
+    public static HttpDiagnosticInfo RandomIntMustBeIntegerArgument(string value)
+    {
+        var id = $"HTTP0018";
+        var severity = DiagnosticSeverity.Error;
+        var messageFormat = """
+            The parameter "{0}" must be a valid integer value.
+            """;
+        return new (id, messageFormat, severity, value);
+    } 
+    public static HttpDiagnosticInfo RandomIntInvalidArguments(string max, string min)
+    {
+        var id = $"HTTP0019";
+        var severity = DiagnosticSeverity.Error;
+        var messageFormat = """
+            The parameter "{0}" must not be greater than the parameter "{1}".
+            """;
+        return new (id, messageFormat, severity, [max, min]);
+    } 
+    public static HttpDiagnosticInfo NoPatternMatch(string text, string pattern) {
+        var id = $"HTTP0020";
+        var severity = DiagnosticSeverity.Error;
+        var messageFormat = """
+            The text "{0}" did not match the pattern, "{1}".
+            """;
+        return new(id,messageFormat, severity, [text, pattern]);
+    }
+
 }
--- a/src/Microsoft.DotNet.Interactive.Http.Parsing/Parsing/HttpDiagnostics.cs
+++ b/src/Microsoft.DotNet.Interactive.Http.Parsing/Parsing/HttpDiagnostics.cs
@@ -105,31 +105,21 @@ internal static class HttpDiagnostics
         return new HttpDiagnosticInfo(id, messageFormat, severity, symbol);
     }
 
-    internal static HttpDiagnosticInfo DateTimePatternMatchError(string datetime)
+    internal static HttpDiagnosticInfo DateTimeFormatError(string expression)
     {
         var id = $"HTTP0013";
         var severity = DiagnosticSeverity.Error;
         var messageFormat =
-            """
-            The supplied argument "xyz" in the expression '{0}' does not follow the correct pattern. The argument should adhere to the following pattern: '{{0} |iso8601|"custom format" [offset option]}}'. See https://aka.ms/http-date-time-format for more details.
-            """;
-        return new HttpDiagnosticInfo(id, messageFormat, severity, datetime);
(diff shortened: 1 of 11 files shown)


src/Microsoft.DotNet.Interactive.Http.Tests/HttpKernelTests.cs
```diff
@@ -320,6 +320,214 @@ public async Task binding_in_variable_value_is_valid()
...
+            POST https://api.example.com/comments HTTP/1.1
+            Content-Type: application/xml
+
+            {
+                "request_id": "{{$guid}}"
+            }
+            """;
+
+        var result = await kernel.SendAsync(new SubmitCode(code));
+        result.Events.Should().NotContainErrors();
```
jonsequitur, line 349: This could use an additional assertion that the we did in fact bind a guid.

src/Microsoft.DotNet.Interactive.Http.Tests/HttpKernelTests.cs
```diff
@@ -320,6 +320,214 @@ public async Task binding_in_variable_value_is_valid()
...
+            POST https://api.example.com/comments HTTP/1.1
+            Content-Type: application/xml
+            
+            {
+                "updated_at": "{{$timestamp}}"
+            }
+            """;
+
+        var result = await kernel.SendAsync(new SubmitCode(code));
+        result.Events.Should().NotContainErrors();
```
jonsequitur, line 377: Ditto. We should verify the binding.

src/Microsoft.DotNet.Interactive.Http.Tests/HttpKernelTests.cs
```diff
@@ -320,6 +320,214 @@ public async Task binding_in_variable_value_is_valid()
...
+            POST https://api.example.com/comments HTTP/1.1
+            Content-Type: application/xml
+            
+            {
+                "created_at": "{{$timestamp -1 d}}"
+            }
+            """;
+
+        var result = await kernel.SendAsync(new SubmitCode(code));
+        result.Events.Should().NotContainErrors();
```
jonsequitur, line 406: Same.

src/Microsoft.DotNet.Interactive.Http.Tests/HttpKernelTests.cs
```diff
@@ -320,6 +320,214 @@ public async Task binding_in_variable_value_is_valid()
...
+            POST https://api.example.com/comments HTTP/1.1
+            Content-Type: application/xml
+            
+            {
+                "review_count": "{{$randomInt 5 200}}"
+            }
+            """;
+
+        var result = await kernel.SendAsync(new SubmitCode(code));
+        result.Events.Should().NotContainErrors();
```
jonsequitur, line 435: Same.

src/Microsoft.DotNet.Interactive.Http.Tests/HttpKernelTests.cs
```diff
@@ -320,6 +320,214 @@ public async Task binding_in_variable_value_is_valid()
...
+            POST https://api.example.com/comments HTTP/1.1
+            Content-Type: application/xml
+            
+            {
+                "custom_date": "{{$datetime 'YYYY-MM-DD'}}"
+            }
+            """;
+
+        var result = await kernel.SendAsync(new SubmitCode(code));
+        result.Events.Should().NotContainErrors();
```
jonsequitur, line 464: Same.

src/Microsoft.DotNet.Interactive.Http/HttpKernel.cs
```diff
@@ -316,8 +319,198 @@ private List<Diagnostic> GetAllDiagnostics(HttpRequestParseResult parseResult)
...
+            return true;
+        }
+        return false;
+    }
+
+    protected string ParseTimestamp(string expression, MatchCollection matches)
+    {
+        var text = expression;
+        var currentDateTimeOffset = DateTimeOffset.UtcNow;
+        for (int i = matches.Count - 1; i >= 0; i--)
```
shyamnamboodiripad, line 374: Why do we expect more than one match? Each embedded expression i.e. `{{ }}` only contain at most one expression, isn't it?

src/Microsoft.DotNet.Interactive.Http/HttpKernel.cs
```diff
@@ -316,8 +319,198 @@ private List<Diagnostic> GetAllDiagnostics(HttpRequestParseResult parseResult)
...
+                {
+                    // If we can't parse the offset, just leave the variable in place
+                    // the error will be reported to the parser via "CanParse"
+                }
+            }
+        }
+        return text;
+    }
+
+    protected string ParseDateTime(string expression, MatchCollection matches, Regex pattern)
```
shyamnamboodiripad, line 401: naming: Since the parsing is happening in the regex, perhaps these should not be prefixed with `Parse`. That would also avoid confusion between the existing `Parse` methods that follow a particular pattern for their signature and the new ones which are based on regex. Perhaps we can use the prefix `Get` instead (i.e., `GetRandomInt`, `GetDateTime` etc.)?

src/Microsoft.DotNet.Interactive.Http/HttpKernel.cs
```diff
@@ -316,8 +319,198 @@ private List<Diagnostic> GetAllDiagnostics(HttpRequestParseResult parseResult)
...
+            return true;
+        }else if(timestampMatches.Count > 0)
+        {
+            expressionResolution = ParseTimestamp(expression, timestampMatches);
+            return true;
+        }
+        return false;
+    }
+
+    protected string ParseTimestamp(string expression, MatchCollection matches)
```
shyamnamboodiripad, line 370: The methods should probably be `private` as opposed to `protected`

src/Microsoft.DotNet.Interactive.Http/HttpKernel.cs
```diff
@@ -316,8 +319,198 @@ private List<Diagnostic> GetAllDiagnostics(HttpRequestParseResult parseResult)
...
+                        && offset != 0)
+                    {
+                        dateTimeOffset = dateTimeOffset.AddOffset(offset, match.Groups["option"].Value);
+                    }
+
+                    text = string.Concat(text.Substring(0, match.Index), dateTimeOffset.ToUnixTimeSeconds().ToString(), text.Substring(match.Index + match.Value.Length));
+                }
+                catch (Exception)
+                {
+                    // If we can't parse the offset, just leave the variable in place
```
shyamnamboodiripad, line 393: Consider delegating the binding failure result creation down to these functions. For example, this function could return `node.CreateBindingSuccess(text);` in the success case and `node.CreateBindingFailure(HttpDiagnostics.SomeDiagnostic(ex.Message));` in the failure case.

src/Microsoft.DotNet.Interactive.Http/HttpKernel.cs
```diff
@@ -316,8 +319,198 @@ private List<Diagnostic> GetAllDiagnostics(HttpRequestParseResult parseResult)
...
+                    // the error will be reported to the parser via "CanParse"
```
shyamnamboodiripad, line 394: nit: Comments probably need to be cleaned up to remove references to functions that don't exist in the current repo.

src/Microsoft.DotNet.Interactive.Http/HttpKernel.cs
```diff
@@ -316,8 +319,198 @@ private List<Diagnostic> GetAllDiagnostics(HttpRequestParseResult parseResult)
...
+                        format = "o";
+                    }
+                    else
+                    {
+                        format = type.Value.Substring(1, type.Value.Length - 2);
+                    }
+
+                    text = string.Concat(text.Substring(0, match.Index), dateTimeOffset.ToString(format, formatProvider), text.Substring(match.Index + match.Value.Length));
+                }
+                catch (Exception)
```
jonsequitur, line 448: It's preferable to catch the specific exception types that we think are possible here. Any other exceptions thrown here are a bug that the current approach would hide.

src/Microsoft.DotNet.Interactive.Http.Tests/HttpKernelTests.cs
```diff
@@ -320,6 +321,250 @@ public async Task binding_in_variable_value_is_valid()
...
+
+        var currentDate = DateTime.UtcNow.ToString("yyyy-MM-dd");
+        var bodyAsString = await request.Content.ReadAsStringAsync();
+        bodyAsString.Should().BeExceptingWhitespace($$"""
+            {
+                "custom_date" : "{{currentDate}}"
+            }
+            """);
```
shyamnamboodiripad, line 494: Consider: Split and compare just the substring after the ':' like in the above text instead (so that we can compare just the date).

src/Microsoft.DotNet.Interactive.Http.Tests/HttpKernelTests.cs
```diff
@@ -320,6 +321,250 @@ public async Task binding_in_variable_value_is_valid()
...
+                "local_custom_date" : "{{$localDatetime 'yyyy-MM-dd'}}"
+            }
+            """;
+
+        var result = await kernel.SendAsync(new SubmitCode(code));
+        result.Events.Should().NotContainErrors();
+
+        var currentDate = DateTime.UtcNow.ToString("yyyy-MM-dd");
+        var bodyAsString = await request.Content.ReadAsStringAsync();
+        bodyAsString.Should().BeExceptingWhitespace($$"""
```
shyamnamboodiripad, line 527: Consider: Similar comment about splitting and comparing just the date part so that you can directly compare with the expected date as opposed to creating a string with the expected date embedded and compating the string.

src/Microsoft.DotNet.Interactive.Http.Tests/HttpKernelTests.cs
```diff
@@ -320,6 +321,250 @@ public async Task binding_in_variable_value_is_valid()
...
+            POST https://api.example.com/comments HTTP/1.1
+            Content-Type: application/xml
+            
+            {
+                "request_id": "{{$guid}}",
+                "updated_at": "{{$timestamp}}",
+                "created_at": "{{$timestamp -1 d}}",
+                "review_count": "{{$randomInt 5 200}}",
+                "custom_date": "{{$datetime 'yyyy-MM-dd'}}",
+                "local_custom_date": "{{$localDatetime 'yyyy-MM-dd'}}"
```
shyamnamboodiripad, line 559: Consider sprinkling some of these in other parts of request (e.g. URL, headers etc.)

src/Microsoft.DotNet.Interactive.Http.Tests/HttpKernelTests.cs
```diff
@@ -347,6 +592,67 @@ public async Task url_in_embedded_expression_is_valid()
...
+            Content-Type: application/xml
+            
+            {
+                "local_custom_date": "{{$localDatetime YYYY-MM-DD'}}"
+            }
+            """;
+
+        var result = await kernel.SendAsync(new SubmitCode(code));
+
+        result.Events.First().Equals(HttpDiagnostics.DateTimePatternMatchError("localDatetime YYYY-MM-DD"));
```
shyamnamboodiripad, line 652: Looks like not all of the new diagnostics in `HttpDiagnostics` have associated tests. Could we add (at least) one test for each error that the kernel catches and report to ensure that the corresponding code paths are covered?

src/Microsoft.DotNet.Interactive.Http/HttpKernelExtension.cs
```diff
@@ -34,3 +35,23 @@ public class HttpKernelExtension
         }
     }
 }
+
+internal static class DateTimeOffsetExtensions
+{
+    public static DateTimeOffset AddOffset(this DateTimeOffset dateTime, int offset, string options)
```
shyamnamboodiripad, line 41: This extension method in this file is intended for use by consumers who load the http kernel via a .dib extension load 
script.

Could you please move this helper as a private (or nested) function inside HttpKernel itself?

shyamnamboodiripad: Alternatively, you could create a new file `DateTimeUtilities.cs` in the current project and move it there.

src/Microsoft.DotNet.Interactive.Http/HttpKernelExtension.cs
```diff
@@ -34,3 +35,23 @@ public class HttpKernelExtension
...
+            "ms" => dateTime.AddMilliseconds(offset),
+            "s" => dateTime.AddSeconds(offset),
+            "m" => dateTime.AddMinutes(offset),
+            "h" => dateTime.AddHours(offset),
+            "d" => dateTime.AddDays(offset),
+            "w" => dateTime.AddDays(offset * 7),
+            "M" => dateTime.AddMonths(offset),
+            "Q" => dateTime.AddMonths(offset * 3),
+            "y" => dateTime.AddYears(offset),
+            _ => throw new ArgumentException("Invalid offset option", nameof(options))
```
shyamnamboodiripad, line 54: minor: Consider renaming `options` to `offsetFormatString`.

```suggestion
            _ => throw new ArgumentException("Invalid offset format string", nameof(offsetFormatString))
```

src/Microsoft.DotNet.Interactive.Http/HttpKernelExtension.cs
shyamnamboodiripad: (It is also generally preferred to have just a single type per C# code file in the current repo)

src/Microsoft.DotNet.Interactive.Http/HttpKernel.cs
```diff
@@ -318,6 +321,209 @@ private List<Diagnostic> GetAllDiagnostics(HttpRequestParseResult parseResult)
...
 
-        return node.CreateBindingFailure(HttpDiagnostics.CannotResolveSymbol(variableName));
+        var bindingResult = MatchExpressionValue(node, expression);
+        if ( bindingResult is not null)
+        {
+            return bindingResult;
+        } else
+        {
+            return node.CreateBindingFailure(HttpDiagnostics.CannotResolveSymbol(variableName));
+        }
```
shyamnamboodiripad, line 331: This block could  be simplified to -
```
else
{
    return MatchExpressionValue(node, expression) ?? node.CreateBindingFailure(HttpDiagnostics.CannotResolveSymbol(variableName));
}
```

src/Microsoft.DotNet.Interactive.Http/HttpKernel.cs
```diff
@@ -318,6 +321,209 @@ private List<Diagnostic> GetAllDiagnostics(HttpRequestParseResult parseResult)
...
         }
 
-        return node.CreateBindingFailure(HttpDiagnostics.CannotResolveSymbol(variableName));
+        var bindingResult = MatchExpressionValue(node, expression);
+        if ( bindingResult is not null)
+        {
+            return bindingResult;
+        } else
+        {
+            return node.CreateBindingFailure(HttpDiagnostics.CannotResolveSymbol(variableName));
```
shyamnamboodiripad, line 330: Since we now support more than just simple variables / symbols inside embedded expression blocks, perhaps we should rename `HttpDiagnostics.CannotResolveSymbol` to `HttpDiagnostics.UnableToEvaluateExpression`. Could you please update the function name (and update the diagnostic message inside the function to match the new name)?

src/Microsoft.DotNet.Interactive.Http/HttpKernel.cs
```diff
@@ -318,6 +321,209 @@ private List<Diagnostic> GetAllDiagnostics(HttpRequestParseResult parseResult)
...
+        {
+            return GetDateTime(node, expression, localDateTimeMatches, localDateTimePattern);
+        } else if(randomIntMatches.Count > 0)
+        {
+            return GetRandInt(node, expression, randomIntMatches);
+        }else if(timestampMatches.Count > 0)
+        {
+            return GetTimestamp(node, expression, timestampMatches);
+        }
+        return null;
```
shyamnamboodiripad, line 367: consider: It may be better if this function also returned the `node.CreateBindingFailure(HttpDiagnostics.UnableToEvaluateExpression(expression));` instead of `null`. That way you could make the return type non-nullable and the caller can just say `return MatchExpressionValue(node, expression);`

src/Microsoft.DotNet.Interactive.Http/HttpKernel.cs
```diff
@@ -318,6 +321,209 @@ private List<Diagnostic> GetAllDiagnostics(HttpRequestParseResult parseResult)
...
     }
+
+    private HttpBindingResult<object?>? MatchExpressionValue(HttpExpressionNode node, string expression)
+    {
+
+        var guidPattern = new Regex(@$"\{"$guid"}", RegexOptions.Compiled);
+        var dateTimePattern = new Regex(@$"\{"$datetime"}\s(?<type>rfc1123|iso8601|'.+'|"".+"")(?:\s(?<offset>-?\d+)\s(?<option>y|M|Q|w|d|h|m|s|ms))?", RegexOptions.Compiled);
+        var localDateTimePattern = new Regex(@$"\{"$localDatetime"}\s(?<type>rfc1123|iso8601|'.+'|"".+"")(?:\s(?<offset>-?\d+)\s(?<option>y|M|Q|w|d|h|m|s|ms))?", RegexOptions.Compiled);
+        var randomIntPattern = new Regex(@$"\{"$randomInt"}(?:\s(?<parameters>-?\d+)){{0,2}}", RegexOptions.Compiled);
+        var timestampPattern = new Regex(@$"\{"$timestamp"}(?:\s(?<offset>-?\d+)\s(?<option>y|M|Q|w|d|h|m|s|ms))?", RegexOptions.Compiled);
```
shyamnamboodiripad, line 343: qq: `$guid` and `$GUID` would mean different things, would they not? In other words, variables are always case-sensitive, aren't they?

src/Microsoft.DotNet.Interactive.Http/HttpKernel.cs
```diff
@@ -318,6 +321,209 @@ private List<Diagnostic> GetAllDiagnostics(HttpRequestParseResult parseResult)
...
+
+        var guidMatches = guidPattern.Matches(expression);
+        var dateTimeMatches = dateTimePattern.Matches(expression);
+        var localDateTimeMatches = localDateTimePattern.Matches(expression);
+        var randomIntMatches = randomIntPattern.Matches(expression);
+        var timestampMatches = timestampPattern.Matches(expression);
+
+        if (guidMatches.Count > 0)
```
shyamnamboodiripad, line 351: Shouldn't we check for exactly one match in all the below cases?

For example, is this legal `{{$guid $guid}}`? Looks like this would end up succeeding currently.

It would be great to add some negative tests for multiple matches.

src/Microsoft.DotNet.Interactive.Http/HttpKernel.cs
shyamnamboodiripad, line 343: Please extract a string constant for `(?:\s(?<offset>-?\d+)\s(?<option>y|M|Q|w|d|h|m|s|ms))?` since it is repeated in 3 of the regexes above.

src/Microsoft.DotNet.Interactive.Http/HttpKernel.cs
```diff
@@ -318,6 +321,209 @@ private List<Diagnostic> GetAllDiagnostics(HttpRequestParseResult parseResult)
...
+        var dateTimePattern = new Regex(@$"\{"$datetime"}\s(?<type>rfc1123|iso8601|'.+'|"".+"")(?:\s(?<offset>-?\d+)\s(?<option>y|M|Q|w|d|h|m|s|ms))?", RegexOptions.Compiled);
+        var localDateTimePattern = new Regex(@$"\{"$localDatetime"}\s(?<type>rfc1123|iso8601|'.+'|"".+"")(?:\s(?<offset>-?\d+)\s(?<option>y|M|Q|w|d|h|m|s|ms))?", RegexOptions.Compiled);
+        var randomIntPattern = new Regex(@$"\{"$randomInt"}(?:\s(?<parameters>-?\d+)){{0,2}}", RegexOptions.Compiled);
+        var timestampPattern = new Regex(@$"\{"$timestamp"}(?:\s(?<offset>-?\d+)\s(?<option>y|M|Q|w|d|h|m|s|ms))?", RegexOptions.Compiled);
+
+        var guidMatches = guidPattern.Matches(expression);
+        var dateTimeMatches = dateTimePattern.Matches(expression);
+        var localDateTimeMatches = localDateTimePattern.Matches(expression);
+        var randomIntMatches = randomIntPattern.Matches(expression);
+        var timestampMatches = timestampPattern.Matches(expression);
```
shyamnamboodiripad, line 349: It may be more efficient to only do a match if a previous one fails. Otherwise, we will end up doing redundant matches even if the first one we tried succeeds.

```
var guidMatches = guidPattern.Matches(expression);
if (guidMatches.Count == 1)
{
    return node.CreateBindingSuccess(Guid.NewGuid().ToString());
}
else if (guidMatches.Count > 0)
{
    return node.CreateBindingFailure(HttpDiagnostics.UnableToEvaluateExpression(expression));
}

var dateTimeMatches = dateTimePattern.Matches(expression);
if (dateTimeMatches.Count == 1)
{
    return GetDateTime(node, expression, dateTimeMatches, dateTimePattern);
}
else if (dateTimeMatches.Count > 0)
{
    return node.CreateBindingFailure(HttpDiagnostics.UnableToEvaluateExpression(expression));
}

...

You could sort the checks such that we try to match the most commonly used types of variables first.

File: src/Microsoft.DotNet.Interactive.Http/HttpKernel.cs [hunk 24]
```diff
@@ -318,6 +321,209 @@ private List<Diagnostic> GetAllDiagnostics(HttpRequestParseResult parseResult)
...
+        } else
+        {
+            return node.CreateBindingFailure(HttpDiagnostics.TimestampFormatError(expression));
+        }
+        
+    }
+
+    private HttpBindingResult<object?> GetDateTime(HttpExpressionNode node, string expression, MatchCollection matches, Regex pattern)
+    {
+        var text = expression;
```
shyamnamboodiripad, line 404: nit: Avoid the new var is redundant. I would recommend just using the parameter `expression` - you could rename the parameter to `expressionText` or `dateTimeExpressionText` if you want for clarity.

src/Microsoft.DotNet.Interactive.Http/HttpKernel.cs
```diff
@@ -318,6 +321,209 @@ private List<Diagnostic> GetAllDiagnostics(HttpRequestParseResult parseResult)
...
+        }else if(timestampMatches.Count > 0)
+        {
+            return GetTimestamp(node, expression, timestampMatches);
+        }
+        return null;
+    }
+
+    private HttpBindingResult<object?> GetTimestamp(HttpExpressionNode node, string expression, MatchCollection matches)
+    {
+        var text = expression;
```
shyamnamboodiripad, line 372: nit: The new var is redundant. I would recommend just using the parameter `expression` - you could rename the parameter to `expressionText` or `dateTimeExpressionText` if you want for clarity.

src/Microsoft.DotNet.Interactive.Http/HttpKernel.cs
```diff
@@ -318,6 +321,209 @@ private List<Diagnostic> GetAllDiagnostics(HttpRequestParseResult parseResult)
...
+        var currentDateTimeOffset = DateTimeOffset.UtcNow;
+        
+        DateTimeOffset dateTimeOffset = currentDateTimeOffset;
```
shyamnamboodiripad, line 375: The second `currentDateTimeOffset` variable also seems redundant. Please consider deleting this (and renaming the other variable to whichever name seems more preferrable).

src/Microsoft.DotNet.Interactive.Http/HttpKernel.cs
```diff
@@ -318,6 +321,209 @@ private List<Diagnostic> GetAllDiagnostics(HttpRequestParseResult parseResult)
...
+    private HttpBindingResult<object?> GetDateTime(HttpExpressionNode node, string expression, MatchCollection matches, Regex pattern)
+    {
+        var text = expression;
+        var currentDateTimeOffset = DateTimeOffset.UtcNow;
+
+        // We just pre-matched the prefix, so we need to re-match the full pattern to get the type, offset, and option
+        matches = pattern.Matches(text);
+
+        string format;
+        DateTimeOffset dateTimeOffset = currentDateTimeOffset;
```
shyamnamboodiripad, line 411: The `currentDateTimeOffset` variable also seems redundant. Please consider deleting this (and renaming the remaining `dateTimeOffset` variable to whichever name seems more preferrable).

src/Microsoft.DotNet.Interactive.Http/HttpKernel.cs
shyamnamboodiripad: I think returning `null` here feels all the more odd considering that the called `Get*` functions already always return a non-nullable result that can be either success or failure.

src/Microsoft.DotNet.Interactive.Http/HttpKernel.cs
```diff
@@ -318,6 +321,209 @@ private List<Diagnostic> GetAllDiagnostics(HttpRequestParseResult parseResult)
...
+    {
+
+        Random random = new();
+        
+        var match = matches.FirstOrDefault();
+        if(match != null)
+        {
+            (int? Min, int? Max)? parameters = ParseParametersFromMatch(match);
+
+            if (parameters is not null)
```
shyamnamboodiripad, line 476: nit: I am probably being pedantic, but I think you may want to use the term `argument(s)` instead of `parameter(s)` here for the variable name. See https://stackoverflow.com/questions/156767/whats-the-difference-between-an-argument-and-a-parameter for some discussion around the distinction.

src/Microsoft.DotNet.Interactive.Http/HttpKernel.cs
```diff
@@ -318,6 +321,209 @@ private List<Diagnostic> GetAllDiagnostics(HttpRequestParseResult parseResult)
...
+
+            return node.CreateBindingSuccess(text);
+        } else
+        {
+            return node.CreateBindingFailure(HttpDiagnostics.CannotResolveSymbol(text));
+        }
+        
+    }
+
+    internal static (int? min, int? max)? ParseParametersFromMatch(Match match)
```
shyamnamboodiripad, line 503: nit: Consider renaming to `ParseArgumentsFromMatch` as discussed in the baove comment.

shyamnamboodiripad, line 503: Consider making this a nested function inside `GetRandInt` above since it is not used for any other case.

src/Microsoft.DotNet.Interactive.Http/HttpKernel.cs
```diff
@@ -318,6 +321,209 @@ private List<Diagnostic> GetAllDiagnostics(HttpRequestParseResult parseResult)
...
+                {
+                    text = string.Concat(text.Substring(0, match.Index), random.Next().ToString(), text.Substring(match.Index + match.Value.Length));
+                }
+                else if (!min.HasValue && max.HasValue)
+                {
+                    text = string.Concat(text.Substring(0, match.Index), random.Next(max.Value).ToString(), text.Substring(match.Index + match.Value.Length));
+                }
+                else if (min.HasValue && max.HasValue)
+                {
+                    text = string.Concat(text.Substring(0, match.Index), random.Next(min.Value, max.Value).ToString(), text.Substring(match.Index + match.Value.Length));
```
shyamnamboodiripad, line 491: I am not sure I understand why the `string.Concat` calls are needed. Shouldn't the text just be `random.Next(min.Value, max.Value).ToString()` so that we return just the number as part of the call to `node.CreateBindingSuccess` below?

shyamnamboodiripad: I am wondering if this is something that we forgot to change after the for loop from the old implementation was removed...

src/Microsoft.DotNet.Interactive.Http/HttpKernel.cs
```diff
@@ -318,6 +321,209 @@ private List<Diagnostic> GetAllDiagnostics(HttpRequestParseResult parseResult)
...
+            try
+            {
+                if (match.Groups["offset"].Success && match.Groups["option"].Success
+                    && int.TryParse(match.Groups["offset"].Value, out int offset)
+                    && offset != 0)
+                {
+                    dateTimeOffset = dateTimeOffset.AddOffset(offset, match.Groups["option"].Value);
+                }
+
+                text = string.Concat(text.Substring(0, match.Index), dateTimeOffset.ToUnixTimeSeconds().ToString(), text.Substring(match.Index + match.Value.Length));
```
shyamnamboodiripad, line 388: I am not sure I understand why the `string.Concat` call is needed here. Shouldn't text just be set to the `dateTimeOffset.ToUnixTimeSeconds().ToString()` (so that `node.CreateBindingSuccess()` returns the corresponding value)?

Is this code that we forgot to change after the for loop from the old implementation was removed?

src/Microsoft.DotNet.Interactive.Http/HttpKernel.cs
```diff
@@ -318,6 +321,209 @@ private List<Diagnostic> GetAllDiagnostics(HttpRequestParseResult parseResult)
...
+            }
+            catch (ArgumentException)
+            {
+                return node.CreateBindingFailure(HttpDiagnostics.InvalidOffset(match.Groups["offset"].Value));
+            }
+            return node.CreateBindingSuccess(text);
```
shyamnamboodiripad, line 394: consider: Move this line inside the `try` block

src/Microsoft.DotNet.Interactive.Http/HttpKernel.cs
shyamnamboodiripad: Same for `(?<type>rfc1123|iso8601|'.+'|"".+"")` which seems to be repeated in 2 places.

src/Microsoft.DotNet.Interactive.Http/HttpKernel.cs
```diff
@@ -318,6 +321,209 @@ private List<Diagnostic> GetAllDiagnostics(HttpRequestParseResult parseResult)
...
+        } else
+        {
+            return node.CreateBindingFailure(HttpDiagnostics.TimestampFormatError(expression));
```
shyamnamboodiripad, line 397: If I'm understanding the code correctly, looks like we would arrive here if the `offset` is not supplied for instance (since we require that there should be 3 groups in the match). But isn't the `offset` portion optional?

src/Microsoft.DotNet.Interactive.Http/HttpKernel.cs
```diff
@@ -318,6 +321,209 @@ private List<Diagnostic> GetAllDiagnostics(HttpRequestParseResult parseResult)
...
+            }
+            catch (ArgumentOutOfRangeException)
+            {
+                return node.CreateBindingFailure(HttpDiagnostics.DateTimePatternMatchError(expression));
+            }
+            return node.CreateBindingSuccess(text);
+        } 
+        else
+        {
+            return node.CreateBindingFailure(HttpDiagnostics.DateTimePatternMatchError(expression));
```
shyamnamboodiripad, line 460: Same question about whether we would incorrectly report an error here in the case where the optional `offset` and `option` parts were not present in the string...

src/Microsoft.DotNet.Interactive.Http/HttpKernel.cs
```diff
@@ -318,6 +321,209 @@ private List<Diagnostic> GetAllDiagnostics(HttpRequestParseResult parseResult)
...
+                    formatProvider = CultureInfo.InvariantCulture;
+                    format = "r";
+                }
+                else if (string.Equals(type.Value, "iso8601", StringComparison.OrdinalIgnoreCase))
+                {
+                    format = "o";
+                }
+                else
+                {
+                    format = type.Value.Substring(1, type.Value.Length - 2);
```
shyamnamboodiripad, line 442: I am confused by this line - could you clarify what is happening here? Looks like user can specify some custom 'type' here - but it is unclear why we need to do `SubString` here...

src/Microsoft.DotNet.Interactive.Http/HttpKernel.cs
```diff
@@ -318,6 +321,209 @@ private List<Diagnostic> GetAllDiagnostics(HttpRequestParseResult parseResult)
...
+                }
+
+                text = dateTimeOffset.ToString(format, formatProvider);
+
+                if(!DateTimeOffset.TryParse(text, out dateTimeOffset) )
+                {
+                    return node.CreateBindingFailure(HttpDiagnostics.DateTimePatternMatchError(expression));
```
shyamnamboodiripad, line 449: Looks like we can only arrive here is the `format` is not recognized. Should we use a diagnostic message that specifically identifies that the problem is with the `format` or `type` portion of the string?

src/Microsoft.DotNet.Interactive.Http/HttpKernel.cs
```diff
@@ -318,6 +321,209 @@ private List<Diagnostic> GetAllDiagnostics(HttpRequestParseResult parseResult)
...
+                }
+            }
+            catch (ArgumentOutOfRangeException)
+            {
+                return node.CreateBindingFailure(HttpDiagnostics.DateTimePatternMatchError(expression));
+            }
+            return node.CreateBindingSuccess(text);
```
shyamnamboodiripad, line 456: minor: Same comment about moving this inside the `try`.

src/Microsoft.DotNet.Interactive.Http/HttpKernel.cs
```diff
@@ -318,6 +321,209 @@ private List<Diagnostic> GetAllDiagnostics(HttpRequestParseResult parseResult)
...
+
+        
+    }
+
+    private HttpBindingResult<object?> GetRandInt(HttpExpressionNode node, string text, MatchCollection matches)
+    {
+
+        Random random = new();
+        
+        var match = matches.FirstOrDefault();
```
shyamnamboodiripad, line 471: consider: Use `SingleOrDefault` instead since we would have already checked that there is only one match before calling this function above.

src/Microsoft.DotNet.Interactive.Http/HttpKernel.cs
```diff
@@ -318,6 +321,209 @@ private List<Diagnostic> GetAllDiagnostics(HttpRequestParseResult parseResult)
...
+            {
+                return (null, max);
+            }
+            else if (group.Captures.Count == 2
+                && int.TryParse(group.Captures[0].Value, out int min)
+                && int.TryParse(group.Captures[1].Value, out max)
+                && min <= max)
+            {
+                return (min, max);
+            }
```
shyamnamboodiripad, line 523: Should we report an error if there are more than 2 matches?

Also consider using a switch statement instead of the if else.

shyamnamboodiripad (commented):
The changes overall look reasonable. However,

* I suspect there are some vestiges of the old logic still lingering in the code that was ported over from the old parser. And seems like this could result in incorrect / missing errors (e.g. in the `{{$guid $guid}}` case). The comments around these functional aspects seem like the most important ones to address first.

* We are also missing tests for some of the new error messages and for some negative cases such as the `{{$guid $guid}}` case above. It would be great to include some tests for these missing cases.

* Other than that, most of the comments are around code hygiene, some improvements for the error messages we report and some improvement for tests. I am ok with fixing these up in follow up PRs (so long as we fix up the error messages during the early preview releases of the new parser integration).

src/Microsoft.DotNet.Interactive.Http/HttpKernel.cs
bleaphar: Correct, they're distinct but we're only allowing lowercase guid. Are you saying that we should support both?

src/Microsoft.DotNet.Interactive.Http/HttpKernel.cs
bleaphar: The way the regex is configured allows for there to be no matches for the optional parameters. In the case that someone puts timestamp then it will still have three groups but two of the groups will be empty.

src/Microsoft.DotNet.Interactive.Http/HttpKernel.cs
bleaphar: Has to do with the configuration of the regex

src/Microsoft.DotNet.Interactive.Http/HttpKernel.cs
bleaphar: We're expecting quotation marks around the string for the provided type so the substring is there to exclude them

src/Microsoft.DotNet.Interactive.Http/HttpKernel.cs
shyamnamboodiripad: Nope was just checking.

Iirc, http primitives such as header names, urls etc. are not case sensitive - but it seems reasonable for variables in .http files to be case sensitive. I was just checking to confirm that this matches the spec (and the other http editors that we have looked at).

src/Microsoft.DotNet.Interactive.Http.Parsing/Parsing/HttpDiagnostics.cs
```diff
@@ -97,11 +97,72 @@ internal static HttpDiagnosticInfo VariableNameExpected()
...
+        var id = $"HTTP0016";
+        var severity = DiagnosticSeverity.Error;
+        var messageFormat = 
+            "The supplied expression '{0}' does not follow the correct pattern. The expression should adhere to the following pattern: '{{$timestamp [offset option]}}'. See https://aka.ms/http-date-time-format for more details.";
+        return new HttpDiagnosticInfo(id, messageFormat, severity, timestamp);
+    }
+
+    internal static HttpDiagnosticInfo InvalidOption(string expression, string option)
+    {
+        var id = $"HTTP0016";
```
shyamnamboodiripad, line 136: Please ensure that the ids are unique.

Also, please sort the errors by id.

shyamnamboodiripad (approved):
The changes look good, and the test coverage is much improved now.

However, I found a couple of bugs and few more cases that need tests. I am fine with merging the PR so long as the corresponding comments are addressed. Please take a look.

src/Microsoft.DotNet.Interactive.Http.Parsing/Parsing/HttpDiagnostics.cs
```diff
@@ -97,11 +97,72 @@ internal static HttpDiagnosticInfo VariableNameExpected()
...
+        var messageFormat = "Unable to evaluate expression '{0}'.";
         return new HttpDiagnosticInfo(id, messageFormat, severity, symbol);
     }
+
+    internal static HttpDiagnosticInfo IncorrectDateTimeFormat(string expression)
+    {
+        var id = $"HTTP0013";
+        var severity = DiagnosticSeverity.Error;
+        var messageFormat =
+            """The supplied expression '{0}' does not follow the correct pattern. The expression should adhere to the following pattern: '{{$datetime |iso8601|"custom format" [offset option]}}' where offset is a valid integer and option is one of the following: ms, s, m, h, d, w, M, Q, y. See https://aka.ms/http-date-time-format for more details.""";
```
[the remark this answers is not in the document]
shyamnamboodiripad: Looks like this comment may have been missed.

shyamnamboodiripad (commented):
Looks like you may have missed some of the comments because GitHub collapses them by default in the discussion view. Let's sync up tomorrow.

shyamnamboodiripad (approved):
👍🏾 Looks great. Thanks for adding test coverage for all the cases!

src/Microsoft.DotNet.Interactive.Http.Tests/HttpKernelTests.cs
```diff
@@ -336,15 +442,832 @@ public async Task url_in_embedded_expression_is_valid()
...
+        var unixValueSubstring = bodyAsString.Split(":").Last().Trim().Substring(1);
+        var unixValueString = unixValueSubstring.Substring(0, unixValueSubstring.IndexOf("\""));
+        var unixValue = DateTimeOffset.FromUnixTimeSeconds(long.Parse(unixValueString));
+        var offsetDaysInteger = int.Parse(offsetDays);
+        var dateTimeOffset = DateTimeOffset.UtcNow.AddDays(offsetDaysInteger);
+        unixValue.Should().BeCloseTo(dateTimeOffset, TimeSpan.FromSeconds(10));
+    }
+
+    [Fact]
+    public async Task cant_bind_timestamp_offset_without_option()
```
shyamnamboodiripad, line 502: This test is failing. It seems redundant based on what the `can_bind_timestamp_with_valid_offset` is doing. I think you can either delete it or add `d` after `-1` below.

PR #3426, opened 2024-01-25
allows empty diagnostics report to flow

colombod:
Fixes the issue were diagnostics are not clear as the user types

### DIFF
--- a/src/Microsoft.DotNet.Interactive.CSharp/CSharpKernel.cs
+++ b/src/Microsoft.DotNet.Interactive.CSharp/CSharpKernel.cs
@@ -471,10 +471,9 @@ public class CSharpKernel :
         var document = _workspace.ForkDocumentForLanguageServices(command.Code);
         var semanticModel = await document.GetSemanticModelAsync(context.CancellationToken);
         var diagnostics = semanticModel.GetDiagnostics(cancellationToken: context.CancellationToken);
-        if (diagnostics.Length > 0)
-        {
-            context.Publish(GetDiagnosticsProduced(command, diagnostics));
-        }
+
+        context.Publish(GetDiagnosticsProduced(command, diagnostics));
+
     }
 
     private bool HasReturnValue =>
--- a/src/Microsoft.DotNet.Interactive/Events/DiagnosticsProduced.cs
+++ b/src/Microsoft.DotNet.Interactive/Events/DiagnosticsProduced.cs
@@ -22,10 +22,10 @@ public class DiagnosticsProduced : KernelEvent
             throw new ArgumentNullException(nameof(diagnostics));
         }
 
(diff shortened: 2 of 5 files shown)


src/Microsoft.DotNet.Interactive/Events/DiagnosticsProduced.cs
```diff
@@ -22,10 +22,10 @@ public class DiagnosticsProduced : KernelEvent
...
         }
 
-        if (!diagnostics.Any())
-        {
-            throw new ArgumentException("At least one diagnostic required.", nameof(diagnostics));
-        }
+        //if (!diagnostics.Any())
+        //{
+        //    throw new ArgumentException("At least one diagnostic required.", nameof(diagnostics));
+        //}
```
shyamnamboodiripad, line 28: From an API standpoint, it seems a bit odd that the system would support two ways to convey absence of diagnostics - a) absence of `DiagnosticsProduced` event in the response and b) `DiagnosticsProduced` event with 0 diagnostics. Iirc we originally made the change to disallow the empty `DiagnosticsProduced` event (in #3002) to avoid clients needing to handle both forms.

Is there an alternate way to fix the IntelliSense issue without introducing the above API issue?

src/Microsoft.DotNet.Interactive.CSharp/CSharpKernel.cs
```diff
@@ -471,10 +471,9 @@ void UpdateScriptOptionsIfWorkingDirectoryChanged()
         var document = _workspace.ForkDocumentForLanguageServices(command.Code);
         var semanticModel = await document.GetSemanticModelAsync(context.CancellationToken);
         var diagnostics = semanticModel.GetDiagnostics(cancellationToken: context.CancellationToken);
-        if (diagnostics.Length > 0)
-        {
-            context.Publish(GetDiagnosticsProduced(command, diagnostics));
-        }
+
+        context.Publish(GetDiagnosticsProduced(command, diagnostics));
```
shyamnamboodiripad, line 475: Can we add a test for the IntelliSense issue that is fixed by this change?

src/Microsoft.DotNet.Interactive/Events/DiagnosticsProduced.cs
jonsequitur: I would regard it as a bug if the event is produced with no actual diagnostics, and now handlers of this event have to be aware of this. This check was here to catch such a bug. The PR overall seems like it's going to hide code defects elsewhere.

colombod: you are right, we closed and abandoned this path as @shyamnamboodiripad  pointed all of this out

PR #3432, opened 2024-01-30
add path for conda setup

colombod:


colombod:
@shibbas what do you think?

shibbas:
> @shibbas what do you think?
We should probably avoid coupling the conda paths to the Jupyter common paths here. 
In this scenario, the `conda activate base` was failing so the jupyter commands didn't run, but perhaps we can check in the conda runtime if the environment is already activated? Using `conda env list` might help here.

PR #3450, opened 2024-02-13
Change to declared variables to support dynamic expressions

bleaphar:
This pr changes the declared variable definition to support dynamic expressions.

**Summary**

- Moved the code for dynamic expression matching into a static class to be used in multiple locations
- Added tests in the kernel and variables to target this scenario 
- Updated the bind function in request node to fix a stack overflow issue that was occurring with the circular reference


src/Microsoft.DotNet.Interactive.Http.Parsing/Parsing/DynamicExpressionUtilities.cs
```diff
@@ -0,0 +1,289 @@
...
+            const string DateTime = "$datetime";
+            const string LocalDateTime = "$localDatetime";
+            const string OffsetRegex = """(?:\s+(?<offset>[-+]?[^\s]+)\s+(?<option>[^\s]+))?""";
+            const string TypeRegex = """(?:\s+(?<type>rfc1123|iso8601|'.+'|".+"))?""";
+
+            var guidPattern = new Regex(@$"^\$guid$", RegexOptions.Compiled);
+            var dateTimePattern = new Regex(@$"^\{DateTime}{TypeRegex}{OffsetRegex}$", RegexOptions.Compiled);
+            var localDateTimePattern = new Regex(@$"^\{LocalDateTime}{TypeRegex}{OffsetRegex}$", RegexOptions.Compiled);
+            var randomIntPattern = new Regex(@$"^\$randomInt(?:\s+(?<arguments>-?[^\s]+)){{0,2}}$", RegexOptions.Compiled);
+            var timestampPattern = new Regex(@$"^\$timestamp{OffsetRegex}$", RegexOptions.Compiled);
```
jonsequitur, line 28: These can be initialized as static fields instead of local variables, since they're effectively immutable. That's also where `RegexOptions.Compiled` will give an advantage.

src/Microsoft.DotNet.Interactive.Http.Parsing/Parsing/DynamicExpressionUtilities.cs
```diff
@@ -0,0 +1,289 @@
...
+
+                if (DateTimeOffset.TryParse(text, out _))
+                {
+                    return node.CreateBindingSuccess(text);
+                }
+            }
+            return node.CreateBindingFailure(HttpDiagnostics.IncorrectDateTimeFormat(expressionText, dateTimeType));
+        }
+
+        private static HttpBindingResult<object?> GetRandInt(HttpExpressionNode node, string text, Match match)
```
jonsequitur, line 190:
```suggestion
        private static HttpBindingResult<object?> GetRandomInt(HttpExpressionNode node, string text, Match match)
```

It's a bit more readable when you use the full word.

src/Microsoft.DotNet.Interactive.Http.Parsing/Parsing/DynamicExpressionUtilities.cs
```diff
@@ -0,0 +1,289 @@
...
+                    }
+                    else if (group.Captures.Count == 1)
+                    {
+                        min = null;
+                        string maxValueString = group.Captures[0].Value;
+                        return TryParseInteger(maxValueString, expression, out max, out diagnostic);
+
+                    }
+
+                    else if (group.Captures.Count == 2)
```
jonsequitur, line 235: Can there be more than two?

src/Microsoft.DotNet.Interactive.Http.Parsing/Parsing/DynamicExpressionUtilities.cs
```diff
@@ -0,0 +1,289 @@
...
+using System.Text.RegularExpressions;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.DotNet.Interactive.Http.Parsing
+{
+    #nullable enable   
+    internal class DynamicExpressionUtilites
+    {
+        internal static HttpBindingResult<object?> MatchExpressionValue(HttpExpressionNode node, string expression)
```
jonsequitur, line 17: The name is a little confusing. Can you come up with a name that makes it clearer what this method is actually doing?

src/Microsoft.DotNet.Interactive.Http.Parsing/Parsing/DynamicExpressionUtilities.cs
bleaphar: No it shouldn't allow more than two

jonsequitur (approved):

PR #3467, opened 2024-02-26
add support for clearing variables from HttpKernel

jonsequitur:


phenning (commented):
This is great, exactly what we need!

colombod (approved):

src/Microsoft.DotNet.Interactive.Http/ClearValues.cs
```diff
@@ -0,0 +1,13 @@
+// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Microsoft.DotNet.Interactive.Commands;
+
+namespace Microsoft.DotNet.Interactive.Http;
+
+public class ClearValues : KernelCommand
```
shyamnamboodiripad, line 8: This seems generally useful. Any reason why this is not defined in the main assembly?

Currently the http editor does not (need to) directly reference the `Interactive.Http` assembly. Adding that reference may not be problematic - but still wanted to double check whether we can move this to the main assembly and avoid the additional dependency.

src/Microsoft.DotNet.Interactive.Http/HttpKernel.cs
```diff
@@ -58,6 +56,12 @@ public class HttpKernel :
         RegisterForDisposal(_client);
     }
 
+    public Task HandleAsync(ClearValues command, KernelInvocationContext context)
```
shyamnamboodiripad, line 59: Could we implement this using explicit interface implementation and avoid the public API similar to what we do for the other commands below?

PR #3469, opened 2024-02-27
Added support for no request nodes and just comments present

bleaphar:
Added a test and support for the absence of a request and just comments


src/Microsoft.DotNet.Interactive.Http.Tests/ParserTests.Comments.cs
```diff
@@ -64,5 +64,18 @@ public void Comment_node_can_immediately_follow_headers()
                   .Should().ContainSingle<HttpCommentNode>()
                   .Which.Text.Should().Be("# This is a comment");
         }
+
+        [Fact]
+        public void Comment_node_without_request_node_does_not_produce_diagnostics()
```
jonsequitur, line 69: This might be a good candidate for a combinatorial test or a theory test. Maybe we already have coverage for some of these but does it behave correctly in these cases?

* One or more comment lines come after a request node.
* One or more comment lines with newlines before, after, or between them.

jonsequitur (approved):

PR #3471, opened 2024-02-29
Add missing command registration to fix deadlock

shyamnamboodiripad:


phenning (approved):

src/Microsoft.DotNet.Interactive.Http/HttpKernel.cs
```diff
@@ -53,6 +54,8 @@ public class HttpKernel :
         _responseDelayThresholdInMilliseconds = responseDelayThresholdInMilliseconds;
         _contentByteLengthThreshold = contentByteLengthThreshold;
 
+        KernelCommandEnvelope.RegisterCommand<ClearValues>();
```
jonsequitur, line 57: As long as the underlying implementation is static, this can be done in a static constructor.

PR #3486, opened 2024-03-14
fix jsonschema dependency

colombod:


colombod:
@markwallace-microsoft this should unblock semantic kernel

markwallace-microsoft (approved):
LGTM and fixes the issue we are facing

shyamnamboodiripad (approved):

PR #3487, opened 2024-03-14
Improve build diagnostics in CSharpProject

ocallesp:


src/Microsoft.DotNet.Interactive.CSharpProject/RoslynWorkspaceUtilities/BuildCacheFileUtilities.cs
```diff
@@ -78,30 +78,40 @@ internal static class BuildCacheFileUtilities
 </Project>
 """;
 
-    internal static async Task<string> BuildAndCreateCacheFileAsync(string csprojFilePath)
+    internal static async Task BuildAndCreateCacheFileAsync(string csprojFilePath)
     {
+        if (string.IsNullOrEmpty(csprojFilePath) || !File.Exists(csprojFilePath))
```
jonsequitur, line 83: I would make these separate conditions. If null is passed, then that's a different category of error, indicating a bug in this codebase, whereas the path not existing is likely a deployment or environment problem.

src/Microsoft.DotNet.Interactive.CSharpProject/RoslynWorkspaceUtilities/BuildCacheFileUtilities.cs
```diff
@@ -78,30 +78,40 @@ internal static class BuildCacheFileUtilities
...
-        File.WriteAllText(tempDirectoryBuildTarget, DirectoryBuildTargetsContent);
+        string tempDirectoryBuildTarget = Path.Combine(Path.GetDirectoryName(csprojFilePath), DirectoryBuildTargetFilename);
 
-        var args = "";
-        if (Path.Exists(csprojFilePath))
+        try
+        {
+            File.WriteAllText(tempDirectoryBuildTarget, DirectoryBuildTargetsContent);
+        }
+        catch (Exception ex)
```
jonsequitur, line 99: It's never ideal to catch `Exception` as it can hide bugs. If there's a specific exception type we know how to handle, let's handle that. Otherwise it's better not to catch it at all.

src/Microsoft.DotNet.Interactive.CSharpProject/RoslynWorkspaceUtilities/BuildCacheFileUtilities.cs
```diff
@@ -78,30 +78,40 @@ internal static class BuildCacheFileUtilities
...
+
         var result = await new Dotnet(directoryInfo).Build(args: args);
 
         if (result.ExitCode != 0)
         {
+            var errorMessage = $"Build failed with exit code {result.ExitCode}. See {lastBuildErrorLogFile.FullName} for details.";
             File.WriteAllText(
                 lastBuildErrorLogFile.FullName,
                 string.Join(Environment.NewLine, result.Error));
+            throw new InvalidOperationException(errorMessage);
```
jonsequitur, line 114: The exception might be the right thing to do but it's not clear to me why should both write the error out to the file and throw.

src/Microsoft.DotNet.Interactive.CSharpProject/RoslynWorkspaceUtilities/BuildCacheFileUtilities.cs
ocallesp: Changed. Only UnauthorizedAccessException and IOException are checked.

src/Microsoft.DotNet.Interactive.CSharpProject/RoslynWorkspaceUtilities/BuildCacheFileUtilities.cs
ocallesp: Changed. Refactored to directly throw an exception upon build failure, facilitating rapid debugging.

jonsequitur (approved):

PR #3488, opened 2024-03-14
Change to have a more accurate url error 

bleaphar:


src/Microsoft.DotNet.Interactive.Http.Parsing/Parsing/HttpRequestParser.cs
```diff
@@ -448,6 +448,10 @@ private void ConsumeCurrentTokenInto(HttpSyntaxNode node)
                         node = new HttpUrlNode(_sourceText, _syntaxTree);
                         node.Add(ParseEmbeddedExpression());
                     }
+                    else if (CurrentToken is { Kind: HttpTokenKind.Punctuation })
+                    {
+                        node = new HttpUrlNode(_sourceText, _syntaxTree);
```
shyamnamboodiripad, line 453: Why is it that we only call ParseLeadingWhitespaceAndComments in the case where the current token is a word (line 444)? Should we also call that here (as well as in the embedded expression case above before line 449)?

src/Microsoft.DotNet.Interactive.Http.Parsing/Parsing/HttpRequestParser.cs
```diff
@@ -448,6 +448,10 @@ private void ConsumeCurrentTokenInto(HttpSyntaxNode node)
                         node = new HttpUrlNode(_sourceText, _syntaxTree);
                         node.Add(ParseEmbeddedExpression());
                     }
+                    else if (CurrentToken is { Kind: HttpTokenKind.Punctuation })
```
shyamnamboodiripad, line 451: Why do we enter the while loop in the case where the current token is a punctuation (on line 436 above)? A url should always start with a word (or an embedded expression) should it not?

src/Microsoft.DotNet.Interactive.Http.Parsing/Parsing/HttpRequestParser.cs
shyamnamboodiripad: In other words, do we drop leading trivia / comments in the two cases where the call is currently missing?

src/Microsoft.DotNet.Interactive.Http.Parsing/Parsing/HttpRequestParser.cs
shyamnamboodiripad: Never mind - I think I understand what is going on here..

shyamnamboodiripad, line 451: So far the above two conditions only represented the positive cases that could start a url. The new third condition is for an error case.

Ideally the parser should not return a url node for something that is not a url - it seems odd that the parser would parse a non-url as a url just to improve the error reporting at bind time...

bleaphar: Well, the rationale is that the text is there where a url should be, the same way that a token that is interpreted as a word but is completely invalid could be in that location and still create a node. I also prefer this approach so that we have the diagnostic on the url node itself as opposed to the root node

shyamnamboodiripad (approved):

src/Microsoft.DotNet.Interactive.Http.Parsing/Parsing/HttpRequestParser.cs
shyamnamboodiripad: (Thanks for fixing - It is interesting that the missing trivia parsing was not caught by the tests since iirc all tests test round-tripping...)

PR #3492, opened 2024-03-19
Bleaphar/comment node investigation

bleaphar:
This pr treats text that is on the same line as a request separator as a part of the node and consumes those tokens into the request separator. Also, it creates an early out for the parsing of url nodes in the event that we parse an embedded expression that is the entirety of the url node. Lastly, added a test based on several requests with request separator nodes and embedded expressions.


src/Microsoft.DotNet.Interactive.Http.Tests/ParserTests.Request.cs
```diff
@@ -263,5 +263,66 @@ public void binding_for_variable_in_header_is_correct()
...
+
+                name=value
+
+                ####
+
+                GET https://httpbin.org/anything
+
+                {name=value}
+                """
+                );
```
shyamnamboodiripad, line 319: I think it may be better to create a more targeted test for the scenario being fixed.

src/Microsoft.DotNet.Interactive.Http.Parsing/Parsing/HttpRequestParser.cs
```diff
@@ -462,6 +474,11 @@ private void ConsumeCurrentTokenInto(HttpSyntaxNode node)
                     }
                 }
 
+                if (!((CurrentToken?.Kind is HttpTokenKind.Word or HttpTokenKind.Punctuation) || (IsAtStartOfEmbeddedExpression())))
```
shyamnamboodiripad, line 477: Was this another bug in the url parsing? Could you please add a separate test for this?

src/Microsoft.DotNet.Interactive.Http.Tests/ParserTests.Request.cs
bleaphar: Namely less requests and intricacies?

shyamnamboodiripad: Yes, a tight test that tests exactly the single case that was not working before and that is fixed now.

src/Microsoft.DotNet.Interactive.Http.Tests/ParserTests.RequestSeparator.cs
```diff
@@ -28,5 +29,19 @@ public void request_separator_at_start_of_request_is_valid()
...
+        public void tokens_on_request_separator_nodes_are_parsed_into_request_separator()
+        {
+            var result = Parse(
+                """
+                ### Slow Response (Json)
+                GET https://httpbin.org/anything?page=2&pageSize=10
+                """
+                );
+
+            var requestNodes = result.SyntaxTree.RootNode.ChildNodes
```
jonsequitur, line 42: Curious about how this is parsed. Is a child comment node added to the request separator node?

src/Microsoft.DotNet.Interactive.Http.Tests/ParserTests.Url.cs
```diff
@@ -160,5 +160,20 @@ public void Punctuation_at_url_start_produces_a_diagnostic()
...
+            var code = """
+                GET {{host}}
+                ###
+                """;
+
+            var result = Parse(code);
+
+            result.GetDiagnostics().Should().BeEmpty();
+
+            result.SyntaxTree.RootNode.ChildNodes.Count().Should().Be(2);
```
jonsequitur, line 176: Can we make a more specific assertion here?

src/Microsoft.DotNet.Interactive.Http.Tests/ParserTests.RequestSeparator.cs
bleaphar: It's parsed as tokens into the http request separator node

jonsequitur (approved):

PR #3525, opened 2024-04-17
force Prebuilds to build in debug mode

dalehhirt:
We need the ability to pass specific arguments to Prebuild.

### DIFF
--- a/src/Microsoft.DotNet.Interactive.CSharpProject/Build/Prebuild.cs
+++ b/src/Microsoft.DotNet.Interactive.CSharpProject/Build/Prebuild.cs
@@ -385,7 +385,8 @@ public class Prebuild
 
         var projectFile = GetProjectFile();
 
-        var args = "";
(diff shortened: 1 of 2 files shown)


src/Microsoft.DotNet.Interactive.CSharpProject/Build/Prebuild.cs
```diff
@@ -440,13 +444,14 @@ public async Task<bool> CreatePrebuildAsync(IPrebuildInitializer initializer)
         return true;
     }
 
-    public static async Task<Prebuild> GetOrCreateConsolePrebuildAsync(bool enableBuild = false)
+    public static async Task<Prebuild> GetOrCreateConsolePrebuildAsync(bool enableBuild = false, string args = null)
```
jonsequitur, line 447: It's better to avoid adding this flexibility until/unless we need it.

PR #3530, opened 2024-04-23
Parsing restructuring

bleaphar:
Change to the parser to search for significant tokens and parsing request separators before parsing requests. In the event of additional whitespaces tokens, they are attached to the root node instead of creating a null request node

### DIFF
--- a/src/Microsoft.DotNet.Interactive.Http.Tests/ParserTests.Comments.cs
+++ b/src/Microsoft.DotNet.Interactive.Http.Tests/ParserTests.Comments.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Linq;
 using System.Management.Automation.Internal;
 using FluentAssertions;
 using Microsoft.DotNet.Interactive.Http.Parsing;
@@ -49,7 +50,7 @@ public partial class HttpParserTests
         [Fact]
         public void Comment_node_can_immediately_follow_headers()
         {
-             var code = """
+            var code = """
                 GET https://example.com
                 Accept: text/plain
                 # This is a comment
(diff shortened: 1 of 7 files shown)


src/Microsoft.DotNet.Interactive.Http.Tests/ParserTests.Comments.cs
```diff
@@ -103,5 +104,25 @@ public void Comment_node_without_request_node_does_not_produce_diagnostics(strin
...
+                GET {{MyRestaurantApi_HostAddress}}/api/Contact
+
+                ###
+
+                # get a specific contact
+                """;
+
+            var result = Parse(code);
+
+            result.SyntaxTree.RootNode.DescendantNodesAndTokens().Should()
```
jonsequitur, line 123: Do we want this assertion to be more specific about where the comment node appears?

src/Microsoft.DotNet.Interactive.Http.Tests/ParserTests.RequestSeparator.cs
```diff
@@ -43,5 +43,34 @@ public void tokens_on_request_separator_nodes_are_parsed_into_request_separator(
                                     .Should().ContainSingle<HttpRequestSeparatorNode>()
                                     .Which.Text.Should().Be("### Slow Response (Json)");
         }
+
+        [Fact]
+        public void tokens_on_request_separator_nodes()
```
jonsequitur, line 48: The test name should be more descriptive of the expectation.

src/Microsoft.DotNet.Interactive.Http.Tests/ParserTests.Variables.cs
```diff
@@ -250,5 +250,21 @@ public void underscores_in_embedded_expressions_are_supported()
...
+                """
+                @host=https://httpbin.org
+                
+                
+                
+                
+                
+                """);
+
+            result.SyntaxTree.RootNode.ChildNodes.Count().Should().Be(1);
```
jonsequitur, line 267: This assertion can be more specific.  It doesn't currently seem related to the test name.

src/Microsoft.DotNet.Interactive.Http.Tests/ParserTests.Comments.cs
bleaphar: Done

src/Microsoft.DotNet.Interactive.Http.Tests/ParserTests.RequestSeparator.cs
bleaphar: Done

src/Microsoft.DotNet.Interactive.Http.Tests/ParserTests.Variables.cs
bleaphar: Done

jonsequitur (approved):

PR #3531, opened 2024-04-24
publish DiagnosticsProduced even when there are no diagnostics

jonsequitur:
This fixes https://github.com/dotnet/try/issues/1173


colombod:
SinguatureHelp, HoverText and Completions needs to be changed too.

colombod (approved):

PR #3537, opened 2024-04-30
1ESPT Compliance updates to our signed build yml.

AbhitejJohn:
Fixing the build and optimizing the number of times compliance tasks are run. 

Here's a succesful build: https://dev.azure.com/dnceng/internal/_build/results?buildId=2441201&view=results.

Please let me know if there is anything else I should be validating.

### DIFF
--- a/azure-pipelines-official.yml
+++ b/azure-pipelines-official.yml
@@ -129,28 +129,30 @@ extends:
                       /p:OfficialBuildId=$(BUILD.BUILDNUMBER)
                       /p:PublishToSymbolServer=true
 
-            outputs:
-              - output: pipelineArtifact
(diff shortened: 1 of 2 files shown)


ocallesp (approved):

AbhitejJohn:
@colombod : This is good to go. if you are alright with it too, I can merge and update our signed pipeline to use the azure-pipelines-official.yml file that was updated here. I'll be on-call for any issues that might cause. I've validate the build is green here though - https://dev.azure.com/dnceng/internal/_build/results?buildId=2441201&view=results. Let me know if there is anything else we'd want to validate.

colombod (approved):

PR #3539, opened 2024-04-30
fix formatted diagnostics to make them required

jonsequitur:
This is part of the fix for https://github.com/dotnet/try/issues/1172


colombod (approved):
I was there

PR #3558, opened 2024-05-28
Changes to request parser to accommodate spaces

bleaphar:
This pr contains a major change to the parser to all for spaces in the url of an HttpUrlNode. This pr is a draft until further design considerations are established.

### DIFF
--- a/src/Microsoft.DotNet.Interactive.Http.Parsing/Parsing/HttpRequestParser.cs
+++ b/src/Microsoft.DotNet.Interactive.Http.Parsing/Parsing/HttpRequestParser.cs
@@ -453,7 +453,8 @@ internal class HttpRequestParser
         {
             HttpUrlNode? node = null;
 
-            while ((CurrentToken?.Kind is HttpTokenKind.Word or HttpTokenKind.Punctuation) || (IsAtStartOfEmbeddedExpression()))
+            var significantToken = GetNextSignificantToken()?.Text.ToLowerInvariant();
+            while (((CurrentToken?.Kind is HttpTokenKind.Word or HttpTokenKind.Punctuation) || (CurrentToken is { Kind: HttpTokenKind.Whitespace } && significantToken is not "http/" or "https/") || (IsAtStartOfEmbeddedExpression())))
             {
                 if (node is null)
                 {
@@ -482,12 +483,23 @@ internal class HttpRequestParser
                     }
                 }
 
-                if (!((CurrentToken?.Kind is HttpTokenKind.Word or HttpTokenKind.Punctuation) || (IsAtStartOfEmbeddedExpression())))
+                if(CurrentToken?.Kind is HttpTokenKind.NewLine)
                 {
                     break;
                 }
 
-                if (IsAtStartOfEmbeddedExpression())
+                if ((CurrentToken?.Kind is HttpTokenKind.Whitespace))
+                {
+                    var nextSignificantToken = GetNextSignificantToken()?.Text.ToLowerInvariant();
+                    if ((nextSignificantToken is not "http" or "https"))
+                    {
+                        ConsumeCurrentTokenInto(node);
+                    } else
+                    {
+                        break;
+                    }
+                }
+                else if (IsAtStartOfEmbeddedExpression())
                 {
                     node.Add(ParseEmbeddedExpression());
                 }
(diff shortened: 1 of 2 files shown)


src/Microsoft.DotNet.Interactive.Http.Parsing/Parsing/HttpRequestParser.cs
```diff
@@ -453,7 +453,8 @@ private void ConsumeCurrentTokenInto(HttpSyntaxNode node)
         {
             HttpUrlNode? node = null;
 
-            while ((CurrentToken?.Kind is HttpTokenKind.Word or HttpTokenKind.Punctuation) || (IsAtStartOfEmbeddedExpression()))
+            var significantToken = GetNextSignificantToken()?.Text.ToLowerInvariant();
+            while (((CurrentToken?.Kind is HttpTokenKind.Word or HttpTokenKind.Punctuation) || (CurrentToken is { Kind: HttpTokenKind.Whitespace } && significantToken is not "http/" or "https/") || (IsAtStartOfEmbeddedExpression())))
```
shyamnamboodiripad, line 457: The `/` char should probably be omitted from the string since it will not be in the same token as the `http` or `https` word.
```suggestion
            while (((CurrentToken?.Kind is HttpTokenKind.Word or HttpTokenKind.Punctuation) || (CurrentToken is { Kind: HttpTokenKind.Whitespace } && significantToken is not "http" or "https") || (IsAtStartOfEmbeddedExpression())))
```

src/Microsoft.DotNet.Interactive.Http.Parsing/Parsing/HttpRequestParser.cs
```diff
@@ -482,12 +483,23 @@ private void ConsumeCurrentTokenInto(HttpSyntaxNode node)
...
                     break;
                 }
 
-                if (IsAtStartOfEmbeddedExpression())
+                if ((CurrentToken?.Kind is HttpTokenKind.Whitespace))
+                {
+                    var nextSignificantToken = GetNextSignificantToken()?.Text.ToLowerInvariant();
+                    if ((nextSignificantToken is not "http" or "https"))
+                    {
+                        ConsumeCurrentTokenInto(node);
```
shyamnamboodiripad, line 496: Does the encoding of whitespace matter? If yes, do we have tests to make sure whitespace is encoded (`%20`) correctly in the resulting `System.Uri` object returned when we bind the url?

src/Microsoft.DotNet.Interactive.Http.Parsing/Parsing/HttpRequestParser.cs
```diff
@@ -482,12 +483,23 @@ private void ConsumeCurrentTokenInto(HttpSyntaxNode node)
...
+                if(CurrentToken?.Kind is HttpTokenKind.NewLine)
                 {
                     break;
                 }
 
-                if (IsAtStartOfEmbeddedExpression())
+                if ((CurrentToken?.Kind is HttpTokenKind.Whitespace))
+                {
+                    var nextSignificantToken = GetNextSignificantToken()?.Text.ToLowerInvariant();
+                    if ((nextSignificantToken is not "http" or "https"))
```
shyamnamboodiripad, line 494: This means I can't have a url that contains the word http separated by spaces :) e.g. `https://www.something.com/query?=tell me more about the http protocol`. It may be interesting to check whether REST editor and other extensions allow this.

src/Microsoft.DotNet.Interactive.Http.Tests/ParserTests.Url.cs
```diff
@@ -161,6 +162,24 @@ public void Punctuation_at_url_start_produces_a_diagnostic()
...
+                GET https://example.com/search?query=valid space here
+                ###
+                """;
+
+            var result = Parse(code);
+            result.GetDiagnostics().Should().BeEmpty();
+
+            var urlNode = result.SyntaxTree.RootNode.DescendantNodesAndTokens().OfType<HttpUrlNode>().Single();
+
+            urlNode.Text.Should().Be("https://example.com/search?query=valid space here");
```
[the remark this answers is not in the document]
shyamnamboodiripad: Also cover negative cases like -

* https://www.something.com/query?=tell me more about the   HTTP protocol
* https://www.something.com/query?=tell me more    about the https/1.1 protocol

shyamnamboodiripad:
@bleaphar Can we abandon this PR based on the outcome of our discussion offline? I think our conclusion was that allowing whitespace in the grammar could be bug prone and that a better approach may be to introduce a refactoring in the IDE to encode URLs with whitespace correctly instead.

PR #3562, opened 2024-06-04
Perf improvements for CSharpProjectKernel

jonsequitur:


colombod (approved):

src/Microsoft.DotNet.Interactive.CSharpProject.Tests/WorkspaceServerTestsCore.cs
```diff
@@ -24,7 +23,7 @@ public abstract class WorkspaceServerTestsCore : IDisposable
         ITestOutputHelper output)
     {
         _prebuildFixture = prebuildFixture;
-        _disposables.Add(output.SubscribeToPocketLogger());
+        // _disposables.Add(output.SubscribeToPocketLogger());
```
shyamnamboodiripad, line 26: Should this be uncommented? Can this be deleted if it not needed?

src/Microsoft.DotNet.Interactive.CSharpProject/Build/BuildResult.cs
```diff
@@ -148,22 +146,25 @@ public AdhocWorkspace CreateWorkspace()
...
             projectName,
             LanguageNames.CSharp,
             filePath: projectBuildInfo.ProjectFilePath,
             outputFilePath: projectBuildInfo.TargetPath,
             documents: GetDocuments(projectBuildInfo, projectId),
-            projectReferences: GetExistingProjectReferences(projectBuildInfo, workspace),
+            projectReferences: GetExistingProjectReferences(projectBuildInfo, solution),
             metadataReferences: GetMetadataReferences(projectBuildInfo),
-            analyzerReferences: GetAnalyzerReferences(projectBuildInfo, workspace),
+            analyzerReferences: null,
```
shyamnamboodiripad, line 167: Is this intentional? Are we ignoring analyzers because while analyzers can block build for projects that otherwise compile, the opposite is not true (i.e. analyzers can't cause projects that fail to build because of other build failures to build successfully)?

jonsequitur: Analyzers are not supported for security reasons.

jonsequitur: Analyzers are not supported for security reasons.

PR #3572, opened 2024-06-12
Fix TryDotnet issue "Compile errors are duplicated in output"

ocallesp:
Fixes https://github.com/dotnet/try/issues/1184

This fixes a bug in TryDotNet that occurs when a user enters invalid code.

- Ensured project diagnostics are added only if their Id is not already present in the main diagnostics collection

image

### DIFF
--- a/src/Microsoft.DotNet.Interactive.CSharpProject/CSharpProjectKernel.cs
+++ b/src/Microsoft.DotNet.Interactive.CSharpProject/CSharpProjectKernel.cs
@@ -291,9 +291,17 @@ public class CSharpProjectKernel :
             projectDiagnostics = projectDiags;
         }
 
-        var allDiagnostics = diagnostics.Concat(projectDiagnostics);
+        var uniqueDiagnostics = diagnostics.ToDictionary(d => d.Id, d => d);
 
-        var finalDiagnostics = allDiagnostics.Select(d => new Diagnostic(new LinePositionSpan(GetLinePositionFromPosition(code, d.Start), GetLinePositionFromPosition(code, d.End)), d.Severity, d.Id, d.Message));
+        foreach (var projDiag in projectDiagnostics)
+        {
+            if (!uniqueDiagnostics.ContainsKey(projDiag.Id))
+            {
+                uniqueDiagnostics.Add(projDiag.Id, projDiag);
+            }
+        }
+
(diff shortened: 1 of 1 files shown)


src/Microsoft.DotNet.Interactive.CSharpProject/CSharpProjectKernel.cs
```diff
@@ -291,9 +291,17 @@ private static IEnumerable<Diagnostic> GetDiagnostics(string code, CompileResult
             projectDiagnostics = projectDiags;
         }
 
-        var allDiagnostics = diagnostics.Concat(projectDiagnostics);
```
jonsequitur, line 294: I wonder if we need the project diagnostics at all. Are there ever diagnostics here that are interesting from the point of view of Try .NET?

ocallesp: `Diagnostics `contain errors introduced by the user, while `ProjectDiagnostics `cover errors across the entire project.

Using only `Diagnostics `works well for our primary focus on user-introduced errors.

For debugging purposes and to catch potential errors from changes to the background project, it might be useful to include `ProjectDiagnostics`. However, since the project is pre-built and we assume deployment doesn't occur with build errors, we can safely remove `ProjectDiagnostics `from `GetDiagnostics`.

src/Microsoft.DotNet.Interactive.CSharpProject/CSharpProjectKernel.cs
```diff
@@ -277,23 +277,14 @@ private async Task<Workspace> AppendCodeToWorkspaceAsync(string code, int positi
     private static IEnumerable<Diagnostic> GetDiagnostics(string code, CompileResult result)
     {
         var diagnostics = Enumerable.Empty<SerializableDiagnostic>();
```
jonsequitur, line 279: This assignment could be moved to an `else` block below now.

jonsequitur (approved):

src/Microsoft.DotNet.Interactive.CSharpProject/CSharpProjectKernel.cs
ocallesp: Good. I made the suggested change. Can you approve this cleanup change https://github.com/dotnet/interactive/pull/3574

PR #3582, opened 2024-06-19
Addition of period into variable declaration

bleaphar:
This PR would update the http parser to support periods in variable declarations. This is a slight feature that has been requested by some users

### DIFF
--- a/src/Microsoft.DotNet.Interactive.Http.Parsing/Parsing/HttpRequestParser.cs
+++ b/src/Microsoft.DotNet.Interactive.Http.Parsing/Parsing/HttpRequestParser.cs
@@ -194,7 +194,7 @@ internal class HttpRequestParser
 
                 while (MoreTokens())
                 {
-                    if (CurrentToken is { Text: "@" } or { Kind: HttpTokenKind.Word } or { Text: "_" })
+                    if (CurrentToken is { Text: "@" } or { Kind: HttpTokenKind.Word } or { Text: "_" } or {Text: "." })
                     {
                         ConsumeCurrentTokenInto(node);
                     }
--- a/src/Microsoft.DotNet.Interactive.Http.Parsing/Parsing/HttpRequestParser.cs
+++ b/src/Microsoft.DotNet.Interactive.Http.Parsing/Parsing/HttpRequestParser.cs
@@ -194,7 +194,7 @@ internal class HttpRequestParser
 
                 while (MoreTokens())
                 {
-                    if (CurrentToken is { Text: "@" } or { Kind: HttpTokenKind.Word } or { Text: "_" } or {Text: "." })
+                    if (CurrentToken is { Kind: HttpTokenKind.Word } or { Text: "@" or "_" or "." })
                     {
                         ConsumeCurrentTokenInto(node);
                     }
--- a/src/Microsoft.DotNet.Interactive.Http.Tests/ParserTests.Request.cs
+++ b/src/Microsoft.DotNet.Interactive.Http.Tests/ParserTests.Request.cs
@@ -255,6 +255,37 @@ public partial class HttpParserTests
         }
 
         [Fact]
+        public void binding_for_variable_with_period_is_correct()
+        {
+            var result = Parse(
+                """             
+                @host.name = httpbin.org
+                @host.full = https://{{host.name}}                     
+
+                POST {{host.full}}/anything HTTP/1.1
+                content-type: application/json
+
+                {
+                    "name": "sample1",
+                }
+
+                
+                """
+                );
+
+            var requestNode = result.SyntaxTree.RootNode.ChildNodes
+                                    .Should().ContainSingle<HttpRequestNode>().Which;
+
+            var bindingResult = requestNode.TryGetHttpRequestMessage(node =>
+            {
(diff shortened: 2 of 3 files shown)


src/Microsoft.DotNet.Interactive.Http.Parsing/Parsing/HttpRequestParser.cs
```diff
@@ -194,7 +194,7 @@ private HttpVariableDeclarationNode ParseVariableDeclaration()
 
                 while (MoreTokens())
                 {
-                    if (CurrentToken is { Text: "@" } or { Kind: HttpTokenKind.Word } or { Text: "_" })
+                    if (CurrentToken is { Text: "@" } or { Kind: HttpTokenKind.Word } or { Text: "_" } or {Text: "." })
```
jonsequitur, line 197:
```suggestion
                    if (CurrentToken is { Kind: HttpTokenKind.Word } or { Text: "@" or "_" or "." })
```

jonsequitur: It looks like this might also allow variable names with `@` in the middle of the word. Is this intended?

jonsequitur (approved):

src/Microsoft.DotNet.Interactive.Http.Parsing/Parsing/HttpRequestParser.cs
```diff
@@ -194,7 +194,7 @@ private HttpVariableDeclarationNode ParseVariableDeclaration()
 
                 while (MoreTokens())
                 {
-                    if (CurrentToken is { Text: "@" } or { Kind: HttpTokenKind.Word } or { Text: "_" })
+                    if (CurrentToken is { Kind: HttpTokenKind.Word } or { Text: "@" or "_" or "." })
```
shyamnamboodiripad, line 197: Non-blocking: Can this cause conflicts for the [reference syntax for named request variables](https://marketplace.visualstudio.com/items?itemName=humao.rest-client#:~:text=The%20request%20variable%20reference%20syntax%20follows%20%7B%7BrequestName.(response%7Crequest).(body%7Cheaders).(*%7CJSONPath%7CXPath%7CHeader%20Name)%7D%7D) down the line?

For example, if we have a first-class variable named `host.response.body` and another named request variable named `host`, then it is ambiguous whether `host.response.body` refers to the variable or to the body of the remembered response for the named request.

This is probably ok so long as we have proper conflict resolution rules in the expression binding (for example, we could always bind to first-class variables before checking for matching named request variables and / or we could look at the expression to see if JSONPath / XPath is involved and is yes prefer the named request variable)...

shyamnamboodiripad (approved):

src/Microsoft.DotNet.Interactive.Http.Parsing/Parsing/HttpRequestParser.cs
bleaphar: I definitely think it could be a discrepancy in the future, but I personally believe that we should prioritize named requests for expression binding and if there is no named request available, then we should look for first-class variables.

jonsequitur: > This is probably ok so long as we have proper conflict resolution rules in the expression binding (for example, we could always bind to first-class variables before checking for matching named request variables and / or we could look at the expression to see if JSONPath / XPath is involved and is yes prefer the named request variable)...

The potential to have two variables where one silently shadows another could lead to some really hard-to-diagnose bugs for users. I'm not convinced that the ability to have a `.` in a variable name adds value.

shyamnamboodiripad: I tend to agree - the more ideal outcome would be to say that `.` is not valid in names. Most other languages don't allow this either.

Do we know if the variable names with `.` were supported in previous versions of the http editor by accident or was this by design? Do we know how broad the break is - is this just a couple of users here and there who happened to use `.` or was there some bigger reason to do so (e.g., advertised in examples in docs / existing convention for env files etc.)?

Also tagging @phenning

bleaphar: It seems like the period in names was supported erroneously before so users are noticing it as a breaking change with the new parser.

phenning: To add, the reason for the explicitly named variable "winning" over the http-environment declaration of the file is because it is the most local declaration. I think the same thing would apply when the named requests are implemented and suggest that a declaration like

@login.response.headers.x-AuthToken = dsgyds87gss

Would be used in place of the value from 

# @name login.

Simply because if the variable is being declared, we should take the customer at face value and assume that is what they wanted since the explicitly declared it. 

The real world scenario here for this would be in cases where I was using the result in a named request in one API, but wanted to quickly test another specific value without rewriting my query. This would also be consistent with the priority between httpenv.json and locally declared variables.

jonsequitur (approved):

PR #3584, opened 2024-06-21
Named Request Support 

bleaphar:
This PR introduces named request support to dotnet interactive. It will allow the kernel to maintain named requests so that users are able to use the properties from its request and response in future requests. 

**Breakdown**

- HttpNamedRequest -> This is the structure of a named request with the ability to resolve a path that the user is navigating to 
- HttpRequestParser -> This file has been modified to support the parsing of the requests in HttpComments 
- HttpCommentRequestNode -> This file represents a named request signifier in an HttpCommentNode, it also has child nodes for the name node which should be @name and the value node which is the actual name of the named request 
- HttpRootSyntaxNode -> Has been edited to support binding from the kernel. This was to ensure that declared variables can be resolved with the results of a named request which previously would not have been feasible. 
- HttpDiagnostics -> Has been updated to reflect the diagnostic messages associated with the named request 
- Tests -> HttpKernelTests for the parsing and using of a named request (incomplete), ParserTests.NamedRequests reflects the parsing of different named request scenarios, ParserTest s.Variables was updated to support the tuple structure that TryGetDeclaredVariables was updated to


src/Microsoft.DotNet.Interactive.Http.Parsing/Parsing/HttpDiagnostics.cs
```diff
@@ -173,4 +175,67 @@ internal static HttpDiagnosticInfo IncorrectDateTimeCustomFormat(string format)
             """The supplied format '{0}' is invalid.""";
         return new HttpDiagnosticInfo(id, messageFormat, severity, format);
     }
+
+    internal static HttpDiagnosticInfo InvalidNamedRequestPath(string expression)
+    {
+        var id = $"HTTP0021";
+        var severity = DiagnosticSeverity.Error;
+        var messageFormat =
+            """The supplied expression '{0}' does not follow the correct pattern. The expression should adhere to following pattern: {{requestName.(response|request).(body|headers).(*|JSONPath|XPath|Header Name)}}.""";
```
jonsequitur, line 184:
```suggestion
            """The supplied expression '{0}' does not follow the correct pattern. The expression should adhere to the following pattern: {{requestName.(response|request).(body|headers).(*|JSONPath|XPath|Header Name)}}.""";
```

src/Microsoft.DotNet.Interactive.Http.Parsing/Parsing/HttpDiagnostics.cs
```diff
@@ -173,4 +175,67 @@ internal static HttpDiagnosticInfo IncorrectDateTimeCustomFormat(string format)
...
+            """The supplied named request '{0}' does not have any headers.""";
+        return new HttpDiagnosticInfo(id, messageFormat, severity, name);
+    }
+
+    internal static HttpDiagnosticInfo InvalidXmlNodeInNamedRequest(string xmlNode)
+    {
+        var id = $"HTTP0024";
+        var severity = DiagnosticSeverity.Error;
+        var messageFormat =
+            """The supplied xml name '{0}' does not exist in the named request.""";
```
jonsequitur, line 220:
```suggestion
            """The supplied XML name '{0}' does not exist in the named request.""";
```

src/Microsoft.DotNet.Interactive.Http.Parsing/Parsing/HttpDiagnostics.cs
```diff
@@ -173,4 +175,67 @@ internal static HttpDiagnosticInfo IncorrectDateTimeCustomFormat(string format)
...
+        return new HttpDiagnosticInfo(id, messageFormat, severity, xmlNode);
+    }
+
+    internal static HttpDiagnosticInfo InvalidContentType(string contentType, string content)
+    {
+        var id = $"HTTP0025";
+        var severity = DiagnosticSeverity.Error;
+        var messageFormat =
+            """The supplied named request has content type of '{0}' which differs from the content type needed of '{1}'.""";
```
jonsequitur, line 229:
```suggestion
            """The supplied named request has content type of '{0}' which differs from the required content type of '{1}'.""";
```

src/Microsoft.DotNet.Interactive.Http.Parsing/Parsing/HttpRequestNode.cs
```diff
@@ -81,31 +86,52 @@ public void Add(HttpBodyNode node)
...
     {
         var originalBind = bind;
-        var declaredVariables = SyntaxTree?.RootNode.GetDeclaredVariables();
-        if (declaredVariables?.Count > 0)
+        var diagnostics = new List<Diagnostic>(base.GetDiagnostics());
+
+        if (SyntaxTree is not null)
         {
-            bind = node =>
+            (Dictionary<string, DeclaredVariable> declaredVariables, List<Diagnostic>? diagnostics) declaredVariableResults = SyntaxTree.RootNode.TryGetDeclaredVariables(originalBind);
```
jonsequitur, line 109: Why is this cast needed?

src/Microsoft.DotNet.Interactive.Http.Parsing/Parsing/HttpRootSyntaxNode.cs
```diff
@@ -62,21 +66,41 @@ public void Add(HttpRequestSeparatorNode separatorNode)
...
                             return DynamicExpressionUtilities.ResolveExpressionBinding(node, node.Text);
                         }
 
                     });
 
-                    if (value is not null && value.Value is not null)
+                    if (value is not null)
                     {
-                        declaredVariables[node.DeclarationNode.VariableName] = new DeclaredVariable(node.DeclarationNode.VariableName, value.Value, value);
+                        if(value.Value is not null)
```
jonsequitur, line 82: The two `if` statements can be combined into one:

```csharp
if (value?.Value != null)
```

src/Microsoft.DotNet.Interactive.Http.Tests/HttpKernelTests.cs
```diff
@@ -2122,15 +2122,169 @@ public async Task when_response_is_slow_and_large_final_displayed_value_includes
...
     {
         // Request Variables
         // Request variables are similar to file variables in some aspects like scope and definition location.However, they have some obvious differences.The definition syntax of request variables is just like a single-line comment, and follows // @name requestName or # @name requestName just before the desired request url. 
 
+        var client = new HttpClient();
+        using var kernel = new HttpKernel(client: client);
 
-        // TODO (responses_to_named_requests_can_be_accessed_as_symbols_in_later_requests) write test
-        throw new NotImplementedException();
+        using var _ = new AssertionScope();
```
jonsequitur, line 2134: In this case you probably don't need the `AssertionScope`. Your first assertion seems like a kind of sanity check and there's no information being lost if the test stops there. `AssertionScope` is most useful when you have more than one assertion that can fail independently of one another, e.g. property checks on a single object.

src/Microsoft.DotNet.Interactive.Http.Tests/ParserTests.NamedRequest.cs
```diff
@@ -0,0 +1,233 @@
...
+
+
+                ###
+                """;
+
+            var result = Parse(code);
+
+            var namedRequest = result.SyntaxTree.RootNode.DescendantNodesAndTokens().OfType<HttpCommentNamedRequestNode>().Single();
+
+            namedRequest.ValueNode.Text.Should().Be("login");
```
jonsequitur, line 65: The node type names could be clearer I think. I would think the "name node" should be referring to `login`. In other variable syntax nodes, the name node refers to the name of the variable, and the value refers to the value of the variable. 

A quick discussion to sort out naming might be useful.

src/Microsoft.DotNet.Interactive.Http.Tests/ParserTests.NamedRequest.cs
```diff
@@ -0,0 +1,233 @@
...
+
+            var result = Parse(code);
+
+            var namedRequest = result.SyntaxTree.RootNode.DescendantNodesAndTokens().OfType<HttpCommentNamedRequestNode>().Single();
+
+            namedRequest.GetDiagnostics().Should().HaveCount(1);   
+        }
+
+        [Fact]
+        public void named_request_with_special_characters_produces_an_error()
```
jonsequitur, line 95: This is just testing one character. It might be useful to make this a theory test with a broader set of disallowed characters.

src/Microsoft.DotNet.Interactive.Http.Tests/ParserTests.NamedRequest.cs
```diff
@@ -0,0 +1,233 @@
...
+
+
+                ###
+                """;
+
+            var result = Parse(code);
+
+            var namedRequest = result.SyntaxTree.RootNode.DescendantNodesAndTokens().OfType<HttpCommentNamedRequestNode>().Single();
+
+            namedRequest.GetDiagnostics().Should().HaveCount(1);
```
jonsequitur, line 117: There are a lot of possibilities for false positives. It's good to assert on the message and position of the diagnostic.

src/Microsoft.DotNet.Interactive.Http.Tests/ParserTests.NamedRequest.cs
```diff
@@ -0,0 +1,233 @@
...
+
+            var result = Parse(code);
+
+            var namedRequest = result.SyntaxTree.RootNode.DescendantNodesAndTokens().OfType<HttpCommentNamedRequestNode>().Single();
+
+            namedRequest.ValueNode.Text.Should().Be("login");
+        }
+
+        [Fact]
+        public void named_request_with_slash_for_comment_start()
```
jonsequitur, line 152: Test names are more informative when they express the expectation. Maybe something like `named_request_comments_can_use_double_slash_prefix`.

src/Microsoft.DotNet.Interactive.Http.Tests/ParserTests.NamedRequest.cs
```diff
@@ -0,0 +1,233 @@
...
+
+            var result = Parse(code);
+
+            var namedRequest = result.SyntaxTree.RootNode.DescendantNodesAndTokens().OfType<HttpCommentNamedRequestNode>().Single();
+
+            namedRequest.ValueNode.Text.Should().Be("login");
+        }
+
+        [Fact]
+        public void comments_before_named_request()
```
jonsequitur, line 176: This one could also convey more about the expected behavior.

src/Microsoft.DotNet.Interactive.Http.Parsing/Parsing/HttpRootSyntaxNode.cs
```diff
@@ -36,11 +38,13 @@ public void Add(HttpRequestSeparatorNode separatorNode)
         AddInternal(separatorNode);
     }
 
-    public Dictionary<string, DeclaredVariable> GetDeclaredVariables()
+    public (Dictionary<string, DeclaredVariable> declaredVariables, List<Diagnostic>? diagnostics) TryGetDeclaredVariables(HttpBindingDelegate? bind = null)
```
jonsequitur, line 41: It's usually good to avoid tuples as return types on public methods. This might also be worth a little discussion.

src/Microsoft.DotNet.Interactive.Http/Model/HttpNamedRequest.cs
```diff
@@ -171,19 +180,24 @@ internal HttpNamedRequest(HttpRequestNode httpRequestNode, HttpResponse response
...
-                    var xmlNodes = xmlDoc.SelectNodes(path[3].Substring(1));
-
-                    if (xmlNodes is not null && xmlNodes.Count == 1)
+                    try
                     {
-                        return node.CreateBindingSuccess(xmlNodes.Item(0)?.Value);
+                        //Remove the leading slash
+                        var xmlNodes = xmlDoc.SelectNodes(path[3].Substring(1));
+
+                        if (xmlNodes is not null && xmlNodes.Count == 1)
```
jonsequitur, line 188:
```suggestion
                        if (xmlNodes is { Count : 1} )
```

This suggestion effectively combines the null check and the property check.

src/Microsoft.DotNet.Interactive.Http/Model/HttpNamedRequest.cs
```diff
@@ -171,19 +180,24 @@ internal HttpNamedRequest(HttpRequestNode httpRequestNode, HttpResponse response
...
+                        {
+                            return node.CreateBindingSuccess(xmlNodes.Item(0)?.Value);
+                        }
+                        else
+                        {
+                            return node.CreateBindingFailure(HttpDiagnostics.InvalidXmlNodeInNamedRequest(path[3]));
+                        }
                     }
-                    else
+                    catch
```
jonsequitur, line 197: Do you have an example of the kinds of exceptions that will be caught here?

It's always preferable to catch specific exceptions to avoid hiding bugs.

src/Microsoft.DotNet.Interactive.Http/HttpKernel.cs
```diff
@@ -202,6 +227,16 @@ async Task SendRequestAndHandleResponseAsync()
         {
             var response = await GetResponseWithTimingAsync(requestMessage, cancellationToken);
 
+            if(requestNode != null)
```
jonsequitur, line 230: The newer syntax `requestNode is not null` is generally preferable because, by avoiding the possibility of an operator overload causing a surprising behavior, it's a bit more robust.

jonsequitur:
The combinatorial tests should probably be updated to include this new syntax.

If you'd like to find some time to pair on it, let me know.

src/Microsoft.DotNet.Interactive.Http.Tests/ParserTests.Combinatorial.cs
```diff
@@ -198,7 +204,22 @@ public static IEnumerable<object[]> GenerateInvalidRequests()
...
+                    new HttpRequestNodeSyntaxSpec(namedRequest, method, url, version, headerSection, bodySection),
+                    generationNumber
+                };
+            }
+
+            foreach (var namedRequest in InvalidNamedRequests())
+            foreach (var method in ValidMethods())
+            foreach (var url in ValidUrls())
+            foreach (var version in ValidVersions())
+            foreach (var headerSection in InvalidHeaderSections())
```
jonsequitur, line 216: If you look at the other examples in this method, each block of nested `foreach` loops includes only a single invalid element. You have both `InvalidNamedRequests` and `InvalidHeaderSections` in the same loop. Is that intentional?

src/Microsoft.DotNet.Interactive.Http/HttpKernel.cs
```diff
@@ -124,7 +124,18 @@ async Task IKernelCommandHandler<SendValue>.HandleAsync(SendValue command, Kerne
 
     private Task SetValueAsync(string valueName, object value, Type? declaredType = null)
     {
-        _variables[valueName] = value;
+        if(value is (HttpExpressionNode))
+        {
+            var bindingResult = BindExpressionValues((HttpExpressionNode) value);
+            if(bindingResult.IsSuccessful && bindingResult.Value is not null)
```
jonsequitur, line 130: Why are we not allowing `null`?

src/Microsoft.DotNet.Interactive.Http/HttpKernel.cs
```diff
@@ -124,13 +124,18 @@ async Task IKernelCommandHandler<SendValue>.HandleAsync(SendValue command, Kerne
 
     private Task SetValueAsync(string valueName, object value, Type? declaredType = null)
     {
-        if(value is (HttpExpressionNode))
+        if(value is (HttpVariableDeclarationAndAssignmentNode))
```
jonsequitur, line 127:
```suggestion
        if(value is HttpVariableDeclarationAndAssignmentNode)
```

jonsequitur (approved):

PR #3586, opened 2024-06-24
Add support for configuring request timeout on HttpKernel

shyamnamboodiripad:


src/Microsoft.DotNet.Interactive.Http/HttpKernel.cs
```diff
@@ -55,13 +57,36 @@ static HttpKernel()
...
+    /// <paramref name="requestTimeoutInMilliseconds"/>. The shorter of the two timeouts will apply in such cases.
+    /// </para>
+    /// <para>
+    /// Note that <see cref="HttpClient.Timeout"/> has a default value of 100 seconds. If you wish to control the
+    /// timeout exclusively via <paramref name="requestTimeoutInMilliseconds"/>, either omit the 
+    /// <see cref="HttpClient"/> parameter in the constructor, or pass in an <see cref="HttpClient"/> with
+    /// <see cref="HttpClient.Timeout"/> set to <see cref="Timeout.InfiniteTimeSpan"/>.
+    /// </para>
+    /// </remarks>
+    public void SetRequestTimeout(int requestTimeoutInMilliseconds)
```
jonsequitur, line 87: It seems like this should be a property with both a getter and a setter and, given the usage below, changed to a `TimeSpan`.

jonsequitur (approved):

PR #3604, opened 2024-07-19
Magic command hover text plus some bug fixes

jonsequitur:
This PR adds support for hover text in magic commands.

It also includes fixes for a couple of issues with the new magic command parser:

- [x] Inputs trigger user prompts when certain language service commands are sent. They should only do this for `SubmitCode`.
- [x] Name the previously-nameless parameter for `#!import` to `--file`
- [x] Fix the description for the `#!import --file`  parameter


src/Microsoft.DotNet.Interactive.Tests/InputsWithinMagicCommandsTests.cs
```diff
@@ -237,6 +237,46 @@ public async Task Additional_properties_of_input_request_are_set_by_input_proper
...
+            requestInputWasSent = true;
+
+            return Task.CompletedTask;
+        });
+
+        var result = await kernel.SendAsync(command);
+
+        result.Events.Should().NotContainErrors();
+
+        requestInputWasSent.Should().BeFalse();
```
shyamnamboodiripad, line 259: consider: Add an AssertionScope since there are multiple asserts

src/Microsoft.DotNet.Interactive.Tests/InputsWithinMagicCommandsTests.cs
```diff
@@ -237,6 +237,46 @@ public async Task Additional_properties_of_input_request_are_set_by_input_proper
         requestInput.InputTypeHint.Should().Be("file");
     }
 
+    [Theory]
+    [MemberData(nameof(LanguageServiceCommands))]
+    public async Task Language_service_commands_do_not_trigger_input_requests(KernelCommand command)
```
shyamnamboodiripad, line 242: If this test fails - will this display the name of the command that is failing? It would be nice if we could do something so that it is easier to know the command (and code) for the failure case

jonsequitur: Yes, test output shows the parameters for the failing case.

src/Microsoft.DotNet.Interactive.Tests/InputsWithinMagicCommandsTests.cs
jonsequitur: Typically that's only useful if the multiple asserts can fail independently. The first assert here is more of a sanity check.

src/Microsoft.DotNet.Interactive/Kernel.cs
```diff
@@ -752,6 +751,64 @@ public void RegisterForDisposal(IDisposable disposable)
...
+                    break;
+
+                case DirectiveParameterNameNode directiveParameterNameNode:
+                    if (directiveParameterNameNode.Parent is DirectiveParameterNode pn &&
+                        pn.TryGetParameter(out var parameter))
+                    {
+                        hoverText = parameter.Description;
+                    }
+
+                    break;
```
shyamnamboodiripad, line 790: Should we add a `defualt:` case to throw for unexpected node type?

src/Microsoft.DotNet.Interactive/Parsing/DirectiveNode.cs
```diff
@@ -382,16 +382,19 @@ void WriteProperty(string propertyName, string value)
                 {
                     var bindingResult = await bind(expressionNode);
 
-                    if (bindingResult.IsSuccessful)
+                    if (bindingResult is not null)
```
shyamnamboodiripad, line 385: When can this return `null`? Should an error be reported (or an exception thrown) when `null` is returned?

src/Microsoft.DotNet.Interactive/Parsing/SubmissionParser.cs
```diff
@@ -481,10 +485,17 @@ private PolyglotParserConfiguration GetParserConfiguration()
 
         if (expressionType is "input" or "password")
         {
-            var parametersNode = expressionNode.ChildNodes.OfType<DirectiveExpressionParametersNode>().SingleOrDefault();
+            if (command is SubmitCode)
```
shyamnamboodiripad, line 488: Is this always going to be a `SubmitCode`? If yes, should this throw an exception if some other command is present?

src/Microsoft.DotNet.Interactive/Parsing/SubmissionParser.cs
```diff
@@ -555,6 +567,8 @@ private PolyglotParserConfiguration GetParserConfiguration()
         bool allowByRef,
         DirectiveExpressionNode expressionNode)
     {
+        // FIX: (RequestSingleValueFromKernel) make sure this is not dispatched on language service commands
```
shyamnamboodiripad, line 570: Did you intend to remove the `// FIX` comments before merge or is this something that needs to be fixed up in follow up PRs?

shyamnamboodiripad (approved):

src/Microsoft.DotNet.Interactive.Tests/InputsWithinMagicCommandsTests.cs
shyamnamboodiripad: Ah was asking if we can tell by looking at the view in the test runner itself without clicking to inspect the output. But not a huge issue either way.

src/Microsoft.DotNet.Interactive/Parsing/DirectiveNode.cs
jonsequitur: Maybe it can be modeled differently but we don't want to perform binding during language service commands, since the resulting directive command won't actually be invoked.

src/Microsoft.DotNet.Interactive/Kernel.cs
jonsequitur: No. It's not an exceptional case. Some node types don't (currently) support hover text.

src/Microsoft.DotNet.Interactive.Tests/InputsWithinMagicCommandsTests.cs
jonsequitur: Yes, most test runners support this.

src/Microsoft.DotNet.Interactive/Parsing/SubmissionParser.cs
jonsequitur: I'll remove this one since it was resolved.

src/Microsoft.DotNet.Interactive/Parsing/SubmissionParser.cs
jonsequitur: Submission splitting happens for many different command types, including language service commands. We don't want the non-`SubmitCode` commands to trigger user prompts though.

Issue #3323, opened 2023-11-15
Auto-recall for input prompts


jonsequitur:
The `@input` and `@password` prompts today don’t have a way to recall values that were entered on past runs. Adding a gesture to instruct the prompt to save its value could be a helpful ergonomic improvement even for values that aren’t secrets. For example, many tutorial notebooks ask for a file path on disk, and for any given user, that file path is probably not going to change from run to run. Restarting the kernel is a common notebook interaction and this flushes the in-memory prompt results, requiring the user to re-enter them. Enabling the prompt to check for a remembered response and re-use it without re-issuing the prompt to the user could help make many notebook workflows faster and smoother.

There are several ways we could approach implementing this. 

Most inputs stored this way would be stored in the user’s profile as plain text files (likely containing JSON), with the exception of values ingested by the `@password` token. The current best candidate for storing these would be the PowerShell [SecretManagement and SecretStore](https://devblogs.microsoft.com/powershell/secretmanagement-and-secretstore-are-generally-available/) modules, but other possibilities can be evaluated. 

Matching could be done based on the complete text of a given magic command. So, for example, the @input prompts for each of these two examples would never access a value stored by the other:

```
#!one --tenant-id @input:"Enter your tenant id" 
#!two --tenant-id @input:"Enter your tenant id"
```

There are a few different UX flows that could work for storage and recall:

•	The user could be prompted to store the value.
•	The user could be prompted to recall the value.
•	There could be a way to make storage and recall the default behavior in order to further reduce redundant interactions. (For example, there could be a separate magic command that sets this preference at the notebook level.)

marckruzik:
It would be interesting for the developer to be able to set a default or suggested value, especially for `input`. Something like:
`#!value --from-value @input:"please enter a filename" --default "readme.txt"`
This way, the prompt already contains a default value, and you can just validate the prompt by pushing Enter.

This behavior already exists with a lot of dialog windows, such as [FileDialog.FileName](https://learn.microsoft.com/en-us/dotnet/api/system.windows.forms.filedialog.filename), to provide a default file name.
Those default values would be superseded by the behavior described in the first post.

I'm talking about that now, as it is close to the subject.
Please tell me if this suggestion needs its own suggestion post.

jonsequitur:
@marckruzik, this is a very good suggestion and I think it's fine to discuss here. 

[...]

PR #3611, opened 2024-07-24
Add support for auto-recall for input prompts

jonsequitur:
This PR addresses #3323. 

⁠I've also removed the `RequestInput.ValueName` property. It is no longer needed. It's a minor breaking change for people directly using the input request APIs, but has no effect on magic command usage.


colombod (commented):
looks good to me so far

colombod (approved):

PR #3620, opened 2024-07-31
Don't treat #r "<assembly path>" as #r "nuget:<package reference>" in the case where <assembly path> happens to contain the word 'nuget'

### LINKED ISSUE #3617: Executing notebook cells reports the following error: DNI120: Unable to parse package reference: "C:/Users/.../.nuget/packages/...Microsoft.DotNet.Interactive.Jupyter.dll"
We are occasionally running into this error when executing notebook cells in latest internal preview bits:

Image

The issue is a race and based on some debugging that @jonsequitur has been doing, the race only seems to repro on release bits.

We have a suspicion that this only happens if we try to execute a cell soon after a notebook is opened / soon after restarting kernels. However, iirc I also ran into it a couple of times after waiting for 1 minute+ after opening a notebook.

This error also causes an initialization failure when opening the http editor in VS if I switch it over to use the latest builds of the dotnet-interactive tool.

shyamnamboodiripad:
This fixes the issue where we end up doing the wrong thing in the case where `<assembly path>` points to a path under the user's '.nuget' folder.

Fixes #3617

### DIFF
--- a/src/Microsoft.DotNet.Interactive.Tests/LanguageKernelPackageTests.cs
+++ b/src/Microsoft.DotNet.Interactive.Tests/LanguageKernelPackageTests.cs
@@ -235,6 +235,52 @@ Formatter.Register<DataFrame>((df, writer) =>
             .Be("(1,4): error DNI210: Unable to parse package reference: \"nuget:\"");
     }
 
+    [Fact]
+    public async Task Pound_r_is_not_treated_as_pound_r_nuget_by_csharp_kernel_if_assembly_path_happens_to_contain_the_string_nuget()
+    {
+        var kernel = CreateCSharpKernel();
+
+        using var events = kernel.KernelEvents.ToSubscribedList();
+
+        await kernel.SubmitCodeAsync("""
+                                     #r "C:\Users\abcde\.nuget\packages\package\1.0.0\package.dll"
+                                     """);
+
+        events
+            .Should()
+            .ContainSingle<CommandFailed>()
+            .Which
+            .Message
+            .Should()
+            .Contain("Metadata file")
+            .And
+            .Contain("could not be found")
+            .And
+            .NotContain("DNI");
+    }
+
+    [Fact]
+    public async Task Pound_r_is_not_treated_as_pound_r_nuget_by_fsharp_kernel_if_assembly_path_happens_to_contain_the_string_nuget()
+    {
+        var kernel = CreateKernel(Language.FSharp);
+
+        using var events = kernel.KernelEvents.ToSubscribedList();
+
+        await kernel.SubmitCodeAsync("""
+                                     #r "C:\Users\abcde\.nuget\packages\package\1.0.0\package.dll"
+                                     """);
+
+        events
+            .Should()
+            .ContainSingle<CommandFailed>()
+            .Which
+            .Message
+            .Should()
+            .Contain("is not a valid assembly name")
+            .And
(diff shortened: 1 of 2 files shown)


src/Microsoft.DotNet.Interactive.Tests/LanguageKernelPackageTests.cs
```diff
@@ -235,6 +235,52 @@ public async Task Pound_r_nuget_disallows_empty_package_specification(Language l
...
+
+        events
+            .Should()
+            .ContainSingle<CommandFailed>()
+            .Which
+            .Message
+            .Should()
+            .Contain("Metadata file")
+            .And
+            .Contain("could not be found")
```
jonsequitur, line 257: You can check for multiple substrings at once using `ContainAll`.

src/Microsoft.DotNet.Interactive.Tests/LanguageKernelPackageTests.cs
```diff
@@ -235,6 +235,52 @@ public async Task Pound_r_nuget_disallows_empty_package_specification(Language l
...
+            .And
+            .NotContain("DNI");
+    }
+
+    [Fact]
+    public async Task Pound_r_is_not_treated_as_pound_r_nuget_by_fsharp_kernel_if_assembly_path_happens_to_contain_the_string_nuget()
+    {
+        var kernel = CreateKernel(Language.FSharp);
+
+        using var events = kernel.KernelEvents.ToSubscribedList();
```
jonsequitur, line 267: It's generally best to avoid testing using `kernel.KernelEvents` and instead check the `KernelCommandResult.Events` for the specific command you're sending.

src/Microsoft.DotNet.Interactive.Tests/LanguageKernelPackageTests.cs
shyamnamboodiripad: Fixed

src/Microsoft.DotNet.Interactive.Tests/LanguageKernelPackageTests.cs
shyamnamboodiripad: Fixed

src/Microsoft.DotNet.Interactive.Tests/LanguageKernelPackageTests.cs
```diff
@@ -235,6 +235,50 @@ public async Task Pound_r_nuget_disallows_empty_package_specification(Language l
...
+
+    [Fact]
+    public async Task Pound_r_is_not_treated_as_pound_r_nuget_by_fsharp_kernel_if_assembly_path_happens_to_contain_the_string_nuget()
+    {
+        var kernel = CreateKernel(Language.FSharp);
+
+        var result =
+            await kernel.SubmitCodeAsync(
+                """
+                #r "C:\Users\abcde\.nuget\packages\package\1.0.0\package.dll"
```
jonsequitur, line 268:
```suggestion
                #r @"C:\Users\abcde\.nuget\packages\package\1.0.0\package.dll"
```

shyamnamboodiripad: Thanks for catching - fixed (and also changed the expected error)

jonsequitur (approved):

PR #3505, opened 2024-04-03
Fix issue where custom datetime format strings result in binding failure

phenning:
Also fix issue with always using UtcNow, even with $localDatetime.

Add tests

PR #3626, opened 2024-08-05
Fix issue with $localDatetime using UTC.

phenning:
Fixes WebTools AZDo issue 2187534

### DIFF
--- a/src/Microsoft.DotNet.Interactive.Http.Tests/ParserTests.DynamicExpressions.cs
+++ b/src/Microsoft.DotNet.Interactive.Http.Tests/ParserTests.DynamicExpressions.cs
@@ -17,8 +17,6 @@ public partial class HttpParserTests
         [Fact]
         public void can_bind_datetime_with_custom_format()
         {
-            using var _ = new AssertionScope();
-
             var expression = "$datetime 'dd-MM-yyyy' 1 d";
             var code = $@"@var = {{{{{expression}}}}}";
 
@@ -33,8 +31,6 @@ public partial class HttpParserTests
         [Fact]
         public void can_bind_datetime_uses_utcnow()
         {
-            using var _ = new AssertionScope();
-
             var expression = "$datetime 'Tzz'";
             var code = $@"@var = {{{{{expression}}}}}";
 
@@ -52,8 +48,6 @@ public partial class HttpParserTests
         [Fact]
         public void can_bind_datetime_with_iso8601_format()
         {
-            using var _ = new AssertionScope();
-
             var expression = "$datetime iso8601";
             var code = $@"@var = {{{{{expression}}}}}""";
 
@@ -68,8 +62,6 @@ public partial class HttpParserTests
         [Fact]
         public void can_bind_datetime_with_iso8601_format_with_offset()
         {
-            using var _ = new AssertionScope();
-
             var expression = "$datetime iso8601 1 y";
             var code = $@"@var = {{{{{expression}}}}}""";
 
@@ -84,8 +76,6 @@ public partial class HttpParserTests
         [Fact]
         public void can_bind_datetime_with_rfc1123_format()
         {
-            using var _ = new AssertionScope();
-
             var expression = "$datetime rfc1123";
             var code = $@"@var = {{{{{expression}}}}}""";
 
@@ -100,8 +90,6 @@ public partial class HttpParserTests
         [Fact]
         public void can_bind_datetime_with_rfc1123_format_with_offset()
         {
-            using var _ = new AssertionScope();
-
             var expression = "$datetime rfc1123 1 d";
             var code = $@"@var = {{{{{expression}}}}}""";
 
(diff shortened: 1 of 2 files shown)


phenning:
@shyamnamboodiripad @bleaphar

jonsequitur:
Can you provide a test to pin the expected behavior?

phenning:
Sure it looks like we already have some that were added in a previous PR, but didn’t use the full time likely due to not being able to get an accurate asset time, so we were hiding this bug. 

I’ll update the test to at least consider the hour in the format string that should reduce the likelihood of random failures as opposed to using a full datetime. 

https://github.com/dotnet/interactive/pull/3505/files

src/Microsoft.DotNet.Interactive.Http.Tests/ParserTests.DynamicExpressions.cs
```diff
@@ -30,6 +30,25 @@ public void can_bind_datetime_with_custom_format()
             binding.Value.As<string>().Should().Be(currentTime.AddDays(1.0).ToString("dd-MM-yyyy").ToString());
         }
 
+        [Fact]
+        public void can_bind_datetime_uses_utcnow()
+        {
+            using var _ = new AssertionScope();
```
jonsequitur, line 36: The `AssertionScope` is not needed here. It's only something we typically do when there are multiple, non-interdependent assertions in the test, so that we can see the results of all of them even if an earlier one fails.

src/Microsoft.DotNet.Interactive.Http.Tests/ParserTests.DynamicExpressions.cs
```diff
@@ -30,6 +30,25 @@ public void can_bind_datetime_with_custom_format()
...
+
+            var result = HttpRequestParser.Parse(code);
+            var currentTime = DateTimeOffset.UtcNow;
+            var node = result.SyntaxTree.RootNode.DescendantNodesAndTokens().OfType<HttpExpressionNode>().Single();
+
+            // Since we want to make sure that the provided expression is evaluated
+            // and .Now or .UtcNow is used as appropriate, we do not pass our own offset, like we do with other unit tests.
+            // we can do this since we are passing a custom format which only includes the time zone.
+            var binding = DynamicExpressionUtilities.ResolveExpressionBinding(node, expression);
+            binding.Value.As<string>().Should().Be(currentTime.ToString("Tzz").ToString());
```
jonsequitur, line 49:
```suggestion
            binding.Value.As<string>().Should().Be(currentTime.ToString("Tzz"));

```

src/Microsoft.DotNet.Interactive.Http.Tests/ParserTests.DynamicExpressions.cs
```diff
@@ -110,6 +129,25 @@ public void can_bind_local_datetime_with_custom_format()
             binding.Value.As<string>().Should().Be(currentTime.AddDays(1.0).ToString("dd-MM-yyyy").ToString());
         }
 
+        [Fact]
+        public void can_bind_local_datetime_uses_now()
+        {
+            using var _ = new AssertionScope();
```
jonsequitur, line 135: Same.

src/Microsoft.DotNet.Interactive.Http.Tests/ParserTests.DynamicExpressions.cs
```diff
@@ -110,6 +129,25 @@ public void can_bind_local_datetime_with_custom_format()
...
+
+            var result = HttpRequestParser.Parse(code);
+            var currentTime = DateTimeOffset.Now;
+            var node = result.SyntaxTree.RootNode.DescendantNodesAndTokens().OfType<HttpExpressionNode>().Single();
+
+            // Since we want to make sure that the provided expression is evaluated
+            // and .Now or .UtcNow is used as appropriate, we do not pass our own offset, like we do with other unit tests.
+            // we can do this since we are passing a custom format which only includes the time zone.
+            var binding = DynamicExpressionUtilities.ResolveExpressionBinding(node, expression);
+            binding.Value.As<string>().Should().Be(currentTime.ToString("Tzz").ToString());
```
jonsequitur, line 148:
```suggestion
            binding.Value.As<string>().Should().Be(currentTime.ToString("Tzz"));

```

phenning: Thanks, looks like the other tests had this as well, fixing them all up.

src/Microsoft.DotNet.Interactive.Http.Tests/ParserTests.DynamicExpressions.cs
phenning: Ah, just followed the pattern in the other tests - is it worth it to remove, or ok to leave?

phenning: I'll go ahead and remove from all the unit tests that dont require it.

jonsequitur (approved):

PR #3642, opened 2024-08-16
Support for saving named requests in variables and improved json parsing

bleaphar:
This PR introduces variable support for the kernel and changes the flow of Json grandchildren parsing 

Highlights: 

- HandleAsync looks if a submit code has the entire associated document in which case it will use that in creating declared variables which enables the support for saving the named request values in variables 
- Changed the chaining of the Json parsing that at the last node to evaluate one of the derived types and then convert to a string 
- Miscellaneous reformatting in the request parser 

PR: #3632

### DIFF
--- a/src/Microsoft.DotNet.Interactive.Http.Parsing/Parsing/HttpRequestParser.cs
+++ b/src/Microsoft.DotNet.Interactive.Http.Parsing/Parsing/HttpRequestParser.cs
@@ -137,7 +137,7 @@ internal class HttpRequestParser
             {
                 if (node is null)
                 {
-                    if (CurrentToken is { Kind: TokenKind.Word } or ({ Kind: TokenKind.Punctuation } and { Text: "/" }))
+                    if (CurrentToken is { Kind: TokenKind.Word } or ({ Kind: TokenKind.Punctuation } and ({ Text: "/" } or {Text: "'" } or { Text: "\""})))
                     {
                         node = new HttpVariableValueNode(_sourceText, _syntaxTree);
 
--- a/src/Microsoft.DotNet.Interactive.Http.Parsing/Parsing/HttpRequestParser.cs
+++ b/src/Microsoft.DotNet.Interactive.Http.Parsing/Parsing/HttpRequestParser.cs
@@ -671,7 +671,7 @@ internal class HttpRequestParser
             while (MoreTokens() && CurrentToken is not { Kind: TokenKind.NewLine })
             {
 
-                if (CurrentToken is not null && (!(CurrentToken is { Kind: TokenKind.Word or TokenKind.Whitespace } or { Text: "_" or "@" or "." }) || CurrentToken is { Kind: TokenKind.Word } && wordParsedOnce))
+                if (CurrentToken is not ({ Kind: TokenKind.Word or TokenKind.Whitespace } or { Text: "_" or "@" or "." }) || CurrentToken is { Kind: TokenKind.Word } && wordParsedOnce)
                 {
                     var diagnostic = CurrentToken.CreateDiagnostic(HttpDiagnostics.InvalidNamedRequestName());
                     node.AddDiagnostic(diagnostic);
--- a/src/Microsoft.DotNet.Interactive.Http.Parsing/Parsing/HttpRequestParser.cs
+++ b/src/Microsoft.DotNet.Interactive.Http.Parsing/Parsing/HttpRequestParser.cs
@@ -668,12 +668,13 @@ internal class HttpRequestParser
                 node.AddDiagnostic(diagnostic);
             }
             bool wordParsedOnce = false;
-            while (MoreTokens() && CurrentToken is not { Kind: TokenKind.NewLine })
+            while (MoreTokens() && CurrentToken is not { Kind: TokenKind.NewLine } or null)
             {
-
-                if (CurrentToken is not ({ Kind: TokenKind.Word or TokenKind.Whitespace } or { Text: "_" or "@" or "." }) || CurrentToken is { Kind: TokenKind.Word } && wordParsedOnce)
+                var currentToken = CurrentToken;
+                if (currentToken is not null && (currentToken is not ({ Kind: TokenKind.Word or TokenKind.Whitespace } or 
+                    { Text: "_" or "@" or "." }) || currentToken is { Kind: TokenKind.Word } && wordParsedOnce))
                 {
-                    var diagnostic = CurrentToken.CreateDiagnostic(HttpDiagnostics.InvalidNamedRequestName());
+                    var diagnostic = currentToken.CreateDiagnostic(HttpDiagnostics.InvalidNamedRequestName());
                     node.AddDiagnostic(diagnostic);
                     wordParsedOnce = false;
                 }
--- a/src/Microsoft.DotNet.Interactive.Http.Parsing/Parsing/HttpRequestParser.cs
+++ b/src/Microsoft.DotNet.Interactive.Http.Parsing/Parsing/HttpRequestParser.cs
@@ -137,7 +137,7 @@ internal class HttpRequestParser
             {
                 if (node is null)
                 {
-                    if (CurrentToken is { Kind: TokenKind.Word } or ({ Kind: TokenKind.Punctuation} and { Text: "/"}))
+                    if (CurrentToken is { Kind: TokenKind.Word } or ({ Kind: TokenKind.Punctuation } and { Text: "/" }))
                     {
                         node = new HttpVariableValueNode(_sourceText, _syntaxTree);
 
@@ -268,7 +268,7 @@ internal class HttpRequestParser
                     foreach (var commentNode in ParseComments())
                     {
                         node.Add(commentNode, addBefore: true);
-                        
+
                     }
                 }
                 else
@@ -668,25 +668,21 @@ internal class HttpRequestParser
                 node.AddDiagnostic(diagnostic);
             }
             bool wordParsedOnce = false;
-            while (MoreTokens() && CurrentToken is not {Kind: TokenKind.NewLine })
-                {
+            while (MoreTokens() && CurrentToken is not { Kind: TokenKind.NewLine })
+            {
 
-                    if (CurrentToken is not null && (!(CurrentToken is { Kind: TokenKind.Word or TokenKind.Whitespace} or { Text: "_" or "@" or "."}) || CurrentToken is {Kind: TokenKind.Word } && wordParsedOnce)) 
-                    {
-                        var diagnostic = CurrentToken.CreateDiagnostic(HttpDiagnostics.InvalidNamedRequestName());
-                        node.AddDiagnostic(diagnostic);
-                        wordParsedOnce = false;
-                    }
-
-                    if (CurrentToken is { Kind: TokenKind.Word })
-                    {
-                        wordParsedOnce = true;
-                    }
-                    ConsumeCurrentTokenInto(node);
-                /*if (CurrentToken is { Kind: TokenKind.Whitespace })
+                if (CurrentToken is not null && (!(CurrentToken is { Kind: TokenKind.Word or TokenKind.Whitespace } or { Text: "_" or "@" or "." }) || CurrentToken is { Kind: TokenKind.Word } && wordParsedOnce))
                 {
-                    ParseTrailingWhitespace(node, stopBeforeNewLine: true);
-                }*/
+                    var diagnostic = CurrentToken.CreateDiagnostic(HttpDiagnostics.InvalidNamedRequestName());
+                    node.AddDiagnostic(diagnostic);
+                    wordParsedOnce = false;
+                }
+
+                if (CurrentToken is { Kind: TokenKind.Word })
+                {
+                    wordParsedOnce = true;
+                }
+                ConsumeCurrentTokenInto(node);
 
             }
 
@@ -697,7 +693,7 @@ internal class HttpRequestParser
         {
             var nextTokenIndicatesName = CurrentTokenPlus(1) != null ? CurrentTokenPlus(1)!.Text.StartsWith("name") : false;
             return (CurrentToken is { Text: "@" } &&
-                nextTokenIndicatesName  &&
+                nextTokenIndicatesName &&
                 CurrentTokenPlus(2) is { Kind: TokenKind.Whitespace }
                     );
         }
--- a/src/Microsoft.DotNet.Interactive.Http.Tests/HttpKernelTests.cs
+++ b/src/Microsoft.DotNet.Interactive.Http.Tests/HttpKernelTests.cs
@@ -112,6 +112,158 @@ public class HttpKernelTests
     }
 
     [Fact]
+    public async Task variables_in_submit_request_are_saved()
+    {
+        HttpRequestMessage request = null;
+        HttpRequestMessage requestTwo = null;
+
+        Boolean firstRequest = true;
+
+        var handler = new InterceptingHttpMessageHandler((message, _) =>
+        {
+            if(firstRequest)
+            {
+                request = message;
+                firstRequest = false;
+                var response = new HttpResponseMessage(HttpStatusCode.OK);
+                return Task.FromResult(response);
+            }
+            else
+            {
+                requestTwo = message;
+                var response = new HttpResponseMessage(HttpStatusCode.OK);
+                return Task.FromResult(response);
+            }
+            
+        });
+        var client = new HttpClient(handler);
+        using var kernel = new HttpKernel(client: client);
+
+        using var _ = new AssertionScope();
+
+        var code = """
+            @hostname=httpbin.org
+            @host=https://{{hostname}}
+
+            Get {{host}}
+            """;
+
+        var result = await kernel.SendAsync(new SubmitCode(code));
+
+        result.Events.Should().NotContainErrors();
+        request.RequestUri.Should().Be($"https://httpbin.org");
+
+        var secondCode = """
+            GET {{host}}
+            """;
+
+        var secondResult = await kernel.SendAsync(new SubmitCode(secondCode));
+        secondResult.Events.Should().NotContainErrors();
+        requestTwo.RequestUri.Should().Be($"https://httpbin.org");
+    }
+
+    [Fact]
+    public async Task variables_submited_in_requests_that_are_not_used_are_valid()
+    {
+        HttpRequestMessage request = null;
+        HttpRequestMessage requestTwo = null;
+
+        Boolean firstRequest = true;
+
+        var handler = new InterceptingHttpMessageHandler((message, _) =>
+        {
+            if (firstRequest)
+            {
+                request = message;
+                firstRequest = false;
+                var response = new HttpResponseMessage(HttpStatusCode.OK);
+                return Task.FromResult(response);
+            }
+            else
+            {
+                requestTwo = message;
+                var response = new HttpResponseMessage(HttpStatusCode.OK);
+                return Task.FromResult(response);
+            }
+
+        });
+        var client = new HttpClient(handler);
+        using var kernel = new HttpKernel(client: client);
+
+        using var _ = new AssertionScope();
+
+        var code = """
+            @hostname=httpbin.org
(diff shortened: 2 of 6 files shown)


src/Microsoft.DotNet.Interactive.Http.Parsing/Parsing/HttpRequestParser.cs
```diff
@@ -668,25 +668,21 @@ private HttpNamedRequestNameNode ParseNamedRequestNameNode()
...
-                    if (CurrentToken is { Kind: TokenKind.Word })
-                    {
-                        wordParsedOnce = true;
-                    }
-                    ConsumeCurrentTokenInto(node);
-                /*if (CurrentToken is { Kind: TokenKind.Whitespace })
+                if (CurrentToken is { Kind: TokenKind.Word })
                 {
-                    ParseTrailingWhitespace(node, stopBeforeNewLine: true);
-                }*/
```
shyamnamboodiripad, line 689: Was this commented for testing something? If yes, please revert before merge. Otherwise, we should probably delete the commented code.

src/Microsoft.DotNet.Interactive.Http/HttpKernel.cs
```diff
@@ -124,28 +125,22 @@ async Task IKernelCommandHandler<SendValue>.HandleAsync(SendValue command, Kerne
 
     private Task SetValueAsync(string valueName, object value, Type? declaredType = null)
```
shyamnamboodiripad, line 126: After this PR this function will only be called for VS-specific variables declared in .env files / things like $projectUrl. Is that correct?

src/Microsoft.DotNet.Interactive.Http/HttpKernel.cs
```diff
@@ -124,28 +125,22 @@ async Task IKernelCommandHandler<SendValue>.HandleAsync(SendValue command, Kerne
...
-        {
-            _variables[valueName] = value;
-        }
+        _variables[valueName] = value;
         return Task.CompletedTask;
     }
 
     async Task IKernelCommandHandler<SubmitCode>.HandleAsync(SubmitCode command, KernelInvocationContext context)
     {
+        string? doc = "";
```
shyamnamboodiripad, line 134: minor: Consider using string.Empty instead of ""

Also, looks like can the `?` be omitted since we only access the value of `doc` if `TryGetValue` succeds?

If yes, consider changing this to `var doc = string.Empty`. Otherwise, if we want to preserve the nullable string, we can probably just set it to null `string? doc = null`.

src/Microsoft.DotNet.Interactive.Http/HttpKernel.cs
```diff
@@ -124,28 +125,22 @@ async Task IKernelCommandHandler<SendValue>.HandleAsync(SendValue command, Kerne
...
+        if (command.Parameters.TryGetValue("Document", out doc))
```
shyamnamboodiripad, line 135: The approach to pass the whole document as a parameter is interesting since that avoids the need to identify which request in the document should be executed.

It does mean that we will be repeating the text for request in the `command` again in the document - but that should hopefully not add too much overhead...

We will however end up reparsing the document and binding all the variables in it before each request, will we not? Or are we being smarter in the http editor and only passing down the document if the document content has changed?

src/Microsoft.DotNet.Interactive.Http/HttpKernel.cs
```diff
@@ -124,28 +125,22 @@ async Task IKernelCommandHandler<SendValue>.HandleAsync(SendValue command, Kerne
...
+        {
+            var parsedDoc = HttpRequestParser.Parse(doc);
+            foreach (DeclaredVariable dv in parsedDoc.SyntaxTree.RootNode.TryGetDeclaredVariables(BindExpressionValues).declaredVariables.Values)
```
shyamnamboodiripad, line 138: The above approach means that the user could see unexpected delays when executing requests in a large document (since we would need to bind all variables including some that are completely unrelated to the user's request).

An alternate approach may be to support a workflow where the code in the `SubmitCode` command can either be the entire document or a single request. We could then include a parameter to indicate which case it was - for example the parameter could be called `Execute`. If the entire document is being passed, we could set `Execute` to `false` and the kernel could parse and bind all the file level variables without executing any request. If `Execute` is set to `true`, we could execute the specified request.

One potential advantage of this approach would be that as the user makes changes to the document, we could asynchronously pass the entire document down to the kernel with `Execute` set to `false` whenever there is a long enough pause in the user's typing. This would allow the kernel to parse and bind the variables in the background after the user has made the file changes but before the user wants to execute a request. Then when the user clicks on `Send request` we could pass down a SubmitCode command with just the request that was selected for execution with `Execute` set to `true`. This could reduce any delays caused due to binding in larger documents.

There may also other ways in which something like above could be implemented (e.g. `SubmitCode` includes an `index` parameter that points to the request that should be executed within a document. If `index` is present, then we only execute the selected request. If `index` is absent we only parse and bind. However, I am not sure I like this since the behavior is implicit and hard to discern intuitively).

src/Microsoft.DotNet.Interactive.Http/Model/HttpNamedRequest.cs
```diff
@@ -233,11 +234,32 @@ internal HttpNamedRequest(HttpRequestNode httpRequestNode, HttpResponse response
     {
         if (currentIndex + 1 == path.Length)
         {
-            var result = responseJSON[path[currentIndex]];
-            return result?.ToString();
+            if(responseJSON.GetType() == typeof(JsonArray))
```
jonsequitur, line 237:
```suggestion
          if (responsJSON is JsonArray)
```

src/Microsoft.DotNet.Interactive.Http/Model/HttpNamedRequest.cs
```diff
@@ -233,11 +234,32 @@ internal HttpNamedRequest(HttpRequestNode httpRequestNode, HttpResponse response
...
+            {
+                var jsonArray = (JsonArray)responseJSON;
```
jonsequitur, line 239:
```suggestion
                
```

jonsequitur: This line isn't needed given the previous suggestion, which declares the correctly-typed variable inline.

src/Microsoft.DotNet.Interactive.Http/HttpKernel.cs
shyamnamboodiripad: I wonder if we should also have a check in the kernel above to only repeat the parsing and binding if the document content has changed. ..

src/Microsoft.DotNet.Interactive.Http/Model/HttpNamedRequest.cs
```diff
@@ -233,11 +234,32 @@ internal HttpNamedRequest(HttpRequestNode httpRequestNode, HttpResponse response
...
+                return jsonArray.FirstOrDefault(n => n?[path[currentIndex]] != null)?.ToJsonString();
+            } 
+            else if(responseJSON.GetType() == typeof(JsonObject))
```
jonsequitur, line 242: You can do similar here. You can also turn the whole block into a `switch`, e.g.

```csharp
switch (responseJson)
{
  case JsonArray jsonArray: //...
  case JsonObject jsonObject: // ...
}
```

src/Microsoft.DotNet.Interactive.Http/Model/HttpNamedRequest.cs
```diff
@@ -233,11 +234,32 @@ internal HttpNamedRequest(HttpRequestNode httpRequestNode, HttpResponse response
...
+            }
         }
 
-        var newResponseJSON = responseJSON[path[currentIndex + 1]];
+        JsonNode? newResponseJSON = null;
+        try
+        {
+            newResponseJSON = responseJSON[path[currentIndex]];
+        } 
+        catch (Exception)
```
jonsequitur, line 258: It's not ideal to catch `Exception` because it can easily hide bugs. If there's a specific exception type you're expecting here, it's better to catch that.

src/Microsoft.DotNet.Interactive.Http/HttpKernel.cs
shyamnamboodiripad, line 135: How are we planning to make this would work in the case of notebooks in VS Code or the dotnet-repl tool? Would we need changes for those clients to repeat the notebook cell contents in both 'Document' parameter as well as in `SubmitCode.Code` to ensure that any variables included in the code are parsed and bound?

Do we need code in the http kernel above to also parse and bind variables present in the `SubmitCode.Code` (in addition to the `Document` parameter) to support the notebooks and dotnet-repl scenarios?

src/Microsoft.DotNet.Interactive.Http/HttpKernel.cs
shyamnamboodiripad: Whichever approach we decide to go with should also work in the case of notebooks and dotnet-repl. So I think we would need to choose an approach where the 'default' implementation naturally supports notebooks and dotnet-repl...

src/Microsoft.DotNet.Interactive.Http/HttpKernel.cs
```diff
@@ -124,28 +125,22 @@ async Task IKernelCommandHandler<SendValue>.HandleAsync(SendValue command, Kerne
...
+            {
+                _variables[dv.Name] = dv.Value;
```
shyamnamboodiripad, line 140: Something to think about... There is a potential for `_variables` to retain bunch of stale variables / named requests over time. Not super high pri - but especially for named requests which can have large payloads, it may be nice to clean up stale entries...

Now that we have a specific code path that re-binds all file-level variables, we may be able to drop all old file-level variables whenever we get a new request to parse and bind the document.

To do this, we would need to separate out the dictionaries for file-level variables, named requests and the remaining (.env) variables. The file-level dictionary can be cleared out whenever we get a new request to bind the document. Any named requests that no longer exist in the document (based on the new syntax tree) could also be cleared out from the dictionary for named requests at this point. But for named requests that still exist in the document, we would need to preserve the old responses.

The dictionary for .env vars would be harder to clean up since we don't have a way to know when .env files change at the moment.

src/Microsoft.DotNet.Interactive.Http.Parsing/Parsing/HttpRequestParser.cs
bleaphar: It wasn't testing anything, I just noticed it was left over from last time and wanted to remvove it

src/Microsoft.DotNet.Interactive.Http/HttpKernel.cs
bleaphar: Yes

src/Microsoft.DotNet.Interactive.Http.Tests/HttpKernelTests.cs
```diff
@@ -2233,6 +2233,51 @@ public async Task responses_to_named_requests_can_be_accessed_as_symbols_in_late
         secondResult.Events.Should().NotContainErrors();
     }
 
+    [Theory]
+    [InlineData("json.response.body.$.slideshow.slides.title")]
+    [InlineData("json.response.body.$.slideshow.slides.type")]
```
shyamnamboodiripad, line 2238: Are we relying on httpbin.org to return a json that looks like this? I think it would be better to use a mock client with our own handler where we can provide our own json response (like we do in some other tests).

Otherwise, we could end up having to deal with failed builds in CI if httpbin.org happens to be down or if the format for the response ends up changing.

shyamnamboodiripad: It would also be nice to see what the response shape looks like in the test so that it is easier to understand what the test is validating...

src/Microsoft.DotNet.Interactive.Http.Parsing/Parsing/HttpRequestParser.cs
shyamnamboodiripad: Ah please remove the commented block then to avoid any confusion

src/Microsoft.DotNet.Interactive.Http.Tests/HttpKernelTests.cs
```diff
@@ -2233,6 +2233,51 @@ public async Task responses_to_named_requests_can_be_accessed_as_symbols_in_late
...
+            {
+                "path" : {{pathContents}}
+            }
+            
+            ###
+            """;
+
+        var secondResult = await kernel.SendAsync(new SubmitCode(secondCode));
+
+        secondResult.Events.Should().NotContainErrors();
```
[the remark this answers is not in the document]
bleaphar: Sounds good, I split this out in the recent commit

src/Microsoft.DotNet.Interactive.Http/HttpKernel.cs
bleaphar: Adding a property to the kernel that tracks the current document and if there are any changes from it then we'll go ahead with the re-parse

bleaphar: Sounds good, I went ahead with the approach to only repeat the parsing if the document has changed

src/Microsoft.DotNet.Interactive.Http/HttpKernel.cs
shyamnamboodiripad: @bleaphar Lets chat about this offline once you are back. I don't think this can be resolved merely by checking whether the file has changed since even minor whitespace changes to one character in the file could end up resulting in a long delay for the subsequent request.

We need some way to amortize some of this cost (especially in large files) - otherwise, users could end up seeing large delays when executing requests in large files and we could end up seeing feedback like the below one even in files that are not as large as the below case -

 https://developercommunity.visualstudio.com/t/Big-HTTP-file-is-not-interpreted-by-Visu/10717885

jonsequitur: A couple of comments.

> The above approach means that the user could see unexpected delays when executing requests in a large document (since we would need to bind all variables including some that are completely unrelated to the user's request).

We should measure this before making a design decision. I strongly suspect that the difference is not perceptible and that, to the extent that it is, serialization is a larger overhead than redundant parsing. But we should measure it.

>  We could then include a parameter to indicate which case it was - for example the parameter could be called Execute.

The intent is different so I would suggest this is no longer a `SubmitCode`. Does the `RequestDiagnostics` command not already do what's needed here?

src/Microsoft.DotNet.Interactive.Http/HttpKernel.cs
jonsequitur: This has the potential to introduce bugs where the kernel state gets out of sync with the editor state. What's the performance impact of resending and reparsing the whole document?

src/Microsoft.DotNet.Interactive.Http/HttpKernel.cs
jonsequitur: I would measure this before complicating the design. Separate dictionaries will increase cyclomatic complexity and the potential for bugs.

An alternative might be to store each variable with some metadata including a scope or origin concept, e.g. to differentiate variables whose values came from responses (*maybe* these are transient) versus those that came from a `SendValue` command (which the kernel shouldn't have a say in clearing.)

Consider this. In the typical REPL or notebook workflow, a single cell can be edited and resubmitted. Let's say I run this in the C# kernel:

```csharp
var x = 123;
```

Then let's say I delete that code and run this:

```csharp
Console.WriteLine(x);
```

It would be incorrect for the variable `x` to be cleared, and probably very surprising if the HTTP kernel had a different behavior. 

It's very typical in REPL and notebook workflows for memory to grow over time, which is why restarting the kernel is a familiar gesture. But if we find there's a performance issue here that's significantly different from other use cases, I'd prefer to address it with an explicit user gesture.

a-t-k:
Looking forward to seeing this feature live!

src/Microsoft.DotNet.Interactive.Http.Parsing/Parsing/HttpRequestParser.cs
```diff
@@ -668,25 +668,21 @@ private HttpNamedRequestNameNode ParseNamedRequestNameNode()
...
+            while (MoreTokens() && CurrentToken is not { Kind: TokenKind.NewLine })
+            {
 
-                    if (CurrentToken is not null && (!(CurrentToken is { Kind: TokenKind.Word or TokenKind.Whitespace} or { Text: "_" or "@" or "."}) || CurrentToken is {Kind: TokenKind.Word } && wordParsedOnce)) 
-                    {
-                        var diagnostic = CurrentToken.CreateDiagnostic(HttpDiagnostics.InvalidNamedRequestName());
-                        node.AddDiagnostic(diagnostic);
-                        wordParsedOnce = false;
-                    }
+                if (CurrentToken is not null && (!(CurrentToken is { Kind: TokenKind.Word or TokenKind.Whitespace } or { Text: "_" or "@" or "." }) || CurrentToken is { Kind: TokenKind.Word } && wordParsedOnce))
```
jonsequitur, line 674:
```suggestion
                if (CurrentToken is not null && (CurrentToken is not ({ Kind: TokenKind.Word or TokenKind.Whitespace } or { Text: "_" or "@" or "." }) || CurrentToken is { Kind: TokenKind.Word } && wordParsedOnce))
```

jonsequitur: Also, a `CurrentToken is { ... }` pattern won't match `null`, so the initial null check shouldn't be necessary.

src/Microsoft.DotNet.Interactive.Http.Tests/ParserTests.Variables.cs
```diff
@@ -283,5 +283,35 @@ public void spaces_after_variable_do_not_produce_diagnostics()
...
+        {
+            var result = Parse(
+                """
+
+                @host="https://httpbin.org"
+                """
+                );
+
+            var variables = result.SyntaxTree.RootNode.TryGetDeclaredVariables().declaredVariables;
+            variables.Should().Contain(n => n.Key == "host").Which.Value.Should().BeOfType<DeclaredVariable>().Which.Value.Should().Be("\"https://httpbin.org\"");
```
jonsequitur, line 313: This seems a little surprising to me that the quotes are part of the variable value.

Maybe related, are spaces in variable values supported like this then?

```
@blah = one two three
```

src/Microsoft.DotNet.Interactive.Http.Tests/HttpKernelTests.cs
```diff
@@ -111,6 +111,158 @@ public async Task can_handle_multiple_request_in_a_single_submission()
...
+            GET {{alternative_host}}
+            """;
+
+        var secondResult = await kernel.SendAsync(new SubmitCode(secondCode));
+        secondResult.Events.Should().NotContainErrors();
+        requestTwo.RequestUri.Should().Be($"https://httpbin.org");
+    }
+
+    [Fact]
+    public async Task variables_submited_in_requests_are_not_valid_across_clients()
```
jonsequitur, line 218: I don't think this is a particularly useful test.

src/Microsoft.DotNet.Interactive.Http.Tests/HttpKernelTests.cs
```diff
@@ -111,6 +111,158 @@ public async Task can_handle_multiple_request_in_a_single_submission()
         requests.Select(r => r.RequestUri.AbsoluteUri).ToArray().Should().BeEquivalentTo(new[] { "https://location1.com:1200/endpoint", "https://location2.com:1200/endpoint" });
     }
 
+    [Fact]
+    public async Task variables_in_submit_request_are_saved()
```
jonsequitur, line 115: The simplest way to verify the variable is to use the `RequestValue` command rather than via a second `SubmitCode`. This approach should make the test simpler and easier to understand.

jonsequitur (approved):

PR #3659, opened 2024-09-03
Drop `custom` property from notebook metadata

DonJayamanne:


src/polyglot-notebooks-vscode-common/src/extension.ts
```diff
@@ -70,13 +70,6 @@ const disposables: (() => void)[] = [];
 let surveryBanner: SurveyBanner;
 
 export async function activate(context: vscode.ExtensionContext) {
-    // Wait for ipynb extension to be ready, temporary.
-    const ipynbExtension = vscode.extensions.getExtension('vscode.ipynb');
-    if (ipynbExtension && !ipynbExtension.isActive) {
-        await ipynbExtension.activate();
-    }
-    metadataUtilities.setUseLegacyMetadata(ipynbExtension && !ipynbExtension.exports.dropCustomMetadata ? true : false);
```
DonJayamanne, line 78: Its been 6 months since the `custom` property was dropped, 
I have removed this completely from VS Code without any support for fallbacks, its now time to remove this fallback from here as well.

DonJayamanne: @colombod /cc

src/polyglot-notebooks-vscode-common/tests/metadataUtilities.test.ts
```diff
@@ -1,1092 +1,1034 @@
...
+        expect(notebookCellMetadata).to.deep.equal({
+            kernelName: "fsharp",
         });
+    });
 
-        it('cell metadata can be extracted from a notebook cell with old metadata', () => {
-            const notebookCell: vscodeLike.NotebookCell = {
-                kind: vscodeLike.NotebookCellKind.Code,
-                metadata: metadataUtilities.useLegacyMetadata() ? {
-                    custom: {
```
brettfo, line 128: I think it would be worthwhile to still allow reading the `custom` property and ensure that the new format is written back out.  This scenario would cover a developer that has an old notebook that only now needs to be updated.

colombod: This code is not dealing with the file, this was a change in object model in vscode api (@DonJayamanne correct right?) so we needed a way to be coexisting with stable and insiders on different object models. Now we can drop this code.

DonJayamanne: The `custom` property no longer exists in VS Code.
The change to move away from `custom` was made late last year, hence I do not se the need to add a backward compatibility anymore.

> so we needed a way to be coexisting with stable and insiders on different object models. Now we can drop this code.

Yes

jonsequitur (approved):

PR #3663, opened 2024-09-08

eosfor:
looks like the `CustomMimeType` parameter was not in use at all. This way we can start using it and thus leverage a set of renderers like the one for vega-light. Example is also included in this pull request

### DIFF
--- a/INTERNAL.md
+++ b/INTERNAL.md
@@ -68,6 +68,13 @@ When you do this locally while authenticated, any packages _not_ on the internal
 
 The signed build produces three versions of the VS Code extension, 2 against Stable VS Code and 1 against Insiders.  Both versions against Stable append a `"0"` character to the extension version number and Insiders appends a `"1"` character.
 
+### Pre-requisite to publish
+
+This is **temporary**. Before you invoke any of the pipelines below, please create a new PAT token with the following scope:
+![image](https://github.com/user-attachments/assets/87f2ad02-594c-440d-bbec-92332b12d864) [here]([url](https://dev.azure.com/dnceng/_usersSettings/tokens)) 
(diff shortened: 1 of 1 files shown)


eosfor:
@dotnet-policy-service agree

eosfor:
this is a very first attempt to send a pull request :D

@colombod, @jonsequitur, pls have a look

jonsequitur:
I was unaware of this property. I think it would be clearer to remove it as well as any constraints on the value of the `MimeType` property, which would be more consistent with how MIME types are used elsewhere.

jonsequitur:
Thanks for the PR! 

I have a couple of requests if you'd like to pursue it. 

* We've largely moved away from trying to document every feature in example notebooks. The usage example is helpful but doesn't need to part of this PR. 

* However, there should be an accompanying test which can serve as an example of expected usage as well as ongoing verification that it's working as expected.

Thanks!

eosfor:
@jonsequitur hopefully i did it right :). Removed the example from the PR and added a test i could come up with. 

But, actually, c# test won't look like a powershell example for many, i guss

src/Microsoft.DotNet.Interactive.PowerShell/Commands/OutDisplayCommand.cs
```diff
@@ -53,8 +55,17 @@ public sealed class OutDisplayCommand : PSCmdlet
...
-        DisplayedValue displayedValue = display(obj, MimeType);
-            
+
+        DisplayedValue displayedValue = default;
+
+        if (ParameterSetName == "MimeType"){
+            displayedValue = display(obj, MimeType);
+        }
+
+        if (ParameterSetName == "CustomMimeType"){
```
jonsequitur, line 65: If we completely remove the `CustomMimeType` property and set `MimeType` in the test (or in the notebook for that matter), does it work as expected?

PR #3666, opened 2024-09-09
Improve input prompt recall messages

jonsequitur:
This makes the messages shown when storing and recalling user input clearer and more informative.

It also incidentally removes the mostly-unused `KernelException` type.


colombod (approved):

src/Microsoft.DotNet.Interactive/KernelException.cs
```diff
@@ -1,14 +0,0 @@
-﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
-// Licensed under the MIT license. See LICENSE file in the project root for full license information.
-
-using System;
-
-namespace Microsoft.DotNet.Interactive;
-
-public class KernelException : Exception
```
colombod, line 8: this will require an update api contract file

PR #3669, opened 2024-09-17

eosfor:
This is a second attempt, i had to close the forst PR attempt and start over to make it cleaner :D.

Now it does not have any formatting problems, i hope.


eosfor:
hmm, still formatting issues ... :(

PR #3670, opened 2024-09-17

eosfor:
another attempt. i pulled all latest from the origin this time.


eosfor:
still some problems, not sure what is happening :(

PR #3671, opened 2024-09-17
fix CustomMimeType parameter

eosfor:
hopefully this time i fixed the formatting issue. vscode was "autoformatting on save"

### DIFF
--- a/INTERNAL.md
+++ b/INTERNAL.md
@@ -68,6 +68,13 @@ When you do this locally while authenticated, any packages _not_ on the internal
 
 The signed build produces three versions of the VS Code extension, 2 against Stable VS Code and 1 against Insiders.  Both versions against Stable append a `"0"` character to the extension version number and Insiders appends a `"1"` character.
 
+### Pre-requisite to publish
+
+This is **temporary**. Before you invoke any of the pipelines below, please create a new PAT token with the following scope:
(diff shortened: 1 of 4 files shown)


eosfor:
@jonsequitur, I had to cancel my previous PRs. Initially I was not able to figure out what was happening with formatting, and then just luckily sopped how VSCode "auto-formatted" on save. Hopefully it works this time

jonsequitur:
Thanks!

This looks better. 

One thing that I might not have been clear about is that removing the `CustomMimeType` property entirely, and just using `MimeType`, would be better still. The extra concept makes things harder to understand, and the `MimeType` property shouldn't be restricted to just a few types. It's intended to open-ended and this is how it works everywhere else in .NET Interactive.

eosfor:
> Thanks!
> 
> This looks better.
> 
> One thing that I might not have been clear about is that removing the `CustomMimeType` property entirely, and just using `MimeType`, would be better still. The extra concept makes things harder to understand, and the `MimeType` property shouldn't be restricted to just a few types. It's intended to open-ended and this is how it works everywhere else in .NET Interactive.

Well, I agree, but for some reason someone did split them into two different options. And behavior is totally different, as far as I understand. What was the reason for splitting them?

eosfor:
@jonsequitur this time I removed CustomMimeType parameter completely

jonsequitur (approved):

jonsequitur:
Thanks, @eosfor!

PR #3672, opened 2024-09-17
Changes to header parsing

bleaphar:
This PR fixes the issue with trying to get values from the headers of an http response. It looks to see if the user has a path of four, assuming they want the first element in the list and if there's a path of five then it gets the first element in the list that matches the value of the string. Additionally, a test for this scenario is included.


src/Microsoft.DotNet.Interactive.Http.Tests/HttpKernelTests.cs
```diff
@@ -2311,6 +2311,73 @@ public async Task responses_to_named_requests_can_be_accessed_as_symbols_in_late
         secondResult.Events.Should().NotContainErrors();
     }
 
+    [Fact]
+    public async Task headers_for_named_requests_can_be_accessed_correctly()
```
jonsequitur, line 2315:
```suggestion
    public async Task Response_headers_for_named_requests_can_be_accessed_correctly()
```

jonsequitur: This name is a little more specific, since the test isn't checking request headers.

src/Microsoft.DotNet.Interactive.Http.Tests/HttpKernelTests.cs
```diff
@@ -2311,6 +2311,73 @@ public async Task responses_to_named_requests_can_be_accessed_as_symbols_in_late
...
 
+    [Fact]
+    public async Task Response_headers_for_named_requests_can_be_accessed_correctly()
+    {
+
+        var responseHandler = new InterceptingHttpMessageHandler((message, _) =>
+        {
+            var response = new HttpResponseMessage(HttpStatusCode.OK);
+            response.RequestMessage = message;
+            var contentString = @"
```
shyamnamboodiripad, line 2322: nit: A `"""` string literal may be better here to avoid escaping double quotes etc.

src/Microsoft.DotNet.Interactive.Http.Tests/HttpKernelTests.cs
```diff
@@ -2311,6 +2311,73 @@ public async Task responses_to_named_requests_can_be_accessed_as_symbols_in_late
...
+        ""X-Amzn-Trace-Id"": ""Root=1-66e9f16e-7afea6f643e754f854c57d85""
+    }
+}
+            ";
+            response.Content = new StringContent(contentString, Encoding.UTF8, "application/json");
+            response.Headers.Add("server", "gunicorn/19.9.0");
+            return Task.FromResult(response);
+        });
+        var client = new HttpClient(responseHandler);
+        // var client = new HttpClient();
```
shyamnamboodiripad, line 2348: nit: Remove commented line

src/Microsoft.DotNet.Interactive.Http.Tests/HttpKernelTests.cs
```diff
@@ -2311,6 +2311,73 @@ public async Task responses_to_named_requests_can_be_accessed_as_symbols_in_late
...
+            ###
+            """;
+
+        var result = await kernel.SendAsync(new SubmitCode(code));
+        result.Events.Should().NotContainErrors();
+
+
+        var secondCode = $$$"""
+            GET https://httpbin.org/headers
+            Detectedserver: {{binHeader.response.headers.Server}}
```
shyamnamboodiripad, line 2366: As it stands the test does not validate the binding for this named request expression (since the message handler always returns the same value).

Should this test somehow validate that the value for the `Detectedserver` header matches the expected value `gunicorn/19.9.0` here? One way to do that may be to update the message handler such that it puts the bound value of any header with name `Detectedserver` in the body of the returned response. The test could then assert that the response body has the correct value...

src/Microsoft.DotNet.Interactive.Http.Tests/HttpKernelTests.cs
```diff
@@ -2350,7 +2417,7 @@ public async Task json_named_requests_with_sub_routes_can_be_accessed_correctly(
             return Task.FromResult(response);
         });
         var client = new HttpClient(responseHandler);
-       // var client = new HttpClient();
+        // var client = new HttpClient();
```
shyamnamboodiripad, line 2420: Consider removing these commented line(s)

shyamnamboodiripad (approved):
Lgtm modulo comments

src/Microsoft.DotNet.Interactive.Http.Tests/HttpKernelTests.cs
jonsequitur: I often take that approach. It can be done as well by just capturing the bound header within the message handler and setting a variable to assert against in the tests's outer scope. I think this is a little more direct.

shyamnamboodiripad, line 2366: Looks like this expression validates the case where path depth is 4

Do we already have other tests that cover cases like below -
* Path depth of 5
* Negative tests for path depth of 6 or more
* Negative test where someone tries to use json path / xpath after `binheader.response.headers` or after `binheader.request.headers.someHeaderName`
* Negative test for path depth of 3 or less. For example,
  * `{{binheader.response.headers}}`
  * `{{binheader.request}}`
  * `{{binheader}}`
* Negative tests for cases where the first, second and third parts contain garbage. For example,
  * 1st part is not a valid named request
  * 1st part maps to a named request, but 2nd part is neither `request` nor `response`, or 
  * 1st and 2nd parts are valid, but the 3rd part is neither `headers` not `body` and so on...

jonsequitur: I had some ideas for a different design for this piece that might better accommodate some of these complexities. It's probably worth finding some time to discuss.

bleaphar: I'll work on a test for the depth of 5, a negative test for the depth of 6, and a test where the depth is correct but the header doesn't exist but all of the other cases are handled by other tests already

src/Microsoft.DotNet.Interactive.Http.Tests/HttpKernelTests.cs
```diff
@@ -2382,6 +2386,210 @@ public async Task Response_headers_for_named_requests_can_be_accessed_correctly(
...
+            ###
+            """;
+
+        var secondResult = await kernel.SendAsync(new SubmitCode(secondCode));
+
+        secondResult.Events.Should().NotContainErrors();
+
+        var returnValue = secondResult.Events.OfType<ReturnValueProduced>().First();
+
+        var response = (HttpResponse)returnValue.Value;
```
jonsequitur, line 2458: It looks like `returnValue` and `response` are no longer needed.

src/Microsoft.DotNet.Interactive.Http.Tests/HttpKernelTests.cs
```diff
@@ -2382,6 +2386,210 @@ public async Task Response_headers_for_named_requests_can_be_accessed_correctly(
...
+
+        headerValue.Should().Be("theme=dark; Path=/; Expires=Wed, 09 Jun 2023 10:18:14 GMT");
+    }
+
+    [Fact]
+    public async Task Response_headers_for_named_requests_with_too_many_additional_headers_will_produce_an_error()
```
jonsequitur, line 2464: I'm not sure the test name quite describes what this is testing.

src/Microsoft.DotNet.Interactive.Http.Tests/HttpKernelTests.cs
```diff
@@ -2382,6 +2386,210 @@ public async Task Response_headers_for_named_requests_can_be_accessed_correctly(
...
+            ###
+            """
+        ;
+        var secondResult = await kernel.SendAsync(new SubmitCode(secondCode));
+        var diagnostics = secondResult.Events.Should().ContainSingle<DiagnosticsProduced>().Which;
+        diagnostics.Diagnostics.First().Message.Should().Be($$$"""The supplied expression 'binHeader.response.headers.Server.gunicorn' does not follow the correct pattern. The expression should adhere to the following pattern: {{requestName.(response|request).(body|headers).(*|JSONPath|XPath|Header Name)}}.""");
+    }
+
+    [Fact]
+    public async Task Response_headers_for_named_requests_that_do_not_exist_will_produce_an_error()
```
jonsequitur, line 2529:
```suggestion
    public async Task Expressions_specifying_response_headers_that_do_not_exist_will_produce_an_error()
```

src/Microsoft.DotNet.Interactive.Http.Tests/HttpKernelTests.cs
```diff
@@ -2382,6 +2386,210 @@ public async Task Response_headers_for_named_requests_can_be_accessed_correctly(
         headerValue.Should().Be("gunicorn/19.9.0");
     }
 
+    [Fact]
+    public async Task Response_headers_for_named_requests_with_additional_header_can_be_accessed_correctly()
```
jonsequitur, line 2390: It's not clear what "additional header" refers to here.

src/Microsoft.DotNet.Interactive.Http.Tests/HttpKernelTests.NamedRequest.cs
```diff
@@ -0,0 +1,1070 @@
...
+            ###
+            """;
+
+                var secondResult = await kernel.SendAsync(new SubmitCode(secondCode));
+
+                secondResult.Events.Should().NotContainErrors();
+
+                var returnValue = secondResult.Events.OfType<ReturnValueProduced>().First();
+
+                var response = (HttpResponse)returnValue.Value;
```
jonsequitur, line 158: It doesn't looks like `returnValue` or `response` are needed.

src/Microsoft.DotNet.Interactive.Http.Tests/HttpKernelTests.NamedRequest.cs
```diff
@@ -0,0 +1,1070 @@
...
```
jonsequitur, line 232: It doesn't looks like `returnValue` or `response` are needed.

src/Microsoft.DotNet.Interactive.Http.Tests/HttpKernelTests.NamedRequest.cs
```diff
@@ -0,0 +1,1070 @@
...
+
+                var secondResult = await kernel.SendAsync(new SubmitCode(secondCode));
+                var diagnostics = secondResult.Events.Should().ContainSingle<DiagnosticsProduced>().Which;
+                diagnostics.Diagnostics.First().Message.Should().Be($$$"""The supplied header name 'Accept' does not exist in the named request.""");
+            }
+
+            [Theory]
+            [InlineData("json.response.body.$.slideshow.slides.title", "Wake up to WonderWidgets!")]
+            [InlineData("json.response.body.$.slideshow.slides.type", "all")]
+            public async Task json_with_additional_syntax_depths_can_be_accessed_correctly(string path, string end)
```
jonsequitur, line 370: Nitpick: Renaming `end` to something more intention-revealing (e.g. `expectedValue`) would help the readability of the test.

src/Microsoft.DotNet.Interactive.Http.Tests/HttpKernelTests.NamedRequest.cs
```diff
@@ -0,0 +1,1070 @@
...
+            {
+                // Request Variables
+                // Request variables are similar to file variables in some aspects like scope and definition location.However, they have some obvious differences.The definition syntax of request variables is just like a single-line comment, and follows // @name requestName or # @name requestName just before the desired request url. 
+
+                var responseHandler = new InterceptingHttpMessageHandler((message, _) =>
+                {
+                    var response = new HttpResponseMessage(HttpStatusCode.OK);
+                    response.RequestMessage = message;
+                    var contentString = @"
```
jonsequitur, line 379: Using a raw string (`"""`) will improve readability.

src/Microsoft.DotNet.Interactive.Http.Tests/HttpKernelTests.NamedRequest.cs
```diff
@@ -0,0 +1,1070 @@
...
+
+                var firstCode = """
+            @baseUrl = https://httpbin.org/json
+
+            # @name json
+            GET {{baseUrl}}
+            Content-Type: application/json
+
+            ###
+            """;
```
jonsequitur, line 416: Indenting raw strings so their left margin is indented further than the main code will improve readability, e.g.: 

```csharp
                var firstCode = """
                    @baseUrl = https://httpbin.org/json

                    # @name json
                    GET {{baseUrl}}
                    Content-Type: application/json

                    ###
                    """;

```

src/Microsoft.DotNet.Interactive.Http.Tests/HttpKernelTests.NamedRequest.cs
```diff
@@ -0,0 +1,1070 @@
...
+
+                var secondResult = await kernel.SendAsync(new SubmitCode(secondCode));
+
+                secondResult.Events.Count().Should().Be(2);
+            }
+
+            [Theory]
+            [InlineData("example.request.headers.*")]
+            [InlineData("example.response.headers.*")]
+            public async Task content_body_after_headers_for_named_request_is_not_supported_and_produces_errors(string path)
```
jonsequitur, line 878:
```suggestion
            public async Task Accessing_content_body_after_headers_is_not_supported_and_produces_errors(string path)
```

src/Microsoft.DotNet.Interactive.Http.Tests/HttpKernelTests.NamedRequest.cs
```diff
@@ -0,0 +1,1070 @@
...
+
+                var diagnostics = secondResult.Events.Should().ContainSingle<DiagnosticsProduced>().Which;
+
+                diagnostics.Diagnostics.First().Message.Should().Be($$$"""The supplied header name '*' does not exist in the named request.""");
+            }
+
+            [Theory]
+            [InlineData("example.request.headers.Content-Type", "Content-Type")]
+            [InlineData("example.response.headers.Authorization", "Authorization")]
+            public async Task accessing_non_existant_header_names_produces_an_error(string path, string headerName)
```
jonsequitur, line 915:
```suggestion
            public async Task accessing_non_existent_header_names_produces_an_error(string path, string headerName)
```

src/Microsoft.DotNet.Interactive.Http/HttpKernel.cs
```diff
@@ -133,17 +133,14 @@ async Task IKernelCommandHandler<SubmitCode>.HandleAsync(SubmitCode command, Ker
...
-
 
         if (command.Parameters.TryGetValue("Document", out var doc))
         {
             var parsedDoc = HttpRequestParser.Parse(doc);
             var lastSpan = parsedDoc.SyntaxTree.RootNode.ChildNodes
                 .OfType<HttpRequestNode>()
                 .FirstOrDefault(n => n.Text == requestNodes.Last().Text)?.Span;
-            if(lastSpan != null)
+            if (lastSpan != null)
```
jonsequitur, line 143:
```suggestion
            if (lastSpan is not null)
```

jonsequitur (approved):

PR #3675, opened 2024-09-21
Create release pipeline to use managed identity

ocallesp:
This pull request introduces a new release pipeline using Managed Identity for authentication, which is intended for experimentation. The goal is to test and validate this new release process without impacting the existing pipeline used for production releases.

**Why this change is necessary:**

- We are required to move all release pipelines to YAML as part of Wave 3 pipeline migrations.

- This pipeline serves as a proof of concept to implement Managed Identity for secure access and deployment.

- It avoids disruptions to the ongoing release pipeline while allowing iterative improvements and testing.

**Next steps:**

- Once the new pipeline has been validated, older release pipelines will be deprecated and removed.


ocallesp:
Testing with internal branch

Issue #3056, opened 2023-06-23
add support for Postgres


jonsequitur:
Add support for connecting Postgres subkernels via `#!connect`.

NikiforovAll:
Hi @jonsequitur, any updates on this one?

NikiforovAll:
Can I use it already?

Something like - `#!connect psql --kernel-name adventureworks \"{connectionString}\"`?

jonsequitur:
I haven't published a package yet. I was going to give it a try before doing so.

Also, I missed this but would `#!connect postgres` make more sense?

jonsequitur:
There should be packages available on the build feed here if you want to try it out: 

https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-tools/nuget/v3/index.json

NikiforovAll:
> We haven't published a package on nuget.org yet. I was going to give it a try before doing so.
> 
> Also, I missed this but would `#!connect postgres` make more sense?

I think `psql` is OK because it resembles `psql` cli, kinda intuitive .

luisquintanilla:
> > We haven't published a package on nuget.org yet. I was going to give it a try before doing so.
> > Also, I missed this but would `#!connect postgres` make more sense?
> 
> I think `psql` is OK because it resembles `psql` cli, kinda intuitive .

This is cool! Personally, since it's not using `psql` (or at least I didn't see it did from the PR), using `psql` might be confusing.

jonsequitur:
I'm happy to make this change.

NikiforovAll:
I'm wondering about syntax highlighting. Should we add it also?

jonsequitur:
It would be nice to have. I would open another issue for that. It's a little complicated and our current solution isn't entirely satisfying.

NikiforovAll:
Found a typo:

![Image](https://github.com/user-attachments/assets/36bda188-4bc3-4c6c-bca7-5c1d91d3a732)

jonsequitur:
That should just work once you have a `TabularDataResource` or even just an `IEnumerable<T>`: https://github.com/colombod/dotnet-interactive-extension-lab/blob/7b8e09e82dfaa21b37688c4f3a5a292dc5cb47c3/src/SandDance.InteractiveExtension/ExploreExtensionsForSandDance.cs

Improving the ergonomics of getting data into those types could use some work in any case.

Issue #3567, opened 2024-06-11
Magic command API redesign


jonsequitur:
**_Update: These changes are now available in VS Code Insiders._**

# Magic command API redesign

## Background and motivations

In the early days of .NET Interactive, providing magic command features as seen in Jupyter required a parser for a magic command grammar that would be independent of the various languages (C#, F#, PowerShell) that .NET Interactive supports. A POSIX command line-style grammar was a fairly clear choice, and System.CommandLine was robust and ready to use, so rather than build something custom, we took a dependency on System.CommandLine. This API has served .NET Interactive well for several years now. But as System.Commandline's design is being [reset](https://github.com/dotnet/command-line-api/issues/2338) for inclusion as a core .NET library, some of the features that .NET Interactive uses are likely to be removed. At the same time, .NET Interactive's input system is undergoing a long-planned set of improvements whose ergonomics can be much better if we support features that are well outside of what a command line parser would include.

The largest impacts of these changes will be on people who've written extensions to .NET Interactive that customize magic commands. We are attempting to minimize impacts on notebook users by maintining backwards compatibility to the greatest extent possible, but it's important to recognize that some breaking changes are unavoidable unless we want to bring the entirety of the System.CommandLine codebase into .NET Interactive, which has a high cost for maintainability and concept complexity.

So what's changing?

## _Options_ and _arguments_ are now just _parameters_

The terms _option_ and _argument_ were chosen based on common Gnu/POSIX naming conventions for the command line. Separating the magic command API from the command line use case is an opportunity to choose the more commonly-recognized term _parameter_. Both named parameters (i.e. _options_) and unnamed parameters (i.e. _arguments_) will now use the same type, `KernelDirectiveParameter`.

## What's being removed?

The following features of System.CommandLine are not currently planned to be reimplemented by the new magic command parser. 

* A magic command can no longer have multiple _unnamed_ parameters (i.e. arguments).

* Optionally, a single unnamed parameter can be allowed if the magic command is configured to allow it. It is not required of API users to configure their magic commands to have a name-optional parameter (`KernelDirectiveParameter.AllowImplicitName`).

    The following two lines are equivalent, assuming `--name` is the parameter whose name is optional:

    ```
    #!fruit --name apple 
    #!fruit apple
    ```

    Similarly, the following are equivalent:

    ```
    #!fruit --color red --name apple
    #!fruit --color red apple
    ```

    Relative ordering between parameters is not significant, so the following are equivalent:

    ```
[...]

PR #3676, opened 2024-09-21
feat: Add PostgreSQL support [WIP]

NikiforovAll:
This is a draft.

Questions:

1. Are we going to update from dotnet5? It is almost the end of support for dotnet 6.
2. Could you please explain what are the requirements for a typical SQL kernel? What kind of features should we expect? E.g: autocomplete, integrations without other plugins, etc.
---

fixes: #3056

### DIFF
--- a/vscode-nuget-release-pipeline.yml
+++ b/vscode-nuget-release-pipeline.yml
@@ -8,9 +8,13 @@ resources:
       branch: main
 
 variables:
-  simulate: $(isSimulated)
-  DotNetNugetApiKey: $(DotNetNugetApiKey-A1)
-  vscodeTargets: $(listVSCodeTargets)
+- group: '.NET Core Nuget API Keys'
+- name : simulate
+  value: $(isSimulated)
+- name : DotNetNugetApiKey
+  value: $(DotNetNugetApiKey-A1)
+- name : vscodeTargets
+  value: $(listVSCodeTargets)
 
 pool:
   vmImage: 'windows-2022'


jonsequitur:
> 1. Are we going to update from dotnet5? It is almost the end of support for dotnet 6.

It should target `net8.0`.

> 2. Could you please explain what are the requirements for a typical SQL kernel? What kind of features should we expect? E.g: autocomplete, integrations without other plugins, etc.

The simplest thing is to implement just `SubmitCode` which is the command sent to execute a query. Other commands can then be added to implement other features. 

* Variable sharing is probably the next most important (`RequestValueInfos`, `RequestValue`, 
and `SendValue`).
* Language services (`RequestCompletions`, `RequestDiagnostics`, `RequestHoverText`) can be harder to implement and are nice to have, but not required. 

A very simple example is the support for [SQLite](https://github.com/dotnet/interactive/blob/main/src/Microsoft.DotNet.Interactive.SQLite/SQLiteKernel.cs).

A thorough example is the [ToolsServiceKernel](https://github.com/dotnet/interactive/blob/main/src/Microsoft.DotNet.Interactive.SqlServer/ToolsServiceKernel.cs), which is the base class supporting both the [MSSQL](https://github.com/dotnet/interactive/blob/main/src/Microsoft.DotNet.Interactive.SqlServer/MsSqlKernel.cs) and [KQL](https://github.com/dotnet/interactive/blob/main/src/Microsoft.DotNet.Interactive.Kql/MsKqlKernel.cs) kernels. It drives a separate executable that actually communicates with the database. This approach is more complex than might be needed if you can get the desired features via the Postgres .NET NuGet packages but the reason this approach was taken is that, at least for MSSQL and Kusto, the language services weren't available via existing libraries.

Hope that helps and I'm happy to help dig further.

src/Microsoft.DotNet.Interactive.PostgreSQL/PostgreSqlKernel.cs
```diff
@@ -0,0 +1,328 @@
...
+            }
+
+            var completionItems = await _serviceClient.ProvideCompletionItemsAsync(_tempFileUri, command);
+            context.Publish(new CompletionsProduced(completionItems, command));
+        }
+
+        protected override ChooseKernelDirective CreateChooseKernelDirective() =>
+            new ChoosePostgreSqlKernelDirective(this);
+
+        private class ChoosePostgreSqlKernelDirective : ChooseKernelDirective
```
jonsequitur, line 301: The `ChooseKernelDirective` and other usage of System.CommandLine has been removed from the `main` branch as of a couple months ago. You can read more about this here: https://github.com/dotnet/interactive/issues/3567.

The new implementation is available in VS Code Insiders.

PR #3679, opened 2024-09-24
Package updates

jonsequitur:


AbhitejJohn (approved):

shyamnamboodiripad:
@jonsequitur @AbhitejJohn Could we please wait a few days before merging this change?

I am not sure if the new versions in this PR are compatible with what VS 17.12 currently ships. Since we have a deadline coming up for the 17.12 release, any package compat issues emerging from this PR on the VS side could end up blocking partner team code flow for any http editor changes that need to be merged before the deadline. This could potentially mean scrambling to fix the package compat issues on the VS side at the last moment which can often end up triggering cascading package updates in other components, looping in multiple teams and additional process overhead for last minute changes.

After the deadline, main would flow into the next VS release (17.13). We would have significantly more flexibility to address any fallout from this change at that point.

colombod:
are not the version locked for vs dependecies?


Dr Diego Colombo PhD

(he/him)

Principal software engineer

.NET Interactive<https://github.com/dotnet/interactive>

Try .NET<https://github.com/dotnet/try>



[Microsoft]

PR #3700, opened 2024-10-07
improve magic command completions

jonsequitur:
This fixes a number of cases with magic command completions.

### DIFF
--- a/src/Microsoft.DotNet.Interactive/Directives/KernelActionDirective.cs
+++ b/src/Microsoft.DotNet.Interactive/Directives/KernelActionDirective.cs
@@ -11,6 +11,7 @@ using System.Text.Json.Serialization;
 using System.Threading.Tasks;
 using Microsoft.CodeAnalysis.Tags;
 using Microsoft.DotNet.Interactive.Events;
+using Microsoft.DotNet.Interactive.Parsing;
 
 namespace Microsoft.DotNet.Interactive.Directives;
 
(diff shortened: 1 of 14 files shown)


src/Microsoft.DotNet.Interactive/Directives/KernelActionDirective.cs
```diff
@@ -81,15 +82,28 @@ private set
...
         var subcommandCompletions = Subcommands.Select(s => new CompletionItem(s.Name, WellKnownTags.Method)
         {
             AssociatedSymbol = s,
             Documentation = s.Description
         });
 
-        return subcommandCompletions.Concat(baseCompletions).ToArray();
+        completions.AddRange(subcommandCompletions);
+
+        if (Parent is KernelDirective parent)
```
shyamnamboodiripad, line 97: Does order in which the completions are added to the list matter? If yes, should the directive-based ones be added first?

src/Microsoft.DotNet.Interactive/Parsing/PolyglotSyntaxParser.cs
```diff
@@ -188,12 +188,16 @@ void AppendNode(SyntaxNode node)
...
                     case DirectiveParameterValueNode valueNode when subcommandNode is null:
                         directiveNode.Add(valueNode);
                         break;
+                    case DirectiveParameterValueNode valueNode when subcommandNode is not null:
+                        subcommandNode.Add(valueNode);
+                        break;
 
-                    case DirectiveNameNode nameNode when subcommandNode is not null:
-                        subcommandNode.Add(nameNode);
+                    case DirectiveParameterNode parameterNode when subcommandNode is null:
```
shyamnamboodiripad, line 199: Is this a duplicate of the `case` on line 188 above?

shyamnamboodiripad (approved):

src/Microsoft.DotNet.Interactive/Directives/KernelActionDirective.cs
jonsequitur: It doesn't matter. At this level, the final list hasn't been assembled, so ordering doesn't matter yet, nor does e.g. filtering for partially-typed completions.

src/Microsoft.DotNet.Interactive/Parsing/PolyglotSyntaxParser.cs
jonsequitur: It is! Nice catch.

PR #3714, opened 2024-10-15
Fix #3710

### LINKED ISSUE #3710: `#!connect jupyter` always uses local `jupyter`, ignores `--url`
Repro is missing because of exact version numbers and [DotnetEvolution discussion](https://discord.com/channels/732297728826277939/832335158925328455/1293075815910604872) with @jonsequitur. Repro can be created if necessary.

### Describe the bug
Latest release `Version: 1.0.522904+cdfa48b2ea1a27dfe0f545c42a34fd3ec7119074`, everything works fine.

Latest `main` `Version: 1.0.550701+71b9dd7a2df38f62dffae6f16d95f47084a84005` `#!connect jupyter` always uses local `jupyter` (`JupyterLocalKernelConnectionOptions`) no matter `--url` and `--token` values, yielding `Error: System.ArgumentException: KernelSpec <kernel-name-here> not found` if the kernel-spec is not also present in local.

`ConnectJupyterKernel connectCommand` is always `null` at [ConnectJupyterKernelDirective.GetJupyterConnection](https://github.com/dotnet/interactive/blob/3895c8cbdad68ecbb0de47658c60e21d43b836ad/src/Microsoft.DotNet.Interactive.Jupyter/ConnectJupyterKernelDirective.cs#L76).

**Which version of .NET Interactive are you using?** (In a notebook, run the `#!about` magic command. ):

- OS
    - Windows 11
 - Frontend
    - Visual Studio Code Insiders

jonsequitur:
This fixes #3710.

It also contains a small completions fix for magic command subcommands.

### DIFF
--- a/src/Microsoft.DotNet.Interactive/Parsing/DirectiveNode.cs
+++ b/src/Microsoft.DotNet.Interactive/Parsing/DirectiveNode.cs
@@ -47,9 +47,9 @@ internal class DirectiveNode : TopLevelSyntaxNode
 
         if (TryGetDirective(out var directive))
         {
-            foreach (var diagnostic1 in GetDiagnosticsForMissingParameters(directive, this))
+            foreach (var diagnostic in GetDiagnosticsForMissingParameters(directive, this))
             {
-                yield return diagnostic1;
+                yield return diagnostic;
             }
         }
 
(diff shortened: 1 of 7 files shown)


src/Microsoft.DotNet.Interactive/Parsing/DirectiveNode.cs
```diff
@@ -47,39 +47,9 @@ internal DirectiveNode(
 
         if (TryGetDirective(out var directive))
         {
-            foreach (var namedParameter in directive.Parameters)
+            foreach (var diagnostic1 in GetDiagnosticsForMissingParameters(directive, this))
```
shyamnamboodiripad, line 50: consider renaming to avoid the `1` suffix :) Would have been nice if C# allowed to omit the var and reuse the `diagnostic` variable

shyamnamboodiripad (approved):

colombod (approved):

Issue #3720, opened 2024-10-18
Microsoft.DotNet.Interactive.NamedPipeConnectorConnector package missing?


michael-hawker:
### Describe the bug

I'm trying to update/get the WPF sample to work, it won't connect via named pipes in the call to `#!connect`, so I tried updating the packages for .NET Interactive itself, but then the `ConnectNamedPipeCommand` was missing. It seems like this got refactored out to its own `Microsoft.DotNet.Interactive.NamedPipeConnectorConnector` project, but then that package isn't available on NuGet?

### Please complete the following:

**Which version of .NET Interactive are you using?** (In a notebook, run the `#!about` magic command. ):

Version: 1.0.522904+cdfa48b2ea1a27dfe0f545c42a34fd3ec7119074

Library version: 1.0.0-beta.24229.4+cdfa48b2ea1a27dfe0f545c42a34fd3ec7119074

Build date: 2024-10-18T17:35:20.9345851Z

- OS
    - [x] Windows 11
    - [ ] Windows 10
    - [ ] macOS
    - [ ] Linux (Please specify distro)
    - [ ] iOS
    - [ ] Android
 - Browser 
    - [ ] Chrome
    - [ ] Edge
    - [ ] Firefox
    - [ ] Safari
 - Frontend
    - [ ] Jupyter Notebook
    - [ ] Jupyter Lab
    - [ ] nteract
    - [x] Visual Studio Code 
    - [ ] Visual Studio Code Insiders
    - [ ] Visual Studio
    - [ ] Other (please specify)

### Screenshots

![Image](https://github.com/user-attachments/assets/7cdf6e5f-7f31-4bd4-b57c-ca05a1009407)

michael-hawker:
I also just noticed that the packageId has a duplicate `Connector` in its name:

https://github.com/dotnet/interactive/blob/3aa24a64f53b81979548d2eb0ad49429e5b0d4b2/src/Microsoft.DotNet.Interactive.NamedPipeConnector/Microsoft.DotNet.Interactive.NamedPipeConnector.csproj#L7

jonsequitur:
The package has now been published on NuGet.org:

https://www.nuget.org/packages/Microsoft.DotNet.Interactive.NamedPipeConnector

PR #3723, opened 2024-10-18
NamedPipeConnector Fixes

michael-hawker:
Contributes to #3720 (not sure why it's not uploaded to NuGet, but noticed some issues while investigating the WPF Connect sample, so figured I'd push these up as a separate PR here so it'd be more ready at least!)

This PR does 3 things:
- Use straight `net8.0` as the TFM. 
  - The `net8.0-windows` TFM, I think is the old way from .NET 5 only? Generally, it should have a version number after it for targeting modern Windows, this happened in .NET 6, see https://learn.microsoft.com/windows/apps/desktop/modernize/desktop-to-uwp-enhance
  - Regardless, from what I can tell anyway, the `NamedPipeClientStream` which is the main part of this code is still just a pure .NET API and _doesn't need the Windows specific TFM_ https://learn.microsoft.com/dotnet/api/system.io.pipes.namedpipeclientstream
  - Having a windows TFM breaks the connect-wpf sample as the notebook in VS Code can't load the windows TFM (I have a draft PR to update the connect-wpf sample, and I think it can connect now with some help, but stalls out, so I think I'm stuck, but this was the first step, so figured would make it an independent PR as the wpf sample is pinned in the past anyway.)
- Update the Package name to match the project name (extra Connector was on there for some reason, copy/paste bug?)
  - Looked later, this may be the root cause of #3720 and the package missing on NuGet as the publish script (see comment below) looks for the expected name
- When loading this extension in a notebook, it would fail as the code in the `extension.dib` wasn't updated alongside the other code in this project
  - This is where I had to guess a bit, but tried to pattern off the SQL Lite extension, and just created a new static method to add the directive to the root when the extension is initialized.

### DIFF
--- a/src/Microsoft.DotNet.Interactive.NamedPipeConnector/ConnectNamedPipeDirective.cs
+++ b/src/Microsoft.DotNet.Interactive.NamedPipeConnector/ConnectNamedPipeDirective.cs
@@ -37,4 +37,13 @@ public class ConnectNamedPipeDirective : ConnectKernelDirective<ConnectNamedPipe
 
         return new Kernel[] { proxyKernel };
     }
+
+    public static void AddToRootKernel()
+    {
+        if (KernelInvocationContext.Current is { } context &&
+            context.HandlingKernel.RootKernel is CompositeKernel root)
+        {
+            root.AddKernelConnector(new ConnectNamedPipeDirective());
+        }
+    }
 }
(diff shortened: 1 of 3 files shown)


jonsequitur (approved):

jonsequitur:
Thanks, @michael-hawker!

michael-hawker:
@jonsequitur my pleasure, thanks for the quick approval! Was going to continue the discussion in #3720, but then realized maybe this PR will fix that issue too... Looking here:

https://github.com/dotnet/interactive/blob/1f11b72e79680e17fba039cfa02aea112645bf01/eng/publish/PublishVSCodeExtension.ps1#L50

It seems like it's part of the publish list, but was using the csproj name (as expected), so since the package name was wrong it was just probably missed... 😅

Now that the name is fixed, maybe it should be picked up on your next publish?

jonsequitur (approved):

PR #3724, opened 2024-10-18
[WIP] Update Connect WPF Sample?

### LINKED ISSUE #2582: Connect-wpf sample app not running because cannot find Microsoft.DotNet.Arcade.Sdk
### Describe the bug
I tried to run the provided sample app in VS Code, but it failed with an Omnisharp error:  
`Unable to find package Microsoft.DotNet.Arcade.Sdk with version (= 8.0.0-beta.22554.2)`

.NET Interactive
© 2020 Microsoft Corporation
Version: 1.0.360602+9bf026dabac44a6256a65168fa93dcb7e2edcca4
Library version: 1.0.0-beta.22606.2+9bf026dabac44a6256a65168fa93dcb7e2edcca4
Build date: 2022-12-29T12:27:44.9947868Z

- OS
    - [x] Windows 11
 - Frontend
    - [x] Visual Studio Code 

### Possible solution is to add the arcade's public nuget source to the nuget.config:
`<add key="dotnet-eng" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-eng/nuget/v3/index.json" />`

michael-hawker:
Requires #3723 and dependent probably on #3720

This is an initial start for updating the Connect WPF sample:

- Add to the main solution and allow building it there alongside the rest of the code (use a dummy props file to prevent the arcade issue) Fixes #2582
- Use ProjectReference to build against the project (at least for now while working on this)
- Updates code to use the new NuGetDirective pattern (grabbed code from the extension in the browser project, not sure why that extension isn't just part of the core library)
- Updates code to use the refactored out NamedPipeConnector project
- Updates to use new KernalActionDirective pattern (seems likes some refactor/change happened to avoid conflicts with cmdline parsing, but this project was never updated)
- Temp? Add packages needed to the notebook for testing

That's as far as I could get, it seems to connect to the WPF app (maybe) and then stalls out? The connect command just continues to 'run' and ignores the request to stop in the notebook from VS Code (with no output/status).

image

I'm out-of-my-depth now trying to just update the code as best as I can with the internals of how .NET Interactive works, but figured putting an update together to share what I had as a starting point for others to jump on would be useful.

Thanks!

### DIFF
--- a/dotnet-interactive.sln
+++ b/dotnet-interactive.sln
@@ -123,6 +123,10 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.DotNet.Interactiv
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.DotNet.Interactive.Parsing.Tests", "src\Microsoft.DotNet.Interactive.Parsing.Tests\Microsoft.DotNet.Interactive.Parsing.Tests.csproj", "{55138BD7-111A-43A5-BFBB-326606C4C1B5}"
 EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "samples", "samples", "{49E6C2EA-FE81-47DF-B206-F818669BEFB1}"
+EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "WpfConnect", "samples\connect-wpf\WpfConnect.csproj", "{98689A19-6B13-4EA9-8C0E-6CB665322531}"
+EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
 		Debug|Any CPU = Debug|Any CPU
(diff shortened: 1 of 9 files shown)


dotnet-interactive.sln
```diff
@@ -123,6 +123,10 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.DotNet.Interactiv
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.DotNet.Interactive.Parsing.Tests", "src\Microsoft.DotNet.Interactive.Parsing.Tests\Microsoft.DotNet.Interactive.Parsing.Tests.csproj", "{55138BD7-111A-43A5-BFBB-326606C4C1B5}"
 EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "samples", "samples", "{49E6C2EA-FE81-47DF-B206-F818669BEFB1}"
```
jonsequitur, line 126: We've been keeping samples out of the solution. Ideally the samples build against NuGet packages, since most people using them won't be building this code.

samples/connect-wpf/WpfConnect.csproj
```diff
@@ -8,7 +8,9 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.DotNet.Interactive.CSharp" Version="1.0.0-beta.23403.1" />
+    <ProjectReference Include="..\..\src\Microsoft.DotNet.Interactive.CSharp\Microsoft.DotNet.Interactive.CSharp.csproj" />
```
jonsequitur, line 11: These should be package references and not project references.

samples/connect-wpf/readme.ipynb
```diff
@@ -59,6 +59,9 @@
         }
       ],
       "source": [
+        "#i \"nuget:I:\\oss\\dotnetinteractive\\artifacts\\packages\\Debug\\Shipping\"\n",
```
jonsequitur, line 62: Referencing a hard-coded path won't work for most people, and we don't want to expect people to build the solution in any case.

PR #3734, opened 2024-10-28
Use return inside ForEach-Object to skip to the next item

ocallesp:
This fixes the issue that stop further processing when a debug package is detected.

### DIFF
--- a/azure-publish-stable-polyglot-notebooks.yml
+++ b/azure-publish-stable-polyglot-notebooks.yml
@@ -19,7 +19,7 @@ variables:
   - name: DotNetNugetApiKey
     value: $(DotNetNugetApiKey-A1)
   - name: vscodeTargets
-    value: stable
+    value: stable,insiders
   - template: /eng/common/templates-official/variables/pool-providers.yml@self
     
 
--- a/dotnet-interactive.sln
+++ b/dotnet-interactive.sln
@@ -123,6 +123,14 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.DotNet.Interactiv
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.DotNet.Interactive.Parsing.Tests", "src\Microsoft.DotNet.Interactive.Parsing.Tests\Microsoft.DotNet.Interactive.Parsing.Tests.csproj", "{55138BD7-111A-43A5-BFBB-326606C4C1B5}"
 EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Solution Items", "Solution Items", "{8EC462FD-D22E-90A8-E5CE-7E832BA40C5D}"
+	ProjectSection(SolutionItems) = preProject
+		azure-pipelines-official.yml = azure-pipelines-official.yml
+		azure-pipelines.yml = azure-pipelines.yml
+		azure-publish-insiders-polyglot-notebooks.yml = azure-publish-insiders-polyglot-notebooks.yml
+		azure-publish-stable-polyglot-notebooks.yml = azure-publish-stable-polyglot-notebooks.yml
+	EndProjectSection
+EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
 		Debug|Any CPU = Debug|Any CPU
--- a/dotnet-interactive.sln
+++ b/dotnet-interactive.sln
@@ -123,14 +123,6 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.DotNet.Interactiv
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.DotNet.Interactive.Parsing.Tests", "src\Microsoft.DotNet.Interactive.Parsing.Tests\Microsoft.DotNet.Interactive.Parsing.Tests.csproj", "{55138BD7-111A-43A5-BFBB-326606C4C1B5}"
 EndProject
-Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Solution Items", "Solution Items", "{8EC462FD-D22E-90A8-E5CE-7E832BA40C5D}"
-	ProjectSection(SolutionItems) = preProject
-		azure-pipelines-official.yml = azure-pipelines-official.yml
-		azure-pipelines.yml = azure-pipelines.yml
-		azure-publish-insiders-polyglot-notebooks.yml = azure-publish-insiders-polyglot-notebooks.yml
-		azure-publish-stable-polyglot-notebooks.yml = azure-publish-stable-polyglot-notebooks.yml
-	EndProjectSection
-EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
 		Debug|Any CPU = Debug|Any CPU
--- a/eng/publish/PublishVSCodeExtensionUsingMI.ps1
+++ b/eng/publish/PublishVSCodeExtensionUsingMI.ps1
@@ -66,7 +66,7 @@ try {
                 if ($nugetPackageName -like '*.symbols.nupkg') {
                     Write-Host "Skipping publishing symbol package $nugetPackagePath"
                     # Use 'continue' to skip to the next iteration
-                    continue
+                    return
                 }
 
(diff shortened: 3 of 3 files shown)


eng/publish/PublishVSCodeExtensionUsingMI.ps1
```diff
@@ -66,7 +66,7 @@ try {
                 if ($nugetPackageName -like '*.symbols.nupkg') {
                     Write-Host "Skipping publishing symbol package $nugetPackagePath"
                     # Use 'continue' to skip to the next iteration
-                    continue
+                    return
                 }
 
                 if ($nugetPackageName -match '(?<=(?<id>.+))\.(?<version>((\d+\.\d+(\.\d+)?))(?<suffix>(-.*)?))\.nupkg')
```
colombod, line 72: why not just fix the regular expression to filter correctly and just process them?

colombod (changes requested):

eng/publish/PublishVSCodeExtensionUsingMI.ps1
```diff
@@ -66,7 +66,7 @@ try {
                 if ($nugetPackageName -like '*.symbols.nupkg') {
                     Write-Host "Skipping publishing symbol package $nugetPackagePath"
                     # Use 'continue' to skip to the next iteration
-                    continue
+                    return
```
colombod, line 69: Also are you sure you are not terminating all? return here will exit the publish script. At the first symbol package this will stop

ocallesp: actually that is the opposite. Return will continue to the next item, and Continue will exit the loop

colombod: ???

eng/publish/PublishVSCodeExtensionUsingMI.ps1
AbhitejJohn: This makes a lot more sense.

dotnet-interactive.sln
```diff
@@ -123,6 +123,14 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.DotNet.Interactiv
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.DotNet.Interactive.Parsing.Tests", "src\Microsoft.DotNet.Interactive.Parsing.Tests\Microsoft.DotNet.Interactive.Parsing.Tests.csproj", "{55138BD7-111A-43A5-BFBB-326606C4C1B5}"
 EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Solution Items", "Solution Items", "{8EC462FD-D22E-90A8-E5CE-7E832BA40C5D}"
```
colombod, line 126: do we need this in the solution file?

azure-publish-stable-polyglot-notebooks.yml
```diff
@@ -19,7 +19,7 @@ variables:
   - name: DotNetNugetApiKey
     value: $(DotNetNugetApiKey-A1)
   - name: vscodeTargets
-    value: stable
+    value: 'stable,insiders'
```
jonsequitur, line 22: Why would `insiders` be a valid value for this pipeline?

dotnet-interactive.sln
jonsequitur, line 126: I'd prefer to keep these out of the solution since VS doesn't do anything with them.

eng/publish/PublishVSCodeExtensionUsingMI.ps1
colombod, line 72: let's fix the filter expression and remove the conditional check

colombod (changes requested):

dotnet-interactive.sln
ocallesp: Do we want to remove them ?

eng/publish/PublishVSCodeExtensionUsingMI.ps1
ocallesp: this is working.

Also this helps the script to be more verbose

https://dev.azure.com/dnceng/internal/_build/results?buildId=2570763&view=logs&j=8b2f3529-484e-5976-a626-d4a8d47f3428&t=18943edd-a66d-52d4-6098-50cd321688d9

ocallesp: This is working with the 'return' https://dev.azure.com/dnceng/internal/_build/results?buildId=2570763&view=logs&j=8b2f3529-484e-5976-a626-d4a8d47f3428&t=18943edd-a66d-52d4-6098-50cd321688d9

And also it makes the script more verbose and easy to understand

eng/publish/PublishVSCodeExtensionUsingMI.ps1
ocallesp: because this is more verbose. Here is a sample log

https://dev.azure.com/dnceng/internal/_build/results?buildId=2570763&view=logs&j=8b2f3529-484e-5976-a626-d4a8d47f3428&t=18943edd-a66d-52d4-6098-50cd321688d9

dotnet-interactive.sln
ocallesp: Removing these files in the next commit

azure-publish-stable-polyglot-notebooks.yml
```diff
@@ -19,7 +19,7 @@ variables:
   - name: DotNetNugetApiKey
     value: $(DotNetNugetApiKey-A1)
   - name: vscodeTargets
-    value: stable
+    value: stable,insiders
```
ocallesp, line 22: @colombod 

stable release now publishes insiders and stable

eng/publish/PublishVSCodeExtensionUsingMI.ps1
ocallesp: @AbhitejJohn what part ?

dotnet-interactive.sln
ocallesp: removed. This just makes things easy when working in vs

eng/publish/PublishVSCodeExtensionUsingMI.ps1
ocallesp: Closing this comment since it looks to be a duplicate

ocallesp: I chose this approach over modifying the regex because any error in the regex could lead to missed packages or unexpected releases. Since this is a release pipeline, we want to minimize the risk of unintended changes

eng/publish/PublishVSCodeExtensionUsingMI.ps1
colombod: why is making filter correct more verbose?

eng/publish/PublishVSCodeExtensionUsingMI.ps1
jonsequitur: It looks like this regex will match both symbols and non-symbols packages.

image

AbhitejJohn: I would consider updating the regex to not match symbols nuget package and log all the matches for the regex in the script. These updates aren't really reading well for the 3 of us reviewing it.

jonsequitur: A simpler and more readable approach is to use a negative lookbehind (e.g. `'.*(?<!.symbols)\.nupkg'`)

image

ocallesp: @jonsequitur could you propose this code change in the regex ?

eng/publish/PublishVSCodeExtensionUsingMI.ps1
```diff
@@ -58,17 +58,10 @@ try {
                 "Microsoft.DotNet.Interactive.SqlServer"
             )
 
-            Get-ChildItem "$artifactsPath\packages\Shipping\Microsoft.DotNet*.nupkg" | ForEach-Object {
+            Get-ChildItem "$artifactsPath\packages\Shipping\Microsoft.DotNet*.nupkg" -Exclude '*.symbols.nupkg' | ForEach-Object {
```
jonsequitur, line 61: Not all package names will necessarily have this prefix.

PR #3740, opened 2024-10-30
Fix #3711

### LINKED ISSUE #3711: Dont throw an exception on duplicate connection name
When connecting to a python kernel, replace error message with simple warning message when user tries to create a connection with the same name 
“Warning: The kernel name or alias [python-kernel] is already in use. Please choose a new name.”  

Image

jonsequitur:
This fixes #3711.

### DIFF
--- a/src/Microsoft.DotNet.Interactive/Commands/KernelDirectiveCommand.cs
+++ b/src/Microsoft.DotNet.Interactive/Commands/KernelDirectiveCommand.cs
@@ -7,5 +7,5 @@ namespace Microsoft.DotNet.Interactive.Commands;
 
 public abstract class KernelDirectiveCommand : KernelCommand
 {
-    public virtual IEnumerable<string> GetValidationErrors() => [];
+    public virtual IEnumerable<string> GetValidationErrors(CompositeKernel kernel) => [];
 }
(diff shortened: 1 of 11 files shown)


src/Microsoft.DotNet.Interactive/Commands/KernelDirectiveCommand.cs
```diff
@@ -7,5 +7,5 @@ namespace Microsoft.DotNet.Interactive.Commands;
 
 public abstract class KernelDirectiveCommand : KernelCommand
 {
-    public virtual IEnumerable<string> GetValidationErrors() => [];
+    public virtual IEnumerable<string> GetValidationErrors(CompositeKernel kernel) => [];
```
shyamnamboodiripad, line 10: The appearance of CompositeKernel in the arguments here feels a bit odd / arbitrary. I wish there were some other way that avoids the need for a command to know the set of other kernels in play, or if there were a way to report this error from a higher layer that already has this context, or if this could be limited to just the `#!connect` command. But I understand this may not be straightforward currently...

Also, would this also work for out of proc scenarios or is that out of scope for `#!connect`? (i.e., The case where there happens to be another kernel with the same name in the broader graph of kernels in the 'environment' potentially outside the current process / current root composite kernel)?

colombod (approved):

shyamnamboodiripad (approved):

src/Microsoft.DotNet.Interactive/Commands/KernelDirectiveCommand.cs
jonsequitur: > The appearance of CompositeKernel in the arguments here feels a bit odd / arbitrary. I wish there were some other way that avoids the need for a command to know the set of other kernels in play, or if there were a way to report this error from a higher layer that already has this context, or if this could be limited to just the #!connect command. But I understand this may not be straightforward currently...

Validation takes place before command dispatch so there's no higher layer that can really provide this validation unless we want to special case it within the `SubmissionParser` (which might be reasonable).

Some sort of context does need to be passed in to make this method more useful, and it happens prior to command dispatch, since validation errors on the command need to prevent command dispatch. This is why `KernelInvocationContext` doesn't really work here.

> Also, would this also work for out of proc scenarios or is that out of scope for #!connect? (i.e., The case where there happens to be another kernel with the same name in the broader graph of kernels in the 'environment' potentially outside the current process / current root composite kernel)?

This works fine. `KernelInfo` is synced across remote kernels and all kernel names are locally scoped. This is how a command's destination is resolved.

PR #3748, opened 2024-11-06
Formatting for VS Code change

bleaphar:
Introducing a change for datetime iso8601 so that it reflects the same format as visual studio code and reduces round tripping difficulties.


src/Microsoft.DotNet.Interactive.Http.Parsing/Parsing/DynamicExpressionUtilities.cs
```diff
@@ -183,6 +184,13 @@ private static DateTimeOffset DefaultGetDateTimeOffset(bool isLocal)
                     else if (string.Equals(type.Value, "iso8601", StringComparison.OrdinalIgnoreCase))
                     {
                         format = "o";
+                        if (currentDateTimeOffset.Offset.TotalMinutes == 0)
+                        {
+                            // for $datetime, format the DateTime in order to eliminate the +00:00 offset and use Z
+                            string text = currentDateTimeOffset.UtcDateTime.ToString(format, formatProvider);
```
shyamnamboodiripad, line 190: There is a try catch around the ToString below, should there be one here too?

src/Microsoft.DotNet.Interactive.Http.Parsing/Parsing/DynamicExpressionUtilities.cs
```diff
@@ -183,6 +184,13 @@ private static DateTimeOffset DefaultGetDateTimeOffset(bool isLocal)
                     else if (string.Equals(type.Value, "iso8601", StringComparison.OrdinalIgnoreCase))
                     {
                         format = "o";
+                        if (currentDateTimeOffset.Offset.TotalMinutes == 0)
+                        {
+                            // for $datetime, format the DateTime in order to eliminate the +00:00 offset and use Z
+                            string text = currentDateTimeOffset.UtcDateTime.ToString(format, formatProvider);
+
+                            return node.CreateBindingSuccess(text);
```
shyamnamboodiripad, line 192: Not a huge deal but could we do something to consolidate the return statements for the success cases? There are many returns in this block and it makes the code a little difficult to read...

For example, consider changing it to something like below where each block computes a value for `text` and there is a single return statement at the end of the try.

```
try
{
    string text;
    if (...)
    {
         text = type.Value.ToString("r", CultureInfo.InvariantCulture;
    }
    else if (...)
    {
        text = type.Value.ToString("o", Thread.CurrentThread.CurrentUICulture);
    }
...
    else
    {
        var format = type.Value.Substring(1, type.Value.Length - 2);
        text = type.Value.ToString(format, Thread.CurrentThread.CurrentUICulture);
    }

    return node.CreateBindingSuccess(text);
}
catch (FormatException)
{
   return node.CreateBindingFailure(HttpDiagnostics.IncorrectDateTimeCustomFormat(format));
}
```

shyamnamboodiripad: It may also make sense to extract a separate function for the try block above if you refactor the code to look like above...

src/Microsoft.DotNet.Interactive.Http.Parsing/Parsing/DynamicExpressionUtilities.cs
```diff
@@ -183,6 +184,13 @@ private static DateTimeOffset DefaultGetDateTimeOffset(bool isLocal)
                     else if (string.Equals(type.Value, "iso8601", StringComparison.OrdinalIgnoreCase))
                     {
                         format = "o";
+                        if (currentDateTimeOffset.Offset.TotalMinutes == 0)
```
shyamnamboodiripad, line 187: Do we have unit tests that cover each of the formatting cases covered in this function?

shyamnamboodiripad (approved):
lgtm modulo comments

src/Microsoft.DotNet.Interactive.Http.Parsing/Parsing/DynamicExpressionUtilities.cs
jonsequitur: Let's set up a little time for a group coding session.

jonsequitur (approved):

PR #3750, opened 2024-11-07
Cleanup release pipelines

ocallesp:
- Separate scripts PublishVSCodeExtensionUsingMI.ps1 into two scripts. One only containing code for stable, and another to only contain code for insiders.


eng/publish/PublishStablePolyglotsNotebooks.ps1
```diff
@@ -0,0 +1,119 @@
...
+        }
+    }
+
+    Write-Host "Publishing extension $extension to VS Code Marketplace using Managed Identity..."
+    if ($simulate) {
+        Write-Host "Simulated command: vsce publish --pre-release --packagePath $extension --noVerify --azure-credential"
+    } else {
+        vsce publish --pre-release --packagePath $extension --noVerify --azure-credential
+    }
+
```
AbhitejJohn, line 37: Any chance we can call into the publishinsiders script you have above in here? That way you wouldn't have to duplicate all this block and you wouldn't have to update in two different places.

AbhitejJohn (approved):

PR #3775, opened 2024-11-21
improve Conda environment discovery and logging

jonsequitur:
This PR includes a significant amount of refactoring and some bug fixes related to the Jupyter subkernel features.

Bugs fixed include:

* Improve support for Conda environment completions and kernel connection without having to launch VS Code from a Conda prompt
* Improve completion support for kernelspecs without having to launch VS Code from a Conda prompt
* Enable connecting Jupyter subkernels without having to launch VS Code from a Conda prompt when `--conda-env` parameter is not specified. (It will now default to `base`.)
* Fix directory separators for Conda environment discovery so that it works on Linux and macOS
* Kill console processes started by Jupyter subkernel integration tests
* Update record/playback expectations in Jupyter tests


src/Microsoft.DotNet.Interactive.Jupyter.Tests/JupyterKernelTestBase.cs
```diff
@@ -22,8 +22,8 @@ public abstract class JupyterKernelTestBase : IDisposable
 {
     protected CompositeDisposable _disposables = new();
 
-    // to re-record the tests for simulated playback with JuptyerTestData, set this to true
-    protected const bool RECORD_FOR_PLAYBACK = false;
+    // to re-record the tests for simulated playback with JupyterTestDataAttribute, set this to true
+    protected const bool RECORD_FOR_PLAYBACK = true;
```
AbhitejJohn, line 26: Curious, are we leaving this value change as is? The comment above makes me think this was meant to be false by default.

src/Microsoft.DotNet.Interactive.Jupyter.Tests/MessagePlayback.cs
```diff
@@ -0,0 +1,124 @@
...
+    private readonly ConcurrentQueue<Message> _processRequests = new();
+    private readonly List<Message> _playbackMessages = new();
+    private readonly CancellationTokenSource _cts = new();
+    public IObservable<Message> Messages => _receivedMessages;
+
+    public MessagePlayback(IReadOnlyCollection<Message> messages)
+    {
+        _playbackMessages.AddRange(messages);
+
+        // FIX: (MessagePlayback) make sure this is awaited
```
AbhitejJohn, line 30: Is this Todo still relevant. Seems like it might be helpful to gracefully terminate the task captured here on Dispose.

src/Microsoft.DotNet.Interactive.ApiCompatibility.Tests/ApiCompatibilityTests.jupyter_api_is_not_changed.approved.txt
```diff
@@ -39,13 +39,13 @@ Microsoft.DotNet.Interactive.Jupyter
...
+    public System.Threading.Tasks.Task<Microsoft.DotNet.Interactive.Utility.CommandLineResult> ExecuteAsync(System.String command, System.String args, System.IO.DirectoryInfo workingDir = null, System.Nullable<System.TimeSpan> timeout = null)
     public System.Diagnostics.Process StartProcess(System.String command, System.String args, System.IO.DirectoryInfo workingDir, System.Action<System.String> output = null, System.Action<System.String> error = null)
   public abstract class IJupyterKernelSpecModule
     public System.IO.DirectoryInfo GetDefaultKernelSpecDirectory()
     public IJupyterEnvironment GetEnvironment()
-    public System.Threading.Tasks.Task<Microsoft.DotNet.Interactive.Utility.CommandLineResult> InstallKernel(System.IO.DirectoryInfo sourceDirectory)
-    public System.Threading.Tasks.Task<System.Collections.Generic.IReadOnlyDictionary<System.String,KernelSpec>> ListKernels()
+    public System.Threading.Tasks.Task<Microsoft.DotNet.Interactive.Utility.CommandLineResult> InstallKernelAsync(System.IO.DirectoryInfo sourceDirectory)
```
AbhitejJohn, line 47: Curious what triggered these changes. From what I can read, it does not seem like there is additional validation/compat tied to these API changes but just wanted to confirm.

AbhitejJohn (approved):
Glad to have the additional logging in place. The record and playback infra is quite interesting.

src/Microsoft.DotNet.Interactive.Jupyter.Tests/JupyterKernelTestBase.cs
jonsequitur: Good point. It's not actually clear to me though that having it set to `false` is useful. If integration tests are running, why not update the recordings? 

There's probably another round of test redesign to be done here when I get more familiar with this.

src/Microsoft.DotNet.Interactive.Jupyter.Tests/MessagePlayback.cs
jonsequitur: This one is a long-running task that's intended to do continuous work in the background. It could use some redesign but awaiting it is actually the wrong approach. I'll remove the comment.

Issue #2871, opened 2023-03-23
Auto-run nuget acquisition for SQL and KQL


sharoncxu:
**Is your feature request related to a problem? Please describe.**
Users have to run an acquisition string, e.g.  `#i "nuget: https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-experimental/nuget/v3/index.json"
#r "nuget: Microsoft.DotNet.Interactive.SqlServer, 1.0.0-beta.22606.2"
` to use R and SQL

**Describe the solution you'd like**
Autorun the nuget acqusition for the user so they don't have to take an extra step

![Image](https://github.com/user-attachments/assets/c19e9e22-300d-415b-b7d3-1fd769a5b333)

AbhitejJohn:
We need to figure out if we can fix this after the GA release of Jupyter subkernel support.

AbhitejJohn:
This is Post-GA work.

jonsequitur:
This is partially addressed by #3810. I'm going to close it because actually running the NuGet acquisition code on the user's behalf requires a much wider design discussion, but #3810 satisfies the discoverability aspect of the issue.

PR #3788, opened 2024-12-05
Infer specific data kernel package version in discoverability kernels

jonsequitur:
This PR contains some refactoring for #2871. 

The `SqlDiscoverabilityKernel` and `KqlDiscoverabilityKernel` now infer specific package versions when displaying instructions:

Before:

```
To connect to a database, first add the SQL extension package by running the following in a C# cell:

    #r "nuget:Microsoft.DotNet.Interactive.SqlServer,*-*"
```

After (for example):

```
To connect to a database, first add the SQL extension package by running the following in a C# cell:

    #r "nuget:Microsoft.DotNet.Interactive.SqlServer,1.0.0-beta.24568.1"
```

These instructions will likely go away entirely as part of #2871, and package acquisition will happen automatically, but the current change will give us an opportunity in VS Code Insiders to verify that the heuristics work correctly with official packages.

### DIFF
--- a/src/dotnet-interactive/Connection/KqlDiscoverabilityKernel.cs
+++ b/src/dotnet-interactive/Connection/KqlDiscoverabilityKernel.cs
@@ -8,7 +8,7 @@ using Microsoft.DotNet.Interactive.Commands;
 using Microsoft.DotNet.Interactive.Formatting;
 using static Microsoft.DotNet.Interactive.Formatting.PocketViewTags;
 
-namespace Microsoft.DotNet.Interactive;
+namespace Microsoft.DotNet.Interactive.App.Connection;
 
 /// <remarks>This kernel is used as a placeholder for the MSKQL kernel in order to enable KQL language coloring in the editor. Language grammars can only be defined for fixed kernel names, but MSKQL subkernels are user-defined via the #!connect magic command. So, this kernel is specified in addition to the user-defined kernel as a kind of "styling" kernel as well as to provide guidance and discoverability for KQL features. </remarks>
 public class KqlDiscoverabilityKernel :
@@ -25,7 +25,7 @@ public class KqlDiscoverabilityKernel :
             "MsKqlKernel"
         };
         KernelInfo.LanguageName = "KQL";
-        KernelInfo.Description = $"""
+        KernelInfo.Description = """
                             Query a Kusto database
                             """;
     }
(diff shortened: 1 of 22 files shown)


src/dotnet-interactive/Connection/KqlDiscoverabilityKernel.cs
```diff
@@ -25,7 +25,7 @@ public KqlDiscoverabilityKernel() : base(DefaultKernelName)
             "MsKqlKernel"
         };
         KernelInfo.LanguageName = "KQL";
-        KernelInfo.Description = $"""
+        KernelInfo.Description = """
                             Query a Kusto database
```
shyamnamboodiripad, line 29: minor nit: This could be a simple string.

src/dotnet-interactive/Connection/PackageAcquisition.cs
```diff
@@ -0,0 +1,30 @@
...
+        {
+            informationalVersion = informationalVersion.Split("+")[0];
+        }
+
+        var splitInformationalVersion = informationalVersion.Split('.');
+
+        var version = splitInformationalVersion.Length is 4
+                          ? string.Join(".", splitInformationalVersion[0..4])
+                          : informationalVersion;
+        return version;
```
shyamnamboodiripad, line 28: Are all the conditions (i.e. versions that contain `+`, versions that contain more than 4 parts etc.) in this function reachable in practice for the versions we ship? It may be a good idea to include comments with an example or two for the (supported) formats we expect to encounter.

It would also be nice to have tests that exercise these cases - but I am guessing this may not be easy and perhaps it will not be worth it once #2871 is addressed fully...

shyamnamboodiripad (approved):
👍🏾

AbhitejJohn (approved):

src/dotnet-interactive/Connection/PackageAcquisition.cs
jonsequitur: The number of variations is endless and in practice the build configuration decides what will actually be here. I've already seen a few variants that I didn't expect but that we don't care about, like `1.0.0.ci-12345.1`. I need to see this running with official bits before deciding the heuristic is acceptable.

src/dotnet-interactive/Connection/KqlDiscoverabilityKernel.cs
jonsequitur: I find this syntax requires less cognitive overhead when reading the code. A lot of these user-facing strings contain characters that require escaping, so the raw string syntax makes it clear what the user will actually see, even if in some specific cases like this, there's no escaping present.

PR #3802, opened 2025-01-06
update test result publishing conditions

jonsequitur:


AbhitejJohn (approved):
Just need to add a similar confition at ln139 of the official yml for windows builds as well.

PR #3803, opened 2025-01-07
clean up Jupyter code and disable some flaky tests to unblock builds

jonsequitur:


src/Microsoft.DotNet.Interactive.Jupyter/Heartbeat.cs
```diff
@@ -25,30 +28,34 @@ public Heartbeat(ConnectionInformation connectionInformation)
...
-                {
-                    var data = _server.ReceiveFrameBytes();
-
-                    // Echoing back whatever was received
-                    _server.TrySendFrame(data);
-                }
-            }
-        }, cancellationToken);
+
+        _startReceiveLoop = Task.Factory.StartNew(ReceiveLoop, creationOptions: TaskCreationOptions.LongRunning);
```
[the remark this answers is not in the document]
jonsequitur: If the loop method never checks the `CancellationToken` that won't do anything.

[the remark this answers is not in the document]
AbhitejJohn: Agreed, it was mostly to avoid having it as a global variable. Was thinking more about lambda closure here.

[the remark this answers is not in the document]
jonsequitur: It's not global, it's an instance field, which is deliberate as it makes it easier to inspect and log the state of when investigating unexpected behaviors.

test-retry-runner.ps1
```diff
@@ -9,21 +9,22 @@ $ErrorActionPreference = "Stop"
 
 if ($IsWindows) {
     $projectsToSkip = @(
-        "Microsoft.DotNet.Interactive.CSharpProject.Tests"
+        "Microsoft.DotNet.Interactive.CSharpProject.Tests",
+        "Microsoft.DotNet.Interactive.Jupyter.Tests"
```
shyamnamboodiripad, line 13: Is it intentional that Jupyter tests are skipped entirely on windows - but only partially otherwise?

jonsequitur: No, that was an experiment. Good catch!

src/Microsoft.DotNet.Interactive.Jupyter/Microsoft.DotNet.Interactive.Jupyter.csproj
```diff
@@ -5,6 +5,12 @@
     <LangVersion>latest</LangVersion>
   </PropertyGroup>
 
+  <ItemGroup>
+    <Compile Remove="Utility\**" />
+    <EmbeddedResource Remove="Utility\**" />
+    <None Remove="Utility\**" />
+  </ItemGroup>
```
shyamnamboodiripad, line 12: minor nit: This may be redundant?

jonsequitur: VS inserted this when I removed "missing" Solution Explorer the node using the UI.

PR #3821, opened 2025-01-23
multitarget CSharp and NamedPipeConnector packages

jonsequitur:


src/Microsoft.DotNet.Interactive.CSharp/Microsoft.DotNet.Interactive.CSharp.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>netstandard2.1</TargetFramework>
+    <TargetFrameworks>netstandard2.0;netstandard2.1</TargetFrameworks>
```
shyamnamboodiripad, line 4: Do we need both? Can it be just 2.0?

shyamnamboodiripad (approved):

PR #3830, opened 2025-02-01
change using to open to be relevant to F#

IvanTheGeek:
Update Importing-packages.ipynb
change using to open to be relevant to F#


jonsequitur (approved):
Nice catch!

PR #3835, opened 2025-02-05
Pipeline consolidation

AbhitejJohn:
Refactoring the pipelines to consolidate tasks across our pipelines. Copilot edits made this task a lot easier!
The tasks at the moment are similar to what we had before except for publishing trx files which was different between the two yml's, so I just used *.* for publish instead of *.trx as this PR from Jon already establishes: https://github.com/dotnet/interactive/pull/3834

Todo:
- [X] Need to figure out how "Output" tasks can be consolidated as well - It does not seem like this is possible today. 1ESPT templates require us to use "output" elements but the public pipeline cannot consume this syntax just yet from what I can tell.


eng/templates/variables-template.yml
```diff
@@ -0,0 +1,26 @@
...
+- name: _BuildConfig
+  value: Release
+- name: _PublishUsingPipelines
+  value: true
+- name: _DotNetArtifactsCategory
+  value: .NETCore
+- name: DisableDockerDetector
+  value: true
+- name: PocketLoggerLogPath
+  value: $(Build.SourcesDirectory)/artifacts/log/$(_BuildConfig)
```
jonsequitur, line 14: I'm trying to move this under the TestResults folder so that they get included in the test artifacts output.

eng/templates/common-tasks.yml
```diff
@@ -0,0 +1,162 @@
...
+
+  - pwsh: ./test-retry-runner.ps1 -buildConfig $env:BUILDCONFIG
+    displayName: Test / Blame
+    workingDirectory: $(Build.SourcesDirectory)
+    env:
+      BUILDCONFIG: $(_BuildConfig)
+    condition: ne(variables['SkipTests'], 'true')
+
+  - pwsh: Get-ChildItem *.dmp -Recurse | Remove-Item
+    displayName: Delete dump files
```
jonsequitur, line 143: We should remove this. If we don't want dumps, we can just change the test runs args so we don't bother generating them in the first place.

eng/templates/common-tasks.yml
```diff
@@ -0,0 +1,162 @@
...
+    env:
+      POCKETLOGGER_LOG_PATH: $(PocketLoggerLogPath)
+      TRYDOTNET_PACKAGES_PATH: $(TryDotNetPackagesPath)
+      DOTNET_INTERACTIVE_FRONTEND_NAME: CI
+
+  - pwsh: ./test-retry-runner.ps1 -buildConfig $env:BUILDCONFIG
+    displayName: Test / Blame
+    workingDirectory: $(Build.SourcesDirectory)
+    env:
+      BUILDCONFIG: $(_BuildConfig)
```
jonsequitur, line 139: I *think* we need PocketLogger logs in this step (but I still can't find log files in the build output, so they might not be propagating correctly.)

eng/templates/common-tasks.yml
```diff
@@ -0,0 +1,162 @@
...
+    condition: ne(variables['SkipTests'], 'true')
+
+  # This is an output artifact in the official yml.
+  # - task: PublishBuildArtifacts@1
+  #   displayName: Publish Test results and Blame dumps
```
jonsequitur, line 47: Why comment this one out?

eng/templates/variables-template.yml
AbhitejJohn: perfect, will make that change. I'll likely look through the changes you have going in your PR and pull them in so they are all intact.

eng/templates/common-tasks.yml
AbhitejJohn: gotcha, will slot them in when we confirm.

eng/templates/common-tasks.yml
AbhitejJohn: I've added these back in the parent yml. They sadly cannot be de-duped today.

AbhitejJohn:
@jonsequitur : Both the linux legs seem to be failing on this PR despite pulling the latest main, do you have any thoughts/hints? It does not seem like the yml changes here could affect that directly. Weird that it's been consistent though. I'm going to retry just in case but I'll leave the link to the previous builds [here](https://dev.azure.com/dnceng-public/public/_build/results?buildId=952092&view=results) and [here](https://dev.azure.com/dnceng-public/public/_build/results?buildId=952092&view=logs&j=22fea640-1099-5f32-ec5d-316ad83f359a&t=4258f96f-d981-559d-e1b0-7732a4f8ce19).

jonsequitur:
This is the same problem I've been working on diagnosing. Something is causing the Jupyter test to hang but I can't repro it outside our CI pipeline. Another attempt at a fix is #3853.

AbhitejJohn:
Abandoned in favor of https://github.com/dotnet/interactive/pull/3866

PR #3845, opened 2025-02-11
A few bug fixes

jonsequitur:
* Improve logging and object disposal in flaky Jupyter tests
* Don't call `BlockingCollection.Count` on possibly disposed instance when logging exceptions in `KernelScheduler.IdleAsync`
* Remove a few extraneous extension methods from the public API
* Update copyright dates in `#!about`
* Fix missing `StandardOutputValueProduced` and `StandardErrorValueProduced` events when emitted from a child command
* Remove Polyglot Notebooks setting for minimum .NET version (since it's misleading and won't actually change which SDK version .NET Interactive can run on.)


src/Microsoft.DotNet.Interactive/KernelExtensions.cs
```diff
@@ -136,7 +137,16 @@ public static async Task LoadAndRunInteractiveDocument(
                 {
                     submitCode.SetParent(parentCommand);
                 }
-                await kernel.RootKernel.SendAsync(submitCode);
+
+                var result = await kernel.RootKernel.SendAsync(submitCode);
+
+                // FIX: (LoadAndRunInteractiveDocument) consider publishing completion events only visible to the inner command so that he API doesn't have this difference in behavior.
+                if (result.Events.Count == 0)
+                {
```
shyamnamboodiripad, line 145: nit: Consolidate the condition to avoid the empty block?

AbhitejJohn (approved):

src/Microsoft.DotNet.Interactive/KernelInvocationContextExtensions.cs
```diff
@@ -97,29 +97,4 @@ public static void DisplayStandardError(
                 command ?? context.Command,
                 formattedValues));
     }
-
-    public static void PublishValueProduced(
-        this KernelInvocationContext context,
-        RequestValue requestValue,
-        object value)
```
shyamnamboodiripad, line 104: This helper seems like it could still be useful to eliminate some boiler plate. Is this being removed to reduce public surface area?

src/Microsoft.DotNet.Interactive/KernelExtensions.cs
jonsequitur: Removing this entirely. It was only here for code coverage feedback.

src/Microsoft.DotNet.Interactive/KernelInvocationContextExtensions.cs
jonsequitur: This helper code is much more verbose than necessary. I compressed it and inlined it. 

In general, helper methods for simple operations should not be part of the public API, especially considering:

* It's specific to one of dozens of possible events, so there's an inconsistency about why this event type and not others have such helpers.
* In IntelliSense, it distracts from the general-purpose `Publish` method.

shyamnamboodiripad (approved):

PR #3846, opened 2025-02-12
Changing Named Request Parsing to NewtonSoft

bleaphar:
This PR reflects a usage dependency on Newtonsoft in the named request parsing. This switch makes it easier to get values from JArrays and has more robust support compared to System.Text.Json. Since the dependency exists in other parts of interactive already, it is not adding a new dependency to the solution.


jonsequitur:
We're generally trying to remove this dependency before stabilizing the API.

Is there a bug or requested feature that requires a capability that NewtonSoft has and System.Text.Json can't be made to do?

PR #3852, opened 2025-02-14
Support in named requests for array indexing

bleaphar:
This PR adds support for resolving JSON path information in named requests by adding a regex pattern and identifying if the users is trying to access and an array explicitly and handling that case.

### DIFF
--- a/src/Microsoft.DotNet.Interactive.Http.Tests/HttpKernelTests.NamedRequest.cs
+++ b/src/Microsoft.DotNet.Interactive.Http.Tests/HttpKernelTests.NamedRequest.cs
@@ -904,6 +904,38 @@ namespace Microsoft.DotNet.Interactive.Http.Tests
             }
 
             [Fact]
+            public async Task accessing_an_index_in_a_json_array_succeeds()
+            {
+                using var kernel = new HttpKernel();
+                var firstRequest = """                    
+                    # @name sampleArray
+                    POST https://httpbin.org/anything
+                    {
+                      "devices": [
+                        {
+                          "id": "5601db0f-32e0-4d82-bc79-251e50fa1407",
+                          "name": "Foo"
+                        },
+                        {
+                          "id": "455301a5-8a6e-49d0-b056-96fb2847be18",
+                          "name": "Bar"
+                        }
+                      ]
+                    }
+                    ###
+                    """;
+                var firstResult = await kernel.SendAsync(new SubmitCode(firstRequest));
+                firstResult.Events.Should().NotContainErrors();
+                var secondRequest = """
+                    GET https://httpbin.org/headers
+                    X-Value: {{sampleArray.response.body.$.json.devices[0].id}}
+                    ###
+                    """;
+                var secondResult = await kernel.SendAsync(new SubmitCode(secondRequest));
+                secondResult.Events.Should().NotContainErrors();
+            }
+
+            [Fact]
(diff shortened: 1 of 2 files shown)


src/Microsoft.DotNet.Interactive.Http.Tests/HttpKernelTests.NamedRequest.cs
```diff
@@ -903,6 +903,38 @@ public async Task accessing_non_existent_header_names_produces_an_error(string p
...
+                    """;
+                var firstResult = await kernel.SendAsync(new SubmitCode(firstRequest));
+                firstResult.Events.Should().NotContainErrors();
+                var secondRequest = """
+                    GET https://httpbin.org/headers
+                    X-Value: {{sampleArray.response.body.$.json.devices[0].id}}
+                    ###
+                    """;
+                var secondResult = await kernel.SendAsync(new SubmitCode(secondRequest));
+                secondResult.Events.Should().NotContainErrors();
```
jonsequitur, line 935: It would be good to assert as well that the expected value is returned.

bleaphar: Good catch, made the change

src/Microsoft.DotNet.Interactive.Http.Tests/HttpKernelTests.NamedRequest.cs
```diff
@@ -903,6 +903,44 @@ public async Task accessing_non_existent_header_names_produces_an_error(string p
...
+                        }
+                      ]
+                    }
+                    ###
+                    """;
+                var firstResult = await kernel.SendAsync(new SubmitCode(firstRequest));
+                firstResult.Events.Should().NotContainErrors();
+                var secondRequest = """
+                    GET https://httpbin.org/headers
+                    X-Value: {{sampleArray.response.body.$.json.devices[0].id}}
```
jonsequitur, line 931: It might be worthwhile to test a case where there's more than one array index in the expression.

src/Microsoft.DotNet.Interactive.Http/Model/HttpNamedRequest.cs
```diff
@@ -263,7 +269,32 @@ internal HttpNamedRequest(HttpRequestNode httpRequestNode, HttpResponse response
...
+                var pathMatch = arrayRegex.Match(path[currentIndex]);
+                if (pathMatch.Success)
+                {
+
+                    // Extract the array name and index
+                    string arrayName = pathMatch.Groups[1].Value;
+                    string indexString = pathMatch.Groups[2].Value;
+
+                    // Cast the index from string to int
+                    int index = int.Parse(indexString);
```
drewnoakes, line 284: `TryParse` here? A very long sequence of digits might not fit within a 32-bit int.

src/Microsoft.DotNet.Interactive.Http/Model/HttpNamedRequest.cs
```diff
@@ -263,7 +269,32 @@ internal HttpNamedRequest(HttpRequestNode httpRequestNode, HttpResponse response
...
+
+                    newResponseJSON = j[arrayName][index];
```
drewnoakes, line 286: Does this need bounds checking? What's the expected behaviour when OOB?

jonsequitur (approved):

PR #3874, opened 2025-02-27
downgrade some dependencies to .NET 8 versions

jonsequitur:


shyamnamboodiripad (approved):
Does this PR update all BCL packages that were bumped to 9.* back to 8.* or is it just a subset for now?

Directory.Build.props
```diff
@@ -19,18 +19,21 @@
...
     <MicrosoftNETTestSdkVersion>17.12.0</MicrosoftNETTestSdkVersion>
     <NewtonsoftJsonVersion>13.0.3</NewtonsoftJsonVersion>
+    <SystemCollectionsImmutableVersion>8.0.0</SystemCollectionsImmutableVersion>
     <SystemDrawingCommonVersion>9.0.0</SystemDrawingCommonVersion>
     <SystemReactiveVersion>6.0.0</SystemReactiveVersion>
+    <SystemReflectionMetadataVersion>8.0.0</SystemReflectionMetadataVersion>
     <SystemSecurityCryptographyPkcsVersion>9.0.0</SystemSecurityCryptographyPkcsVersion>
     <SystemSecurityCryptographyXmlVersion>9.0.0</SystemSecurityCryptographyXmlVersion>
-    <SystemTextJsonVersion>9.0.0</SystemTextJsonVersion>
+    <SystemTextJsonVersion>8.0.0</SystemTextJsonVersion>
```
shyamnamboodiripad, line 36: I think this may trigger component governance alerts. I am guessing it may be best to go back to the minor version that we were on before we bumped everything to 9.*

jonsequitur:
> Does this PR update all BCL packages that were bumped to 9.* back to 8.* or is it just a subset for now?

It's just a subset, based on issues @jfleisher was having embedding a kernel in VS. 

Bumping all versions back would prevent C# 13 from being used in .NET Interactive.

shyamnamboodiripad:
> > Does this PR update all BCL packages that were bumped to 9.* back to 8.* or is it just a subset for now?
> 
> It's just a subset, based on issues @jfleisher was having embedding a kernel in VS.
> 
> Bumping all versions back would prevent C# 13 from being used in .NET Interactive.

Ok hopefully this is the set that matters, and we won't have to peel the onion further😄

PR #3887, opened 2025-03-19
Embedded kernel sample and DbContext generation fixes

jonsequitur:
This fixes a few issues:

* The WpfConnect sample showing how to embed a kernel and attach to it using a named pipe was outdated.
* An issue where `SecretManager` throwing an exception when updating the recent connections list is not useful since it's likely not an end-user problem, and could cause execution of a `#!connect` to appear to fail when it had actually succeeded. I changed this to just log an error. (The underlying failure is possibly specific to my environment and I'm still investigating.)
* The move to central package management broke the codegen used in EntityFramework `DbContext` generation to reference required package versions. This is fixed and better tested now.

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -5,6 +5,14 @@
     
     <RoslynVersion>4.12.0</RoslynVersion>
   </PropertyGroup>
+  
+  <PropertyGroup Label="Properties used for generating package references (#r nuget) at runtime. Inlining these might cause some tests to fail.">
+    <HumanizerVersion>2.14.1</HumanizerVersion>
+    <MicrosoftDataSqlClientVersion>5.2.2</MicrosoftDataSqlClientVersion>
+    <MicrosoftIdentityClientVersion>4.66.2</MicrosoftIdentityClientVersion>
+    <MicrosoftEntityFrameworkVersion>9.0.3</MicrosoftEntityFrameworkVersion>
+  </PropertyGroup>
+  
 
   <ItemGroup>
     <PackageVersion Include="Assent" Version="2.3.2" />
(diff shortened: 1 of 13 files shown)


Directory.Packages.props
```diff
@@ -96,7 +104,7 @@
     <PackageVersion Include="System.Security.Cryptography.Pkcs" Version="9.0.0" />
     <PackageVersion Include="System.Security.Cryptography.X509Certificates" Version="4.3.2" />
     <PackageVersion Include="System.Security.Cryptography.Xml" Version="9.0.0" />
-    <PackageVersion Include="System.Text.Json" Version="9.0.0" />
+    <PackageVersion Include="System.Text.Json" Version="9.0.3" />
```
shyamnamboodiripad, line 107: I hope these versions have been bumped in VS too :) Are all the bumps here required to fix the problem or can any of them wait a bit until we flow the latest interactive changes into VS?

samples/connect-wpf/NuGet.config
```diff
@@ -3,8 +3,13 @@
   <solution>
     <add key="disableSourceControlIntegration" value="true" />
   </solution>
+  
   <packageSources>
     <clear />
-    <add key="dotnet-public" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-public/nuget/v3/index.json" />
+    <add key="nuget.org" value="https://api.nuget.org/v3/index.json" />
```
shyamnamboodiripad, line 9: Is this change intentional or just something you were testing? cc @AbhitejJohn in case there is any policy around feeds that should be used in the dotnet repo.

shyamnamboodiripad (approved):

samples/connect-wpf/NuGet.config
jonsequitur: I was testing out a package version that's not in the dotnet-public repo but this can be reverted.

AbhitejJohn (approved):

samples/connect-wpf/NuGet.config
jonsequitur: It's a sample and the key reason it needs to be here is because I wanted to override and decouple from the Directory.Packages.props in the root folder.

PR #3910, opened 2025-04-10
Adds `--http-local-only` option to bypass networking probing

### LINKED ISSUE #3903: Add fallback to `NetworkInterface.GetAllNetworkInterfaces()`
When running on an proot/chroot Android enviroment, non-rooted apps cannot access /sys/class/net/. It's restricted by SELiux policy

Change the life-cycle of `Microsoft.DotNet.Interactive.Http.HttpProbingSettings` to be able to read command line arguments
Add a command line argument for bypassing HttpProbing and providing a port directly

So far, since the SELiux policy change on Android, the `NetworkInterface.GetAllNetworkInterfaces()` wont work unless the device is rooted.
I heard that Android offers APIs as an alternative to trying to read `/sys/class/net/` directly, preharps it would be possible to create an issue documenting the restriction on the arm64 linux runtime.
For now, Im considering using other REPL alternatives, but the lack of Jupyter integration is going to be missed.
Im also considerimg taking this Issue to the proot/chroot on anroid community instead and create a daemon that populates the `/sys/class/net` instead.

Luiz-Ossinho:
# Add fallback to `NetworkInterface.GetAllNetworkInterfaces()` [Fixes #3903]

- All commits should be squashed before merge.

### Content
I’ve applied the same logic used in other "server startup" related options.  
Please let me know if you spot any issues or inconsistencies in the current implementation.

### Styling
I believe I’m following the [specified coding guidelines](https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/coding-style.md).  
Let me know if you notice anything that needs improvement.

### DIFF
--- a/src/dotnet-interactive/CommandLine/CommandLineParser.cs
+++ b/src/dotnet-interactive/CommandLine/CommandLineParser.cs
@@ -171,9 +171,15 @@ public static class CommandLineParser
                 description: LocalizationResources.Cli_dotnet_interactive_jupyter_install_http_port_range_Description(),
                 isDefault: true);
 
+            var httpLocalOnlyOption = new Option<bool>(
+                "--http-local-only",
+                description: LocalizationResources.Cli_dotnet_interactive_jupyter_http_local_only_Description()
+            );
+
             var jupyterCommand = new Command("jupyter", LocalizationResources.Cli_dotnet_interactive_jupyter_Description())
             {
                 defaultKernelOption,
+                httpLocalOnlyOption,
                 httpPortRangeOption,
                 new Argument<FileInfo>
                 {
@@ -272,6 +278,11 @@ public static class CommandLineParser
                     return new HttpPort(portNumber);
                 });
 
+            var httpLocalOnlyOption = new Option<bool>(
+                "--http-local-only",
+                description: LocalizationResources.Cli_dotnet_interactive_jupyter_http_local_only_Description()
+            );
+
             var kernelHostOption = new Option<Uri>(
                 "--kernel-host",
                 parseArgument: x => x.Tokens.Count == 0 ? KernelHost.CreateHostUriForCurrentProcessId() : KernelHost.CreateHostUri(x.Tokens[0].Value),
@@ -292,6 +303,7 @@ public static class CommandLineParser
                 defaultKernelOption,
                 httpPortRangeOption,
                 httpPortOption,
+                httpLocalOnlyOption,
                 kernelHostOption,
                 previewOption,
                 workingDirOption
--- a/src/dotnet-interactive/CommandLine/CommandLineParser.cs
+++ b/src/dotnet-interactive/CommandLine/CommandLineParser.cs
@@ -278,11 +278,6 @@ public static class CommandLineParser
                     return new HttpPort(portNumber);
                 });
 
-            var httpLocalOnlyOption = new Option<bool>(
-                "--http-local-only",
-                description: LocalizationResources.Cli_dotnet_interactive_jupyter_http_local_only_Description()
-            );
-
             var kernelHostOption = new Option<Uri>(
                 "--kernel-host",
                 parseArgument: x => x.Tokens.Count == 0 ? KernelHost.CreateHostUriForCurrentProcessId() : KernelHost.CreateHostUri(x.Tokens[0].Value),
@@ -303,7 +298,6 @@ public static class CommandLineParser
                 defaultKernelOption,
                 httpPortRangeOption,
                 httpPortOption,
-                httpLocalOnlyOption,
                 kernelHostOption,
                 previewOption,
                 workingDirOption
--- a/src/dotnet-interactive/Http/HttpProbingSettings.cs
+++ b/src/dotnet-interactive/Http/HttpProbingSettings.cs
@@ -11,50 +11,37 @@ namespace Microsoft.DotNet.Interactive.Http;
 
 internal class HttpProbingSettings
 {
-    public Uri[] AddressList { get; private set; }
+    public IEnumerable<string> AddressList { get; private set; }
 
-    public static HttpProbingSettings Create(int? port)
+    public static HttpProbingSettings Create(int? httpPort, bool httpLocalOnly)
     {
+        HashSet<string> ipAddress = [IPAddress.Loopback.ToString()];
+
+        if (!httpLocalOnly)
+        {
+            ipAddress = NetworkInterface.GetAllNetworkInterfaces()
+                .Where(ni => ni.OperationalStatus == OperationalStatus.Up)
+                .SelectMany(ni => ni.GetIPProperties().UnicastAddresses)
+                .Select(x => x.Address.ToString())
+                .Concat(ipAddress)
+                .ToHashSet();
+        }
+
+        var uriAddresses = ipAddress
+            .Select(AddHttpPort(httpPort));
+
         return new HttpProbingSettings
         {
-            AddressList = GetProbingAddressList(port)
+            AddressList = uriAddresses
         };
     }
 
-    private static Uri[] GetProbingAddressList(int? httpPort)
+    private static Func<string, string> AddHttpPort(int? httpPort)
     {
-        var sources = new List<string>();
-        foreach (var ni in NetworkInterface.GetAllNetworkInterfaces())
-        {
-            if (ni.OperationalStatus == OperationalStatus.Up)
-            {
-                foreach (var ip in ni.GetIPProperties().UnicastAddresses.Select(a => a.Address.ToString()))
-                {
-
-                    if (ip != IPAddress.Loopback.ToString())
-                    {
-                        sources.Add(ip);
-                    }
-                }
-            }
-        }
-
-        sources.Add(IPAddress.Loopback.ToString());
-
-        var addresses = sources
-            .Where(s => !string.IsNullOrWhiteSpace(s))
-            .Select(s =>
-            {
-                var uriString = httpPort is not null ? $"http://{s}:{httpPort}/" : $"http://{s}/";
-                if (Uri.TryCreate(uriString, UriKind.Absolute, out var uri))
-                {
-                    return uri;
-                }
-
-                return null;
-            })
+        if (httpPort is null)
+            return ipAddress => $"http://{ipAddress}/";
             .Where(u => u is not null)
-            .ToArray();
+        return ipAddress => $"http://{ipAddress}:{httpPort}/";
 
         return addresses;
     }
--- a/src/dotnet-interactive/Http/HttpProbingSettings.cs
+++ b/src/dotnet-interactive/Http/HttpProbingSettings.cs
@@ -13,19 +13,14 @@ public class HttpProbingSettings
 {
     public IEnumerable<string> AddressList { get; private set; }
 
-    public static HttpProbingSettings Create(int? httpPort, bool httpLocalOnly)
+    public static HttpProbingSettings Create(int? httpPort, Func<NetworkInterface[]> getAllNetworkInterfaces)
     {
-        HashSet<string> ipAddress = [IPAddress.Loopback.ToString()];
-
-        if (!httpLocalOnly)
-        {
-            ipAddress = GetAllNetworkInterfaces()
+        var ipAddress = getAllNetworkInterfaces()
                 .Where(ni => ni.OperationalStatus == OperationalStatus.Up)
                 .SelectMany(ni => ni.GetIPProperties().UnicastAddresses)
                 .Select(x => x.Address.ToString())
-                .Concat(ipAddress)
+                .Append(IPAddress.Loopback.ToString())
                 .ToHashSet();
-        }
 
         var uriAddresses = ipAddress
             .Select(AddHttpPort(httpPort));
@@ -42,16 +37,5 @@ public class HttpProbingSettings
             return ipAddress => $"http://{ipAddress}/";
 
         return ipAddress => $"http://{ipAddress}:{httpPort}/";
-    }
-
-    // Delegate that matches the signature of GetAllNetworkInterfaces
-    public delegate NetworkInterface[] GetAllNetworkInterfacesDelegate();
-
-    // Property to replace the implementation for testing
-    public static GetAllNetworkInterfacesDelegate GetAllNetworkInterfacesImpl { get; set; } = NetworkInterface.GetAllNetworkInterfaces;
-
-    private static NetworkInterface[] GetAllNetworkInterfaces()
-    {
-        return GetAllNetworkInterfacesImpl();
     }
 }
--- a/src/dotnet-interactive/Http/HttpProbingSettings.cs
+++ b/src/dotnet-interactive/Http/HttpProbingSettings.cs
@@ -9,7 +9,7 @@ using System.Net.NetworkInformation;
 
 namespace Microsoft.DotNet.Interactive.Http;
 
-internal class HttpProbingSettings
+public class HttpProbingSettings
 {
     public IEnumerable<string> AddressList { get; private set; }
 
@@ -19,7 +19,7 @@ internal class HttpProbingSettings
 
         if (!httpLocalOnly)
         {
-            ipAddress = NetworkInterface.GetAllNetworkInterfaces()
+            ipAddress = GetAllNetworkInterfaces()
                 .Where(ni => ni.OperationalStatus == OperationalStatus.Up)
                 .SelectMany(ni => ni.GetIPProperties().UnicastAddresses)
                 .Select(x => x.Address.ToString())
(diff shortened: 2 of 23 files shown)


Luiz-Ossinho:
> @Luiz-Ossinho please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> @dotnet-policy-service agree [company="{your company}"]
> ```
> 
> > Options:
> > 
> > * (default - no company specified) I have sole ownership of intellectual property rights to my Submissions and I am not making Submissions in the course of work for my employer.
> > 
> > ```
> > @dotnet-policy-service agree
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > * (when company given) I am making Submissions in the course of work for my employer (or my employer has intellectual property rights in my Submissions by contract or applicable law). I have permission from my employer to make Submissions and enter into this Agreement on behalf of my employer. By signing below, the defined term “You” includes me and my employer.
> > 
> > ```
> > @dotnet-policy-service agree company="Microsoft"
> > ```
> 
> Contributor License Agreement

@dotnet-policy-service agree

src/dotnet-interactive/Http/HttpProbingSettings.cs
```diff
@@ -13,39 +13,38 @@ internal class HttpProbingSettings
...
+                        }
                     }
                 }
             }
-        }
 
-        sources.Add(IPAddress.Loopback.ToString());
-
-        var addresses = sources
-            .Where(s => !string.IsNullOrWhiteSpace(s))
```
jonsequitur, line 45: Was it intentional to remove the `Where` call and if so, what's the reasoning?

src/dotnet-interactive/xlf/Resources.cs.xlf
```diff
@@ -52,6 +52,11 @@
         <target state="translated">Nainstaluje specifikace jádra do zadaného adresáře.</target>
         <note />
       </trans-unit>
+      <trans-unit id="Cli_dotnet_interactive_jupyter_local_only_network_interfaces_Description">
+        <source>Exposes ports only on local network interfaces</source>
+        <target state="new">Exposes ports only on local network interfaces</target>
+        <note />
+      </trans-unit>
```
jonsequitur, line 59: It's not necessary to update any of the `.xlf` files.

src/dotnet-interactive/CommandLine/CommandLineParser.cs
```diff
@@ -171,9 +171,16 @@ Command Jupyter()
                 description: LocalizationResources.Cli_dotnet_interactive_jupyter_install_http_port_range_Description(),
                 isDefault: true);
 
+            var localOnlyNetworkInterfacesOption = new Option<bool>(
+                "--local-only-network-interfaces",
+                description: LocalizationResources.Cli_dotnet_interactive_jupyter_local_only_network_interfaces_Description(),
+                getDefaultValue: () => false
```
jonsequitur, line 177: There's no need to specify a default value since `false` is already the default.

src/dotnet-interactive/CommandLine/CommandLineParser.cs
```diff
@@ -171,9 +171,16 @@ Command Jupyter()
                 description: LocalizationResources.Cli_dotnet_interactive_jupyter_install_http_port_range_Description(),
                 isDefault: true);
 
+            var localOnlyNetworkInterfacesOption = new Option<bool>(
+                "--local-only-network-interfaces",
```
jonsequitur, line 175: Maybe `--http-local-only` to align better with the `--http-port-range` option?

src/dotnet-interactive/Http/HttpProbingSettings.cs
Luiz-Ossinho: Im changing this entire block to be more fluent, and refactoring code that references it to accept `IEnumerable<string>` instead of `Uri[]`. 
I believe this improves readability, and perfomance to a lesser extent.

Every time this property was referenced, it used `Select(x => x.AbsoluteUri)`, making it pointless to be in an Uri constructed with `UriKind.Absolute` anyway.

I removed the null or empty check when dealing with the address of network interfaces because:
```
The Address property of a UnicastIPAddressInformation object is of type IPAddress, which is a non-nullable reference type in this context.
When you call ToString() on an IPAddress object, it will always return a formatted string representation of the IP address (like "192.168.1.1" for IPv4 or "2001:db8::1" for IPv6). The IPAddress.ToString() method is implemented to always return a valid string representation.

The only ways this query could fail would be:
- If GetAllNetworkInterfaces() throws an exception (e.g., due to insufficient permissions)
- If any of the network interfaces' GetIPProperties() throws an exception
- If access to UnicastAddresses throws an exception

But in terms of the values contained in the resulting IEnumerable<string>, you will not get null or empty string items. The collection itself could be empty if there are no network interfaces or no unicast addresses, but the individual items will always be valid IP address strings.
```

As Im no longer working with `Uri`, it made no sense to use `Uri.TryCreate`.
I thought about adding `.Where(uriAddress => Uri.IsWellFormedUriString(uriAddress, UriKind.Absolute))` in its place, but it seemed weird to verify the validity of a output that is hardcoded, in this case, either "http://{ipAddress}:{httpPort}/" or "http://{ipAddress}/", considering that the ipAddress itself is a valid, as previously justifyied.

In that same spirit, I also went and 
- implemented the [CA1837](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1837) suggestion at `HttpApiBootstrapper.cs, line 15`
- changed interpolated the values directly instead of using an expensive `.Replace()`.
- changed the string building at line `HttpApiBootstrapper.cs, line 109` to a simpler `JsonSerializer.Serialize(probingUris)`

Luiz-Ossinho:
Im changing the [entire block around AddressList](https://github.com/dotnet/interactive/pull/3910/files#diff-1d1e7e86074064fc1b7ec38f9292a98a467345b0be8681b4a3b3fcd98a2afe33) to be more fluent, and refactoring code that references it to accept IEnumerable<string> instead of Uri[].
I believe this improves readability, and perfomance to a lesser extent.

Every time this property was referenced, it used Select(x => x.AbsoluteUri), making it pointless to be in an Uri constructed with UriKind.Absolute anyway.

I removed the null or empty check when dealing with the address of network interfaces because:

The Address property of a UnicastIPAddressInformation object is of type IPAddress, which is a non-nullable reference type in this context.
When you call ToString() on an IPAddress object, it will always return a formatted string representation of the IP address (like "192.168.1.1" for IPv4 or "2001:db8::1" for IPv6). The IPAddress.ToString() method is implemented to always return a valid string representation.

The only ways this query could fail would be:
- If GetAllNetworkInterfaces() throws an exception (e.g., due to insufficient permissions)
- If any of the network interfaces' GetIPProperties() throws an exception
- If access to UnicastAddresses throws an exception

But in terms of the values contained in the resulting IEnumerable<string>, you will not get null or empty string items. The collection itself could be empty if there are no network interfaces or no unicast addresses, but the individual items will always be valid IP address strings.
As Im no longer working with Uri, it made no sense to use Uri.TryCreate.
I thought about adding .Where(uriAddress => Uri.IsWellFormedUriString(uriAddress, UriKind.Absolute)) in its place, but it seemed weird to verify the validity of a output that is hardcoded, in this case, either "http://{ipAddress}:{httpPort}/" or "http://{ipAddress}/", considering that the ipAddress itself is a valid, as previously justifyied.

In that same spirit, I also went and

implemented the [CA1837](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1837) suggestion at HttpApiBootstrapper.cs, line 15
changed interpolated the values directly instead of using an expensive .Replace().
changed the string building at line HttpApiBootstrapper.cs, line 109 to a simpler JsonSerializer.Serialize(probingUris)

Luiz-Ossinho:
I was wondering why some tests use a common TestConsole while others create a new one:
```
// CommandLineParserTests.cs 

[Fact]
public async Task It_parses_log_output_directory()
{
  var logPath = new DirectoryInfo(Path.GetTempPath());

  await _parser.InvokeAsync($"jupyter --log-path {logPath} {_connectionFile}", _console);

  _startOptions
      .LogPath
      .FullName
      .Should()
      .Be(logPath.FullName);
}
  
[Fact]
public async Task jupyter_command_returns_error_if_connection_file_path_does_not_exist()
{
    var expected = "not_exist.json";

    var testConsole = new TestConsole();
    await _parser.InvokeAsync($"jupyter {expected}", testConsole);

    testConsole.Error.ToString().Should().ContainAll("File does not exist", "not_exist.json");
}
```

Luiz-Ossinho:
My build is failing due to: `Error reading resource 'Microsoft.DotNet.Interactive.App.resources.dotnet-interactive.js' -- 'Could not find a part of the path 'C:\[REDACTED]\ms-dotnet-interactive-issue3903\src\polyglot-notebooks-browser\dist\dotnet-interactive.js'.'`

image

But I think it wont be a problem, thanks to the CI pipeline

src/dotnet-interactive/xlf/Resources.cs.xlf
Luiz-Ossinho: I did redefined it to back to what it was before any changes introduced in my commits, but the CI pipeline failed with:
---
.packages/microsoft.dotnet.xlifftasks/1.0.0-beta.23475.1/build/Microsoft.DotNet.XliffTasks.targets(84,5): error : (NETCORE_ENGINEERING_TELEMETRY=Build) 'xlf/Resources.cs.xlf' is out-of-date with 'Resources.resx'. Run `msbuild /t:UpdateXlf` to update .xlf files or set UpdateXlfOnBuild=true to update them on every build, but note that it is strongly discouraged to set UpdateXlfOnBuild=true in official/CI build environments as they should not modify source code during the build.

jonsequitur: My mistake. I haven't worked in this repo for a while.

jonsequitur: My mistake. I haven't added one of these strings in a while.

jonsequitur:
> My build is failing due to: `Error reading resource 'Microsoft.DotNet.Interactive.App.resources.dotnet-interactive.js' -- 'Could not find a part of the path 'C:\[REDACTED]\ms-dotnet-interactive-issue3903\src\polyglot-notebooks-browser\dist\dotnet-interactive.js'.'`
> 
> image
> 
> But I think it wont be a problem, thanks to the CI pipeline

You can fix this by running `ensure-symlinks.ps1` as described in the developer guide: https://github.com/dotnet/interactive/blob/570b9b578ced4e9d65863ea7121476d3e2a45f9d/DEVELOPER-GUIDE.md#L97

src/dotnet-interactive/Http/HttpProbingSettings.cs
```diff
@@ -9,53 +9,49 @@
...
-            .Select(s =>
-            {
-                var uriString = httpPort is not null ? $"http://{s}:{httpPort}/" : $"http://{s}/";
-                if (Uri.TryCreate(uriString, UriKind.Absolute, out var uri))
-                {
-                    return uri;
-                }
+    // Delegate that matches the signature of GetAllNetworkInterfaces
+    public delegate NetworkInterface[] GetAllNetworkInterfacesDelegate();
```
jonsequitur, line 48: The custom delegate isn't needed here since it's such a simple type, e.g. it has a single use and isn't a complex enough delegate type to provide some self-documentation benefits from named parameters, etc. 

So it's preferable to use `Func<NetworkInterface[]>`. We're generally trying to reduce the number of types on the way to GA.

src/dotnet-interactive/CommandLine/CommandLineParser.cs
```diff
@@ -292,6 +303,7 @@ Command StdIO()
                 defaultKernelOption,
                 httpPortRangeOption,
                 httpPortOption,
+                httpLocalOnlyOption,
```
jonsequitur, line 306: Are you actually using this in `stdio` mode?

We're planning to remove the HTTP options from this mode entirely.

src/dotnet-interactive/Http/HttpProbingSettings.cs
```diff
@@ -9,53 +9,49 @@
...
 
-                return null;
-            })
-            .Where(u => u is not null)
-            .ToArray();
+    // Property to replace the implementation for testing
+    public static GetAllNetworkInterfacesDelegate GetAllNetworkInterfacesImpl { get; set; } = NetworkInterface.GetAllNetworkInterfaces;
```
jonsequitur, line 51: Rather than introduce static mutable state here, I'd recommend passing this as a parameter to `Create`.

jonsequitur:
> I was wondering why some tests use a common TestConsole while others create a new one:
> 
> ```
> // CommandLineParserTests.cs 
> 
> [Fact]
> public async Task It_parses_log_output_directory()
> {
>   var logPath = new DirectoryInfo(Path.GetTempPath());
> 
>   await _parser.InvokeAsync($"jupyter --log-path {logPath} {_connectionFile}", _console);
> 
>   _startOptions
>       .LogPath
>       .FullName
>       .Should()
>       .Be(logPath.FullName);
> }
>   
> [Fact]
> public async Task jupyter_command_returns_error_if_connection_file_path_does_not_exist()
> {
>     var expected = "not_exist.json";
> 
>     var testConsole = new TestConsole();
>     await _parser.InvokeAsync($"jupyter {expected}", testConsole);
> 
>     testConsole.Error.ToString().Should().ContainAll("File does not exist", "not_exist.json");
> }
> ```

I don't know that there's any particular reason. 

Using local variables rather than fields in a test does make it a little easier to understand what the test is doing, assuming the code is concise like this example is. Also, in test frameworks that reuse class instances it can help reduce shared state flakiness between tests. (Xunit doesn't have this problem.)

jonsequitur:
> refactoring code that references it to accept IEnumerable<string> instead of Uri[] 
> I believe this improves readability, and perfomance to a lesser extent.

The refactoring makes sense and I didn't spot any issues. On this one point, there are pitfalls with using `IEnumerable<T>`:

* Deferred execution means the values can change on subsequent calls, depending on what the caller passed, and sometimes surprising bugs can hide there. The array (or, e.g. `IReadOnlyList<T>` is more constrained and safer.
* `IEnumerable` forces people to use an iterator (e.g. a `foreach` loop), whereas arrays and lists allow `for`, which saves an allocation.

I don't think either of these issues is critical here since this is only evaluated once at startup, but they're good general design considerations.

src/dotnet-interactive/CommandLine/CommandLineParser.cs
Luiz-Ossinho: Nope.

Removing it...

src/dotnet-interactive/Http/HttpProbingSettings.cs
Luiz-Ossinho: Storing the Func in the StartupOptions class was the only way I found where I could both  have testable assertions and change its behaviour according to the flag.

Let me know if you meant something else.

Luiz-Ossinho:
The windows build failed with:


---
```
Starting test execution, please wait...
A total of 1 test files matched the specified pattern.
Blame: Attaching crash dump utility to process testhost (5776).
The active test run was aborted. Reason: Test host process crashed
Data collector 'Blame' message: The specified inactivity time of 10 minutes has elapsed. Collecting hang dumps from testhost and its child processes.
Data collector 'Blame' message: Dumping 5776 - testhost.
Results File: D:\a\_work\1\s\artifacts\TestResults\Release\cloudtest_38ff2a48c000004_2025-04-14_20_21_52.trx

Passed!  - Failed:     0, Passed:     6, Skipped:     0, Total:     6, Duration: 3 s - Microsoft.DotNet.Interactive.Tests.dll (net9.0)
Test Run Aborted.

The active Test Run was aborted because the host process exited unexpectedly. Please inspect the call stack above, if available, to get more information about where the exception originated from.

The test running when the crash occurred: 
Microsoft.DotNet.Interactive.Tests.MultipleInputsWithinMagicCommandsTests.When_multiple_inputs_are_enabled_then_RequestInput_is_not_sent_for_a_magic_command_that_uses_custom_binding

This test may, or may not be the source of the crash.
```
---

I have no idea why this happend. 
I know all code is conected but i believe my changes shouldnt fail that test.
Also, it runs on my windows machine:
image

I dont know if the process crashed due to a timeout, or it crashed and then timed-out.

jonsequitur:
This isn't related to your changes.

We have an occasional flakiness in the test infrastructure and when it crashes it looks like this. It leads to a deadlock and the forced timeout is what crashes the test runner. 

I requeued it.

jonsequitur (approved):

jonsequitur:
Thanks @Luiz-Ossinho!

PR #3912, opened 2025-04-13
Improve PowerShell kernel output: support rich display and standard output separation

eosfor:
**Summary**

This PR improves how output from PowerShell cells is handled in .NET Interactive notebooks. The changes ensure that:

- Plain string values are sent to standard output, consistent with PowerShell behavior.
- Rich .NET objects (like DataFrame) are published as objects, enabling proper rendering using registered formatters.
- System types like SecureString are displayed via formatter infrastructure, and related tests are adjusted accordingly.

**Changes**

- Removed the usage of Out-Default2. Instead, output is collected via Pwsh.Invoke(...).
- Introduced conditional handling:
    - If the result is a string, it’s published using StandardOutputValueProduced.
    - Otherwise, it’s published via KernelInvocationContext.Display(...).
    - Updated the test sends_InputRequest_message_when_submission_requests_user_password_in_powershell to expect rich object display instead of a plain text stream.

**Notes**

- Enables correct rendering of Microsoft.Data.Analysis.DataFrame in PowerShell notebooks.
- Keeps expected console-like behavior for simple string output.
- Uses BaseObject unwrapping from PSObject to support type-based formatter selection.

**Motivation**

These changes align PowerShell kernel behavior more closely with how the C# kernel treat result values, providing a more consistent and powerful notebook experience.


samples/notebooks/powershell/Samples/Using-Mocrosoft-DataFrame.dib
```diff
@@ -0,0 +1,77 @@
+#!meta
```
jonsequitur, line 1: You have a typo in the file name: `Mocrosoft`

You might also consider making this a `.ipynb` and saving it with the outputs. GitHub will display this nicely and people can understand the sample without having to run it.

src/Microsoft.DotNet.Interactive.PowerShell/PowerShellKernel.cs
```diff
@@ -389,14 +389,29 @@ internal bool RunLocally(string code, out string errorMessage, bool suppressOutp
...
+            var result = Pwsh.InvokeAndClearWithResult();
+
+            foreach (var item in result)
+            {
+                var value = item?.BaseObject ?? item;
+
+                if (value is string str)
+                {
+                    var formatted = new FormattedValue("text/plain", str + Environment.NewLine);
+                    KernelInvocationContext.Current?.Publish(
```
jonsequitur, line 407: It's best to avoid calling `KernelInvocationContext.Current` from within .NET Interactive code. This method is really meant as a convenience for notebook users but if possible it's better to use a `KernelInvocationContext` object directly.

One is passed into `HandleAsync` for the main code path, so passing it here as an optional parameter will solve that.

`SecretManager` also calls this method but it's not expecting to display anything directly, so I would say if context isn't passed, don't bother trying to display anything.

src/Microsoft.DotNet.Interactive.Tests/KernelCommandNestingTests.KernelCommandResult_Events.cs
```diff
@@ -209,11 +209,11 @@ public async Task Commands_sent_within_the_code_of_another_command_publish_Stand
             returnedValueFromCSharp
                 .Should().BeOfType<List<KernelEvent>>()
                 .Which
-                .Should().ContainSingle<StandardOutputValueProduced>()
```
jonsequitur, line 212: The test is specifically about making sure that `StandardOutputValueProduced` is propagated correctly when one kernel is executing within another's context.

Maybe change the return value on line 200 to a `string` so that the resulting event will still be `StandardOutputValueProduced`?

samples/notebooks/powershell/Samples/Using-Mocrosoft-DataFrame.dib
eosfor: I think I fixed it

src/Microsoft.DotNet.Interactive.PowerShell/PowerShellKernel.cs
eosfor: updated, added a context as a parameter and using it now

src/Microsoft.DotNet.Interactive.Tests/KernelCommandNestingTests.KernelCommandResult_Events.cs
eosfor: updated it too, pls have a look if that works.

samples/notebooks/powershell/Samples/Using-Microsoft-DataFrame.ipynb
```diff
@@ -0,0 +1,260 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "befbe206",
+   "metadata": {
+    "polyglot_notebook": {
+     "kernelName": "pwsh"
+    }
+   },
+   "outputs": [],
```
jonsequitur, line 57: Feel free to save this with outputs after running it.

src/Microsoft.DotNet.Interactive.PowerShell/PowerShellExtensions.cs
```diff
@@ -27,6 +28,21 @@ public static void InvokeAndClear(this PowerShell pwsh)
         }
     }
 
+    public static Collection<PSObject> InvokeAndClearWithResult(this PowerShell pwsh)
```
jonsequitur, line 31: Since this method is identical behaviorally to `InvokeAndClear`, I would suggest just changing the signature of `InvokeAndClear` to return the result. This will reduce some duplication and the risk that these two methods might eventually diverge in behavior.

src/Microsoft.DotNet.Interactive.Tests/KernelCommandNestingTests.KernelCommandResult_Events.cs
```diff
@@ -195,25 +195,25 @@ public async Task Commands_sent_within_the_code_of_another_command_publish_Stand
...
+            //                                     .Should()
+            //                                     .ContainSingle<ReturnValueProduced>()
+            //                                     .Which.Value;
 
-            returnedValueFromCSharp
-                .Should().BeOfType<List<KernelEvent>>()
-                .Which
+            result.Events
+                // .Should().BeOfType<List<KernelEvent>>()
+                // .Which
```
jonsequitur, line 211: This looks good. Thanks!

You can remove the commented out code.

eosfor: done

src/Microsoft.DotNet.Interactive.PowerShell/PowerShellExtensions.cs
eosfor: running tests

samples/notebooks/powershell/Samples/Using-Microsoft-DataFrame.ipynb
eosfor: well, feels like GitHub still does not display them correctly :(

jonsequitur: Here's an example:

https://github.com/dotnet/interactive/blob/main/samples/connect-wpf/readme.ipynb

jonsequitur (approved):

jonsequitur (approved):

jonsequitur:
Thanks @eosfor!

PR #3954, opened 2025-05-27
Add monthly scheduled build on the 17th for Component Governance scans

ocallesp:


jonsequitur:
Why does this require a separate build?

ocallesp:
we don't need this

PR #3975, opened 2025-06-20
package updates

jonsequitur:


src/Microsoft.DotNet.Interactive/KernelScheduler.cs
```diff
@@ -107,7 +107,7 @@ public async Task IdleAsync()
         }
         catch (InvalidOperationException exception)
         {
-            Log.Warning(exception, _currentlyRunningOperation);
+            Log.Warning("Exception while awaiting child operations", exception, _currentlyRunningOperation);
```
shyamnamboodiripad, line 110: Do we typically localize log messages?

Directory.Packages.props
```diff
@@ -80,35 +76,33 @@
...
+    <PackageVersion Include="System.Runtime.Caching" Version="9.0.6" />
     <PackageVersion Include="System.Runtime.InteropServices" Version="4.3.0" />
     <PackageVersion Include="System.Runtime.Loader" Version="4.3.0" />
-    <PackageVersion Include="System.Security.Cryptography.Pkcs" Version="9.0.5" />
+    <PackageVersion Include="System.Security.Cryptography.Pkcs" Version="9.0.6" />
     <PackageVersion Include="System.Security.Cryptography.X509Certificates" Version="4.3.2" />
-    <PackageVersion Include="System.Security.Cryptography.Xml" Version="9.0.5" />
-    <PackageVersion Include="System.Text.Json" Version="9.0.5" />
+    <PackageVersion Include="System.Security.Cryptography.Xml" Version="9.0.6" />
+    <PackageVersion Include="System.Text.Json" Version="9.0.6" />
```
shyamnamboodiripad, line 103: I am hoping VS has already moved to this version - otherwise we may run into problems when we insert these changes into VS (for the case where we load MS.DN.Interactive and the relevant dependencies in the devenv process as part of setting up the proxy kernels in proc). Is this something you have already tried?

shyamnamboodiripad (approved):
👍🏾 modulo the concerns mentioned in the comments

src/Microsoft.DotNet.Interactive/KernelScheduler.cs
jonsequitur: Nope.

PR #4007, opened 2025-08-01
Sign official builds of the vscode extensions

JoeRobich:
Adds extension signing as outlined in https://eng.ms/docs/cloud-ai-platform/devdiv/vs-services-dougam/vs-marketplace-skofman/visual-studio-marketplace/extension-publisher-guides/publisher-signing-as-microsoft#add-signing-workloads

Used [dotnet/vscode-runtime](https://github.com/dotnet/vscode-dotnet-runtime/tree/main/msbuild) as a guide.

It is possible that your pipeline may need to be granted permission to additional resources.

### DIFF
--- a/azure-pipelines-official.yml
+++ b/azure-pipelines-official.yml
@@ -51,7 +51,7 @@ extends:
   template: v1/1ES.Official.PipelineTemplate.yml@1esPipelines
   parameters:
     sdl:
-      sourceAnalysisPool: 
+      sourceAnalysisPool:
         name: $(DncEngInternalBuildPool)
         image: 1es-windows-2022
         os: windows
@@ -152,6 +152,7 @@ extends:
             - template: /eng/templates/build-and-test-tasks.yml@self
               parameters:
                 platform: windows
+                signType: $(_SignType)
 
       - template: /eng/common/templates-official/jobs/jobs.yml@self
         parameters:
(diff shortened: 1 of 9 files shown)


.devcontainer/devcontainer.json
```diff
@@ -9,7 +9,7 @@
       // Append -bullseye or -focal to pin to an OS version.
       "VARIANT": "9.0",
       // Options
-      "NODE_VERSION": "16"
```
jonsequitur, line 12: We've tended to stick with LTS because it's a little more stable. Latest LTS is 22.

JoeRobich: Since I needed to update it, I just thought go all the way. Will downgrade.

src/polyglot-notebooks-vscode-common/src/lineReader.ts
```diff
@@ -15,12 +15,12 @@ export class LineReader {
 
         let i = this.current.indexOf('\n');
         while (i >= 0) {
-            let line = this.current.substr(0, i);
+            let line = this.current.substring(0, i);
```
JoeRobich, line 18: `substr(start, length?)` is deprecated. However all use of it translated to `substring(start, end?)` directly.

src/polyglot-notebooks-vscode-insiders/tsconfig.json
```diff
@@ -1,10 +1,10 @@
 {
   "compilerOptions": {
     "module": "commonjs",
-    "target": "es6",
+    "target": "ES2018",
```
JoeRobich, line 4: This was necessary for the named regex capture group. This version of ES is long supported in node and vscode.

JoeRobich:
@jonsequitur Ready for a review

PR #4036, opened 2025-09-09
Disable powershell strictmode when invoking npm

JoeRobich:
After [reading about strictmode](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/set-strictmode?view=powershell-7.5#example-2-turn-on-strict-mode-as-version-2-0), I do not think it is problematic to disable it when invoking npm.


AbhitejJohn (approved):

AbhitejJohn:
Merging to see if we can unblock signed builds. Happy to address any concerns in a follow-up.

PR #4111, opened 2025-12-16
Add globalconfig for npm

dbreshears:
Adding pointer to globalconfig for npm install and setting always-auth to true


AbhitejJohn:
@dbreshears : Can we make the same change on networkisolation in [this yml](https://github.com/dbreshears/interactive/blob/9b9ace902bc4660ab3b0cff22d90e057567e7a02/azure-pipelines.yml) file as well, so we can catch failures earlier in PR builds as opposed to CI builds? That'll also help validate the changes being made here

AbhitejJohn (approved):

Issue #4130, opened 2026-01-22
Polyglot notebook stuck on installing NET interactive


dcstevenbrixie-arch:
### Describe the bug
I updated VS code a day ago. Since then, the Polyglot notebooks will not run. I get a notification that the NET Interactive version is installing. I've installed the NET 9 SDK as suggested in on-line forms. I've restarted my computer multiple times, un-installed and installed the extension. It's not working. 
Here's the version of VS Code:
Version: 1.108.1 (system setup)
Commit: 585eba7c0c34fd6b30faac7c62a42050bfbc0086
Date: 2026-01-14T14:55:44.241Z
Electron: 39.2.7
ElectronBuildId: 12953945
Chromium: 142.0.7444.235
Node.js: 22.21.1
V8: 14.2.231.21-electron.0
OS: Windows_NT x64 10.0.26200

### Please complete the following:

**Which version of .NET Interactive are you using?** (In a notebook, run the `#!about` magic command. ):

- OS
    - [x ] Windows 11
    - [ ] Windows 10
    - [ ] macOS
    - [ ] Linux (Please specify distro)
    - [ ] iOS
    - [ ] Android
 - Browser 
    - [x ] Chrome
    - [ ] Edge
    - [ ] Firefox
    - [ ] Safari
 - Frontend
    - [ ] Jupyter Notebook
    - [ ] Jupyter Lab
    - [ ] nteract
    - [x ] Visual Studio Code 
    - [ ] Visual Studio Code Insiders
    - [ ] Visual Studio
    - [ ] Other (please specify)

### Screenshots
If applicable, please add screenshots of the entire VS Code window.

CapAbleMirek:
Just had the same issue.

I had to downgrade Polyglot Notebooks extension to version 
1.0.6323010
 
and disable auto updates.
The latest version released yesterday hangs during installation.

DawMatt:
OK, a number of recent commits in this repo appear to be moving dotnet-interactive across to .NET 10 SDK, but the latest released version in the repo is dotnet-interactive v1.0.6160010 and the latest extension version is trying to install v1.0.707101. 

Even though v1.0.707101 doesn't seem to be officially released yet, the released version of Polyglot is trying to use it and having trouble installing it. Is something to do with the dotnet-interactive release process for the new version causing this issue?

seansweeney:
> UPDATE: After installing .NET 10 SDK and restarting VS Code, this problem has been resolved for me.

This worked for me as well.

PR #4151, opened 2026-01-30
Resolve create tool manifest issue #1430 using --force

### LINKED ISSUE #1430: Start-Transcript in Powershell .NET Interactive notebook hides future output
[PowerShellTranscriptFailure.zip](https://github.com/dotnet/interactive/files/6672285/PowerShellTranscriptFailure.zip)

Once Start-Transcript appears on a PowerShell notebook, any attempt to output to the notebook fails.
See attached notebook for a repro.

geertjanthomas:
Fixes dotnet/interactive#1430 Polyglot notebook stuck on installing NET interactive.

https://github.com/dotnet/interactive/issues/4130

The issue is that, at least for some users, during extension activation the command "dotnet new tool-manifest" failed because this file already exists and/or could not be overwritten:

c:\Users\****\AppData\Roaming\Code\User\globalStorage\ms-dotnettools.dotnet-interactive-vscode\dotnet-tools.json

By adding the '--force' argument this is resolved.

@jonsequitur

### DIFF
--- a/src/polyglot-notebooks-vscode-common/src/commands.ts
+++ b/src/polyglot-notebooks-vscode-common/src/commands.ts
@@ -61,7 +61,7 @@ export async function registerAcquisitionCommands(context: vscode.ExtensionConte
     }));
 
     async function createToolManifest(dotnetPath: string, globalStoragePath: string): Promise<void> {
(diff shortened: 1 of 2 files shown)


geertjanthomas:
@twcrews Thanks for your remark. I saw that other PR as well and to be honest, while both are fixing the issue, neither identify the root cause and why it does not affect all users. As a user I just want the issue to go away because it is really annoying.

